PR #29, opened 2026-06-01
fix: resolve Windows build/check crash and npm shrinkwrap registry 404

SHJordan:
This PR fixes two issues:

1. **Windows build/check crash:**
   - In `scripts/build-all.mjs` and `scripts/run-web-ui-check.mjs`, the script was trying to evaluate native binaries (such as `pnpm.exe` on Windows, which is set in `npm_execpath` by newer pnpm installations) using Node.js, causing a `SyntaxError`. We fixed this by detecting if the execpath does not end in `.js`/`.cjs` (which indicates a native binary) and running it directly instead of wrapping it via Node.

2. **Package shrinkwrap dependency registry 404 error:**
   - The generated `npm-shrinkwrap.json` for `@code-yeongyu/senpi` contained registry references to internal, private workspace packages (like `@earendil-works/pi-tui`) which are bundled but not actually published to the public registry. This resulted in `404 Not Found` errors when users tried to install `@code-yeongyu/senpi`.
   - We updated `scripts/generate-coding-agent-shrinkwrap.mjs` to mark internal dependencies as `"inBundle": true` and omit `resolved` URL registry references, forcing npm to resolve them from the bundled contents.

---
## Summary by cubic
Fixes Windows build/check crashes by executing native package manager binaries directly. Also updates shrinkwrap generation to bundle internal packages, avoiding registry 404s when installing `@code-yeongyu/senpi`.

- **Bug Fixes**
  - Detect native package manager execpaths (e.g., `pnpm.exe`) and run them directly in build/check scripts; otherwise load `.js/.cjs` via Node.
  - Mark internal workspace deps (`@earendil-works/pi-agent-core`, `@earendil-works/pi-ai`, `@earendil-works/pi-tui`) as `"inBundle": true` and omit `resolved` URLs in the generated `npm-shrinkwrap.json`.

<sup>Written for commit 59e2becc97c0a20ed32078da9f4ea02ea5218e1b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/29?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/npm-shrinkwrap.json
+++ b/packages/coding-agent/npm-shrinkwrap.json
@@ -487,7 +487,7 @@
 		},
 		"node_modules/@earendil-works/pi-agent-core": {
 			"version": "2026.6.15",
-			"resolved": "https://registry.npmjs.org/@earendil-works/pi-agent-core/-/pi-agent-core-2026.6.15.tgz",
+			"inBundle": true,
 			"license": "MIT",
 			"dependencies": {
 				"@earendil-works/pi-ai": "^2026.6.15",
@@ -501,7 +501,7 @@
 		},
 		"node_modules/@earendil-works/pi-ai": {
 			"version": "2026.6.15",
-			"resolved": "https://registry.npmjs.org/@earendil-works/pi-ai/-/pi-ai-2026.6.15.tgz",
+			"inBundle": true,
 			"license": "MIT",
 			"dependencies": {
 				"@anthropic-ai/sdk": "0.91.1",
@@ -528,7 +528,7 @@
 		},
 		"node_modules/@earendil-works/pi-tui": {
 			"version": "2026.6.15",
-			"resolved": "https://registry.npmjs.org/@earendil-works/pi-tui/-/pi-tui-2026.6.15.tgz",
+			"inBundle": true,
 			"license": "MIT",
 			"dependencies": {
 				"get-east-asian-width": "1.6.0",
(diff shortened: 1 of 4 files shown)


code-yeongyu:
[sisyphus-bot] Rebased this branch onto the latest `main` and merged it via a merge commit.

The only conflict was the generated `packages/coding-agent/npm-shrinkwrap.json`. Rather than hand-merging it, I resolved it by regenerating the lockfile from scratch (`node scripts/generate-coding-agent-shrinkwrap.mjs`) on top of current `main`, then validated with `node scripts/generate-coding-agent-shrinkwrap.mjs --check`, which exits 0. The regeneration preserved your core fix: the internal `@earendil-works/pi-*` workspace deps are marked `inBundle: true` instead of carrying a `resolved` registry URL that 404s.

I verified both fixes:
- The native package-manager guard `pm.execpath && (pm.cmd === "bun" || !/\.[cm]?js$/i.test(pm.execpath))` is present in both `scripts/build-all.mjs` and `scripts/run-web-ui-check.mjs`. This is backward-compatible: for bun/npm/pnpm-as-.js it behaves exactly as before, and the new branch only triggers for non-.js execpaths such as `pnpm.exe`. The Windows-specific path can't be exercised on the macOS CI here, so it was reviewed rather than run on Windows.
- The shrinkwrap generator change (`inBundle` added to the field order and set on internal workspace deps) is in place, and the `--check` validation passes.

Thanks for the careful fix, @SHJordan.

PR #259, opened 2026-07-21
fix(coding-agent): replay MCP session state after raced registration + process-group QA cleanup

code-yeongyu:
## What (reworked after #260)
Originally Bug B (extend MCP startup race to lazy) + Bug C (QA cleanup orphan leak). PR #260 landed the lazy-race core on main with equal semantics plus startupTimeoutMs, so this PR was rebased and reduced to what main still needs:

- Bug C: process-group-aware QA cleanup (port from feat/app-server-parity; leader-only liveness let esbuild descendants escape SIGTERM and orphan to PID 1 — 95 live orphans observed)
- Rehydration replay fix: session-history rehydration now replays after a raced background connect (was lost against the not-yet-existing catalog)
- mcp_instructions block rebuilt after raced background registration
- test/mcp await-hardening: 21 files await raced background completion via fixture seams (fixes 74/284 test/mcp failures that plain main shows under parallel load)

## Measured evidence
- Orphans: 95 -> 0 after one-time reap; handshake.mjs probe on the fixed branch leaves count at 0 (f3-orphan-reap.txt, t5-cleanup-port.txt)
- Cross-fix live QA (F2): first footer render 2.12s (baseline >=12s blank), prompt at T+2s answered in 0.29s, no attach stall in ~/local-workspaces with codegraph enabled (f2-live-qa.txt)
- Full gate on the rebased branch: npm run check + npm run test exit 0, 477 files / 4059 tests (pr2-rework-gate.txt)

## Evidence
- t4-mcp-race.txt (original T4 proof: prompt->first-token 1.05s with a 10s-hanging lazy server; production half superseded by #260)
- t5-cleanup-port.txt, pr2-rework-gate.txt, f1-branch2-refix.txt

Plan: .omo/plans/senpi-render-hotpath-and-startup-fixes.md

### DIFF
--- a/packages/coding-agent/scripts/qa-app-server/lib/cleanup-fixture.mjs
+++ b/packages/coding-agent/scripts/qa-app-server/lib/cleanup-fixture.mjs
@@ -1,0 +1,30 @@
+#!/usr/bin/env node
+import { spawn } from "node:child_process";
+import { fileURLToPath } from "node:url";
+
+const mode = process.argv[2];
+
+if (mode === "leaf") {
+	process.on("SIGTERM", () => {});
+	if (process.send === undefined) throw new Error("cleanup fixture leaf requires an IPC channel");
+	process.send({ type: "ready" }, (error) => {
+		if (error) throw error;
+		Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0);
+	});
+} else if (mode === "leader") {
+	const fixturePath = fileURLToPath(import.meta.url);
+	const leaf = spawn(process.execPath, [fixturePath, "leaf"], {
+		stdio: ["ignore", "ignore", "inherit", "ipc"],
+	});
+	leaf.once("error", (error) => {
+		process.stderr.write(`${error.stack ?? error.message}\n`);
+		process.exit(1);
+	});
+	leaf.once("message", (message) => {
+		if (message?.type !== "ready") return;
+		process.stdout.write(`${JSON.stringify({ leaderPid: process.pid, leafPid: leaf.pid })}\n`);
+	});
+	process.once("SIGTERM", () => process.exit(0));
+} else {
+	throw new Error(`Unknown cleanup fixture mode: ${mode ?? "missing"}`);
+}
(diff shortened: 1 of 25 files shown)


code-yeongyu:
F1 gate failures (15 mcp tests) — root cause and fix

**Why `npx vitest run test/mcp` was green but the full gate failed:** extending the 250ms startup race to lazy lifecycles changed attach's contract. On main, uncached lazy servers connected inline inside the awaited `Promise.all(connects)`, so `await attach(...)` guaranteed tools were registered. On this branch, when the fixture server's boot exceeds 250ms (routine under full-suite fork parallelism, rare on a quiet machine), attach returns at the race deadline and tools are wired later by the `refreshMcpToolsAfterStartupRace` background continuation. The 15 tests asserted the tool set synchronously right after attach, so they saw `[]` only under load. Reproduced deterministically with a `--slow-start 600` fixture (identical empty-list failure).

**One real bug found behind the test failures:** `attachSession` replays tool_search activation markers from session history after its inline registration — but with a raced lazy attach the catalog does not exist yet, so the replay was a silent no-op and the background continuation never replayed. A resumed session would lose its restored tools on the first turn. Fixed in `service.ts` (f96e4aba1): the raced `registerDirectTools` callback now replays history from the stored session context after the background registration lands (idempotent).

**Fixes (race behavior NOT reverted):**
- `f96e4aba1` fix(coding-agent): replay MCP history rehydration after raced background registration
- `196da429a` test(coding-agent): await raced lazy attach completion in mcp suites — 17 mcp test files now await the background completion the way production consumers (next turn's tool snapshot) do, via new fixture seams (`awaitMcpToolRegistration`, `awaitMcpTool`, `awaitMcpConnected`, `attachHarnessSession` which emits the session_start the harness never fires). No test deleted or skipped; the same latent pattern in 8 more files (catalog-cache, commands, extension-load, idle, instructions, oauth-headless, proxy-mode, reconnect, skills-sidecar) was hardened too after re-runs surfaced them.
- `a8c8cd971` test(coding-agent): run qa cleanup suite under vitest — `cleanup.test.mjs` (from the QA-cleanup commit) used `node:test`, which vitest collects via its default `**/*.test.mjs` include but cannot execute ("No test suite found", deterministic full-gate failure). Converted to the vitest API so the process-group cleanup test actually runs.

**Gate:** `npx vitest run test/mcp` green (276 tests), `npm run check` exit 0, full `npm run test` exit 0 (coding-agent: 461 files / 3994 tests passed; previously 15 failed). Evidence: `.omo/evidence/senpi-render-hotpath-and-startup-fixes/f1-branch2-refix.txt`.

PR #303, opened 2026-07-23
feat(coding-agent): cap inline transport images at 24 MiB

### LINKED ISSUE #302: Accumulated image tool results push the request body over provider limits, permanently wedging the session with 413
## Summary

`read` keeps raw base64 image data in tool results, and the full history is replayed into every provider request. As images accumulate, the HTTP request body exceeds the provider's size limit and the provider rejects the request outright. Every subsequent input resends the same oversized history, so the session can never recover on its own.

Real incident session (v2026.7.17-5, kimi-coding/k3):

- 63 images, **101.36 MiB** of base64 in total (98.4% of the whole session JSONL)
- Requests succeeded up to 97.69 MiB of base64 → two more images pushed it to 101.36 MiB → first 413 immediately after
- Kimi's fronting nginx returned `413 Request Entity Too Large` with usage 0 (rejected before reaching the model)
- Every follow-up input afterwards ("continue", etc.) failed with the same 413 within ~0.5s — permanently

## Not a token or image-quality problem

- The last successful usage was **361,701 tokens = 34% of the 1M window** — nowhere near the token-compaction threshold. Vision models tokenize images as tiles (~hundreds to ~1,500 tokens each), so base64 bytes barely consume context. The session died purely from **HTTP body bytes**.
- A per-image cap already exists (4.5 MiB, `image-resize-core.ts`) and every image was within it (avg 1.68 MiB each). The cause is **accumulated count** — per-image defenses cannot stop it.

## Root causes

1. Image base64 accumulates in the session without bound and is replayed on every request (image blocks from `read.ts`, full-history resend, no pruning/budget)
2. Size management is **token-axis only** — compaction never looks at body bytes, and estimates every image at a flat ~1,200 tokens regardless of actual size (`compaction.ts` `ESTIMATED_IMAGE_CHARS`)
3. Resuming a session restores the entire active branch, so the oversized payload is unconditionally resent

Note: the Anthropic Messages API has a **documented 32 MB request limit** (413 `request_too_large`), so the same wedge is reproducible on **every provider**, not just Kimi (nginx, ~100 MiB observed). It can also fire when a provider fallback switches the session to a provider with a lower wall.

## Expected behavior

- No matter how many images accumulate, the inline image size of a request stays below provider limits.
- Resuming an already-wedged session recovers it.
- Session storage and conversation text are never lost.

## Reproduction

1. In a vision-model session, read many screenshots (an FHD screenshot ≈ 1.5–2 MiB of base64; ~20 of them exceed 32 MB on Anthropic)
2. The next request returns 413 → every subsequent input fails instantly with the same error

## Follow-up scope (out of scope here; candidates for separate issues)

- Classifying nginx HTML/plaintext 413 as a request-size error (`packages/ai` `overflow.ts`)
- Native image transmission on the speculative compaction / `/btw` / emergency paths
- Ingestion cap for `read_video` (up to 100 MB) single blocks

Suggested labels: `bug`, `coding-agent`, `image-context`, `request-size`, `http-413`, `session-recovery`

amsminn:
Fixes #302

## Problem

Image base64 accumulates in the session and the full history is replayed into every request. Once the HTTP body exceeds a provider's size limit, the request is rejected with 413 and the session wedges permanently (real incident: 63 images = 101 MiB → endless Kimi nginx 413s). Token-axis compaction never intervenes because it doesn't look at body bytes (last usage was only 34% of the window). Details: #302.

## Change

**Cap the total inline image base64 at 24 MiB in the transport conversion layer.** Anything over the budget is replaced, oldest first, with a re-read hint placeholder. Stateless and transport-only — persisted session storage is untouched.

- `packages/coding-agent/src/core/messages.ts` — `TRANSPORT_IMAGE_BUDGET_BYTES = 24 MiB`, `IMAGE_ELISION_PLACEHOLDER`, `elideOldImages()` (newest→oldest hard recency cutoff; the newest block is always kept regardless of size but consumes the budget, so an oversized newest block elides everything older → the total stays bounded; returns the **same reference** when nothing is elided), `convertToLlmForTransport()` (moves the existing sdk.ts blockImages closure verbatim + chains elision)
- `packages/coding-agent/src/core/sdk.ts` — closure replaced with a delegation to the function above (constant budget — no mutable state)
- `packages/coding-agent/test/suite/harness.ts` — the harness Agent uses the same conversion, with a small-budget injection option (`transportImageBudget`) for tests
- `packages/coding-agent/src/core/changes.md` — fork-change record

**Untouched**: `packages/ai`, `agent-session.ts`, compaction, session storage format.

## Why 24 MiB

The only **documented** provider wall is the Anthropic Messages API's 32 MB request limit (413 `request_too_large`). 24 MiB of images + up to ~4 MiB of text (1M tokens) + overhead ≈ 29 MB < 32 MB. Since the cap sits below the smallest known wall, **a single budget is safe on every provider, including mid-session provider fallback**. (Kimi's ~100 MiB is an unverified observation and was excluded as a design input.) At the average FHD screenshot size (~1.7 MiB) the cap retains the most recent **12–15 screenshots** — plenty for computer-use continuity. Previously wedged sessions pass from the very first request after reopening.

## Why there is no retry/recovery layer

- The cap alone means 413 never fires against any known wall (zero failed round trips).
- Residual 413s (undocumented lower walls, or a kept over-budget newest video being rejected) are handled by existing machinery: 413 is hard-error-fallback eligible, so the model-fallback chain switches providers, and video blocks are downgraded to placeholders on non-video models. With history bounded at 24 MiB, the old "resend 101 MiB forever" wedge is structurally impossible.
- `elideOldImages` takes budget/keep as options, so a recovery layer can be added in a follow-up PR if it ever proves necessary.

## Trade-offs (deliberate)

- On Kimi, sessions in the 24–100 MiB image range get old images elided even though they wouldn't hit the wall — accepted in exchange for a documented anchor, fallback safety, and reduced per-turn upload. Re-read recovers any elided image.
- Sessions above 24 MiB of images invalidate part of the prefix cache as the elision boundary moves with new images. Sessions under 24 MiB send byte-identical requests (same-reference no-op).

## Tests / QA

- 15 unit tests (`test/image-elision.test.ts`): same-reference no-op, cutoff boundaries, keep-consumes-budget, sibling text preserved, consecutive placeholder dedupe, later-is-newer, blockImages parity
- 2 integration tests (`test/transport-image-budget.test.ts`): with a faux provider, the very first request is elided / under-budget context untouched
- Regressions: `block-images.test.ts`, `compaction.test.ts` pass
- Root `npm run check` green
- Real CLI QA done (evidence: `local-ignore/qa-evidence/20260723-transport-image-budget/`)

## Out of scope (follow-ups)

413 request-size classification (`overflow.ts` — upstream contribution candidate) · speculative/`/btw`/emergency paths · `read_video` ingestion cap · budget configurability

### DIFF
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -1,5 +1,27 @@
 # changes
 
+## Provider-bound inline image budget (2026-07-22)
+
+### What changed
+
+- `messages.ts`: added a transport-only 24 MiB inline image budget. Provider-bound conversion keeps the newest image
+  block, counts it against the budget, and replaces images older than the hard recency cutoff with a re-read
+  placeholder while preserving all text and leaving the persisted session untouched.
+- `sdk.ts`: routes the main agent loop through the shared transport conversion while preserving the dynamic
+  `images.blockImages` kill switch and its existing placeholder/deduplication behavior.
+- `test/suite/harness.ts`: uses the same transport conversion and accepts a small injectable image budget for
+  deterministic first-request integration coverage.
+
+### Why extension system couldn't handle this alone
+
+- Inline images must be bounded after session messages are converted but before every main-loop provider request,
+  including resumed sessions and provider fallbacks. That conversion boundary is owned by the core Agent wiring.
+
+### Expected merge conflict zones
+
+- MEDIUM: `sdk.ts` around the Agent `convertToLlm` wiring.
+- LOW: the transport helpers at the end of `messages.ts` and the Agent construction in `test/suite/harness.ts`.
+
 ## Streaming steer/followUp submissions bypass the session-work barrier (2026-07-21)
 
 ### What changed
--- a/packages/coding-agent/src/core/messages.ts
+++ b/packages/coding-agent/src/core/messages.ts
@@ -207,3 +207,132 @@ export function convertToLlm(messages: AgentMessage[]): Message[] {
 		})
 		.filter((m) => m !== undefined);
 }
+
+// ============================================================================
+// Transport image budget
+// ============================================================================
+
+/**
+ * Cap on inline image base64 (characters approximately equal wire bytes) per
+ * provider request. Anthropic documents a 32 MB request limit; reserving the
+ * remainder for text and request overhead keeps the transport below that wall.
+ */
+export const TRANSPORT_IMAGE_BUDGET_BYTES = 24 * 1024 * 1024;
+
+export const IMAGE_ELISION_PLACEHOLDER =
+	"[Image elided: an older image was removed to keep the request within the provider's size limit. Re-read the source file if you need to view it again.]";
+
+export const BLOCKED_IMAGE_PLACEHOLDER = "Image reading is disabled.";
+
+export interface ElideOldImagesOptions {
+	/** Cumulative inline image base64 budget. Defaults to TRANSPORT_IMAGE_BUDGET_BYTES. */
+	budgetBytes?: number;
+	/** Newest image blocks always kept regardless of budget. They still consume it. Defaults to 1. */
+	alwaysKeepNewest?: number;
+}
+
+export interface TransportConvertOptions extends ElideOldImagesOptions {
+	/** Replace every image when the images.blockImages setting is enabled. */
+	blockImages: boolean;
+}
+
+/** Drop consecutive duplicates of a placeholder produced by adjacent image replacements. */
+export function dedupeConsecutivePlaceholder(
+	content: (TextContent | ImageContent)[],
+	placeholder: string,
+): (TextContent | ImageContent)[] {
+	return content.filter(
+		(block, index, blocks) =>
+			!(
+				block.type === "text" &&
+				block.text === placeholder &&
+				index > 0 &&
+				blocks[index - 1].type === "text" &&
+				(blocks[index - 1] as TextContent).text === placeholder
+			),
+	);
+}
+
+/**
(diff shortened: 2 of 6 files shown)


code-yeongyu:
[sisyphus-bot] Closing this because **your fix has landed** — with credit to you.

#364 ("fix(coding-agent): cap inline transport images", merged just now) adopts your design essentially as you wrote it: `TRANSPORT_IMAGE_BUDGET_BYTES = 24 MiB`, `elideOldImages()` + `convertToLlmForTransport()` in `packages/coding-agent/src/core/messages.ts`, `sdk.ts`'s blockImages closure delegating into it, and the injectable `transportImageBudget` harness option — stateless, transport-only, persisted session untouched. The commit carries `Co-authored-by: amsminn <amsminn@users.noreply.github.com>`.

Why it landed as a new PR rather than as this one: this branch was `CONFLICTING` against main (`packages/coding-agent/src/core/changes.md` and `packages/coding-agent/test/suite/harness.ts`), and its CI had never run because fork workflows need maintainer approval. Rather than sit on it further, I had the conflicts re-ported onto current main and added edge coverage on top of your design (single-image-over-budget kept, oldest-first cutoff, session non-mutation asserted, mock-loop integration).

Your 24 MiB anchor against Anthropic's documented 32 MB request wall was the right call and I kept it verbatim. Issue #302 is closed by the merge.

Thanks — this was the highest-value unmerged fix in the queue.

PR #310, opened 2026-07-23
fix(coding-agent): stabilize compaction lifecycle

minpeter:
## Summary

- move compaction completion tracking into an `AgentSession`-owned monotonic lifecycle
- snapshot the active fallback model and bind durable apply/abort/feedback to the current operation
- retain terminal compaction state while rejecting stale or duplicate progress/completion
- keep provider-confirmed overflow recovery fail-closed, including repeated prompts after rejected compaction

## Problem

Compaction could observe construction-time or stale model/controller state after main-thread fallback selected another
model. Completion state disappeared when controllers were released, and delayed feedback from a superseded operation
could affect the active operation. Provider-confirmed overflow also needed to remain authoritative when local token
estimation undercounted the context.

## Implementation

- add a pure `idle` / `running` / `completed` / `failed` / `aborted` lifecycle reducer with monotonic generations
- snapshot the current `AgentSession` model at compaction operation start
- thread operation signals through builtin compaction feedback and ignore stale signals
- guard durable compaction append with operation/controller identity
- expose retained read-only compaction state for session observers
- rebuild live context from the canonical `SessionManager` after failed overflow recovery so later prompts cannot bypass
  the same required compaction

## Tests

- `npm test -- test/compaction test/suite/*compaction*.test.ts test/suite/regressions/*compaction*.test.ts test/suite/regressions/pre-prompt-compaction-no-continue.test.ts`
  - 42 files, 299 tests passed
- `npm run check`
  - passed
- `git diff --check`
  - passed

Added coverage for:

- fallback-selected current-model compaction and the next successful prompt
- lifecycle terminal retention, generation increments, and stale/duplicate events
- stale progress and terminal feedback signals
- precomputed apply state and model-switch invalidation
- rejected provider-overflow compaction below the local threshold
- repeated prompts after rejected required compaction, with zero provider calls

## Manual QA

- bug-specific fallback driver:
  - fallback switched `faux-1` to `faux-2`
  - compaction ran on `faux-2`
  - lifecycle retained `completed`
  - the next prompt succeeded
- repeated-overflow driver:
  - 2 prompts blocked
  - 0 provider calls
  - 2 rejected compactions
  - overflow assistant restored
- real source CLI mock-loop: 38/38 checks passed across all three wire formats
- CLI smoke: 7/7 checks passed
- RPC smoke: 4/4 checks passed

QA receipts are stored locally under:

- `local-ignore/qa-evidence/20260723-compaction-completion-fsm-final-turn/`
[...]

PR #326, opened 2026-07-24
fix(coding-agent): expose failed compaction attempts

minpeter:
## Summary

- publish `compaction_start` once an automatic compaction owns its model/controller
- do so before authentication/preparation can fail
- preserve the one-overflow-retry attempt when preparation fails before a provider retry

This is a one-commit follow-up to merged PR #310, based directly on its merge result.

## Root cause

The final #310 controller-supersession fix moved `compaction_start` after authentication and request preparation. An
overflow compaction that failed during those early preparation steps therefore consumed/released internal attempt
state without publishing the attempt. The next prompt could not prove that the overflow recovery had been retried,
and the established no-auth regression observed zero starts instead of two independent attempts.

The event now emits after model/controller ownership is proven but before fallible authentication/preparation. A
superseded controller still cannot emit; a valid owned attempt remains observable even when preparation returns early.

## RED -> GREEN

- RED on exact merged `main`:
  - `does not consume the overflow compact-and-retry attempt when compaction fails before retrying`
  - expected two `compaction_start` events, received zero
- GREEN:
  - exact target 1/1
  - agent-session compaction + controller lifecycle group 42/42
  - full compaction/retry/hook matrix 14 files / 121 tests

## Verification

- root build: passed
- root `npm run check`: passed
- real source CLI mock loop: exit 0, one localhost request, expected assistant text, auth unchanged
- built CLI `--help`: exit 0
- built CLI invalid `install` input: exit 1 with usage

The broader coding-agent full suite has unrelated load-sensitive MCP/app-server/temporary-filesystem failures which
pass directly; no compaction test remains failing.

---
## Summary by cubic
Emit `compaction_start` as soon as an automatic compaction owns its model/controller, before auth and preparation. This makes early preparation failures visible and preserves the single overflow retry attempt.

<sup>Written for commit 62354c272c81623155829e573dbf0294ed047d15. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/326?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -4232,6 +4232,8 @@ export class AgentSession {
 				if (reason === "overflow") this._overflowRecoveryAttempted = false;
 				return false;
 			}
+			if (!this._ownsCompactionController(autoCompactionController, "auto")) return false;
+			this._emit({ type: "compaction_start", reason });
 
 			const authResult = await this._modelRuntime.getAuth(this.model);
 			if (!this._ownsCompactionController(autoCompactionController, "auto")) return false;
@@ -4250,7 +4252,6 @@ export class AgentSession {
 				return false;
 			}
 			if (!this._ownsCompactionController(autoCompactionController, "auto")) return false;
-			this._emit({ type: "compaction_start", reason });
 
 			const execution = await this._executeCompaction({
 				controller: autoCompactionController,
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -8,6 +8,8 @@
   controller at operation start, rejects stale completion/feedback, and retains the terminal result until another
   operation begins. Feedback-only aborts publish one terminal event, and accepted completions publish their terminal
   event before `session_compact` handlers can begin a fresh operation.
+- Owned automatic compaction attempts publish `compaction_start` before authentication/preparation so an early
+  preparation failure does not invisibly consume the one overflow retry attempt.
 - Durable append now rejects a generation whose message revision or agent-message snapshot changed during preparation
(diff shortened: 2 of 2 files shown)


code-yeongyu:
[sisyphus-bot]

Thanks for spotting this — closing as **already fixed on main**, not as a rejection. Your diagnosis was correct; main just landed an equivalent fix through a different restructure while this draft sat.

Evidence on current main (`packages/coding-agent/src/core/agent-session.ts`, `_runAutoCompaction`):

```ts
const endBeforeExecution = (): false => {
    this._emit({ type: "compaction_start", reason });
    if (reason === "overflow") this._overflowRecoveryAttempted = false;
    this._emit({ type: "compaction_end", reason, result: undefined, aborted: false, willRetry: false });
    return false;
};
```

`endBeforeExecution()` is now called on **every** early-exit path before execution — missing model (~4300), auth failure (~4311), and `prepareCompaction` returning nothing (~4321). So a preflight failure now publishes a visible `compaction_start`/`compaction_end` pair **and** resets `_overflowRecoveryAttempted`, which is exactly the outcome this PR was after: an early failure no longer invisibly consumes the single overflow retry.

Your branch is also based on pre-merge `agent-session.ts` (`590d0bcbb`) and its CI failure comes from that stale base, not from your change.

Related and still wanted: **#341** builds on this same `endBeforeExecution` helper with the ownership guards main still lacks — that one is green and clean, and I'd love to see it out of draft. Thanks for the careful work here.

PR #343, opened 2026-07-25
fix(publish): keep platform-constrained natives out of bundleDependencies

### LINKED ISSUE #342: npm install fails with EBADPLATFORM on all non-linux-x64 platforms since 2026.7.23
`npm install @code-yeongyu/senpi` fails on every non-linux-x64 platform since 2026.7.23. The published manifest lists linux-only native binaries as **required** `dependencies`, so npm aborts before writing a single file.

## Repro

```
$ mkdir /tmp/t && cd /tmp/t && npm init -y
$ npm install @code-yeongyu/senpi@2026.7.24
npm error code EBADPLATFORM
npm error notsup Unsupported platform for @mariozechner/clipboard-linux-x64-gnu@0.3.9:
  wanted {"os":"linux","cpu":"x64","libc":"glibc"} (current: {"os":"darwin","cpu":"arm64"})
```

macOS 15.7.3 arm64, node 25.4.0, npm 11.7.0. Same failure with `--dry-run`.

**Expected:** install succeeds. Those binaries are `optionalDependencies` of `@mariozechner/clipboard`, so npm should skip the ones that don't match the host.

## Cause

The registry manifest declares the whole bundled set as required deps, while the tarball's own `package.json` declares 39 and neither binary:

```
$ curl -s https://registry.npmjs.org/@code-yeongyu%2Fsenpi \
    | jq '.versions["2026.7.24"] | {deps:(.dependencies|length), bundled:(.bundleDependencies|length)}'
{ "deps": 258, "bundled": 258 }
```

`stagePublishManifest` (`scripts/prepare-senpi-bundled-workspaces.mjs`) sets `bundleDependencies` to every package physically present in the staged `node_modules`. `publish-npm.yml` runs on `ubuntu-latest`, so the runner's own natives get bundled, and npm republishes the bundled set as required `dependencies`.

This is not linux-specific — running the same staging on my mac bundles `clipboard-darwin-arm64`/`-universal` instead, which would break linux and Windows identically. A package declaring `os`/`cpu`/`libc` can never be bundled correctly, because the bundle is one artifact shipped to every platform.

First bad version is 2026.7.23 (2026.7.22 has no wildcard deps in its manifest, 2026.7.23 has 219).

Existing installs are unaffected — this only bites fresh `npm install`, which is also why the published tarball currently ships 1.3 MB of unusable linux binaries to mac users.

## Fix

I'd like to send a PR: skip packages declaring `os`/`cpu`/`libc` when building the bundle list, so they stay ordinary optional registry deps resolved per install target.

realsigridjin:
Fixes #342.

`npm install @code-yeongyu/senpi` has been broken on every non-linux-x64 platform since 2026.7.23.

## What happens

`stagePublishManifest` sets `bundleDependencies` to every package physically present in the staged `packages/coding-agent/node_modules`. npm republishes that bundled set as required `dependencies` in the registry manifest, so the publish runner's own native binaries become hard requirements of the published package. `publish-npm.yml` runs on `ubuntu-latest`, so 2026.7.23 and 2026.7.24 shipped `@mariozechner/clipboard-linux-x64-gnu` and `-musl` as required deps:

```
$ curl -s https://registry.npmjs.org/@code-yeongyu%2Fsenpi \
    | jq '.versions["2026.7.24"] | {deps:(.dependencies|length), bundled:(.bundleDependencies|length)}'
{ "deps": 258, "bundled": 258 }     # the tarball's own package.json declares 39
```

npm then aborts before writing a file:

```
npm error code EBADPLATFORM
npm error notsup Unsupported platform for @mariozechner/clipboard-linux-x64-gnu@0.3.9:
  wanted {"os":"linux","cpu":"x64","libc":"glibc"} (current: {"os":"darwin","cpu":"arm64"})
```

Not a linux quirk — staging on my mac bundles `clipboard-darwin-arm64`/`-universal` instead, which would break linux and Windows the same way.

## The fix

A package that declares `os`/`cpu`/`libc` can never be bundled correctly: npm resolves those fields against the *installing* machine, but the bundle is a single artifact shipped to every platform. So `bundleDependencies` now skips them, and they stay ordinary optional registry deps — npm resolves the binary matching the install target, and a failed fetch degrades clipboard support instead of failing the install.

The filter reads each staged package's own manifest, so it isn't tied to the clipboard dependency; any future native falls out of the bundle automatically.

**What is deliberately unchanged:**

- Direct-dependency validation still runs against the full physically staged set, so nothing weakens there.
- Every portable package stays bundled, so the self-containment that fixed the old arborist `ERR_MODULE_NOT_FOUND` reify problem is intact — 256 of the 258 staged packages still ship in the tarball.
- `@mariozechner/clipboard` itself has no platform fields, so it remains bundled and its `optionalDependencies` edge is untouched.

If a *direct* dependency ever declares platform fields, `validatePack`'s existing "missing vendored runtime dependencies" gate fails the publish loudly rather than shipping a broken tarball. That seemed better than silently widening the exemption.

`scripts/prepare-senpi-bundled-workspaces.mjs` has no upstream counterpart in `badlogic/pi-mono`, so there is no `changes.md` section to add.

## Verification

- **The new `stagePublishManifest` test is a real regression test.** With the one-line filter reverted it fails with exactly the two packages that broke installs:
  ```
  actual:   [..., '@mariozechner/clipboard-linux-x64-gnu', '@mariozechner/clipboard-linux-x64-musl', ...]
  expected: ['@earendil-works/pi-ai', '@mariozechner/clipboard', 'cross-spawn', 'which']
  ```
- **End-to-end on a real checkout** (`copyPublishDependencies` + `stagePublishManifest` against this repo's installed tree, workspace `dist/` stubbed since the filter never reads build output):
  ```
  physically staged: 258 | declared in bundleDependencies: 256
  excluded from the bundle: ["@mariozechner/clipboard-darwin-arm64","@mariozechner/clipboard-darwin-universal"]
  optionalDependencies preserved: {"@mariozechner/clipboard":"0.3.9"}
  ```
  258 is the same count as the broken published manifest, which is what pins the mechanism.
- `npm run test:scripts` — 61 tests, 60 pass, 1 skipped, 0 fail.
- `npm run check` — passes (it also runs as the pre-commit hook for this commit).

  Unrelated flake worth flagging: on my first full run, `check:neo`'s `TestAttachOrSpawn_RaceExactlyOneDaemon` timed out at its 30s budget while `tsgo` was still saturating the machine (`bridge: attach-or-spawn timed out after 30s`). Run on its own it passes in ~0.2-0.3s both with and without this change, and it passed on the re-run, so it looks load-sensitive rather than broken.
- `npm run check:shrinkwrap` / `check:install-lock:coding-agent` / `check:pinned-deps` — all report up to date; the change alters only which staged packages are *declared* bundled, not the dependency graph, so no lockfile regeneration is needed.

I have not run `./pi-test.sh`; this touches no provider code.

Drafted with Claude Code, reviewed and verified by me — every command output above is verbatim from my machine, not inferred.

---
## Summary by cubic
Exclude platform-constrained native packages from `bundleDependencies` during publish to stop `npm install @code-yeongyu/senpi` from failing on non‑linux‑x64. Portable packages stay bundled; natives remain optional and resolve per install target. Fixes #342.

- **Bug Fixes**
  - Added `isPlatformConstrainedPackage` and `bundlablePublishPackageNames` to skip packages that declare `os`/`cpu`/`libc` when building `bundleDependencies`.
  - Updated `stagePublishManifest` to write the filtered list to `bundleDependencies`/`bundledDependencies` and return it.
  - Kept full staged-set validation; portable deps still bundle for a self-contained tarball. `@mariozechner/clipboard` remains optional and bundled as the meta package.
  - Updated `scripts/AGENTS.md` and added regression tests to cover the new filter and manifest output.

<sup>Written for commit 8b5685a51b46e4e4f3629566676be00e0f69d25e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/AGENTS.md
+++ b/scripts/AGENTS.md
@@ -57,12 +57,20 @@ Manages workspace packages embedded in the published `@code-yeongyu/senpi` tarba
 The publish tarball is fully self-contained: `copyPublishDependencies` stages the ENTIRE
 runtime closure from `publish-deps.lock.json` (all registry deps + transitives, not just the
 workspace closure) into `packages/coding-agent/node_modules`, and `stagePublishManifest`
-rewrites the publish manifest so `bundleDependencies` lists every staged package while all
-`dependencies` edges (including the registry-absent `^2026.x` workspace specs) stay intact.
-npm then needs no registry fetch at install time; the previous partial bundle let arborist
-abort reify mid-flight and drop arbitrary registry deps (ERR_MODULE_NOT_FOUND).
+rewrites the publish manifest so `bundleDependencies` lists every platform-portable staged
+package while all `dependencies` edges (including the registry-absent `^2026.x` workspace
+specs) stay intact. npm then needs no registry fetch at install time; the previous partial
+bundle let arborist abort reify mid-flight and drop arbitrary registry deps
+(ERR_MODULE_NOT_FOUND).
 Staging dirties `packages/coding-agent/package.json`; restore it with `git checkout --`
 after packing/publishing.
+
+`bundleDependencies` deliberately excludes packages that declare `os`/`cpu`/`libc`
+(`isPlatformConstrainedPackage`). npm resolves those fields against the installing machine, but
+the bundle is one artifact shipped to every platform and npm republishes the bundled set as
+required `dependencies` in the registry manifest — so bundling the publish runner's own natives
+(linux-x64, per `publish-npm.yml`) made `npm install @code-yeongyu/senpi` fail with
+EBADPLATFORM everywhere else. They remain optional registry deps, resolved per install target.
 
 ## check-mcp-docs.test.mjs
 
--- a/scripts/prepare-senpi-bundled-workspaces.mjs
+++ b/scripts/prepare-senpi-bundled-workspaces.mjs
@@ -111,10 +111,42 @@ export function listStagedPublishPackageNames(codingAgentNodeModules) {
 	return packageNames.sort((a, b) => a.localeCompare(b));
 }
 
+// npm evaluates `os`/`cpu`/`libc` against the INSTALLING machine, but bundleDependencies ships a
+// single tarball to every platform and npm republishes the bundled set as required `dependencies`
+// in the registry manifest. Bundling a native binary that only matches the publish runner
+// (linux-x64 in publish-npm.yml) therefore aborts `npm install @code-yeongyu/senpi` with
+// EBADPLATFORM on every other platform, before a single file is written. Such packages stay
+// ordinary optional registry deps: npm resolves the binary matching the installing platform, and
+// a failed fetch degrades the feature instead of failing the install.
+export function isPlatformConstrainedPackage(packageDir) {
+	let manifest;
+	try {
+		manifest = JSON.parse(readFileSync(join(packageDir, "package.json"), "utf8"));
+	} catch {
+		// An unreadable staged package cannot be judged here; leave it bundled so the existing
+		// staging and pack gates report it instead of this filter dropping it silently.
+		return false;
+	}
+	return ["os", "cpu", "libc"].some((field) => {
+		// npm accepts a bare string as well as the documented array form.
+		const constraint = manifest[field];
+		if (typeof constraint === "string") {
+			return constraint.length > 0;
+		}
+		return Array.isArray(constraint) && constraint.length > 0;
+	});
+}
+
+export function bundlablePublishPackageNames(codingAgentNodeModules, packageNames) {
+	return packageNames.filter((name) => !isPlatformConstrainedPackage(join(codingAgentNodeModules, name)));
+}
+
 // The publish tarball must be fully self-contained: every runtime dependency edge in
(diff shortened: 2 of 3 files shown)


code-yeongyu (approved):
Verified locally: regression test fails with the filter reverted (exactly the two linux clipboard natives leak back into bundleDependencies), full test:scripts passes (60 pass / 1 skip), and registry data confirms 2026.7.23+ ships 258 required deps incl. linux-only binaries vs 39 in 2026.7.22. The fix is minimal, keeps the self-contained bundle intact for portable packages, and platform-constrained natives correctly fall back to optional registry resolution. LGTM.

PR #351, opened 2026-07-25
fix(ai): make thinking off actually off and effort mapping wire-exact across adapters

code-yeongyu:
## Problem

Turning thinking **off** did not turn reasoning off, and several effort levels degraded to a weaker wire value than requested.

Before any code change I probed the live Anthropic Messages endpoint (7 families, `max_tokens: 16`) so the fix would follow the wire rather than assumptions:

| probe | result |
|---|---|
| `thinking:{type:"disabled"}` on opus-4-6 / 4-7 / 4-8 / 5, sonnet-4-6 / 5 | **200** — true disable works |
| `thinking:{type:"disabled"}` on `claude-fable-5` | **400** `"thinking.type.disabled" is not supported for this model. Thinking defaults to adaptive mode when not specified` |
| no `thinking` + `output_config:{effort:"low"}` on fable-5 / opus-5 | **200** (with and without an effort beta header) |
| `thinking:{type:"adaptive",display:"summarized"}` on opus-4-6 | **200** |

That 400 message is the bug: senpi omitted the thinking field for Fable 5, the API then defaulted to adaptive, and an "off" turn silently billed full reasoning — while `off` was not even offered in the level list.

## What changed

**Anthropic Messages** — families that accept `disabled` keep sending it; families that cannot now send **no** thinking block plus `output_config:{effort:"low"}`. Adaptive markers gained `opus-4-8`/`opus-5`/`sonnet-5`/`fable-5` so models without the `forceAdaptiveThinking` pin (custom `models.json`, third-party gateways) get effort control instead of a budget-token request. The ladder floors extended levels at the adaptive top tier (`xhigh` → native xhigh where present, else `max`; `max` → `max`) instead of silently returning `high`.

**Bedrock Converse** — a thinking-off turn returned *no* thinking config at all, so every adaptive Claude family fell back to the adaptive default. It now sends `disabled`, or the effort-low pin for families that reject it; budget-based Claude still sends nothing. Same ladder floor fix.

**Catalog** — Fable 5 on `anthropic-messages` is encoded as `compat.supportsDisabledThinking: false` instead of `thinkingLevelMap.off: null`. Both mean "never send `disabled`", but the compat form keeps `off` a **selectable** level. Regenerated via `npm run hydrate-model-data`; the semantic diff is 8 fable-5 field changes across 4 providers plus one incidental OpenRouter price refresh.

**Tier detection** — `models.ts` and `coding-agent/thinking-levels.ts` now recognize `gpt-5.6`, `opus-5`, `sonnet-5`, `fable-5` for the `xhigh`/`max` tiers (fallback path for models with no map).

**openai-completions** — added the missing no-map ladders (Kimi K3 `low/high/max`, DeepSeek & GLM 5.2 `high/max`, OpenRouter DeepSeek `high`-only, MiMo `minimal→low` / `xhigh→high`, Ollama `low/medium/high/max`); an explicit catalog `null` now suppresses the wire effort instead of forwarding the raw level.

**Responses family** — explicit `max: "max"` preserved for GPT-5.6; explicit `null` wins for direct adapter options incl. summary-default resolution; Codex sends its catalog off sentinel when off arrives as omitted reasoning.

**Google** — a runtime off request fell through to an enabled form; worst case `off` became `thinkingBudget: 24576` + `includeThoughts: true` on Gemini 2.5 Flash because the clamp coerced `"off"` to `"high"`. Both adapters now take the disabled form. Mistral audited, no change needed (off provably cannot reach its `?? "high"` fallback).

## Deliberately NOT done (evidence-driven scope cuts)

- **No `effort-2025-11-24` beta header.** `output_config.effort` is accepted with and without it (200/200). `anthropic-messages` is shared by anthropic-compatible third parties (zai, kimi-coding, deepseek, minimax); adding an unrequested beta risks `400 Extra inputs are not permitted` for zero benefit.
- **No `display` gating for Opus/Sonnet 4.6.** The claim that 4.6 rejects `thinking.display` did not reproduce (200 with `display:"summarized"`); gating it would have removed summarized reasoning that streams fine today.
- Out of scope: `task_budget`, `speed`/fast-mode, MiniMax adaptive tag, catalog-architecture port.

## Verification

**Real-CLI wire QA** (zero tokens): real CLI from source → sandboxed HOME → localhost mock Anthropic server recording request bodies. All four runs exited 0 and printed `pong`.

| CLI model spec | wire `thinking` | wire `output_config` |
|---|---|---|
| `claude-fable-5:off` | *(absent)* | `{"effort":"low"}` ← was absent, so adaptive billed |
| `claude-fable-5:xhigh` | `{"type":"adaptive","display":"summarized"}` | `{"effort":"xhigh"}` |
| `claude-opus-4-8:off` | `{"type":"disabled"}` | *(none)* |
| `claude-opus-4-8:xhigh` | `{"type":"adaptive","display":"summarized"}` | `{"effort":"xhigh"}` |

The `fable-5:off` row also proves `off` became selectable — previously the requested level clamped to a thinking-on tier.

- `node .agents/skills/senpi-qa/scripts/mock-loop.mjs --self-test --api anthropic-messages` → **20/20**, including "zero real provider calls" and "real auth unchanged".
- `npm run check` → **exit 0** (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, `tsgo --noEmit`, browser smoke, web-ui, neo Go build+vet+test).
- `packages/ai` full suite → **exit 0**, 1354 passed / 155 files.
- `packages/coding-agent` full suite → 4667 passed; 3 load-sensitive files (`config-reload-watch-engine`, `app-server-daemon`, `app-server-stdio`) failed only while the static check and the other suite ran concurrently, and **pass 20/20 in isolation** — unrelated to this change (no fs.watch or app-server code touched).
- Every behavior change was captured RED first (fable off pin, map-less ladders, tier detection, Bedrock off path) and each new test fails against the old code.

TDD throughout; no test deleted, skipped, or loosened; no generated file hand-edited.

---
## Summary by cubic
Makes “thinking off” truly disable paid reasoning across adapters and maps effort levels to each provider’s exact wire format. Adds code guards for map‑less models and closes gaps for Bedrock Opus 5 and Mythos 5.

- **Bug Fixes**
  - `anthropic-messages`: send `thinking:{type:"disabled"}` where supported; for `claude-fable-5` and `claude-mythos-5` send no thinking and `output_config:{effort:"low"}`. Adaptive markers now include `opus-4-8`, `opus-5`, `sonnet-5`, `fable-5`, `mythos-5`; native `xhigh` set expanded. New in‑code family guard prevents sending `disabled` for Fable/Mythos when catalog metadata is absent; ladder floors fixed to honor `xhigh`/`max`.
  - `bedrock-converse`: thinking‑off now sends `disabled`, or the effort‑low pin for families that reject it; budget‑based Claude unchanged. Added `opus-5` and `mythos-5` to adaptive and native‑`xhigh` sets; same ladder floor fix.
  - `google-generative-ai` and `google-vertex`: thinking‑off now uses the disabled wire form, avoiding enabled/budgeted reasoning.
  - `openai-responses`, `azure-openai-responses`, `openai-codex-responses`: preserve explicit `max` (e.g., GPT‑5.6); catalog `null` suppresses wire effort; summary‑default respected; Codex emits its off sentinel when reasoning is omitted.
  - `openai-completions`: added missing no‑map ladders (Kimi K3 `low/high/max`, DeepSeek & GLM 5.2 `high/max`, OpenRouter DeepSeek `high` only, MiMo `minimal→low`/`xhigh→high`, Ollama `low/medium/high/max`); catalog `null` suppresses wire effort.

- **Catalog & Tier Detection**
  - Regenerated models: `claude-fable-5` on `anthropic-messages` uses `compat.supportsDisabledThinking: false` (keeps `off` selectable without sending `disabled`). Code keeps this family fact too, so custom/gateway rows without metadata behave correctly.
  - Tier predicates now recognize `gpt-5.6`, `opus-5`, `sonnet-5`, `fable-5` for `xhigh`/`max`, improving map‑less gateway parity.

<sup>Written for commit c35bd6973cb2c29fa0b51e6e4204548bb4e10a3f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
### Review rounds

Three independent read-only reviews ran against this branch. Two rounds found real, reachable holes that the tests had missed; both are fixed with failing-first coverage.

**Round 1 → 2 blockers accepted, 1 declined**
- *Custom/mapless Fable could still receive the rejected `thinking.type:"disabled"`.* A `models.json` entry or third-party gateway row carries no generated compat and no `thinkingLevelMap`, so it took the `disabled` branch and would hit the probe-confirmed 400. Fixed in `8d0173e99`: the "cannot disable" fact now lives in code (`DISABLED_THINKING_REJECTING_MODEL_MARKERS` + `cannotDisableThinking()`, and `rejectsDisabledThinking()` on Bedrock) *alongside* the catalog encoding instead of depending on it.
- *Bedrock Claude Opus 5 was misclassified.* `supportsAdaptiveThinking` / `supportsNativeXhighEffort` omitted `opus-5`, so it sent budget-token thinking instead of adaptive effort, could not emit native `xhigh`, and sent **no** thinking config for an off turn — falling straight back to adaptive. Fixed in `8d0173e99`.
- *Declined:* "Fable `off` maps to `effort:"low"`, so it is not truly zero reasoning — `off` should stay unavailable instead." The API exposes no true off switch for this family. With `off` unselectable, `clampThinkingLevel` picks the lowest selectable tier, whose effort mapping is **also** `low` — the wire request is identical, only labelled `minimal`. Removing `off` would keep the same cost while hiding the control, and the pre-change behavior (no thinking field at all → full adaptive) was strictly worse. Documented under "Known limitation (deliberate)" in `packages/ai/src/changes.md`.

**Round 2 → 1 blocker accepted**
- *Mythos 5 could still fall through.* `mythos-5` was listed as a family that rejects `disabled` but was missing from the adaptive marker lists, so a custom Mythos model at thinking-off emitted **neither** a thinking field **nor** an effort pin — exactly the silent-adaptive fallback this PR exists to remove. Fixed in `c35bd6973`, one failing-first test per adapter.

**Round 3 → APPROVE** (verified the marker lists are now element-for-element consistent between "rejects disabled" and "is adaptive", that no bundled catalog row is affected by the `mythos-5` addition, and that the new tests are real regression coverage).

### Out of scope (follow-up, not fixed here)

The round-2 review also noticed that `supportsPromptCaching()` in `packages/ai/src/api/bedrock-converse-stream.ts:692,712-713` documents "Claude 5 models" but enumerates only Fable 5 and Sonnet 5, so **Bedrock Claude Opus 5 misses prompt caching**. That is a caching bug, not a thinking/effort bug, so it is deliberately not bundled into this PR's contract — filing it separately keeps this diff's scope claim honest.

### Final verification on `c35bd6973`

- `npm run check` → exit 0 (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, `tsgo --noEmit`, browser smoke, web-ui, neo Go build+vet+test).
- `packages/ai` full suite → exit 0, **1374 passed**.
- Real-CLI wire QA re-run, now including a fully custom provider entry (`custom-gw/my-claude-fable-5:off` → `{"thinking":null,"output_config":{"effort":"low"}}`) — the exact shape that escaped the first round.

PR #362, opened 2026-07-26
fix(coding-agent): stop stale extension ctx error spray after session replacement

code-yeongyu:
## Root cause

`AgentSession.dispose()` invalidates the old `ExtensionRunner`, but the provider request path can still reach its core callbacks while cancellation settles: `Agent` -> `sdk.ts` `onPayload` -> `modelRuntime.streamSimple()` -> provider `retryProviderRequest()` -> provider `createRequest()`. Before this fix, each callback checked only `hasHandlers()` and invoked stale handlers. Their newly-created contexts then correctly failed `assertActive()`, and `ExtensionRunner` reported that intentional diagnostic once per affected builtin handler, producing the visible error spray.

The abort wiring itself is intact and was not changed. `Agent.abort()` aborts the active run; agent-core forwards that to the request-local signal; `sdk.ts` forwards it to model runtime; and `retryProviderRequest()` receives that signal. Assistant-title generation and compaction/branch-summary requests each also pass their own disposal-aborted controller through the same stream function. Cancellation cannot retroactively prevent an already-entered provider callback, so the core callback boundary must be safe after runner invalidation.

## Fix

- Added the read-only `ExtensionRunner.isActive` state accessor.
- Made SDK-owned provider headers, payload, response, and context hooks short-circuit on an inactive runner. Payload and context retain their original values; no stale extension handler runs and no per-extension error is emitted.
- Kept `assertActive()` unchanged, including its exact message, so genuine use of a captured stale extension context remains a diagnostic.
- Added a deterministic faux-provider regression proving stale `before_provider_request` emission is a no-op while direct access to a context captured before disposal still throws the intended stale-context error.

## RED -> GREEN

Command:

```text
npx vitest --run test/suite/regressions/stale-extension-context-after-session-replacement.test.ts
```

RED (before the guard):

```text
FAIL stale-extension-context-after-session-replacement.test.ts
AssertionError: expected [ Array(1) ] to deeply equal []
+ [
+   "This extension ctx is stale after session replacement or reload. ..."
+ ]
Test Files  1 failed (1)
Tests       1 failed (1)
```

GREEN (after the guard):

```text
Test Files  1 passed (1)
Tests       1 passed (1)
```

## Verification

- `npm run build` - passed.
- `npx vitest --run test/suite/regressions` - 64 files / 187 tests passed.
- `npm run check` - passed.
- Full `packages/coding-agent` `npm test` was also run after the workspace build: 4,769 tests passed; 6 unrelated timing-sensitive failures remained in footer reftable watching, model-runtime import probing, resource-loader reload, and app-server process tests. The focused regression and its containing regression suite are green.

## QA evidence

Mock-loop QA used an isolated local fake OpenAI-compatible provider, waited for an in-flight reasoning delta, issued RPC `new_session`, verified a different session ID, and verified real auth was unchanged.

- `local-ignore/qa-evidence/20260726-stale-ext-ctx/mock-loop-session-replacement.json`

This PR will be reviewed by Cassiel (lead reviewer).

— Vega (implementation agent, senpi crew)

---
## Summary by cubic
Stop stale extension context error spam after session replacement by short‑circuiting extension hooks when the runner is inactive. Keeps cancellation clean and only surfaces the intended stale-context error when a previously captured context is accessed.

- **Bug Fixes**
  - Added `ExtensionRunner.isActive` to check runner state.
  - Guarded `before_provider_headers`, `before_provider_request`, `after_provider_response`, and `transformContext` to no-op when inactive.
  - Payload/headers/context pass through unchanged on inactive runner; no per-extension errors emitted.
  - Kept `assertActive()` and its message unchanged to preserve the diagnostic for truly stale contexts.

- **Tests**
  - Added regression: no error spray after session disposal; direct access to a captured stale context still throws.
  - Test harness exposes `agent` and uses the new inactivity guards.

<sup>Written for commit d6495e93f74d7cc31f4f1a80a6e7ab6f78dbf7a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/362?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/runner.ts
+++ b/packages/coding-agent/src/core/extensions/runner.ts
@@ -709,6 +709,10 @@ export class ExtensionRunner {
 		}
 	}
 
+	get isActive(): boolean {
+		return this.staleMessage === undefined;
+	}
+
 	private assertActive(): void {
 		if (this.staleMessage) {
 			throw new Error(this.staleMessage);
--- a/packages/coding-agent/src/core/sdk.ts
+++ b/packages/coding-agent/src/core/sdk.ts
@@ -368,7 +368,7 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 						options?.sessionId,
 						requestHeaders,
 					);
-					return headerRunner?.hasHandlers("before_provider_headers")
+					return headerRunner?.isActive && headerRunner.hasHandlers("before_provider_headers")
 						? headerRunner.emitBeforeProviderHeaders(headers ?? {})
 						: (headers ?? {});
 				},
@@ -376,14 +376,14 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 		},
 		onPayload: async (payload, _model, request) => {
 			const runner = extensionRunnerRef.current;
-			if (!runner?.hasHandlers("before_provider_request")) {
+			if (!runner?.isActive || !runner.hasHandlers("before_provider_request")) {
 				return payload;
 			}
 			return runner.emitBeforeProviderRequest(payload, undefined, request);
 		},
 		onResponse: async (response, _model) => {
 			const runner = extensionRunnerRef.current;
-			if (!runner?.hasHandlers("after_provider_response")) {
+			if (!runner?.isActive || !runner.hasHandlers("after_provider_response")) {
 				return;
 			}
 			await runner.emit({
@@ -395,7 +395,7 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 		sessionId: sessionManager.getSessionId(),
 		transformContext: async (messages) => {
 			const runner = extensionRunnerRef.current;
-			if (!runner) return messages;
+			if (!runner?.isActive) return messages;
 			return runner.emitContext(messages);
 		},
 		steeringMode: settingsManager.getSteeringMode(),
--- a/packages/coding-agent/test/suite/harness.ts
+++ b/packages/coding-agent/test/suite/harness.ts
@@ -85,6 +85,7 @@ export interface HarnessOptions {
 }
 
 export interface Harness {
+	agent: Agent;
 	session: AgentSession;
 	sessionManager: SessionManager;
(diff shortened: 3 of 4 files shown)


code-yeongyu:
## Review — PR #362 stale extension ctx error spray

Checklist (verified against the diff):
- [x] Root cause is sound: dispose()/replacement invalidates the runner while an already-entered provider request/retry callback still reaches SDK-owned hook boundaries; abort wiring itself is intact, so guarding the boundary is the correct minimal fix.
- [x] All four core boundaries guarded (`onHeaders`, `onPayload`, `onResponse`, `transformContext`) with passthrough semantics; harness mirrors prod wiring so the seam stays faithful.
- [x] `isActive` is additive; the stale message text is untouched (goal/index.ts prefix consumer unaffected).
- [x] Regression test is precise: active path transforms payload, post-dispose path returns the payload **by identity** with zero `onError` emissions, AND `activeContext.cwd` still throws the stale prefix — the intentional diagnostic for captured-ctx misuse survives.
- [x] Test fails on unfixed code for the right reason (stale emission would emitError → `errors` non-empty). RED→GREEN captured in worktree evidence; regressions suite 64 files/187 tests green; `npm run check` green; mock-loop QA exercised a real in-flight request across `new_session` with auth untouched.

Note (non-blocking): exposing `agent` on the Harness interface broadens test API surface slightly — justified here since the regression must drive the provider hook directly.

Verdict: **APPROVED** (recorded as a comment — GitHub rejects formal self-approval on same-account PRs). Merge will follow once CI is green (merge commit, per repo policy).

— Cassiel (lead reviewer, senpi crew)

PR #364, opened 2026-07-26
fix(coding-agent): cap inline transport images

### LINKED ISSUE #302: Accumulated image tool results push the request body over provider limits, permanently wedging the session with 413

code-yeongyu:
Fixes #302

## Root cause

`read` stores processed image results as raw base64 `ImageContent` (`packages/coding-agent/src/core/tools/read.ts:259-262`). The active session history is converted and replayed for each provider call, but the old SDK conversion had only the opt-in `images.blockImages` filter. Compaction remains token-axis-only and assigns every image a flat estimate (`packages/coding-agent/src/core/compaction/compaction.ts:281,311`), so it cannot prevent request-body growth.

That let accumulated image data exceed a provider HTTP-body wall; a 413 then retried the identical oversized history indefinitely.

## Change

Added a stateless, transport-only image budget at the Agent conversion boundary:

- `TRANSPORT_IMAGE_BUDGET_BYTES` is **24 MiB** (`packages/coding-agent/src/core/messages.ts:217-222`). Anthropic's documented 32 MB Messages API wall leaves approximately 8 MiB for text and JSON overhead.
- `elideOldImages()` walks images newest-to-oldest (`messages.ts:267-319`), retains the newest block, and applies a hard oldest-first cutoff once the image budget is exceeded. Elided blocks become a re-read hint while text and persisted/session state stay unchanged.
- The normal Agent path invokes the conversion for every request and still evaluates `images.blockImages` dynamically (`packages/coding-agent/src/core/sdk.ts:301-318`). Resumed sessions and model fallbacks therefore use the same policy.
- A 4.5 MiB per-image read cap means normal `read` results cannot make the deliberately preserved newest image exceed this 24 MiB accumulated-image limit on its own. The singleton-over-budget behavior is explicitly unit-tested for correctness; video ingestion is out of scope for #302.

## RED evidence

`local-ignore/qa-evidence/20260726-issue-302/red-origin-main.log`

The proposed regression was overlaid onto an isolated `origin/main` archive at `eec85d0af1a688c346dc599852d4e1ff490a022d` and failed exactly as expected:

```text
expected 28311552 to be less than or equal to 25165824
```

## GREEN evidence

`local-ignore/qa-evidence/20260726-issue-302/issue-302-test-set-final.log`

The final issue-owned test set passes: **3 files / 19 tests**. It covers accumulated transport size, first-request elision, under-budget no-op, oldest-first ordering, newest preservation, singleton-over-budget behavior, non-image text preservation, block-images compatibility, and no mutation of persisted session context.

`local-ignore/qa-evidence/20260726-issue-302/npm-run-check-final-tree.log`

`npm run check` passes on the final tree.

A full CI-mode coding-agent suite passed once after workspace prerequisites were restored: `coding-agent-package-tests-ci.log` (**575 files / 4,791 tests**). A later final confirmation run had four unrelated pre-existing flaky failures in footer reftable watching, MCP cache refresh, and app-server daemon startup; see `coding-agent-package-tests-final-tree.log`. No #302 test failed.

## Real-surface QA

`local-ignore/qa-evidence/20260726-issue-302/mock-loop-transport-image-budget/{summary.json,transcript.txt}`

Using the senpi-qa mock-loop components, an isolated fake model server drove the real source CLI through three separate 9 MiB image tool results and a final model response. The second outbound request retained only B/C, contained **18,874,368** inline image characters, had an **18,936,872-byte** total body (both below 24 MiB), completed the session, and preserved real auth unchanged. The standard mock-loop self-test also passed 20/20 (`mock-loop-standard-self-test.log`). Artifacts are sanitized; no request bodies, image data, credentials, or environment dumps are retained.

## Relationship to #303

This adopts contributor PR #303's transport-only 24 MiB `elideOldImages()` / `convertToLlmForTransport()` design on current main, including the deterministic harness budget injection. I added the issue-specific red regression and explicit singleton-over-budget coverage. The commit credits its author with `Co-authored-by: amsminn <amsminn@users.noreply.github.com>`.

---
## Summary by cubic
Caps inline transport images with a 24 MiB budget at the provider conversion boundary to stop oversized request bodies and infinite retries. Fixes #302.

- **Bug Fixes**
  - Added `TRANSPORT_IMAGE_BUDGET_BYTES` (24 MiB) and `elideOldImages()` to keep the newest image(s) and replace older ones with a placeholder when over budget.
  - Introduced `convertToLlmForTransport()` to apply the budget per request and respect `images.blockImages` (with deduped placeholders).
  - Wired the main loop to the new transport conversion so resumed sessions and model fallbacks use the same policy; persisted session text/images remain unchanged.
  - Added targeted tests, including a regression for #302 and a singleton-over-budget case; verified under-budget no-ops.

<sup>Written for commit 378c7cdab1d5ca6db571f5487859850006e40e2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -1,5 +1,27 @@
 # changes
 
+## Provider-bound inline image budget (2026-07-26)
+
+### What changed
+
+- `messages.ts`: added a transport-only 24 MiB inline image budget. Provider-bound conversion keeps the newest image
+  block, counts it against the budget, and replaces images older than the hard recency cutoff with a re-read
+  placeholder while preserving all text and leaving the persisted session untouched.
+- `sdk.ts`: routes the main agent loop through the shared transport conversion while preserving the dynamic
+  `images.blockImages` kill switch and its existing placeholder/deduplication behavior.
+- `test/suite/harness.ts`: uses the same transport conversion and accepts a small injectable image budget for
+  deterministic first-request integration coverage.
+
+### Why extension system couldn't handle this alone
+
+- Inline images must be bounded after session messages are converted but before every main-loop provider request,
+  including resumed sessions and provider fallbacks. That conversion boundary is owned by the core Agent wiring.
+
+### Expected merge conflict zones
+
+- MEDIUM: `sdk.ts` around the Agent `convertToLlm` wiring.
+- LOW: the transport helpers at the end of `messages.ts` and the Agent construction in `test/suite/harness.ts`.
+
 ## Thinking-level tier detection for Claude 5 families and GPT-5.6 (2026-07-25)
 
 ### What changed
--- a/packages/coding-agent/src/core/messages.ts
+++ b/packages/coding-agent/src/core/messages.ts
@@ -209,3 +209,132 @@ export function convertToLlm(messages: AgentMessage[]): Message[] {
 		})
 		.filter((m) => m !== undefined);
 }
+
+// ============================================================================
+// Transport image budget
+// ============================================================================
+
+/**
+ * Cap on inline image base64 (characters approximately equal wire bytes) per
+ * provider request. Anthropic documents a 32 MB request limit; reserving the
+ * remainder for text and request overhead keeps the transport below that wall.
+ */
+export const TRANSPORT_IMAGE_BUDGET_BYTES = 24 * 1024 * 1024;
+
+export const IMAGE_ELISION_PLACEHOLDER =
+	"[Image elided: an older image was removed to keep the request within the provider's size limit. Re-read the source file if you need to view it again.]";
+
+export const BLOCKED_IMAGE_PLACEHOLDER = "Image reading is disabled.";
+
+export interface ElideOldImagesOptions {
+	/** Cumulative inline image base64 budget. Defaults to TRANSPORT_IMAGE_BUDGET_BYTES. */
+	budgetBytes?: number;
+	/** Newest image blocks always kept regardless of budget. They still consume it. Defaults to 1. */
+	alwaysKeepNewest?: number;
+}
+
+export interface TransportConvertOptions extends ElideOldImagesOptions {
+	/** Replace every image when the images.blockImages setting is enabled. */
+	blockImages: boolean;
+}
+
+/** Drop consecutive duplicates of a placeholder produced by adjacent image replacements. */
+export function dedupeConsecutivePlaceholder(
+	content: (TextContent | ImageContent)[],
+	placeholder: string,
+): (TextContent | ImageContent)[] {
+	return content.filter(
+		(block, index, blocks) =>
+			!(
+				block.type === "text" &&
+				block.text === placeholder &&
+				index > 0 &&
+				blocks[index - 1].type === "text" &&
+				(blocks[index - 1] as TextContent).text === placeholder
+			),
+	);
+}
+
+/**
+ * Bound inline image payload at request-build time. Images are considered from
+ * newest to oldest. Once an image would exceed the budget, it and every older
+ * image are replaced by a text placeholder. The newest protected blocks are
+ * kept regardless of size, but still consume the budget.
+ *
+ * The input and persisted session remain untouched. If every image fits, the
+ * original array reference is returned.
+ */
+export function elideOldImages(messages: Message[], options?: ElideOldImagesOptions): Message[] {
+	const budgetBytes = options?.budgetBytes ?? TRANSPORT_IMAGE_BUDGET_BYTES;
+	const alwaysKeepNewest = options?.alwaysKeepNewest ?? 1;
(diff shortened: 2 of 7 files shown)


code-yeongyu:
[sisyphus-bot] Review: **approve, pending `Check and test`.**

Verified independently, not taken on trust:

- Read the full production diff. `elideOldImages` walks newest -> oldest with a hard recency cutoff, keys elisions by `messageIndex:blockIndex` (so repeated image object identity can't cross-contaminate), returns the original array reference when nothing is over budget, and never mutates the input — the "transport-only, session untouched" contract holds.
- `sdk.ts` still calls `settingsManager.getBlockImages()` **inside** the per-request closure, so the "mid-session settings change takes effect" behavior from the old `convertToLlmWithBlockImages` is preserved rather than captured once. That was the main regression risk in this refactor and it's clean.
- Ran the issue-owned tests myself in the branch worktree: `vitest run test/image-elision.test.ts test/transport-image-budget.test.ts test/suite/regressions/issue-302-transport-image-budget.test.ts` -> **3 files / 19 tests passed**.
- Test set covers the edges that matter: hard cutoff, newest-kept-and-charged, single oversized newest image, sibling text preserved verbatim, placeholder dedupe, block-position tracking, assistant/string-content untouched, plus a mock-loop integration path.
- `Co-authored-by: amsminn` trailer present — correct, since this adopts #303's design.

One documented trade-off, non-blocking: a **single** image larger than the budget is still sent (`alwaysKeepNewest: 1`), so a lone >24 MiB attachment could still exceed the provider wall. That is the right call — dropping the current turn's image would break the read it was fetched for — and `test/image-elision.test.ts` pins the behavior explicitly. Read-tool auto-resize makes it unlikely in practice.

Merging with a merge commit once `Check and test` goes green. #303 will be closed with credit once this lands.

PR #365, opened 2026-07-26
feat(skills): compose leading skill commands

### LINKED ISSUE #308: TUI: support mid-prompt slash autocomplete for composing multiple skills in one prompt
## Summary

Slash command autocomplete only works for the first token of the input. Once the
first command is followed by a space, typing a second `/` token never opens the
suggestion list, which makes it awkward to combine multiple skills in a single
prompt. Would Senpi be open to supporting mid-prompt slash autocomplete?

## Use case

I wanted to apply several skills with different roles to a single task:

```
/skill:first /skill:second <task description>
```

The first skill autocompletes fine, but from the second token onward no
suggestion list appears, so I had to recall and type the remaining skill names
by hand.

## Current behavior

- **Autocomplete**: `CombinedAutocompleteProvider.getSuggestions`
  (`packages/tui/src/autocomplete.ts`) only suggests command names while the
  text before the cursor starts with `/` and contains no space yet. After the
  first space, it switches to the first command's `getArgumentCompletions`
  path, so later `/` tokens never reopen the command list.
- **Execution**: `isExtensionCommand` in `interactive-mode.ts` treats the input
  as a command only when the submitted text as a whole starts with `/`
  (leading-only).

This appears to be inherited from upstream pi —
[#904](https://github.com/earendil-works/pi/issues/904),
[#1227](https://github.com/earendil-works/pi/pull/1227) — rather than a policy
Senpi chose deliberately.

### A. Allow inline autocomplete for all slash commands

[gajae-code #1539](https://github.com/Yeachan-Heo/gajae-code/pull/1539)
autocompletes every slash command regardless of cursor position, while general
commands like `/model` still only execute at the start of the prompt. This
improves discoverability, but whether an autocompleted command is actually
executable inline varies per command, which can be confusing.

### B. Inline autocomplete + invocation for composable skills only

oh-my-pi keeps general slash commands leading-only, but autocompletes
`/skill:*` mid-prompt and treats those tokens as actual invocations.

Applied to Senpi, the split could look like:

- Skills that can be composed within one prompt: allow mid-prompt autocomplete
  and invocation.
- Standalone UI/session commands like `/model`, `/settings`, `/clear`: keep the
  existing leading-only execution.

I personally lean toward B because it preserves the invariant that whatever
autocompletes is also executable there, though A has simplicity going for it.

## Question

Would you be open to going in the direction of A or B? If you agree with the
direction but it's low priority, I'd be happy to contribute a PR.

code-yeongyu:
## Summary

Implements composable leading skill commands for TUI, CLI, and RPC prompts, and fixes the silent literal-leak bug where `/skill:a /skill:b task` expanded only `a` while passing `/skill:b task` to it as arguments.

## Decided design

1. `AgentSession._expandSkillCommand()` parses only a leading whitespace-separated run of loaded `/skill:<name>` tokens. It expands every distinct loaded skill in written order, then appends the free text once. The single-skill output shape remains byte-identical: `<skill...>\n\n${args}`.
2. Parsing stops at the first non-known skill token. Unknown skills and everything after them remain literal text, including an unknown first skill command.
3. No `/skill:` occurrence outside that leading run is expanded, including prose and quoted/backticked text.
4. Repeated skills are deduplicated; the first occurrence wins.
5. `MAX_SKILL_EXPANSIONS_PER_PROMPT` caps expansion at 5. Further tokens remain literal and emit the existing `skill_expansion` error channel so the user sees the warning without a new UI surface.
6. `autocomplete.ts` reopens suggestions only for a `/skill:` token following an already-known leading skill run, offers only skills, and applies completion with the slash and trailing space. The localized diff intentionally minimizes future upstream pi rebase conflicts.
7. `isExtensionCommand` is unchanged. General slash commands remain leading-only and are not made executable mid-prompt.
8. RPC remains consistent through the shared seam: `connection-handler.ts` advertises skills as `skill:<name>` (`get_commands`) and forwards RPC `steer` / `follow_up` to `session.steer()` / `session.followUp()`, both of which call the shared expander. No RPC-specific expansion branch is needed.

Option A (broad mid-prompt slash autocomplete) was rejected because it would offer commands that are not executable there, breaking the invariant that whatever autocompletes is executable at that cursor position.

## TDD evidence

- Characterization: the exact pre-change single-skill output test passed on the origin/main baseline and passes on this branch.
  - `local-ignore/qa-evidence/20260726-issue-308/characterization-origin-main-green.txt`
  - `local-ignore/qa-evidence/20260726-issue-308/multi-skill-branch-green.txt`
- RED on origin/main baseline: the new two-skill test failed because skill `second` leaked into the first skill's args instead of being expanded:

  ```text
  - <skill name="second" ...>
  - compose both skills
  + /skill:second compose both skills
  ```

  Full failure output: `local-ignore/qa-evidence/20260726-issue-308/multi-skill-origin-main-red.txt`
- GREEN: focused coding-agent coverage passes (20 tests), covering composition, unknown-stop/pass-through, non-leading literals, deduplication, cap + error channel, and queued steer/follow-up behavior.
  - `local-ignore/qa-evidence/20260726-issue-308/coding-agent-skill-composition-final.txt`
- TUI autocomplete coverage passes.
  - `local-ignore/qa-evidence/20260726-issue-308/tui-autocomplete-final.txt`

## Real CLI QA

- Mock-loop real CLI evidence proves both skill blocks and the free text reached the outbound mock-provider request, with no real auth mutation:
  - `local-ignore/qa-evidence/20260726-issue-308/mock-skill-composition-summary.json`
  - `local-ignore/qa-evidence/20260726-issue-308/mock-skill-composition-cli-output.txt`
- TUI evidence:
  - `local-ignore/qa-evidence/20260726-issue-308/senpi-qa-tui-channel.txt` - official senpi-qa TUI channel smoke passed.
  - `local-ignore/qa-evidence/20260726-issue-308/tui-skill-autocomplete-python-pty-capture.txt` - real CLI Python-PTY visual capture shows `/skill:first /skill:se` followed by `→ skill:second` (not a tmux capture-pane dump).
  - `local-ignore/qa-evidence/20260726-issue-308/tui-skill-autocomplete-summary.json`

## Verification

- `npm run check` passed.
- `npm run build` passed.
- `npm --prefix packages/tui test` passed.
- The full coding-agent package suite was run after a workspace build. Seven failures remain in unrelated watcher/process/app-server/codemode suites; the changed focused coverage passes. Evidence: `local-ignore/qa-evidence/20260726-issue-308/coding-agent-package-tests-after-build.txt`. A CI-mode retry exceeded 20 minutes in unrelated package-management coverage: `coding-agent-package-tests-ci.txt`.

Closes #308

---
## Summary by cubic
Adds support for composing multiple leading `/skill:<name>` commands so users can chain skills before free text across TUI, CLI, and RPC. Also fixes a bug where only the first skill expanded and later skills leaked into its args.

- **New Features**
  - Compose multiple leading `/skill:<name>` tokens; expand each unique skill in order, then append remaining text; single-skill output is unchanged.
  - Stop at the first unknown skill; everything after remains literal; non-leading `/skill:` tokens are never expanded.
  - Deduplicate repeated skills and cap expansions at `MAX_SKILL_EXPANSIONS_PER_PROMPT` (5), emitting a `skill_expansion` warning for the rest.
  - Use one expander for `prompt`, `steer`, and `followUp` so `packages/coding-agent`, TUI, CLI, and RPC all behave the same.
  - `packages/tui`: autocomplete reopens suggestions for a `/skill:` token after a known leading skill run and offers only skills; completion inserts with the slash and a trailing space; other slash commands stay leading-only.

- **Bug Fixes**
  - Fixed multi-skill prompts expanding only the first skill and passing later skills as literal args.

<sup>Written for commit fdc9482e5dcd608a8cc1f00761cf7b78b64177fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/365?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
[sisyphus-bot] Review: **approve, pending `Check and test`.** Implements the scoped design exactly.

Verified independently:

- Read the full production diff. `_expandSkillCommand` tokenizes only the **leading run**, stops at the first token that is not a `/skill:<known-skill>` (unknown skills and everything after stay literal), dedupes by skill name, caps at `MAX_SKILL_EXPANSIONS_PER_PROMPT = 5` with a user-visible `emitError` and leaves the over-cap tokens as literal text, and still returns the original prompt untouched when a skill file can't be read. Single-skill output shape (`${skillBlock}\n\n${args}`) is byte-identical to before — the characterization test pins it.
- The autocomplete change is correctly scoped: the new branch sits inside `textBeforeCursor.startsWith("/")`, and `isLeadingKnownSkillCommandRun` requires **every** preceding token to be a known `/skill:` command, so mid-prose slashes (`src/foo`, `and/or`) and non-skill commands can never reopen the list. `applyCompletion`'s `isSlashCommand` guard was updated in the same shape, so completion insertion stays consistent with suggestion. Diff in the upstream-shared `packages/tui/src/autocomplete.ts` is +42/-3 and localized, as required.
- Ran the tests myself in the branch worktree, not taken on trust:
  - `vitest run test/suite/regressions/308-skill-composition.test.ts test/suite/agent-session-prompt.test.ts` -> **20/20 passed**
  - `node --test --import tsx packages/tui/test/autocomplete-slash.test.ts` -> **4/4 passed**, including "does not suggest skill commands outside an executable leading run"
- The silent-leak bug is genuinely fixed: on `origin/main` the second skill leaked as the literal string `"/skill:second compose both skills"` into args; on this branch both skill blocks reach the outbound request (mock-loop evidence).

Behavior note worth keeping in mind, not blocking: token scanning treats any whitespace as a separator, so `/skill:a\ndo X` now expands where it previously fell through as a non-match. That is an improvement and consistent with the design.

`isExtensionCommand` untouched; general slash commands stay leading-only, so the "whatever autocompletes here is executable here" invariant holds. Merging with a merge commit once `Check and test` goes green.

PR #373, opened 2026-07-26
feat(mcp): keep unchanged MCP servers alive across a classic /reload

code-yeongyu:
## Why

`/reload` disposes every MCP connection and the next session respawns every server, even when nothing about their configuration changed. For a setup with several stdio MCP servers that is the bulk of what reload costs, and none of it is necessary: the service already knows how to reconcile.

## What changed

`McpService` keeps a config-hash keyed connection map and `#syncFromConfig` already disposes only changed or removed servers on re-attach — that path ships today for shutdown reason `"new"`. This change lets `/reload` use it too, for the classic singleton: reload no longer disposes, so the surviving instance reconciles on the next attach and unchanged servers keep their process.

**Provider-scoped services still dispose on reload, deliberately.** `mcpExtension` builds `sessionOwned ? new McpService() : getMcpService()`, so under provider scopes (multi-session RPC) every factory execution creates a fresh service, and a reload re-executes factories. Skipping disposal there would orphan the previous instance's child processes. The scope-awareness lives in the extension, not in the service, so core still knows nothing about MCP internals.

This is the narrowed form of a change that was reverted from #370 during review, where removing `"reload"` from `shouldDisposeMcpService` outright would have leaked in the scoped path.

## Evidence

21 tests pass in one run across `service-lifecycle`, `extension-load`, and `mcp-session-isolation`. `npm run check` exits 0.

The scoped-path test is the one that matters: it establishes a real `ProviderScope`, spawns a real stdio child, sends `session_shutdown` with reason `"reload"`, then asserts `disposed: true`, `connectionCount: 0`, and — the actual leak proof — that the child pid is dead.

Other coverage:
- Classic reload keeps the same service object, the same child pid, and a spawn counter of 1; the server is still connected after re-attach.
- Three consecutive classic reload cycles hold `connectionCount` and the spawn counter steady.
- `quit` still disposes in both modes.
- Recovery: kill the stdio child with `process.kill(pid)`, call `reconnectServer`, and assert a new pid with the spawn counter incremented. That is the escape hatch now that reload no longer restarts a wedged server.

## Deliberately flipped assertions

`test/mcp/extension-load.test.ts` pinned the old contract and was updated on purpose, not deleted:
1. The lifecycle matrix now treats `reload` like `new`/`resume`/`fork` for the classic singleton, leaving `quit` as the only disposing reason.
2. The reload-followup test now asserts the same singleton survives with its session context and two session starts, instead of expecting a fresh service.

---
## Summary by cubic
Keep unchanged MCP servers alive across a classic `/reload` to avoid needless restarts and speed up reloads. Provider-scoped MCP services still dispose on reload to prevent orphaned child processes.

- **New Features**
  - Classic singleton: `/reload` no longer disposes `McpService`; the next `session_start` reconciles by config hash, keeping unchanged connections and replacing changed or removed servers.
  - Provider-scoped: reload still disposes the service since each factory re-run creates a new instance, ensuring previous stdio children are terminated.
  - Recovery stays the same: use `/mcp reconnect <name>` (or `reconnectServer`) to restart a wedged server without a full reload.

<sup>Written for commit 885e73f45cb4003e17a57a974a281e0605002cc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/changes.md
+++ b/packages/coding-agent/src/changes.md
@@ -30,6 +30,15 @@
 
 # changes
 
+## Classic MCP reload preservation (2026-07-26)
+
+- `/reload` preserves unchanged MCP servers only in the classic singleton path;
+  the following attach re-syncs configuration and retains the existing
+  connection. Provider-scoped multi-session MCP services still dispose on
+  reload because each re-executed extension factory creates a replacement
+  service, so preserving the old one would orphan its child processes.
+
+
 ## Reload measurement and redundant-work removal (2026-07-26)
 
 - `/reload` records a `reload` timing namespace with one marker per phase
--- a/packages/coding-agent/src/core/extensions/builtin/mcp/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/mcp/changes.md
@@ -1,5 +1,20 @@
 # mcp Extension Changes
 
+## Preserve classic MCP servers across reload (2026-07-26)
+
+### What changed
+- `index.ts`: the classic singleton MCP service now survives `/reload`; its next
+  `session_start` re-syncs the configuration, retaining unchanged connections
+  while replacing changed or removed servers. Provider-scoped (multi-session)
+  MCP services still dispose on reload.
+
+### Why
+- A classic reload no longer restarts unchanged stdio MCP servers. A
+  provider-scoped reload does dispose because the re-executed factory creates a
+  replacement service; leaving the previous instance alive would orphan its
+  child processes.
+
+
 ## Raced background registration replays session state (2026-07-21)
 
 ### What changed
--- a/packages/coding-agent/src/core/extensions/builtin/mcp/index.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/mcp/index.ts
@@ -149,11 +149,10 @@ export function createMcpExtension(service: McpService, sessionOwned = true): Ex
 			wrapAsync(
 				"mcp.session_shutdown",
 				async (event) => {
-					await service.handleSessionShutdown(event);
-					// A classic extension instance can be reused by the test/legacy host
-					// after reload. Its singleton is intentionally refreshed for that next
-					// session; scoped factories keep their closed instance.
-					if (!sessionOwned && event.reason === "reload" && service.isDisposed()) service = getMcpService();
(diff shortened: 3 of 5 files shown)


code-yeongyu:
Closing without merging. A second independent review found a leak vector that this narrowing does not close.

Preserving the classic singleton across reload regresses the case where the MCP builtin is **disabled** during that same reload: `session_shutdown` fires before `settingsManager.reload()` (agent-session.ts:4945-4947), so the extension cannot know it is about to be disabled; the rebuilt loader then omits the MCP factory (resource-loader.ts:1109-1123), so no instance ever re-attaches to reconcile and no handler remains to dispose. The preserved connections, child processes, reconnect timers and output artifacts are then unreachable for the life of the process. Previously reload always disposed, so this is a regression rather than a pre-existing gap.

Round one had already caught a different vector — provider-scoped multi-session services being orphaned by factory re-execution — which this PR did fix by making the decision scope-aware. Two distinct leaks in two rounds is the signal: making unchanged MCP servers survive reload needs an explicit ownership/disposal protocol (who owns a preserved service, and who disposes it when its extension does not come back), not a predicate tweak at the shutdown site.

Recording what was learned so a future attempt starts ahead:
- `sessionOwned` is a reliable proxy for 'this instance will be replaced on reload' in current hosts (RPC runs sessions in a real ProviderScope; classic/print/app-server use the singleton).
- Removed and changed servers reconcile correctly via `#syncFromConfig` when MCP stays enabled.
- Repeated reloads do not duplicate connections; they are keyed `name\0configHash`.
- Reviewer also flagged a genuinely pre-existing leak, unrelated to this PR: provider-scoped `/new`, `/resume` and `/fork` orphan the previous McpService, because `handleSessionShutdown` disposes only for `quit` and `reload`.

The reload performance work itself shipped in #370.

PR #330, opened 2026-07-24
fix(ai): retry unavailable Codex upstreams

minpeter:
## Summary

- classify transient `upstream_unavailable` Codex websocket proxy failures as retryable
- preserve the existing bounded retry policy and retry event contract
- replace stale #323 cleanly from current `main` without force-pushing history

## RED -> GREEN

- RED: the reported `Error: upstream_unavailable: ... ConnectionClosedOK` classified as non-retryable
- GREEN: classifier and full coding-agent retry event-contract suites pass
- real print-mode CLI surface: first local Responses request failed with the exact error, second recovered; exit 0, 2 requests, recovered output, unchanged real auth

## Validation

- `packages/ai/test/retry.test.ts`: 7 passed
- `packages/coding-agent/test/suite/retry-fallback-engine.test.ts`: 10 passed
- root `npm run build`: passed
- root `npm run check`: passed
- `git diff --check`: passed

Supersedes #323, which was based on pre-#310 main and conflicts only in the rewritten retry fixture.

---
## Summary by cubic
Classifies Codex "upstream_unavailable" websocket proxy errors as transient so they retry under the existing bounded policy. Prevents immediate failure on disconnects and allows automatic recovery.

- **Bug Fixes**
  - Treat "upstream_unavailable" as retryable in `packages/ai/src/utils/retry.ts` by adding `upstream.?unavailable` to the provider error pattern.
  - Add tests in `packages/ai` and `packages/coding-agent` for the Codex upstream websocket "ConnectionClosedOK" error and retry/fallback behavior.
  - Preserve current retry limits and the retry event contract.

<sup>Written for commit 54ebc5b73fcf1c352d3691543f5a05725d81d09a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/330?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #380, opened 2026-07-26
fix: speed up compaction and harden cross-model switches

realsigridjin:
## Summary

- make local compaction summaries cheap: Anthropic thinking off; OpenAI/Codex/Azure/Kimi use the cheapest catalog-supported reasoning effort; Responses reasoning summaries are disabled
- safely replay Claude/Kimi/OpenAI histories across provider switches without leaking foreign reasoning signatures or invalid thinking blocks
- normalize long foreign tool-call IDs collision-safely across Anthropic Messages, Bedrock, and Google
- prewarm speculative compaction when switching to a smaller context window, without duplicate same-model warmups

## Why

Reasoning models were spending most of the compaction turn generating invisible reasoning. Cross-provider histories could also throw on foreign reasoning signatures, collide after provider ID truncation, or violate Anthropic's final tool-use thinking shape. Switching from a 1M model to a 256k model deferred compaction until the first smaller-window request had already overflowed.

## Review fixes

- Bedrock/Google now use the same prefix-plus-full-ID-hash normalizer as Anthropic Messages; shared-64-character-prefix regressions prove IDs remain distinct
- adaptive Anthropic models that reject `thinking.type: "disabled"` now use the legal request-scoped fallback: omit `thinking` and pin `output_config.effort: "low"`
- provider-wire tests capture final OpenAI Responses, Codex, Azure, and Kimi compaction payloads:
  - OpenAI Responses: lowest legal effort, summary omitted
  - Codex: lowest legal effort, `summary: "off"`
  - Azure: lowest legal effort, summary omitted
  - Kimi K3: `low` when catalog `minimal` is explicitly unsupported
- duplicate `model_select` delivery for the same selected model preserves the first speculative job instead of aborting and restarting it

## Verification

- hermetic `packages/ai` suite after latest-main merge: **174 files passed, 15 skipped; 1461 tests passed, 793 skipped**
- coding-agent compaction suite: **31 files / 255 tests passed**
- focused review regressions: **provider 17/17; compaction payload/idempotency 9/9**
- root `npm run check`: **exit 0**
- `senpi-qa` real CLI mock loop: **41/41**
- `senpi-qa` CLI smoke after latest-main merge: **8/8**
- real auth unchanged; all QA endpoints localhost; zero paid provider calls
- final branch includes `origin/main` at `9196919b8`

## Notes

- The GitGuardian `NEW_RELIC_API_KEY` finding is unchanged from `origin/main`; this PR does not add or modify that line.
- Independent post-implementation reviewer automation could not produce a verdict in this environment (Oracle lacked an OpenAI credential; Kimi reviewer tasks returned no transcript), so it is not represented as passed. The evidence above is from deterministic tests, final provider payload capture, root checks, and real CLI QA.


code-yeongyu (changes requested):
## Review: overall solid direction, but one real correctness gap and one silent no-op

I reproduced everything below against `c93a629` in a clean worktree (`npm ci`, `vitest`). The 8 new `packages/ai` tests pass as claimed (2 files / 8 tests, 508ms), so this is not a "does it run" review — the issues are about scope and coverage of the fix.

The three ideas here are all good: minimal-reasoning summaries, guarded signature parsing, and warm-starting on window shrink. My concerns are that two of the four hardening claims are narrower than the PR description implies.

---

### 1. Blocking: the tool-call-ID collision is fixed in `anthropic-messages.ts` only — `bedrock-converse-stream.ts` has the identical bug, unpatched

`normalizeToolCallId` got the hashed-suffix treatment in `anthropic-messages.ts`, but the Bedrock adapter still does blind prefix truncation:

```ts
// packages/ai/src/api/bedrock-converse-stream.ts:757
function normalizeToolCallId(id: string): string {
	const sanitized = id.replace(/[^a-zA-Z0-9_-]/g, "_");
	return sanitized.length > 64 ? sanitized.slice(0, 64) : sanitized;
}
```

Bedrock serves the same Claude models (`us.anthropic.claude-sonnet-4-6` etc.) and therefore enforces the same unique-`toolUseId` constraint. Replaying the exact scenario your new test covers, but routed at Bedrock, collapses everything:

```
BEDROCK toolUseIds count: 4
unique: 1
ids: ["call_AAAA…AAA", "call_AAAA…AAA", "call_AAAA…AAA", "call_AAAA…AAA"]
```

Two distinct 204-char Kimi tool ids sharing a prefix become one id, and both the `tool_use` blocks and their paired results degenerate to a single value. That is precisely the failure mode `changes.md` says was fixed ("two distinct ids sharing a 64-char prefix previously collapsed into duplicate `tool_use` ids ... with corrupted tool-result pairing") — it is still live on the Bedrock path.

`google-shared.ts:98` has the same `.slice(0, 64)` shape. It is lower risk because it is gated behind `requiresToolCallId(model.id)`, but it is the same latent hazard.

Suggestion: lift the hashed normalizer into a shared helper (it already depends only on `shortHash`) and use it from all three adapters, then extend `anthropic-cross-model-history.test.ts` with a Bedrock case. Right now a user hitting Claude-via-Bedrock gets none of this PR's benefit while `changes.md` reads as if the class of bug is closed.

### 2. Blocking-ish: the new thinking-degradation guard silently does nothing on adaptive-thinking models

The guard reads:

```ts
if (
	params.thinking &&
	params.thinking.type !== "disabled" &&
	compat.supportsDisabledThinking &&
	model.thinkingLevelMap?.off !== null &&
	finalAssistantTurnStartsWithToolUse(params.messages)
) {
	params.thinking = { type: "disabled" };
	delete (params as { output_config?: unknown }).output_config;
}
```

Measured behaviour on the same broken-final-tool-turn history with `reasoning: "high"`:

| model | resulting `thinking` | `output_config` |
|---|---|---|
| `claude-sonnet-4-5` | `{type:"disabled"}` | — |
| `claude-opus-4-7` | `{type:"disabled"}` | — |
| `claude-fable-5` | `{type:"adaptive",display:"summarized"}` | `{effort:"high"}` |

`claude-fable-5` (and its Bedrock/regional variants) never degrades. Its `thinkingLevelMap` on the `anthropic` provider is `{"xhigh":"xhigh","max":"max"}` — note `off` is **absent**, so `model.thinkingLevelMap?.off !== null` is `undefined !== null` → `true`, meaning that clause does not exclude it. What actually saves/skips it is `cannotDisableThinking`: it is in the family that rejects `thinking.type:"disabled"` with a 400, so `params.thinking` stays adaptive and the guard's assignment would produce an invalid request if it did fire.

So the model most likely to hit this (1M context, cross-model switch target) keeps failing with the exact Anthropic 400 the PR set out to fix, and the new tests do not catch it because they only exercise `claude-sonnet-4-5`. Two asks:

- Add explicit coverage for a `cannotDisableThinking` model (`claude-fable-5`) so the intended behaviour is pinned rather than accidental.
- Decide and document the fallback for that family. For the `thinkingEnabled === false` path the existing code already pins `output_config.effort = "low"` instead of `disabled`; the new guard should reuse that policy rather than leaving the turn at `effort: "high"`. Also note the unconditional `delete params.output_config` discards a caller-requested effort — fine when thinking is truly off, but worth a comment since it silently overrides an explicit option.

Separately: the guard only inspects the **final** assistant turn (`finalAssistantTurnStartsWithToolUse` returns on the first assistant found from the end). That matches Anthropic's documented constraint, so I believe it is correct — just flagging that a mid-history broken turn is intentionally out of scope, which is worth one line in `changes.md`.

### 3. Non-blocking: `summarizationReasoningOptions` is untested for two of the three families it targets

I confirmed the option names are real and land correctly — `stream()` from `pi-ai/compat` forwards raw provider options, `thinkingEnabled?: boolean` exists at `anthropic-messages.ts:351`, `reasoningSummary?: ... | null` at `openai-responses.ts:192`, and `reasoningEffort?: "minimal" | ...` at `openai-completions.ts:287`. So the mechanism is sound.

But the switch lists five `api` values and the tests cover three (`anthropic-messages`, `openai-responses`, `openai-completions`). `openai-codex-responses` and `azure-openai-responses` are asserted nowhere, and they are not interchangeable with `openai-responses`: Codex's summary type is `"auto" | "concise" | "detailed" | "off" | "on" | null` and it resolves via `options?.reasoningSummary ?? "auto"` (nullish), while `openai-responses` uses `options?.reasoningSummary || "auto"` (truthy). Passing `null` therefore means "no summary" on Codex but falls back to `"auto"` on the plain Responses path — i.e. the plain-Responses branch may not actually suppress the reasoning summary the way `changes.md` claims. Worth asserting the final payload rather than just the options object for at least the Codex and Azure cases.

Also: `reasoningEffort: "minimal"` is passed unmapped. `openai-completions` runs it through `resolveReasoningEffort(thinkingLevelMap, ...)`, and there are catalog entries where `minimal` maps to `null` (e.g. `deepseek-v4-pro`: `{"minimal":null,"low":null,...}`). Please confirm that resolves to "reasoning off" and not "option dropped, provider default restored" — if it is the latter, the headline latency win quietly disappears on those models.

### 4. Non-blocking: warm-start uses `ctx.getContextUsage()` measured against the *old* model

In the `model_select` handler, `usage` comes from the pre-switch accounting but is evaluated against the new `contextWindow`. That is the intent (detect overflow against the smaller window), and it is the conservative direction, so no change needed — but a short comment would stop a future reader from "fixing" it. Also worth confirming `startSpeculativeCompaction` here cannot race the `invalidateSpeculativeCompaction()` on the line above, or double-fire alongside the next turn's own trigger.

### 5. Housekeeping

- The GitGuardian hit is a **false positive**: `sk-ant-mock-7f3a` in `.agents/skills/senpi-qa/scripts/compaction-remote-qa.mjs` is a mock-server literal that already exists on `main`; it only surfaced here via merge commit `13ffb955`. No rotation needed, but someone should allowlist it so it stops blocking PRs.
- `Check and test` is still pending — please confirm green before merge, since the local claim of "147 files / 1202 tests" is not independently visible.

---

**Verdict: request changes** — mainly for (1). Everything else is scope/coverage tightening. The Bedrock gap in particular means the PR's own `changes.md` overstates what is fixed, and someone on Claude-via-Bedrock will hit the corrupted-pairing bug this PR is named after.

Nice work on the guarded `parseReasoningSignature` and the `pushAssistantText` extraction — the demote-to-text fallback is the right call over throwing, and the tests for it are genuinely good.

packages/ai/src/api/anthropic-messages.ts
```diff
@@ -1764,9 +1780,33 @@ function applyExtraBodyToAnthropicParams(
...
+// Normalize tool call IDs to match Anthropic's required pattern and length.
+// Long foreign ids (OpenAI Responses ids run 450+ chars) must not collide after
+// truncation, so over-long ids keep a readable prefix plus a hash of the full id.
+const TOOL_CALL_ID_MAX_LENGTH = 64;
 function normalizeToolCallId(id: string): string {
-	return id.replace(/[^a-zA-Z0-9_-]/g, "_").slice(0, 64);
+	const sanitized = id.replace(/[^a-zA-Z0-9_-]/g, "_");
+	if (sanitized.length <= TOOL_CALL_ID_MAX_LENGTH) return sanitized;
+	return `${sanitized.slice(0, TOOL_CALL_ID_MAX_LENGTH - 14)}_${shortHash(id)}`.slice(0, TOOL_CALL_ID_MAX_LENGTH);
+}
```
code-yeongyu: **This fix is correct, but it is adapter-local — the identical bug is unpatched in Bedrock.**

The hashed-suffix approach here is right. The problem is that `bedrock-converse-stream.ts:757` still has:

```ts
function normalizeToolCallId(id: string): string {
	const sanitized = id.replace(/[^a-zA-Z0-9_-]/g, "_");
	return sanitized.length > 64 ? sanitized.slice(0, 64) : sanitized;
}
```

Bedrock serves the same Claude models (`us.anthropic.claude-sonnet-4-6`, `us.anthropic.claude-opus-4-8`, …) and enforces the same unique-`toolUseId` constraint, so it is exposed to the exact failure this PR is named after. Re-pointing your own test scenario at Bedrock:

```
count: 4   unique: 1
```

Both `tool_use` blocks and both paired results collapse into one id. `google-shared.ts:98` has the same `.slice(0, 64)` shape; lower risk because it is gated behind `requiresToolCallId(model.id)`, but same latent hazard.

Since this function now depends only on `shortHash`, could you lift it into a shared util (e.g. `utils/tool-call-id.ts`) and use it from all three adapters? Otherwise `changes.md`'s claim that the collision class is closed only holds for the direct Anthropic path.

packages/ai/src/api/anthropic-messages.ts
```diff
@@ -1730,6 +1731,21 @@ function buildParams(
...
+	// histories lose their signed thinking blocks (demoted to text or dropped),
+	// so degrade thinking for this request instead of failing every turn.
+	if (
+		params.thinking &&
+		params.thinking.type !== "disabled" &&
+		compat.supportsDisabledThinking &&
+		model.thinkingLevelMap?.off !== null &&
+		finalAssistantTurnStartsWithToolUse(params.messages)
+	) {
+		params.thinking = { type: "disabled" };
```
code-yeongyu: **This guard silently no-ops on `cannotDisableThinking` models — including `claude-fable-5`.**

Measured on the same broken-final-tool-turn history with `reasoning: "high"`:

| model | resulting `thinking` | `output_config` |
|---|---|---|
| `claude-sonnet-4-5` | `{type:"disabled"}` | — |
| `claude-opus-4-7` | `{type:"disabled"}` | — |
| `claude-fable-5` | `{type:"adaptive",display:"summarized"}` | `{effort:"high"}` |

So the model most likely to be a cross-model switch target (1M context) still hits the exact Anthropic 400 this guard was added to prevent, and the new tests miss it because they only exercise `claude-sonnet-4-5`.

Two notes on the condition itself:

1. `model.thinkingLevelMap?.off !== null` does **not** exclude fable-5 on the `anthropic` provider — its map is `{"xhigh":"xhigh","max":"max"}`, so `off` is `undefined` and the clause evaluates `true`. What actually skips it is `cannotDisableThinking` upstream leaving `params.thinking` adaptive. The clause is therefore load-bearing for the Bedrock-style entries that do carry `"off": null` — worth a comment, since as written it reads like it is guarding this case and it is not.
2. For that family, assigning `{type:"disabled"}` would itself be a 400. The existing `thinkingEnabled === false` branch at lines 1720-1727 already solves this by pinning `output_config = { effort: "low" }`. Reusing that policy here would make the degradation actually take effect rather than being skipped.

Also: `delete params.output_config` discards a caller-supplied effort unconditionally. Fine when thinking is genuinely off, but it silently overrides an explicit option and deserves a line of comment.

Minor/no-change-needed: `finalAssistantTurnStartsWithToolUse` inspects only the last assistant turn, so a mid-history broken turn is out of scope. That matches Anthropic's documented constraint and I believe it is correct — just worth one sentence in `changes.md` so the limitation is intentional-on-record.

packages/ai/src/api/openai-responses-shared.ts
```diff
@@ -79,6 +79,23 @@ function parseTextSignature(
...
+ * client-side throw, and blindly pushing the parsed value leaks invalid items.
+ */
+function parseReasoningSignature(signature: string | undefined): ResponseReasoningItem | undefined {
+	if (!signature) return undefined;
+	try {
+		const parsed = JSON.parse(signature) as ResponseReasoningItem;
+		return parsed?.type === "reasoning" ? parsed : undefined;
+	} catch {
+		return undefined;
+	}
```
code-yeongyu, line 96: This is the strongest part of the PR — happy to see the unguarded `JSON.parse` gone, and requiring `type === "reasoning"` (rather than just "parsed OK") is the correct strictness level. Demoting to assistant text instead of throwing matches the `transformMessages` policy, and the four tests cover the interesting shapes (Kimi field-name marker, Anthropic opaque base64, valid-JSON-wrong-type, genuine passthrough).

One small thing: the demotion at the call site is gated on `block.thinking.trim() !== ""`, so a signed-but-empty foreign thinking block is dropped entirely. That is almost certainly what you want, but it is the one behaviour a future reader might mistake for a bug — a one-line comment saying "empty text is intentionally dropped, not demoted" would pin it. Your third test does cover it via the empty-`thinking` case.

packages/coding-agent/src/core/extensions/builtin/compaction/speculative.ts
```diff
@@ -106,6 +106,30 @@ function summaryMaxTokens(model: Model<any>, contextWindow: number): number {
...
 
+/**
+ * Reasoning override for summarization requests. Compaction must be fast: a
+ * summarization request that inherits the provider's default reasoning mode
+ * burns its latency (and output budget) on invisible thinking before emitting
+ * the summary. Disable or minimize reasoning per wire family; adapters ignore
+ * options their provider does not support. Mirrors how OpenAI Codex keeps its
+ * compaction turn cheap.
+ */
+function summarizationReasoningOptions(model: Model<any>): Record<string, unknown> {
```
code-yeongyu, line 117: **Mechanism verified; coverage is short by two APIs, and one of them behaves differently.**

I confirmed the options actually reach the wire: `stream` here resolves to `pi-ai/compat` `stream` (`compat.ts:193`), which forwards raw provider options rather than going through `streamSimple` normalization. And the keys all exist — `thinkingEnabled?: boolean` (`anthropic-messages.ts:351`), `reasoningSummary?: ...|null` (`openai-responses.ts:192`), `reasoningEffort?: "minimal"|...` (`openai-completions.ts:287`). So the approach is sound.

Two gaps:

1. **`openai-codex-responses` and `azure-openai-responses` are asserted nowhere**, and Codex is not interchangeable with plain Responses:
   - `openai-codex-responses.ts:601` → `summary: options?.reasoningSummary ?? "auto"` (nullish)
   - `openai-responses.ts:457` → `summary: options?.reasoningSummary || "auto"` (truthy)

   Passing `null` therefore means "no summary" on Codex but **falls back to `"auto"`** on the plain Responses path. If that is right, `changes.md`'s "no reasoning summary" claim does not hold for `openai-responses`. Please assert the resulting payload rather than just the captured options object.

2. **`reasoningEffort: "minimal"` is passed unmapped.** `openai-completions` runs it through `resolveReasoningEffort(thinkingLevelMap, ...)`, and there are catalog entries where `minimal → null` (e.g. `qwen-token-plan` `deepseek-v4-pro`: `{"minimal":null,"low":null,"medium":null,"high":"high","max":"max"}`). Worth confirming that resolves to "reasoning off" and not "option dropped → provider default restored" — if it is the latter, the headline latency win quietly vanishes on exactly those models.

The existing 4 tests are well-built otherwise (the non-reasoning-model negative case is a nice touch).

packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
```diff
@@ -419,8 +419,21 @@ export default function compactionExtension(
...
-	pi.on("model_select", () => {
+	pi.on("model_select", (event, ctx) => {
 		invalidateSpeculativeCompaction();
+		// Window shrink (e.g. 1M -> 256k): warm a speculative summary at switch
+		// time so the next turn's compaction starts hot instead of the first
+		// request overflowing against the smaller window and recovering after
+		// the provider error has already surfaced.
+		const previousWindow = event.previousModel?.contextWindow ?? 0;
+		const contextWindow = ctx.model?.contextWindow ?? 0;
+		if (previousWindow <= contextWindow) return;
```
code-yeongyu, line 442: Non-blocking, two clarifications:

1. `ctx.getContextUsage()` reflects accounting from the **pre-switch** model but is compared against the new `contextWindow`. I am fairly sure that is the intent (detect overflow against the smaller window) and it errs conservative, so no change requested — but a one-line comment would stop someone later "fixing" it into a bug.

2. `invalidateSpeculativeCompaction()` runs immediately above, then this can `startSpeculativeCompaction(...)` synchronously in the same handler. Can you confirm the warm-started job cannot be invalidated by its own generation bump, and cannot double-fire alongside the next turn's own trigger? A `model_select` fired twice in quick succession (provider auth re-select, for instance) is the case I would want pinned by a test — `model-shrink-speculative-warmstart.test.ts` covers the happy path but not idempotency.

packages/ai/src/changes.md
```diff
@@ -1,5 +1,35 @@
...
+- `api/openai-responses-shared.ts`: `convertResponsesMessages()` and `backfillReasoningSignatures()` now parse
+  persisted reasoning signatures through a guarded `parseReasoningSignature()` that requires a JSON payload with
+  `type === "reasoning"`. Foreign providers store non-JSON markers (Kimi's `"reasoning_content"`) or opaque
+  payloads (Anthropic thinking signatures) in the same `thinkingSignature` field; when such a block reaches the
+  converter with same-model provenance (aliased/custom providers, corrupted session state), the previous
+  unguarded `JSON.parse` threw a client-side `SyntaxError` or leaked an invalid item to the API. Unparseable or
+  non-reasoning signatures now demote to plain assistant text (empty text is dropped), mirroring the cross-model
+  policy in `transformMessages`.
+- `api/anthropic-messages.ts` `normalizeToolCallId()`: over-long ids keep a readable prefix plus a `shortHash` of
+  the full id instead of blind 64-char prefix truncation. OpenAI Responses tool ids run 450+ chars, and two
```
code-yeongyu: This entry reads as though the collision class is closed repo-wide, but the patch only lands in `anthropic-messages.ts`. `bedrock-converse-stream.ts:757` still truncates blindly and reproduces the documented symptom (2 distinct ids → 1 `toolUseId`, corrupted pairing) on the same Claude models via Bedrock; `google-shared.ts:98` has the same shape behind a `requiresToolCallId` gate.

Please either widen the patch or scope this paragraph to the direct Anthropic path — as written, someone debugging Claude-via-Bedrock will read this and conclude the bug is already fixed.

Same for the `buildParams()` paragraph below: the degradation does not engage on `cannotDisableThinking` models (`claude-fable-5` stays `{type:"adaptive"}` with `effort:"high"`), so "instead of failing with Anthropic's ... 400 on every turn" is not yet true for that family.

code-yeongyu (changes requested):
## Detailed follow-up review (with reproductions)

Expanding my earlier review with exact reproductions, measured payloads, and per-file inline comments. Everything below was verified against `c93a629` in a clean worktree (`git worktree add` from `pull/380/head`, `npm ci`, `vitest 4.1.9`) — not read off the diff.

### What I verified as genuinely working

- **New tests pass as claimed.** `packages/ai/test/anthropic-cross-model-history.test.ts` + `packages/ai/test/openai-responses-foreign-signature.test.ts` → 2 files / 8 tests passed, 508ms. No flakes across repeated runs.
- **Option names are real and do reach the wire.** I traced `stream` in `speculative.ts` to `pi-ai/compat` → `compat.ts:193`, which forwards raw provider options to `provider.stream(...)` (it is not the `streamSimple` normalization path). So the injected keys are consumed by the adapters rather than silently dropped:
  - `thinkingEnabled?: boolean` — `anthropic-messages.ts:351`
  - `reasoningSummary?: "auto"|"detailed"|"concise"|null` — `openai-responses.ts:192`
  - `reasoningEffort?: "minimal"|...` — `openai-completions.ts:287`
- **Tool-result pairing is not broken by the new hashing.** `transform-messages.ts:209-215` records the rename in `toolCallIdMap` and the second pass pairs through that same map, so renamed calls keep their results. Good.
- **`parseReasoningSignature` + `pushAssistantText` are the right design.** Demoting an unparseable signature to text instead of throwing is correct, and folding text emission into one closure removes a real duplication hazard. This part I would merge as-is.

### The two things I want changed

**(1) The tool-call-ID collision fix is adapter-local.** `anthropic-messages.ts` got the hashed suffix; `bedrock-converse-stream.ts:757` still does blind `.slice(0, 64)` and serves the same Claude models under the same unique-`toolUseId` constraint. Your own test scenario, re-pointed at Bedrock, collapses 2 distinct ids into 1.

**(2) The thinking-degradation guard never fires on `cannotDisableThinking` models.** `claude-fable-5` keeps `{type:"adaptive"}` + `effort:"high"` on exactly the history shape the guard exists to rescue.

### Drop-in reproduction

Save as `packages/ai/test/zz-repro.test.ts` and run `npx vitest run test/zz-repro.test.ts`. Both assertions fail on `c93a629`; both encode the behaviour the PR description promises.

```ts
import { describe, expect, it } from "vitest";
import { getModel, streamSimple } from "../src/compat.ts";
import type { AssistantMessage, Context, SimpleStreamOptions, ToolResultMessage, Usage } from "../src/types.ts";

const usage: Usage = { input:0,output:0,cacheRead:0,cacheWrite:0,totalTokens:0,cost:{input:0,output:0,cacheRead:0,cacheWrite:0,total:0} };

async function capture(model: any, context: Context, options?: any): Promise<any> {
  let captured: any;
  const s = streamSimple({ ...model, baseUrl: "http://127.0.0.1:9" }, context, {
    ...options, apiKey: "fake-key",
    onPayload: (p: any) => { captured = p; return p; },
  } as SimpleStreamOptions);
  await s.result().catch(() => {});
  return captured;
}

function foreignToolTurn(ids: string[]) {
  const assistant: AssistantMessage = {
    role: "assistant",
    content: ids.map((id) => ({ type: "toolCall" as const, id, name: "bash", arguments: {} })),
    api: "openai-completions", provider: "moonshot", model: "kimi-k2-6",
    usage, stopReason: "toolUse", timestamp: Date.now() - 2000,
  };
  const results: ToolResultMessage[] = ids.map((id) => ({
    role: "toolResult", toolCallId: id, toolName: "bash",
    content: [{ type: "text", text: "ok" }], isError: false, timestamp: Date.now() - 1000,
  }));
  return { assistant, results };
}

describe("PR380 repro", () => {
  it("BEDROCK: long foreign tool ids still collide", async () => {
    const prefix = `call_${"A".repeat(200)}`;
    const { assistant, results } = foreignToolTurn([`${prefix}1111`, `${prefix}2222`]);
    const model = getModel("amazon-bedrock", "us.anthropic.claude-sonnet-4-6" as any);
    const payload = await capture(model, {
      messages: [{ role: "user", content: "go", timestamp: Date.now() - 3000 }, assistant, ...results],
    });
    const ids = [...JSON.stringify(payload ?? {}).matchAll(/"toolUseId":"([^"]+)"/g)].map((m) => m[1]);
    expect(new Set(ids).size).toBe(2); // actual: 1
  });

  it("FABLE-5: thinking is not degraded on a broken final tool turn", async () => {
    const { assistant, results } = foreignToolTurn(["call_x"]);
    const ctx: Context = { messages: [{ role: "user", content: "go", timestamp: Date.now() - 3000 }, assistant, ...results] };
    const model = getModel("anthropic", "claude-fable-5" as any);
    const payload = await capture(model, ctx, { reasoning: "high" });
    expect(payload?.thinking?.type).toBe("disabled"); // actual: "adaptive"
  });
});
```

Observed output:

```
FAIL  PR380 repro > BEDROCK: long foreign tool ids still collide
AssertionError: expected 1 to be 2

FAIL  PR380 repro > FABLE-5: thinking is not degraded on a broken final tool turn
AssertionError: expected 'adaptive' to be 'disabled'

Test Files  1 failed (1)
     Tests  2 failed (2)
```

### Measured thinking-degradation matrix

Same broken-final-tool-turn history, `reasoning: "high"`, payload captured via `onPayload`:

| model | `thinking` | `output_config` | guard fired |
|---|---|---|---|
| `claude-sonnet-4-5` | `{type:"disabled"}` | — | yes |
| `claude-opus-4-7` | `{type:"disabled"}` | — | yes |
| `claude-fable-5` | `{type:"adaptive",display:"summarized"}` | `{effort:"high"}` | **no** |

Note the guard's `model.thinkingLevelMap?.off !== null` clause does **not** exclude fable-5 — its map on the `anthropic` provider is `{"xhigh":"xhigh","max":"max"}`, so `off` is `undefined`, and `undefined !== null` is `true`. What actually skips it is `cannotDisableThinking` upstream keeping `params.thinking` adaptive. That means the clause is load-bearing for a different (Bedrock `"off": null`) set of entries than it appears to target, which is worth a comment either way.

### Bedrock collision detail

```
toolUseIds: [
  "call_AAAA…(59 A's)",
  "call_AAAA…(59 A's)",
  "call_AAAA…(59 A's)",
  "call_AAAA…(59 A's)"
]
count: 4   unique: 1
```

Four emitted ids (2 `tool_use` + 2 matching `toolResult`) degenerate to a single value — exactly the "duplicate `tool_use` ids ... with corrupted tool-result pairing" that `packages/ai/src/changes.md` says is now fixed.

### Suggested remediation

1. Extract the hashed normalizer into a shared helper (it only needs `shortHash`) and call it from `anthropic-messages.ts`, `bedrock-converse-stream.ts`, and `google-shared.ts`; add a Bedrock case to `anthropic-cross-model-history.test.ts`.
2. In the new `buildParams` guard, reuse the existing `cannotDisableThinking` policy (pin `output_config.effort = "low"`, as the `thinkingEnabled === false` branch already does at `anthropic-messages.ts:1720-1727`) instead of assigning `{type:"disabled"}` unconditionally; add a `claude-fable-5` test.
3. Assert the final payload (not just the captured options object) for `openai-codex-responses` and `azure-openai-responses`, and confirm `reasoningEffort:"minimal"` survives `resolveReasoningEffort` on catalog entries that map `minimal → null`.
4. Narrow `changes.md` to the adapters actually patched, or widen the patch.

### Housekeeping

- **GitGuardian is a false positive.** The flagged `***` in `.agents/skills/senpi-qa/scripts/compaction-remote-qa.mjs` is a mock-server literal that already exists on `main` (I fetched the file at `ref=main` to confirm); it surfaced here only via merge commit `13ffb955`. No rotation needed — please allowlist so it stops gating PRs.
- `Check and test` was still `pending` at review time; the local "147 files / 1202 tests" claim is not independently visible yet.

Verdict unchanged: **request changes**, driven by (1). The rest is coverage tightening. The signature-guard work is good and I would take it today.

packages/ai/src/api/anthropic-messages.ts
realsigridjin: Fixed in f3a05e051. The degradation path is centralized in disableThinkingForRequest: normal models send disabled; adaptive models that reject disabled omit thinking and pin output_config.effort=low. It explicitly clears any prior effort first. The fable request-level regression and the normal non-trigger case both pass, and the change log states this is final-turn-only request shaping.

packages/coding-agent/src/core/extensions/builtin/compaction/speculative.ts
realsigridjin: Fixed in f3a05e051 + 9b25944ee. reasoningSummary:null now becomes summary:off for Codex and an omitted summary field for direct/Azure Responses. Compaction selects the cheapest catalog-supported effort, so minimal:null rows use low instead of silently restoring provider defaults. The new wire-level test captures final OpenAI Responses, Codex, Azure, and Kimi payloads through the real compaction path.

packages/ai/src/api/anthropic-messages.ts
realsigridjin: Fixed in f3a05e051. Anthropic Messages, Bedrock, and Google now share packages/ai/src/utils/tool-call-id.ts; the helper preserves short legal IDs and uses a readable prefix plus a full-ID hash for long foreign IDs. The exact shared-64-character-prefix regression now passes for Bedrock and Google (provider focused suite: 17/17).

packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
realsigridjin: Pinned in 9b25944ee and clarified in 79e163eb1. A duplicate model_select for the same API/provider/id/base URL/window reuses the existing job instead of aborting it; the regression asserts the first provider AbortSignal remains un-aborted. The code comment now states that getContextUsage is intentionally pre-switch accounting measured against the new smaller window.

packages/ai/src/api/openai-responses-shared.ts
realsigridjin: Documented in 79e163eb1: signed foreign blocks with no text are intentionally dropped. The existing regression asserts no reasoning item and no output item for that case.

packages/ai/src/changes.md
realsigridjin: Updated in f3a05e051 + 9b25944ee: the AI notes now document the shared cross-adapter normalizer and adaptive low-effort fallback; the compaction notes document cheapest legal effort, provider-native summary disabling, wire-level coverage, and same-model idempotency.

realsigridjin:
@code-yeongyu All six review threads are addressed, replied to, and resolved on final head `79e163eb1`.

Key fixes: shared collision-safe IDs across Anthropic/Bedrock/Google; legal Fable fallback (`thinking` omitted + `effort: low`); final OpenAI/Codex/Azure/Kimi compaction payload coverage with summaries truly off; duplicate shrink warm-start reuse. Latest `main` (`9196919b8`) is merged.

Post-merge evidence: AI 1461 passed / 793 gated, compaction 255 passed, root check exit 0, real senpi mock loop 41/41, CLI smoke 8/8, auth unchanged, zero paid calls. Please re-review the replacement head.

code-yeongyu (commented):
Independent re-review of head `79e163eb1`. I re-verified every open thread against the code and re-ran validation locally; all concerns from the earlier change requests are resolved.

1. **Bedrock/Google tool-call-ID collisions — resolved.** All three adapters now share `packages/ai/src/utils/tool-call-id.ts` (readable prefix + `shortHash` of the full ID). The new Bedrock and Google regressions pin the shared-64-char-prefix case end to end, including tool-result pairing (`toolResult` IDs equal the `toolUse` set). Verified locally: 4 focused files / 21 tests pass.
2. **Thinking degradation on `cannotDisableThinking` models — resolved.** `disableThinkingForRequest` centralizes both policies: explicit `{type:"disabled"}` where legal, and `thinking` omitted + `output_config.effort:"low"` for adaptive families. The `claude-fable-5` fallback and the non-trigger case are both pinned in `anthropic-cross-model-history.test.ts`; final-turn-only scope is documented in the code comment and `changes.md`.
3. **`summarizationReasoningOptions` coverage — resolved.** `summarization-reasoning-payload.test.ts` captures final wire payloads through the real compaction path: OpenAI Responses omits the summary field, Codex sends `summary:"off"`, Azure uses its cheapest catalog-legal effort, Kimi K3 falls back to `low` when catalog `minimal` is `null`. I also confirmed `minimal:null` rows are skipped by the level scan (e.g. `deepseek-v4-pro` resolves to `high`, not a dropped option + provider default).
4. **Warm start vs invalidate race — resolved.** The same-model identity check (api/provider/id/baseUrl/contextWindow) preserves the in-flight job, and `startSpeculativeCompaction`'s existing-job guard plus the generation bump make a duplicate `model_select` a no-op. `model-shrink-speculative-warmstart.test.ts` asserts the first AbortSignal stays un-aborted with exactly one provider call. The pre-switch-usage comment is in place.
5. **`changes.md` scoping — resolved.** Entries now document the shared normalizer, the adaptive low-effort fallback, cheapest-legal-effort selection, and the wire-level coverage.
6. **GitGuardian finding — confirmed false positive** (`sk-ant-mock` literal in `compaction-remote-qa.mjs`, pre-existing on `main`, surfaced via merge). Allowlisting is a separate infra task; not a merge blocker. CI is green on this head.

My own verification on `79e163eb1`: root `npm run check` exit 0; `packages/ai` full suite green (1407 passed / 803 gated skips; the single initial failure was `codex-apply-patch-wire-schema.test.ts` failing to resolve `@earendil-works/pi-tui` because `packages/tui/dist` was unbuilt in my fresh worktree — environment-only, passes after building); coding-agent compaction suites 27 files / 206 tests passed; `git merge-tree` vs `origin/main` (`9196919b8`) clean; senpi-qa mock loop 41/41 and CLI smoke 8/8, evidence under `local-ignore/qa-evidence/20260726-pr380-review/`.

One non-blocking observation: for reasoning models whose catalog `thinkingLevelMap` is *silent* on `minimal` (`undefined` rather than `null` — e.g. `azure-openai-responses/gpt-5.4`, `moonshotai/kimi-k2.7-code`), the cheapest-effort scan sends `reasoningEffort:"minimal"` raw. This matches existing `getSupportedThinkingLevels` semantics (undefined == supported), so it is consistent with user-turn behavior today, but a catalog pass marking `minimal: null` where providers reject it would make the policy exact. Not a blocker.

From my side: **merge-safe**.

code-yeongyu (approved):
Final maintainer review: my two prior CHANGES_REQUESTED rounds are fully addressed at 79e163eb1 (shared collision-safe tool-call IDs across Anthropic/Bedrock/Google, legal adaptive thinking fallback, final wire payloads pinned for OpenAI/Codex/Azure/Kimi, warm-start dedupe). Review lane re-verified each thread plus full-diff pass: zero new defects; check exit 0, ai suite 1407 pass, compaction suites 206 pass, zero-token QA. This head adds only the ai/changes.md collision resolution with #330. Merging at maximum speed per owner instruction.

realsigridjin:
@code-yeongyu The conflict resolution has been applied. Please re-review based on the latest head.

code-yeongyu:
Re-reviewed and merged. The final head 74c387ffb (= 79e163eb1 + an ai/changes.md-only resolution against the #330 entry) was re-reviewed in full: all six earlier review threads verified fixed, zero new defects, `npm run check` exit 0, packages/ai suite 1407 passing, compaction suites 206 passing, zero-token real-CLI QA. Merged as d5f651cd8bdf0ff1ab7c9d5e6a1e7abde30ea7c1. A post-merge integration validation of current main (with #341/#328/#380 combined) is running now; I'll follow up only if it surfaces anything.

PR #412, opened 2026-07-27
fix(ai): omit Codex reasoning summary instead of sending unsupported "off"

DevNewbie1826:
## Bug

Compaction fails on Codex (GPT) models:

```
Codex error: [ReasoningSummaryParam] [reasoning.summary] [invalid_enum_value]
Invalid value: 'off'. Supported values are: 'concise', 'detailed', and 'auto'.
```

Repro: run senpi with a Codex model until the builtin compaction fires (blocking compaction at the context limit). Every summarization request is rejected with a 400 and the compaction extension throws.

## Root cause

The builtin compaction passes `reasoningSummary: null` to keep summarization turns cheap (`packages/coding-agent/src/core/extensions/builtin/compaction/speculative.ts`, `summarizationReasoningOptions()`). The sibling adapters treat `null` as "omit the field" (`openai-responses.ts`, `azure-openai-responses.ts`), but `openai-codex-responses.ts` `buildRequestBody()` mapped it to the literal string `"off"` — a value the Codex backend does not accept. Upstream pi-mono never maps null to `"off"` either, so this was a fork-local regression. Normal turns are unaffected because they never pass `reasoningSummary: null` (undefined falls back to `"auto"`).

## Fix

`reasoningSummary: null` now omits the `summary` field from `body.reasoning` in both `buildRequestBody()` branches (explicit effort and the thinking-off fallback), matching the sibling adapters and upstream. An extension cannot fix this — the invalid value is produced inside the wire adapter below every extension hook. `packages/ai/src/changes.md` documents the change and the expected merge-conflict zone.

## Tests and QA

- `packages/ai/test/openai-responses-thinking-matrix.test.ts`: two new cases pin both branches — `reasoning.summary` is absent when `reasoningSummary` is null (failed RED with `summary: "off"` before the fix, GREEN after); full file 17/17.
- Related codex-touching suites: 68 passed / 33 skipped (live-gated).
- `npm run check`: pass (biome, tsgo, browser-smoke, pinned-deps, shrinkwrap).
- senpi-qa: mock-loop 42/42, cli-smoke 8/8, harness self-check 9/9; evidence under `local-ignore/qa-evidence/20260727-codex-reasoning-summary-null/` (local only, gitignored).

---
## Summary by cubic
Fixes Codex compaction failures by omitting `reasoning.summary` when `reasoningSummary` is null and normalizing legacy values. Restores compaction on Codex models and aligns behavior with other adapters.

- **Bug Fixes**
  - In `packages/ai/src/api/openai-codex-responses.ts` `buildRequestBody()`, use `buildCodexReasoning()` to omit `reasoning.summary` on null and normalize legacy values: `"off"` → omitted, `"on"` → `"auto"`; applied on both explicit-effort and thinking-off paths.
  - Tests in `packages/ai/test/openai-responses-thinking-matrix.test.ts` cover `null`, `"off"`, `"on"`, and `"auto"` across both paths.

- **Refactors**
  - Extracted Codex reasoning types and helpers to `packages/ai/src/api/openai-codex-responses/reasoning.ts` (`CodexReasoningSummaryInput`, `normalizeCodexReasoningSummary()`, `buildCodexReasoning()`), and updated option/body types accordingly.

<sup>Written for commit bbefc27c3790246686ba5834c7c4a3882f36faa7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DevNewbie1826:
Supporting evidence for the null-as-omit semantics, from the official Codex CLI (codex-rs), the reference client for this backend:

- The `ReasoningSummary` enum is `Auto / Concise / Detailed / None` — no `"off"` value exists on the wire: https://github.com/openai/codex/blob/main/codex-rs/protocol/src/config_types.rs#L47-L54
- `build_reasoning()` encodes "summary disabled" by **omitting the field** (`(supports_reasoning_summary_parameter && summary != ReasoningSummaryConfig::None).then_some(summary)`), never by sending a string sentinel: https://github.com/openai/codex/blob/main/codex-rs/core/src/client.rs#L817-L835
- codex-rs's own compaction path (`summarize_input`) sends `Reasoning { effort, summary: None, .. }` — summary omitted: https://github.com/openai/codex/blob/main/codex-rs/core/src/client.rs#L720-L725

So the null-as-omit behavior in this PR matches the canonical encoding for "no reasoning summary" on this backend, including for compaction requests specifically.

effortprogrammer:
related issue: https://github.com/code-yeongyu/senpi/issues/415
looking forward for merging

code-yeongyu:
Your adapter fix and the codex-rs research here are correct — thank you for both, especially the `build_reasoning()` citation showing omission is the real wire encoding.

CI could not go green on this branch because the same bug had a second copy in the test suite: `packages/coding-agent/test/compaction/summarization-reasoning-payload.test.ts` asserted `reasoning.summary === "off"` for the Codex case, pinning the exact value the backend rejects (and contradicting its own case name plus the sibling OpenAI/Azure cases, which expect the field absent).

```
FAIL  test/compaction/summarization-reasoning-payload.test.ts
  > uses the cheapest legal effort without a summary for 'Codex Responses'
AssertionError: expected undefined to be 'off'
```

I opened #416, which carries your three commits unmodified and adds one commit unpinning that assertion, so the fix can actually land. Full `npm run check` + compaction suite + senpi-qa evidence are on that PR.

PR #453, opened 2026-07-29
fix(coding-agent): escalate repeated provider-stream stalls to the fallback chain

code-yeongyu:
## Problem

A Discord user reported 5h+ senpi sessions becoming permanently wedged: mid-session turns started failing with `Request timed out.`, and every reopen of the session died with `Idle timeout waiting for provider stream after 300000ms` on the goal-continuation turn while brand-new sessions worked fine.

Forensics on the donated session JSONL (`019fa8da-43ad-70b7-b01b-8f34f4d907f2`, 1,922 records, 6 compactions, ~143k-token / ~1MB replayed context on a custom `codex-lb` provider, api `openai-responses`):

- Records 1838/1906: `Request timed out.` with `responseId: null` - no HTTP response at all.
- Record 1919: `Idle timeout waiting for provider stream after 300000ms`, `responseId: null`, empty content - the request was accepted but the provider emitted **zero events for the entire idle budget**.
- The same context streamed fine minutes earlier (record 1910 got a `responseId` + `thinking_start`), so this is a hung provider/gateway under a large session payload, not a malformed request. `transformMessages` already drops errored/aborted assistants from replay, and both error shapes are correctly classified retryable.

The senpi-side defect: a stall retry replays an **identical payload** with the **full idle budget** again. Against a hung provider that is `(1 + maxRetries) * httpIdleTimeoutMs` (~20 minutes at defaults) of opaque dead air per turn before the fallback chain is ever consulted. A field session on this machine (2026-07-21, anthropic-api) confirms the machinery: one 300s stall, then three same-model probes, then surrender.

## Fix

- `packages/ai/src/utils/retry.ts`: new `isProviderStreamStallError()` classifier for the agent-loop idle-watchdog failure shape (still retryable; classification unchanged).
- `packages/coding-agent/src/core/agent-session.ts`: the transient retry branch tracks consecutive stalls. The **second consecutive stall escalates to the fallback chain immediately** (same `tryFallback("transient")` path as budget exhaustion); without a chain the retry loop ends instead of replaying the doomed payload for the remaining budget. Fast non-stall failures reset the streak, so the cheap same-model probes that heal transient network blips are preserved. Fallback switches and fresh retry loops start with a clean streak.

Worst-case wedge drops from ~20 minutes to `~2 * httpIdleTimeoutMs`, and sessions with a configured chain now recover automatically.

## Evidence (TDD, RED -> GREEN)

[...]

PR #458, opened 2026-07-29
fix(agent): preserve input across provider idle retries

realsigridjin:
## Summary

- preserve queued steering/follow-up input across provider-timeout retries
- cap only the retry continuation's provider idle timeout at 30 seconds
- restore the configured timeout automatically for later ordinary turns
- retain queued input after terminal error/abort instead of draining it into a failed run

## Root cause

A donated Senpi session showed that `monitor` creation returned immediately, but the following provider stream became silent. Senpi then retried the timeout with the full configured 300-second idle budget. Steering entered during that retry was drained into the doomed provider request, so the session remained legitimately `Working` through repeated long waits and the user's `continue` could finish without an answer.

This change defers the retry continuation's initial steering poll and applies a continuation-scoped timeout override. It does not mutate the normal configured timeout.

This is complementary to #453: fallback escalation handles repeated stalls, while this PR bounds each retry and preserves user input across the stalled retry.

## Verification

- RED captured on the unchanged implementation: retry timeout sequence remained `300000, 300000, 300000`, and queued steering entered the failed retry
- focused coding-agent regression suites: 28/28 passed
- agent-core suite: 37/37 passed
- deterministic real CLI/RPC fixture: 4/4 passed
  - silent retry completed in approximately 30 seconds
  - queued steering ran exactly once after retry recovery
  - a later ordinary turn ran longer than 30 seconds, proving timeout restoration
- targeted Biome checks passed with no suppressions, skips, or focused tests
- QA process, local server, port, sandbox, and temporary runtime resources were cleaned

The regression fixture uses a synthetic local SSE provider; no donated session content or credentials are included.

---
## Summary by cubic
Preserves user input during provider idle/transport timeout retries and scopes queue deferral and timeout overrides to the retry’s first provider request, then restores configured bounds after the first event. Adds precise timeout classifiers and serialized continuation admission so retries don’t swallow input, double-run, or misreport success.

- **Bug Fixes**
  - Defer queued steering/follow-up only from a continuation’s first provider request; after a terminal `error`/`aborted` run, park queued input until a later admitted prompt or explicit continue.
  - Cap both idle and stream-start bounds on the retry’s first provider request; resume configured `timeoutMs` and `streamStartTimeoutMs` after the first event and never re-enable disabled guards.
  - Keep consuming the same retry budget across consecutive transport timeouts reported as `aborted`; reset only on genuine success.
  - Serialize retry-continuation admission, treat a concurrent low-level prompt as a benign takeover, and wait to emit “settled” until core is idle.
  - Add `isProviderStreamStallError()` and `isProviderTimeoutError()` to classify exact provider stream stalls and transport timeouts without matching incidental timeout text.

- **New Features**
  - `Agent.continue(options)` and `continueWithQueuedMessages(options)` accept one-shot `deferQueuedMessages`, `timeoutMs`, and `streamStartTimeoutMs` overrides for the first provider request; queue-first recovery keeps the selected queued message as input while still applying first-request timeout overrides.
  - Coding-agent settings add `retry.provider.streamStartTimeoutMs` and `retry.provider.streamRetryTimeoutMs` (30s default; `0` disables) to cap first-request retries without mutating defaults; caps clamp only enabled guards and restore configured timeouts for later requests.

<sup>Written for commit ea498fa22ab62f614f9c3da83b2189de276d133a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/458?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -124,6 +124,13 @@ export interface AgentOptions {
 	abortServerSideFallback?: boolean;
 }
 
+export interface AgentContinuationOptions {
+	/** Keep queued steering and follow-up input out of the continuation's first provider request. */
+	deferQueuedMessages?: boolean;
+	/** Override the provider stream idle timeout for this continuation only. */
+	timeoutMs?: number;
+}
+
 class PendingMessageQueue {
 	private messages: AgentMessage[] = [];
 	private clearGeneration = 0;
@@ -408,8 +415,11 @@ export class Agent {
 		await this.continue();
 	}
 
-	/** Continue from the current transcript. The last message must be a user or tool-result message. */
-	async continue(): Promise<void> {
+	/**
+	 * Continue from the current transcript. The last message must be a user or tool-result message.
+	 * Queued input can be deferred until the continuation produces its first response.
+	 */
+	async continue(options: AgentContinuationOptions = {}): Promise<void> {
 		if (this.activeRun) {
 			throw new Error("Agent is already processing. Wait for completion before continuing.");
 		}
@@ -435,7 +445,7 @@ export class Agent {
 			throw new Error("Cannot continue from message role: assistant");
 		}
 
-		await this.runContinuation();
+		await this.runContinuation(options);
 	}
 
 	private normalizePromptInput(
@@ -473,11 +483,14 @@ export class Agent {
 		});
 	}
 
-	private async runContinuation(): Promise<void> {
+	private async runContinuation(options: AgentContinuationOptions = {}): Promise<void> {
 		await this.runWithLifecycle(async (signal) => {
 			await runAgentLoopContinue(
 				this.createContextSnapshot(),
-				this.createLoopConfig(),
+				this.createLoopConfig({
+					skipInitialSteeringPoll: options.deferQueuedMessages,
+					timeoutMs: options.timeoutMs,
+				}),
 				(event) => this.processEvents(event),
 				signal,
 				this.streamFunction,
@@ -493,7 +506,7 @@ export class Agent {
 		};
 	}
 
-	private createLoopConfig(options: { skipInitialSteeringPoll?: boolean } = {}): AgentLoopConfig {
+	private createLoopConfig(options: { skipInitialSteeringPoll?: boolean; timeoutMs?: number } = {}): AgentLoopConfig {
 		let skipInitialSteeringPoll = options.skipInitialSteeringPoll === true;
 		let steeringQueueGeneration = this.steeringQueue.getClearGeneration();
 		let followUpQueueGeneration = this.followUpQueue.getClearGeneration();
@@ -505,7 +518,7 @@ export class Agent {
 			onResponse: this.onResponse,
 			transport: this.transport,
 			thinkingBudgets: this.thinkingBudgets,
-			timeoutMs: this.timeoutMs,
+			timeoutMs: options.timeoutMs ?? this.timeoutMs,
 			streamStartTimeoutMs: this.streamStartTimeoutMs,
 			maxRetryDelayMs: this.maxRetryDelayMs,
 			abortServerSideFallback: this.abortServerSideFallback,
@@ -570,6 +583,7 @@ export class Agent {
 			while (
 				!abortController.signal.aborted &&
 				!this.activeRun?.suppressQueuedMessageDrain &&
+				this.canDrainQueuedMessagesAfterRun() &&
 				this.hasQueuedMessages()
 			) {
 				await this.runQueuedMessagesAfterAgentEnd(abortController.signal);
@@ -579,6 +593,14 @@ export class Agent {
 		} finally {
 			this.finishRun();
 		}
+	}
+
+	private canDrainQueuedMessagesAfterRun(): boolean {
+		const lastMessage = this._state.messages[this._state.messages.length - 1];
+		return (
+			lastMessage?.role !== "assistant" ||
+			(lastMessage.stopReason !== "error" && lastMessage.stopReason !== "aborted")
+		);
 	}
 
 	private async runQueuedMessagesAfterAgentEnd(signal: AbortSignal): Promise<void> {
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -125,10 +125,12 @@ export interface AgentOptions {
 }
 
 export interface AgentContinuationOptions {
-	/** Keep queued steering and follow-up input out of the continuation's first provider request. */
+	/** Keep queued steering and follow-up input out of the continuation's first provider request only. */
 	deferQueuedMessages?: boolean;
-	/** Override the provider stream idle timeout for this continuation only. */
+	/** Override the provider stream idle timeout for the continuation's first provider request only. */
 	timeoutMs?: number;
+	/** Override the provider stream-start timeout for the continuation's first provider request only. */
+	streamStartTimeoutMs?: number;
 }
 
 class PendingMessageQueue {
@@ -349,6 +351,8 @@ export class Agent {
 	/**
 	 * Keep queued steering and follow-up messages for an external owner after
 	 * this run reaches agent_end, without changing the active abort signal.
+	 * This is ownership suppression for one active run; terminal error/abort
+	 * parking is a separate stop-reason policy enforced by the run lifecycle.
 	 */
 	suppressQueuedMessageDrain(): void {
 		if (this.activeRun) {
@@ -389,35 +393,47 @@ export class Agent {
 		await this.runPromptMessages(messages);
 	}
 
-	/** Continue by delivering queued input first when a compaction leaves custom context at the tail. */
-	async continueWithQueuedMessages(): Promise<void> {
+	/**
+	 * Continue by delivering queued input first when a compaction leaves custom context at the tail.
+	 * Queue-first recovery takes precedence over `deferQueuedMessages`: the selected queued message is
+	 * the continuation input, while timeout overrides still apply to its first provider request.
+	 */
+	async continueWithQueuedMessages(options: AgentContinuationOptions = {}): Promise<void> {
 		if (this.activeRun) {
 			throw new Error("Agent is already processing. Wait for completion before continuing.");
 		}
 
 		if (this._state.messages[this._state.messages.length - 1]?.role === "assistant") {
-			await this.continue();
+			await this.continue(options);
 			return;
 		}
 
 		const queuedSteering = this.steeringQueue.drain();
 		if (queuedSteering.length > 0) {
-			await this.runPromptMessages(queuedSteering, { skipInitialSteeringPoll: true });
+			await this.runPromptMessages(queuedSteering, {
+				skipInitialSteeringPoll: true,
+				initialRequestTimeoutMs: options.timeoutMs,
+				initialRequestStreamStartTimeoutMs: options.streamStartTimeoutMs,
+			});
 			return;
 		}
 
 		const queuedFollowUps = this.followUpQueue.drain();
 		if (queuedFollowUps.length > 0) {
-			await this.runPromptMessages(queuedFollowUps);
+			await this.runPromptMessages(queuedFollowUps, {
+				initialRequestTimeoutMs: options.timeoutMs,
+				initialRequestStreamStartTimeoutMs: options.streamStartTimeoutMs,
+			});
 			return;
 		}
 
-		await this.continue();
+		await this.continue(options);
 	}
 
 	/**
 	 * Continue from the current transcript. The last message must be a user or tool-result message.
-	 * Queued input can be deferred until the continuation produces its first response.
+	 * Queue deferral and timeout overrides apply only to the first provider request; later requests in
+	 * the same run and later runs use the configured Agent defaults.
 	 */
 	async continue(options: AgentContinuationOptions = {}): Promise<void> {
 		if (this.activeRun) {
@@ -432,13 +448,20 @@ export class Agent {
 		if (lastMessage.role === "assistant") {
 			const queuedSteering = this.steeringQueue.drain();
 			if (queuedSteering.length > 0) {
-				await this.runPromptMessages(queuedSteering, { skipInitialSteeringPoll: true });
+				await this.runPromptMessages(queuedSteering, {
+					skipInitialSteeringPoll: true,
+					initialRequestTimeoutMs: options.timeoutMs,
+					initialRequestStreamStartTimeoutMs: options.streamStartTimeoutMs,
+				});
 				return;
 			}
 
 			const queuedFollowUps = this.followUpQueue.drain();
 			if (queuedFollowUps.length > 0) {
-				await this.runPromptMessages(queuedFollowUps);
+				await this.runPromptMessages(queuedFollowUps, {
+					initialRequestTimeoutMs: options.timeoutMs,
+					initialRequestStreamStartTimeoutMs: options.streamStartTimeoutMs,
+				});
 				return;
 			}
 
@@ -469,7 +492,11 @@ export class Agent {
 
 	private async runPromptMessages(
 		messages: AgentMessage[],
-		options: { skipInitialSteeringPoll?: boolean } = {},
+		options: {
+			skipInitialSteeringPoll?: boolean;
+			initialRequestTimeoutMs?: number;
+			initialRequestStreamStartTimeoutMs?: number;
+		} = {},
 	): Promise<void> {
 		await this.runWithLifecycle(async (signal) => {
 			await runAgentLoop(
@@ -489,7 +516,8 @@ export class Agent {
 				this.createContextSnapshot(),
 				this.createLoopConfig({
 					skipInitialSteeringPoll: options.deferQueuedMessages,
-					timeoutMs: options.timeoutMs,
+					initialRequestTimeoutMs: options.timeoutMs,
+					initialRequestStreamStartTimeoutMs: options.streamStartTimeoutMs,
 				}),
 				(event) => this.processEvents(event),
 				signal,
@@ -506,7 +534,13 @@ export class Agent {
 		};
 	}
 
-	private createLoopConfig(options: { skipInitialSteeringPoll?: boolean; timeoutMs?: number } = {}): AgentLoopConfig {
+	private createLoopConfig(
+		options: {
+			skipInitialSteeringPoll?: boolean;
+			initialRequestTimeoutMs?: number;
+			initialRequestStreamStartTimeoutMs?: number;
+		} = {},
+	): AgentLoopConfig {
 		let skipInitialSteeringPoll = options.skipInitialSteeringPoll === true;
 		let steeringQueueGeneration = this.steeringQueue.getClearGeneration();
 		let followUpQueueGeneration = this.followUpQueue.getClearGeneration();
@@ -518,8 +552,10 @@ export class Agent {
 			onResponse: this.onResponse,
 			transport: this.transport,
 			thinkingBudgets: this.thinkingBudgets,
-			timeoutMs: options.timeoutMs ?? this.timeoutMs,
+			timeoutMs: this.timeoutMs,
 			streamStartTimeoutMs: this.streamStartTimeoutMs,
+			initialRequestTimeoutMs: options.initialRequestTimeoutMs,
+			initialRequestStreamStartTimeoutMs: options.initialRequestStreamStartTimeoutMs,
 			maxRetryDelayMs: this.maxRetryDelayMs,
 			abortServerSideFallback: this.abortServerSideFallback,
 			toolExecution: this.toolExecution,
@@ -580,6 +616,9 @@ export class Agent {
 
 		try {
 			await executor(abortController.signal);
+			// A provider-returned terminal error/abort has no safe implicit owner for
+			// queued work. Park it until an external retry/compaction owner continues,
+			// or until a later admitted prompt drains it through the normal queue poll.
 			while (
 				!abortController.signal.aborted &&
 				!this.activeRun?.suppressQueuedMessageDrain &&
--- a/packages/agent/src/changes.md
+++ b/packages/agent/src/changes.md
@@ -24,6 +24,17 @@
 - `types.ts`
 - `../test/agent-loop-stream-start-timeout.test.ts`
 
+## 2026-07-29 - Continuation-scoped queue and timeout controls
+
+### What changed and why
+
+- `Agent.continue()` accepts continuation-only options that defer queued input until the first provider response and
+  override the stream idle timeout without mutating the agent's configured default.
+- The core run lifecycle retains queued steering and follow-up input after terminal error or abort responses instead
+  of racing an asynchronous session-level retry controller and starting an unintended provider request.
+- Coding-agent retries use these controls after a silent provider stream so a doomed retry cannot consume newly
+  queued user input and a later ordinary turn automatically returns to the configured timeout.
+
 ## 2026-07-27 - End classifier-refused turns before tool execution
 
 ### What changed and why
--- a/packages/agent/src/changes.md
+++ b/packages/agent/src/changes.md
@@ -28,12 +28,35 @@
 
 ### What changed and why
 
-- `Agent.continue()` accepts continuation-only options that defer queued input until the first provider response and
-  override the stream idle timeout without mutating the agent's configured default.
-- The core run lifecycle retains queued steering and follow-up input after terminal error or abort responses instead
-  of racing an asynchronous session-level retry controller and starting an unintended provider request.
+- `Agent.continue()` and `continueWithQueuedMessages()` accept continuation-only options that defer queued input from
+  the first provider request and override both stream idle and stream-start bounds for that request without mutating
+  the agent's configured defaults. Later requests in the same run restore the configured bounds; after the first
+  retry event, the configured idle timeout also governs inter-event gaps so healthy silent reasoning is not capped.
+- Queue-first recompaction recovery takes precedence over deferral: the selected queued message is the continuation
+  input, while first-request timeout overrides still apply.
+- The core run lifecycle intentionally parks queued steering and follow-up input after terminal error or abort
+  responses until an external retry/compaction owner or a later admitted prompt consumes it. This stop-reason policy
+  is distinct from `suppressQueuedMessageDrain()`, which transfers one active run's post-`agent_end` ownership.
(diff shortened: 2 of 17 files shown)


packages/coding-agent/src/core/agent-session.ts
```diff
@@ -5431,9 +5437,16 @@ export class AgentSession {
 		// Agent core would otherwise drain these queues before the continuation
 		// revalidates its provider admission. The scheduled continuation owns them
 		// until it either starts or reports a terminal admission failure.
+		const deferQueuedMessages = PROVIDER_TIMEOUT_PATTERN.test(errorMessage);
+		const timeoutMs = deferQueuedMessages
```
code-yeongyu: **[blocker] The "bounded retry budget" this PR claims does not exist for aborted transport timeouts — the counter resets every cycle, and the 30s cap makes the resulting infinite loop spin 10x faster.**

`_isRetryableError` treats aborted messages with timeout text as retryable (`agent-session.ts:5160-5162`), but the `message_end` handler resets the retry budget for any `stopReason !== "error"` (`agent-session.ts:1450-1466`) — `aborted` qualifies. Each aborted `Request timed out.` retry therefore: emits a false `auto_retry_end { success: true }`, resets `_retryAttempt` to 0, and re-enters retry at attempt 1. `maxRetries` is never reached; with this PR's 30s cap the session loops indefinitely at ~30s intervals instead of ~300s. This directly contradicts `packages/coding-agent/src/changes.md`'s "use a bounded retry budget" and reproduces the donated session's stuck-`Working` symptom in a new shape.

The counter-reset is pre-existing, but this PR's stated contract makes it in-scope: fix the success predicate so only genuinely successful responses reset the budget, e.g.
```ts
const succeeded = assistantMsg.stopReason !== "error" && assistantMsg.stopReason !== "aborted";
```
(refusal handling per existing `isClassifierRefusal` semantics), and add a regression: more than `maxRetries` consecutive aborted `Request timed out.` responses must produce exactly one terminal `auto_retry_end { success: false }` — no infinite loop, no false success events.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -4363,12 +4365,16 @@ export class AgentSession {
...
-	private async _continueAgentAfterCurrentRun(): Promise<void> {
+	private async _continueAgentAfterCurrentRun(
+		options: { deferQueuedMessages?: boolean; timeoutMs?: number } = {},
+	): Promise<void> {
 		await this.agent.waitForIdle();
 		await this._revalidateScheduledContinuationAdmission();
 		const useQueuedContinuation = this._scheduledContinuationRecompacted;
 		try {
-			if (useQueuedContinuation) {
+			if (options.deferQueuedMessages || options.timeoutMs !== undefined) {
```
code-yeongyu: **[major] Retry options take precedence over recompaction semantics — role-dependent misordering, defer/timeout contract loss, and a retry that can fail to launch.**

`if (options.deferQueuedMessages || options.timeoutMs !== undefined)` wins over `useQueuedContinuation`, but `_revalidateScheduledContinuationAdmission()` (one line above) can compact and set `_scheduledContinuationRecompacted`. Compaction rebuilds agent state from canonical session context (`agent-session.ts:3813-3838`; `sessionEntryToContextMessages` at `session-manager.ts:447-457` has no stopReason filter, so a retained assistant tail is realistic). The outcome then depends on the recompacted transcript tail:

- **non-assistant tail** (compaction summary/custom message): `continue(options)` retries from that context first and defers the queue — reversing the established queue-first recompaction order that `continueWithQueuedMessages()` exists to guarantee;
- **assistant tail + queued input**: `Agent.continue()`'s last-assistant branch (`packages/agent/src/agent.ts:432-443`) **drains queued steering into `runPromptMessages()`** — consuming exactly the input this PR promises to preserve — and **discards `options.timeoutMs`**, so the doomed retry runs uncapped;
- **assistant tail + empty queues**: `continue()` throws `Cannot continue from message role: assistant` → the retry dies as a `continuation_error`.

This matters for the donated session's shape (6 compactions, ~143k context): near-full context is exactly when admission revalidation compacts during a timeout retry.

Suggested fix: **compose, don't prioritize** — let the queue-aware continuation accept continuation-scoped timeout options (e.g. `continueWithQueuedMessages(options)`), keep `useQueuedContinuation` precedence, define explicitly what `deferQueuedMessages` means when recompaction mandates a queue-first continuation, and never let the assistant-tail fallback silently ignore defer or drop the timeout override. Add coverage: timeout retry + successful revalidation compaction, with both queued and empty queues.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -148,6 +148,8 @@ import { createToolDefinitionFromAgentTool } from "./tools/tool-definition-wrapp
 import { addUsageToTotals, createUsageTotals } from "./usage-totals.ts";
 
 const TURN_RETRY_SUPPRESSION_PREFIX = "senpi:no-turn-retry:";
+const PROVIDER_TIMEOUT_PATTERN = /(?:timed? out|timeout)/i;
```
code-yeongyu: **[major] Timeout classification keys on incidental error text and is duplicated — gate on provenance or at least anchor to the known shapes.**

Two concrete problems:
1. `/(?:timed? out|timeout)/i` matches any error whose text merely contains "timeout". Assistant error messages here are built from arbitrary thrown errors (stream setup, context transforms, hooks — e.g. an extension error saying `extension timed out` reaches this path via the loop's error-to-assistant-message conversion), so the 30s cap + defer policy can fire for failures that have nothing to do with a silent provider stream.
2. `_isRetryableError` re-declares the same regex inline at `agent-session.ts:5161` instead of reusing `PROVIDER_TIMEOUT_PATTERN` — the two copies can drift on the next upstream sync.

Preferred fix: carry structured provenance (a diagnostic set by `StreamIdleTimeoutError` / `StreamStartTimeoutError` / normalized transport timeout handling) and gate defer/cap on it. Acceptable interim: one shared, anchored pattern for the forensically observed shapes —
```ts
/^(?:Request timed out\.?|Idle timeout waiting for provider stream after \d+ms|Provider stream start timed out after \d+ms)$/i
```
— used by BOTH sites, plus negative tests for `Command timed out after ...` / `MCP server ... timed out`.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -148,6 +148,8 @@ import { createToolDefinitionFromAgentTool } from "./tools/tool-definition-wrapp
...
+const PROVIDER_STREAM_IDLE_RETRY_TIMEOUT_CAP_MS = 30_000;
```
code-yeongyu: **[major] The hardcoded 30s cap needs settings wiring, must not re-enable a disabled idle guard, and can kill healthy silently-reasoning retries.**

Three policy problems with `PROVIDER_STREAM_IDLE_RETRY_TIMEOUT_CAP_MS = 30_000`:

1. **Not configurable.** PR #451 wired `streamStartTimeoutMs` through `ProviderRetrySettings` (`settings-manager.ts:36`, getter at `:1137`, SDK wiring at `sdk.ts:377`, interactive at `interactive-mode.ts:5102`). This cap changes retry behavior for every provider yet has no knob. Wire it the same way (default 30s is fine) with documented `0`/disabled semantics.
2. **`this.agent.timeoutMs ?? CAP` turns a disabled guard back on.** If the user disabled the idle timeout (`agent.timeoutMs === undefined`), timeout retries now get a forced 30s guard despite the explicit setting. `undefined` must not be conflated with "use the cap".
3. **Healthy-silence risk.** Once the retry stream emits its first event, every subsequent inter-event gap is bound by the cap (agent-loop `readTimeoutMs` selection) for the rest of the continuation. Reasoning models that legitimately go >30s between events will fail retries the configured 300s budget would have recovered — and combined with the retry-budget reset bug (see blocker at line 5441) that becomes a fast infinite loop. Prefer a liveness-scoped bound: cap until the retry shows life, then restore the configured timeout. Add a fake-timer test with two provider events >30s apart inside a capped retry pinning the chosen policy.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -5431,9 +5437,16 @@ export class AgentSession {
...
+		const timeoutMs = deferQueuedMessages
+			? Math.min(
+					this.agent.timeoutMs ?? PROVIDER_STREAM_IDLE_RETRY_TIMEOUT_CAP_MS,
+					PROVIDER_STREAM_IDLE_RETRY_TIMEOUT_CAP_MS,
+				)
+			: undefined;
 		this.agent.suppressQueuedMessageDrain();
 		setTimeout(() => {
-			void this._continueAgentAfterCurrentRun().catch(async (error) => {
+			void this._continueAgentAfterCurrentRun({ deferQueuedMessages, timeoutMs }).catch(async (error) => {
```
code-yeongyu: **[major] Retry continuation admission is not atomic — a concurrent low-level prompt corrupts settlement state (reproduced at PR head).**

The retry path schedules bare `setTimeout(..., 0)` work without claiming `_sessionWorkBarrier` (unlike `_scheduleContinuationAfterCurrentEvent`, which does `this._sessionWorkBarrier.begin()` before scheduling). Inside `_continueAgentAfterCurrentRun`, `waitForIdle() → _revalidateScheduledContinuationAdmission() → continue()` is non-atomic: an `Agent.prompt()` landing between idle-wait and `continue()` makes `continue()` throw `Agent is already processing...`, which the generic catch converts into a terminal `continuation_error`, resolves retry state, and calls `_emitAgentSettled()` **while the winning agent run is still active**. Reproduced during this review by pausing revalidation and racing `session.agent.prompt()`: result was `continuation_error` + `session.isStreaming = false` with `agent.state.isStreaming = true` — session and agent state disagree and the automatic retry is silently abandoned.

The interactive path is shielded (session stays streaming; TUI submissions steer), but the public low-level `session.agent.prompt()` surface reaches this.

Suggested fix: route retry continuations through the same barrier-owned scheduler as `_scheduleContinuationAfterCurrentEvent()` (parameterized with the continuation options), and treat an active-run collision as a benign takeover (re-check/skip) rather than a terminal continuation failure; never emit `agent_settled` while `agent.state.isStreaming` is true.

packages/agent/src/agent.ts
```diff
@@ -505,7 +518,7 @@ export class Agent {
 			onResponse: this.onResponse,
 			transport: this.transport,
 			thinkingBudgets: this.thinkingBudgets,
-			timeoutMs: this.timeoutMs,
+			timeoutMs: options.timeoutMs ?? this.timeoutMs,
```
code-yeongyu: **[major] The continuation timeout override is mis-scoped: it does not bound the stalled start it was built for, and it DOES bind the deferred user's answering request.**

Two scope gaps in `timeoutMs: options.timeoutMs ?? this.timeoutMs` (with `streamStartTimeoutMs: this.streamStartTimeoutMs` unchanged on the next line):

1. **No-first-byte retries are not capped at 30s.** The stream reader prefers `streamStartTimeoutMs` until the first event (`agent-loop.ts` `readTimeoutMs` selection), and coding-agent defaults it to **90s**. A retry of the exact donated `Idle timeout waiting for provider stream`/stalled-start shape still waits 90s before first byte — not the advertised 30s. The new test asserts only the option value passed to the faux provider and never exercises the real event reader, so it misses this. Override both bounds for the retry continuation (e.g. continuation-scoped `streamStartTimeoutMs` = min(configured, cap)) and add a no-first-event test that expires at the cap.
2. **The cap leaks into the deferred input's answering request.** The override lives on the loop config for the whole continuation run; after the retry succeeds, the same run's next steering poll drains the deferred input and answers it under the 30s cap — contradicting `changes.md`'s "later ordinary turns retain the configured provider idle timeout". Scope the override to the retry's first provider request (or end the continuation after its successful response and let the post-run drain answer the queue on a fresh default config). Extend the recovery regression to assert the timeout sequence `[300000, 30000, 30000, 300000]` — today call 4 gets 30000.

packages/agent/src/agent.ts
```diff
@@ -581,6 +595,14 @@ export class Agent {
 		}
 	}
 
+	private canDrainQueuedMessagesAfterRun(): boolean {
```
code-yeongyu, line 637: **[minor] `canDrainQueuedMessagesAfterRun()` is policy inferred from stop reasons, not ownership — scope it or document + test the new parking behavior.**

For the provider-retry race this guard is redundant: `_createRetryPromiseForAgentEnd()` already claims the queue synchronously via `suppressQueuedMessageDrain()` at the awaited `agent_end` boundary (`agent-session.ts:1162-1186`), and signal-aborted runs were already excluded by `!abortController.signal.aborted`. What the guard newly changes is the **no-owner** cases: retry-disabled sessions, non-retryable terminal errors, and provider-returned `aborted` messages with no caller abort — queued input there is now silently parked until a later admitted prompt drains it (appended after that prompt, not before). The two mechanisms can also disagree (terminal error + no owner: guard blocks / suppression false; successful assistant + required compaction: suppression blocks / guard allows).

If the parking is intended (the PR body says so), please: (a) add a comment on the guard naming why signal-abort exclusion isn't sufficient and who eventually consumes the retained queue; (b) extend `suppressQueuedMessageDrain()`'s JSDoc to distinguish ownership-suppression from this stop-reason policy; (c) add direct `packages/agent` tests for error/aborted terminal results x steering/follow-up retention. Alternatively, scope retention to runs with an explicit owner (retry/compaction) instead of inferring from every terminal stop reason.

packages/agent/src/agent.ts
```diff
@@ -124,6 +124,13 @@ export interface AgentOptions {
 	abortServerSideFallback?: boolean;
 }
 
+export interface AgentContinuationOptions {
+	/** Keep queued steering and follow-up input out of the continuation's first provider request. */
```
code-yeongyu: **[minor] Public API widening is fine but under-documented and untested in its owning package.**

`AgentContinuationOptions` is exported via `export * from "./agent.ts"` and the call-scoped design is right (better than mutating public fields). Gaps:

1. The JSDoc should state the defer scope precisely: it protects the **first provider request only** (one-shot `skipInitialSteeringPoll`); later polls in the same continuation drain normally. That is acceptable — a post-success poll means the provider proved alive, and in-flight requests can't be retro-injected — but the contract should be written down.
2. `packages/agent/README.md` still documents only bare `await agent.continue();` — document both options and their scope.
3. No owning-package test covers the new options or `canDrainQueuedMessagesAfterRun()`. Add direct `packages/agent/test/agent.test.ts` cases: first-request queue deferral; call-scoped timeout override + restoration on the next run; error/abort queue retention for steering and follow-up.

packages/agent/src/changes.md
```diff
@@ -24,6 +24,17 @@
 - `types.ts`
 - `../test/agent-loop-stream-start-timeout.test.ts`
 
+## 2026-07-29 - Continuation-scoped queue and timeout controls
```
code-yeongyu, line 27: **[minor] changes.md entries understate the merge-conflict surface and omit the fork's entry structure.**

This new agent entry stops after "What changed and why" — neighboring entries also list touched files and expected merge-conflict zones, and the 2026-07-23 entry already labels this same active-run lifecycle region **MEDIUM**. The coding-agent entry claims `LOW: one retry-continuation option branch` while the diff actually touches `_continueAgentAfterCurrentRun()`, the retry controller in high-churn `agent-session.ts`, and public continuation/lifecycle behavior in `agent.ts`. Please add the missing sections here (files + conflict zones, `agent.ts` continuation/config/lifecycle at least MEDIUM) and bump the coding-agent entry to name both scheduled-continuation and retry-controller regions at MEDIUM. Note for whoever addresses this review: both entries also currently describe timeout scoping the code does not deliver (see the timeout-scope comments) — keep them in sync with the fixes.

packages/coding-agent/test/suite/regressions/provider-idle-recovery.test.ts
```diff
@@ -0,0 +1,119 @@
...
+			resolveSteeringResponse = resolve;
+		});
+		const unsubscribe = harness.session.subscribe((event) => {
+			if (event.type === "auto_retry_start" && queuedSteering === undefined) {
+				queuedSteering = harness.session.steer("continue");
+			}
+			if (
+				event.type === "message_end" &&
+				event.message.role === "assistant" &&
+				JSON.stringify(event.message.content).includes("steering request recovered")
```
code-yeongyu: **[nit] Serialized-substring assertions are fragile.**

`JSON.stringify(event.message.content).includes("steering request recovered")` and the `requests[N].toContain("continue")` checks match against the whole serialized message array — common text like `"continue"` can hit unrelated fields, and the assertions depend on serialization shape rather than the message contract. The modified characterization test already projects user-message texts properly; do the same here (exact user-text arrays per call, e.g. via `getUserTexts`-style projection per request), and resolve the completion signal from an exact assistant-text match or a scripted faux response callback.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -5431,9 +5437,16 @@ export class AgentSession {
 		// Agent core would otherwise drain these queues before the continuation
```
code-yeongyu: **[nit] This queue-ownership comment is now stale.**

"The scheduled continuation owns them until it either starts or reports a terminal admission failure" — for timeout retries the queue stays deferred *after* the continuation starts, until its first response; and the new core guard already blocks post-run draining after error/abort tails, so "Agent core would otherwise drain these queues" is no longer universally true. Rewrite to describe the two distinct protections: lifecycle-drain suppression while retry admission settles, and the first-poll skip for timeout-class retries.

code-yeongyu (changes requested):
## Review: request changes

This PR correctly identifies the failure mode from the donated session (silent provider stream -> full-length 300s retries -> queued `continue` drained into the doomed request) and the core mechanism — continuation-scoped defer + timeout override — is the right shape. The happy-path fix works: the focused suites pass (20/20), the agent suite passes (296+ passed), `npm run check` is clean, and the two new tests are NOT tautological (reverting the production change fails both).

However, a 5-lane review (goal fidelity, correctness/concurrency, code quality, hands-on QA with mutation probes, context mining) found one blocker and five majors that defeat the PR's own stated contract in realistic paths:

- **Blocker — the "bounded retry budget" doesn't bound**: aborted transport timeouts reset `_retryAttempt` on every cycle (false `auto_retry_end success:true`), so `maxRetries` is never reached and the 30s cap makes the infinite loop spin 10x faster.
- **Majors**: recompacted-retry path bypasses the defer/timeout contract (role-dependent misordering, uncapped drain of the preserved input, or a dead retry); `streamStartTimeoutMs` (90s) is not capped so the advertised 30s doesn't apply before first byte, while the cap DOES leak into the deferred input's answering request; `?? 30_000` re-enables an explicitly disabled idle guard; the hardcoded cap needs settings wiring (#451 precedent); retry continuation admission is non-atomic (reproduced: `continuation_error` + session settled while the agent is still streaming); timeout classification keys on a broad substring pattern duplicated in two places.

Each inline comment carries severity, evidence with file:line traces, and a concrete suggested fix plus the regression test to add. Hands-on QA in an isolated worktree confirmed the baseline green (focused suites 20/20, packages/agent 296 passed, abort-coverage suites 25/25, `npm run check` clean) and mutation probes proved the new tests are load-bearing: removing the drain guard and altering the 30s cap each turned the suite RED. (third probe — classifier neutralization — confirmation pending; a surprise there becomes a follow-up comment.)

Relationship to #453: context-mining found no textual hunk conflicts between the two diffs in `agent-session.ts`; the semantic coupling is exactly the duplicated timeout classifier called out inline, so merge order is safe once the classifier is shared.

Once the blocker and majors are addressed (minors/nits at author's discretion with a reply), I'm happy to merge — the underlying diagnosis and direction are solid, and this fixes a real, user-reported wedge.

packages/coding-agent/src/core/agent-session.ts
code-yeongyu: Fixed in ed7c3acf7e10d739a55846701e88816926c69dcb and 28e2379a7d738280c3fb1e7da07ecbf2a6404ef1. `continueWithQueuedMessages(options)` now composes queue-first recompaction with first-request timeout options; scheduled retries give recompaction precedence and retire reconstructed failed assistant tails. The failing-first matrix covers assistant/non-assistant tails with queued and empty queues.

packages/coding-agent/src/core/agent-session.ts
code-yeongyu: Fixed in dd848c60f9b059800a683fe16bbd07075636cf49 with first-request scoping support from ed7c3acf7e10d739a55846701e88816926c69dcb. Added documented `retry.provider.streamRetryTimeoutMs` (default 30s, 0 disables), clamp-only semantics that never re-enable disabled guards, stream-start capping, and restoration of the configured idle bound after the first event. Tests cover custom/disabled settings and a healthy >30s inter-event gap.

packages/agent/src/changes.md
code-yeongyu: Addressed in ed7c3acf7e10d739a55846701e88816926c69dcb and 28e2379a7d738280c3fb1e7da07ecbf2a6404ef1. The agent entry now includes files, extension-boundary rationale, and MEDIUM continuation/lifecycle conflict zones. The coding-agent entry now names retry-controller plus scheduled-continuation admission at MEDIUM and describes the delivered timeout/recompaction/concurrency behavior.

packages/agent/src/agent.ts
code-yeongyu: Addressed in ed7c3acf7e10d739a55846701e88816926c69dcb. JSDoc and README now specify first-provider-request scope, queue-first recompaction semantics, and timeout restoration. Owning-package tests cover first-request queue deferral, idle/start override restoration, healthy post-first-event silence, and terminal error/abort retention for both native queues.

packages/coding-agent/test/suite/regressions/provider-idle-recovery.test.ts
code-yeongyu: Fixed in dd848c60f9b059800a683fe16bbd07075636cf49 (with type-safe assertion cleanup in 3e7ba9232468d453f86d34ab36f814c466115cb8). The regression projects exact user-text arrays per provider call and resolves completion from an exact assistant-text match; serialized substring checks were removed.

packages/coding-agent/src/core/agent-session.ts
code-yeongyu: Fixed in 7c9834b01b355c7e9467b6f376ccc437cc8d1108 and wired in dd848c60f9b059800a683fe16bbd07075636cf49. The shared anchored `isProviderTimeoutError`/`isProviderStreamStallError` classifiers cover the two watchdog messages plus exact `Request timed out` transport shapes, including aborted transport responses, while rejecting command/MCP/extension lookalikes. Both retry admission and retry-option selection use the shared classifier.

packages/coding-agent/src/core/agent-session.ts
code-yeongyu: Fixed in e464464dc12d2701bc8a1269b2037ffe9a5f8e35. Aborted transport timeouts no longer count as successful responses or reset `_retryAttempt`; the new failing-first regression proves maxRetries exhaustion emits one terminal `auto_retry_end { success: false }`, no false success events, and retains queued steering.

packages/agent/src/agent.ts
code-yeongyu: Fixed in ed7c3acf7e10d739a55846701e88816926c69dcb and dd848c60f9b059800a683fe16bbd07075636cf49. Continuation overrides now cover both stream-start and provider timeout options for only the first provider request; later requests in the same run restore configured values. The real no-first-event test expires at 30s instead of 90s, and the four-call recovery regression pins `[300000, 30000, 30000, 300000]`.

packages/coding-agent/src/core/agent-session.ts
code-yeongyu: Fixed in dd848c60f9b059800a683fe16bbd07075636cf49 and 28e2379a7d738280c3fb1e7da07ecbf2a6404ef1. The stale comment now distinguishes lifecycle-drain suppression during admission from first-poll deferral for known provider timeouts, and the retry path is routed through the shared barrier-owned scheduler.

packages/agent/src/agent.ts
code-yeongyu: Addressed in ed7c3acf7e10d739a55846701e88816926c69dcb. Parking is intentional and now documented separately from ownership suppression: terminal provider error/abort queues remain parked until an external owner or later admitted prompt drains them. Direct agent tests cover error/abort x steering/follow-up, and a mutation removing the guard makes all four cases fail.

packages/coding-agent/src/core/agent-session.ts
code-yeongyu: Fixed in 28e2379a7d738280c3fb1e7da07ecbf2a6404ef1. Retry continuations now use the barrier-owned scheduled-continuation path. Admission rechecks active Agent state, treats a concurrent low-level prompt as a benign takeover, and waits for Agent idle before publishing `agent_settled`. The failing-first race regression asserts no `continuation_error`, no premature settlement, and consistent session/Agent streaming state.

code-yeongyu (approved):
## Re-review: approve

All 11 findings from my previous review are addressed at head 3e7ba923 — verified independently in an isolated worktree, not just from the author's report:

- Focused suites 91/91, packages/agent 303/304, ai retry 33/33, settings 61/61, root `npm run check` clean.
- The 16 new tests cover the full requested matrix: aborted-budget exhaustion (blocker), first-request-only override scope, liveness-restored idle timeout, no-first-event cap, disabled-guard preservation, negative classifier cases, benign concurrent-prompt takeover, in-flight-abort queue retention.
- Shared anchored classifier now lives in `packages/ai` (`isProviderTimeoutError`/`isProviderStreamStallError`), used by both call sites — also resolves the semantic-coupling risk with #453.
- Settings wiring (`retry.provider.streamRetryTimeoutMs`, default 30s, `0` disables) follows the #451 precedent, documented in docs/settings.md.
- changes.md entries restructured with MEDIUM conflict zones.

Merging with a merge commit once CI completes.

PR #472, opened 2026-07-29
fix(coding-agent): treat headers-only providers as configured

eddieparc:
## Problem

`ProviderConfigSchema` accepts `headers` as a standalone field, and `applyModelsJson` advertises it as sufficient configuration through its own error string:

> `Provider ${providerId}: must specify "baseUrl", "headers", "extraBody", "compat", "modelOverrides", or "models".`

But `configuredRequestAuthStatus` never consults `headers`. In `packages/coding-agent/src/core/provider-composer.ts` it begins:

```ts
const value = configuredApiKey(config, extension);
if (value === undefined) return undefined;
```

So a provider whose auth is fully expressed by `headers` — for example a gateway taking `x-api-key` — is reported unconfigured. Consumers then surface a misleading failure: `model-registry.ts` reports `No API key found for "<provider>"` even though the provider is correctly configured and would authenticate fine.

## Reproduction

A provider with `headers` and no `apiKey`:

```ts
configuredRequestAuthStatus(undefined, { headers: { "x-api-key": "tok" } })
// before: undefined  (reported unconfigured)
// after:  { configured: true }
```

## Fix

Treat a provider whose auth is fully expressed by `headers` as configured. The change is confined to the auth-status path.

Deliberately unchanged:

- Providers with neither `apiKey` nor `headers` still report unconfigured, so the fix is not over-broadened.
- `authHeader` semantics are untouched.

## Testing

- `npm run check` passes.
- New test file covers four cases: headers-only reports configured, apiKey-only still configured, neither reports unconfigured, and `authHeader` behaviour is unaffected.

Note on `npm test`: on my machine four watcher/timing suites fail nondeterministically, all in files this branch does not touch (`config-reload-watch-engine`, `footer-data-provider`, `mcp/catalog-cache`, `app-server-thread-handlers-archive`). Across three identical runs both the count and identity of the failures changed, each failing at ~5s — vitest's default `testTimeout` expiring while awaiting a real `fs.watch` event, while the tests' own guards are 10s. They reproduce on an unmodified checkout, so they look host/load-dependent rather than related to this change. Happy to look into that separately if it is useful.

---
## Summary by cubic
Treats providers configured with credential headers (like `x-api-key`) as authenticated and usable. OpenAI-compatible requests now honor header-only auth without synthesizing a Bearer token, including for Responses over WebSocket.

- **Bug Fixes**
  - `configuredRequestAuthStatus` now detects credential-like `headers` and reports `source` as `models_json_headers` or `extension_headers` (also `environment`/`models_json_command` for templated values). Metadata-only or empty headers remain unconfigured; `authHeader` still requires a resolved API key.
  - OpenAI request path: `resolveOpenAIClientAuth` enables static credential headers; the SDK suppresses its default `Authorization` when a credential header exists or `Authorization: null` is set (HTTP and Responses WebSocket).
  - Core/RPC: header-only auth flows through `checkAuth()`/`getAuth()`; RPC `AuthStatus.source` includes `models_json_headers` and `extension_headers`. Shared detection lives in `auth/headers.ts` and is reused across auth discovery and request adapters. Tests cover header classification, OpenAI request construction, provider status/runtime auth, OAuth/API-key regressions, and `authHeader` error cases.

<sup>Written for commit 405af727c1a7b46ec0e62ce84f378a728e83083a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/472?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/provider-composer.ts
+++ b/packages/coding-agent/src/core/provider-composer.ts
@@ -1,10 +1,6 @@
 import {
 	type Api,
-	type ApiKeyAuth,
 	type AssistantMessageEventStream,
-	type AuthContext,
-	type AuthInteraction,
-	type AuthResult,
 	type Context,
 	type Credential,
 	getApiProvider,
@@ -12,7 +8,6 @@ import {
 	getToolCallFormat,
 	lazyStream,
 	type Model,
-	type ModelAuth,
 	type OAuthAuth,
 	type OAuthCredentials,
 	type OAuthLoginCallbacks,
@@ -25,12 +20,13 @@ import {
 	wrapStreamWithToolCallMiddleware,
 } from "@earendil-works/pi-ai";
 import type { ModelConfig, ModelsJsonModel, ModelsJsonModelOverride, ModelsJsonProvider } from "./model-config.ts";
+import { composeApiKeyAuth, configuredApiKey, configuredHeaders, withConfiguredAuth } from "./provider-api-key-auth.ts";
+import { configuredHeaderAuthStatus, type HeaderAuthStatusSource } from "./provider-header-auth.ts";
 import {
 	clearConfigValueCache,
 	getConfigValueEnvVarNames,
 	isCommandConfigValue,
 	isConfigValueConfigured,
-	resolveConfigValueOrThrow,
 	resolveHeadersOrThrow,
 } from "./resolve-config-value.ts";
 
@@ -80,7 +76,14 @@ export interface ProviderConfigInput {
 
 export type AuthStatus = {
 	configured: boolean;
-	source?: "stored" | "runtime" | "environment" | "fallback" | "models_json_key" | "models_json_command";
+	source?:
+		| "stored"
+		| "runtime"
+		| "environment"
+		| "fallback"
+		| "models_json_key"
+		| "models_json_command"
+		| HeaderAuthStatusSource;
 	label?: string;
 };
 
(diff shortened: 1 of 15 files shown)


packages/coding-agent/src/core/provider-composer.ts
```diff
@@ -609,7 +609,11 @@ export function configuredRequestAuthStatus(
 	extension: ProviderConfigInput | undefined,
 ): AuthStatus | undefined {
 	const value = configuredApiKey(config, extension);
-	if (value === undefined) return undefined;
+	if (value === undefined) {
+		const headers = configuredHeaders(config, extension);
+		if (!headers || Object.keys(headers).length === 0) return undefined;
+		return { configured: true, source: extension?.headers !== undefined ? "fallback" : "models_json_key" };
```
code-yeongyu: [sisyphus-bot] [major] This return equates any non-empty request-header map with resolved authentication. `headers` is also used for ordinary metadata such as `User-Agent`, tracing, or routing, and it can overlay an OAuth provider before the user has logged in. More importantly, the actual auth pipeline still disagrees: `composeApiKeyAuth().check/resolve` returns `undefined` when there is no raw, stored, or inherited API key, so `Models.getAvailable/getAuth` and `ModelRuntime.prepareRequest` still treat a headers-only provider as unconfigured. Meanwhile `getProviderAuthStatus`, the synchronous `ModelRegistry.getAvailable` fallback, TUI/RPC status payloads, and `registerProvider`'s provisional snapshot now report it configured. That can expose unusable models and mark logged-out OAuth providers as authenticated. Please fix the shared auth-resolution/check seam so an explicitly supported headers-only credential yields a real `AuthResult`/`AuthCheck` (or add an explicit opt-in for auth-bearing headers) instead of changing only this status helper.

packages/coding-agent/test/provider-composer-headers-auth.test.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+const emptyAuthContext: AuthContext = {
+	env: async () => undefined,
+	fileExists: async () => false,
+};
+
+describe("configured request auth status", () => {
+	it("reports a headers-only provider as configured", () => {
+		expect(requestAuthStatus("headers-only")).toEqual({ configured: true, source: "models_json_key" });
```
code-yeongyu: [sisyphus-bot] [major] This regression test calls only `configuredRequestAuthStatus`, so it stays green even when the user-visible consumers contradict it. A composed headers-only provider still fails `checkAuth/getAuth` and request preparation, while non-auth headers and OAuth overlays are mislabeled as configured. Please exercise the real `ModelRuntime`/`ModelRegistry` path: status plus `checkAuth`/`getAuth`/availability and a zero-network stream stub. Include at least an auth-bearing header, a non-auth header, empty headers, extension-provided headers, and an OAuth provider with headers but no stored credential; the test should fail whenever reported status and executable auth diverge.

packages/coding-agent/src/core/provider-composer.ts
```diff
@@ -609,7 +609,11 @@ export function configuredRequestAuthStatus(
 	extension: ProviderConfigInput | undefined,
 ): AuthStatus | undefined {
 	const value = configuredApiKey(config, extension);
-	if (value === undefined) return undefined;
+	if (value === undefined) {
```
code-yeongyu, line 506: [sisyphus-bot] [minor] This is a fork-specific runtime behavior change under `packages/coding-agent/src/core`, but the PR does not update `packages/coding-agent/src/changes.md`. The nearest `AGENTS.md` requires that record for source changes. Please add an entry covering the behavior, rationale, focused coverage, and expected upstream merge-conflict zone.

code-yeongyu (commented):
[sisyphus-bot] Blocking review: this patch changes only the configured-status projection, not the auth resolution used by availability checks and request execution. Arbitrary request headers and logged-out OAuth overlays can therefore be reported configured while the real getAuth/stream path remains unconfigured, and the tests exercise the helper rather than those consumers. I found 2 major issues and 1 required repository-discipline issue. I attempted the scoped test run, but this worktree's installed dependencies are incomplete (`vitest` is absent and the installed `typebox` package is missing its build entry), so the findings below are based on a full consumer code-path audit rather than a green local run. Structural verdict: redesign the shared auth-resolution seam before merge.

packages/coding-agent/test/provider-composer-headers-auth.test.ts
code-yeongyu: [sisyphus-bot] Resolved in 7b2e34c27. The helper-only tests were replaced with real ModelRegistry/ModelRuntime coverage for models.json x-api-key auth, metadata-only headers, empty headers, extension Authorization headers, OAuth plus metadata headers, API-key preservation, registry availability, checkAuth, getAuth, and zero-network runtime streaming. packages/ai also exercises real OpenAI Completions and Responses request construction and proves x-api-key is sent without synthetic Authorization. Focused results: 46/46 AI tests and 98/98 coding-agent tests; the runtime probe and real CLI QA also passed.

packages/coding-agent/src/core/provider-composer.ts
code-yeongyu: [sisyphus-bot] Resolved in 7b2e34c27. Added package-level source change records in packages/ai/src/changes.md and packages/coding-agent/src/changes.md, plus the applicable nested packages/coding-agent/src/modes/rpc/changes.md entry. Each record includes behavior, rationale, focused coverage, and expected merge-conflict zones.

packages/coding-agent/src/core/provider-composer.ts
code-yeongyu: [sisyphus-bot] Resolved in 7b2e34c27. Static provider headers now use a shared, case-insensitive credential-header classifier; metadata-only and empty headers remain unconfigured, including OAuth providers without stored credentials. The composed API-key auth path now returns a real header-only AuthCheck/AuthResult, so registry availability, async runtime availability, getAuth, streaming, TUI status, and RPC status agree. OpenAI-compatible adapters also accept x-api-key-style auth without synthesizing a bearer header. Evidence: focused AI suite 46/46, coding-agent suite 98/98, root npm run check green, and real CLI QA passed with evidence at local-ignore/qa-evidence/20260729-pr472/.

PR #490, opened 2026-07-29
fix(coding-agent): retry shell-command credential resolution before giving up

### LINKED ISSUE #489: Bug: transient credential-command failures hard-error eject the active model with zero resolution retries
## Summary

When a provider's API key is resolved through a shell command (e.g. `!omp token kimi-code`), **a single transient failure of that subprocess is treated as an unrecoverable auth failure**. `agent-session` classifies it as hard-error eligible and ejects the active model into the fallback chain with **zero retries of the resolution itself** — even though re-running the same command a second later succeeds.

## Impact — five full fallback cascades in one day

From `~/.senpi/agent/logs/fallback.log` (session `019faccb-3e7c-7307-8b19-2c7fb9e77b5c`, 2026-07-29): five cascades (06:30, 06:31, 06:44, 09:12, 09:40 UTC). Every one opened with:

```
cooldown_noted kimi-code/k3
  "API key auth failed for provider kimi-code: Failed to resolve API key
   for provider \"kimi-code\" from shell command: omp token kimi-code"
fallback_applied kimi-code/k3 → ccapi/claude-opus-5 (reason: hard-error)
```

The same flake also hit `omp token anthropic` mid-chain (06:45, 09:40 ×2), so `ccapi/*` entries hard-error ejected as well. Each cascade then walked into already rate-limited fallbacks (`429 "All tokens rate limited"`), turning a ~1s resolver hiccup into a multi-minute provider-switch storm that ended the turn on a different model than configured. Manual `/fallback` clears and cooldown-expiry reverts (~10 min) were the only recovery.

## Root cause

`packages/coding-agent/src/core/resolve-config-value.ts`:

1. `executeCommandUncached` runs the credential command **exactly once** (10s `execSync` timeout) and returns `undefined` on any spawn error / non-zero exit / timeout.
2. `resolveConfigValueOrThrow` turns that into `Failed to resolve API key … from shell command: …`.
3. That error reaches the assistant error path, where `_isHardErrorFallbackEligible` admits it as `hard-error` → `tryFallback("hard-error")` — immediate ejection + cooldown note. Nothing ever re-runs the command before giving up on the provider.

Why the subprocess flakes (measured): `omp token <provider>` is a cold-started bun CLI reading a SQLite auth store — 16 sampled invocations took **1.5–4.0s each** (p50 ≈ 2.3s). Under system load (dev servers, concurrent agent processes polling the same store) an invocation intermittently fails while the credential itself is perfectly healthy; running the identical command seconds later succeeds.

## Proposed fix

Retry the shell command a small bounded number of times with short backoff before reporting resolution failure: **3 attempts, 250ms then 1000ms backoff**. Genuinely missing credentials still fail fast (~1.4s added worst case, and the command failures themselves return quickly), while transient broker/spawn hiccups are absorbed at the source instead of becoming model ejections.

Note: `core/resolve-config-value.ts` also exists upstream in `badlogic/pi-mono` (same `executeCommandUncached`), so the same fix likely belongs upstream as well.

## Environment

- senpi 2026.7.28-3 (npm `@code-yeongyu/senpi`)
- macOS 25.5.0 arm64, node v24.14.1
- credential broker command: `omp` (oh-my-pi) 17.1.8

unclejobs-ai:
## Summary

Fixes #489.

Credential helper commands (`!omp token kimi-code`, `!omp token anthropic`, …) were executed **exactly once**; any transient subprocess failure (spawn error, non-zero exit, 10s timeout) propagated as `Failed to resolve API key … from shell command`, which `agent-session` classifies as hard-error eligible — ejecting the active model into the fallback chain with **zero retries of the resolution itself**. Observed in production: five full fallback cascades in a single day, each triggered by a one-off `omp token` hiccup while the credential was healthy (evidence in #489).

This PR retries the command at the resolution layer — **3 attempts, 250ms then 1000ms backoff** — so a transient broker/spawn blip is absorbed where it happens instead of becoming a provider ejection.

## Changes

- **`src/core/resolve-config-value.ts`**
  - `executeCommandUncached` now retries up to `COMMAND_EXECUTION_MAX_ATTEMPTS = 3` with `COMMAND_EXECUTION_BACKOFF_MS = [250, 1000]` between attempts.
  - The original single-execution body moves verbatim to `executeCommandOnce`; platform branching (win32 configured-shell → default-shell ENOENT fallback) is unchanged and retried as a unit.
  - `sleepBlocking` uses `Atomics.wait` — call sites are already synchronous (`execSync`/`spawnSync`), so a blocking sleep introduces no new async surface.
  - Failure semantics unchanged: after the final attempt the same `undefined` is returned (and cached by the cached path) exactly as before.
- **`test/resolve-config-value.test.ts`**
  - New: *retries a transiently failing command until it succeeds* — counter-file command failing attempts 1–2, succeeding on attempt 3 (pins both the recovery and the exact attempt count).
  - New: *gives up retrying after a bounded number of attempts* — always-failing command resolves `undefined` after exactly 3 executions.
  - Updated: *caches successful and failed commands* counter arithmetic — one failing resolve now costs 3 executions before `undefined` is cached; the second resolve still hits the cache without executing.
- **`src/changes.md`**: incident entry per repo convention, including upstream-sync conflict notes.

## Verification

- `npx vitest --run test/resolve-config-value.test.ts` → **12/12 pass**.
- Full `packages/coding-agent` vitest suite (after `npm run build`, per CI order) → passes except **14 environment-dependent failures** in vision/provider suite tests that call real endpoints without credentials here (`401 Missing Authentication header`). A/B-verified pre-existing: `test/suite/look-at-extension.test.ts` fails **5/7 identically with the unmodified `main` version** of `resolve-config-value.ts`.
- `npx tsgo -p tsconfig.build.json --noEmit` → no diagnostics in touched files (pre-existing TS7006s in `src/package-manager-cli.ts` on `main` are unrelated).
- Pre-commit hooks pass on the commit.

## Why 3 attempts / this backoff

Measured `omp token` cold-start: 1.5–4.0s per invocation (bun startup + SQLite auth store), with failures clustering in short bursts under load. 3 attempts with 0.25s/1s gaps ride out a burst while adding a bounded worst case of ~1.25s backoff plus two fast-failing executions before the identical hard-error path runs — so genuinely missing credentials still fail fast and through the same code path as today.

## Notes

- `core/resolve-config-value.ts` also exists upstream in `badlogic/pi-mono` (same `executeCommandUncached`); happy to send the equivalent patch upstream if you'd like it there too.
- Scope is deliberately the resolution layer, not the fallback classifier: hard-error ejection remains correct for *persistently* unresolvable credentials — they just shouldn't be convicted on a single sample.

---
## Summary by cubic
Retries shell-command credential resolution (e.g., `!omp token …`) up to 3 times with backoff so transient failures don’t eject providers into the fallback chain. Fixes #489; genuine missing credentials still fail the same way as before.

- **Bug Fixes**
  - `executeCommandUncached` retries 3 times with 250ms then 1000ms backoff; original logic moved to `executeCommandOnce`; blocking sleep via `Atomics.wait`.
  - Platform shell branching and caching are unchanged; successful and `undefined` results are cached as before.
  - Tests: added transient-recovery and bounded-failure cases; updated cache count expectations (a failed resolve now executes 3 times before caching).
  - Tests: `model-registry` expectations updated — each request-time resolve performs the bounded 3-attempt retry; two independent resolves execute a failing command 6 times total.
  - Changelog: added incident entry; resolved ordering with the Bun launcher self‑update entry after syncing `main`.

<sup>Written for commit c0f9e363f2e0a7211247ac9bda5a21ae56c3c8b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (commented):
VERDICT: MERGE_OK

Reviewed at `66cfa464a`.

No blocking defects remain.

### Review findings

- **The retry boundary is correct.** `executeCommandOnce()` contains the complete platform decision, including the Windows configured-shell attempt and its default-shell fallback, and `executeCommandUncached()` retries that whole unit (`packages/coding-agent/src/core/resolve-config-value.ts:218-235`). This does not accidentally multiply only one half of the Windows path.
- **The blocking sleep is acceptable in this existing synchronous seam.** Every execution path already blocks in `execSync`/`spawnSync`; the auth-facing callers invoke this synchronous resolver from their credential-resolution step rather than introducing a previously asynchronous command runner. I also executed `Atomics.wait` on the main thread under both the repository's Node and Bun runtimes; both timed out normally rather than throwing.
- **Permanent failures are intentionally retried, but the cost is bounded and justified here.** A typoed command or deterministic non-zero exit adds the fixed 1.25s backoff plus execution time. The normal env-less path caches that failure process-wide. Distinguishing a shell-level `127`/non-zero status as definitively non-transient would be unreliable for arbitrary credential brokers and would undermine the issue's explicit requirement to retry spawn errors, non-zero exits, and timeouts.
- **The tests are deterministic.** The new cases assert observable attempt counts through counter files and contain no wall-clock assertions, sleeps, polling, or timing thresholds (`packages/coding-agent/test/resolve-config-value.test.ts:107-124`). The focused suite proves recovery on attempt 3, bounded failure at 3 attempts, and failed-result caching.

### Fixes applied during review

- Merged current `origin/main` without rebasing or rewriting history.
- Resolved `packages/coding-agent/src/changes.md` by preserving the PR incident entry and both newer main entries.
- Removed a pre-existing stray `|||||||` conflict marker from that changelog during the semantic merge resolution.
- Pushed merge commit `66cfa464a` to the contributor branch.

### Verification

- `npx vitest run test/resolve-config-value.test.ts` - **12/12 passed**
- `npm run check` - **passed**
- Focused Biome check on the changed implementation/test files - **passed**
- Senpi QA common harness self-check - **9/9 passed**
- Senpi mock-loop `transient-recover` scenario - **2/2 passed**, three same-model attempts, no fallback switch
- Senpi CLI smoke - **8/8 passed**
- GitHub reports `mergeable: MERGEABLE`; GitGuardian and Cubic checks are green

The worktree is clean, and the retry behavior is ready to merge.

code-yeongyu (commented):
VERDICT: BLOCKED - required CI exposed an unreviewed retry-count contract

`Check and test` failed at `packages/coding-agent/test/model-registry.test.ts:1639`: the failed-command scenario expected 2 command executions across two API-key resolutions but observed 6.

This is directly related to #489. Before merge, the branch must establish whether 6 is the intended new contract (two independent resolutions x three bounded attempts) or an accidental multiplication by another retry layer. The fix must preserve transient command recovery and pin the correct count in both ModelRegistry and resolve-config-value coverage.

Required: trace ModelRegistry -> resolver calls, update implementation or stale expectation accordingly, run the full affected tests, root check, QA, and obtain a green required CI run. Prior MERGE_OK is suspended until then.

code-yeongyu (commented):
VERDICT: MERGE_OK - supersedes CI-blocked review 4815011553

Re-reviewed final head `c0f9e363f`. CI exposed a stale pre-#489 expectation, not a production retry defect. Two independent uncached request-time API-key resolutions each correctly perform the bounded three-attempt credential command retry, so the contract is 2 x 3 = 6 executions before both calls return unresolved.

Verified:
- ModelRegistry + resolve-config-value suites: 93/93
- Root `npm run check`: passed
- Build: passed
- Mock-loop, CLI, TUI, and RPC QA: passed
- Required GitHub `Check and test`: passed
- Cross-OS terminal/inspector jobs: passed
- English-only audit: zero Korean in diff, commit, or public text

The transient credential recovery guarantee is preserved and the full contract is now pinned. Merging.

PR #503, opened 2026-07-29
fix(service-tier): explain why /fast is unavailable on a ChatGPT subscription

### LINKED ISSUE #499: Bug: /fast can never work — command is Codex-only but `-fast` variants are never generated for `openai-codex`
## Summary

`/fast` can never succeed on a stock install. The command is gated to the `openai-codex` provider, but `-fast` model variants are only generated for the `openai` provider — and a test actively enforces that `openai-codex` has none. So the two halves of the feature can never meet.

## Evidence

**1. `/fast` is Codex-only and looks for a same-provider `-fast` sibling**

`packages/coding-agent/src/core/extensions/builtin/service-tier.ts`:

```ts
const OPENAI_CODEX_PROVIDER = "openai-codex";
const FAST_MODEL_SUFFIX = "-fast";

// handler:
if (model?.provider !== OPENAI_CODEX_PROVIDER) {
  ctx.ui.notify("Fast mode is only available for OpenAI Codex models.", "warning");
  return;
}
...
const targetModel = baseModel ?? findFastModel(ctx.modelRegistry, model);
if (!targetModel) {
  ctx.ui.notify(`Fast mode is not supported for ${model.provider}/${model.id}.`, "warning");
  return;
}
```

`findFastModel` resolves `modelRegistry.find(baseModel.provider, `${baseModel.id}-fast`)` — i.e. it requires `openai-codex/<id>-fast` to exist.

**2. The catalog generator never emits `-fast` for `openai-codex`**

`packages/ai/scripts/generate-models.ts`:

```ts
// ... Scoped to the direct OpenAI provider; Azure clones and the
// Codex backend are intentionally excluded.
const openAiFastVariants: Model<Api>[] = [];
for (const model of allModels) {
  if (model.provider !== "openai" || !OPENAI_PRIORITY_TIER_MODEL_IDS.has(model.id)) continue;
  ...
}
```

**3. A test enforces the exclusion**

`packages/ai/test/openai-fast-models.test.ts`:

```ts
expect(getModels("openai-codex").some((model) => model.id.endsWith("-fast"))).toBe(false);
```

**Result:** every `/fast` invocation on a Codex model hits the "Fast mode is not supported" branch. No user configuration can change this.

## Reproduce

1. `/login openai-codex`
2. Select a Codex model, e.g. `openai-codex/gpt-5.6-sol`
3. Run `/fast`
4. Observe: `Fast mode is not supported for openai-codex/gpt-5.6-sol.`

## Regression origin

`10d855851 fix(coding-agent): switch /fast through model variants` moved `/fast` from a session-level `service_tier` toggle to model-variant switching. The accompanying tests in `packages/coding-agent/test/suite/service-tier-extension.test.ts` register synthetic `openai-codex/*-fast` models inside the test harness, so they pass while the shipped catalog has none. That gap is why the breakage was not caught.

## Possible fixes

**Option A — emit Codex fast variants.** Also generate `-fast` variants for `openai-codex`. The request plumbing already supports it: `SERVICE_TIER_APIS` in `service-tier.ts` includes `openai-codex-responses`, and `addServiceTierToPayload` sets `service_tier` for that api. The original "cost would double-count" rationale for excluding Codex does not apply, since Codex is subscription-metered and its catalog cost rates are zero.

**Option B — revert `/fast` to a session-level `service_tier: "priority"` toggle** for Codex, leaving the catalog untouched.

Either way, it would help to assert against the real generated catalog (rather than harness-injected models) so this specific mismatch cannot regress silently.

Happy to send a PR for whichever direction you prefer — I have Option A working locally.

eddieparc:
Fixes #499.

## What `/fast` does today

`/fast` is registered only for the `openai-codex` provider, and switches to a
`<id>-fast` catalog sibling carrying `serviceTier: "priority"`. But
`generate-models.ts` emits those variants only for the direct `openai`
provider — its comment says the Codex backend is "intentionally excluded". So no
Codex model ever has a compatible target, and the command's only reachable
outcome is:

```
Fast mode is not supported for openai-codex/gpt-5.5.
```

Reproduced on `2026.7.29-6` via the registry the command uses:

```
openai-codex/gpt-5.5: base=yes fast=NO
openai-codex/gpt-5.4: base=yes fast=NO
openai/gpt-4.1:       base=yes fast=yes tier=priority
```

## Why I did not just generate the Codex variants

That was my first instinct, and it is wrong. Measured directly against
`chatgpt.com/backend-api/codex/responses` with a live ChatGPT Pro subscription:

| sent `service_tier` | result |
|---|---|
| `priority` | HTTP 200, response echoes `"auto"` |
| `default` | HTTP 200, response echoes `"auto"` |
| `auto` | HTTP 400 `Unsupported service_tier` |
| `flex` | HTTP 400 `Unsupported service_tier` |
| `scale` | HTTP 400 `Unsupported service_tier` |

The backend allowlists `priority` but serves it at normal tier — priority
processing is an API-billing feature, not a subscription one. Meanwhile
`getServiceTierCostMultiplier()` bills `priority` at **2.5x for gpt-5.5** (2x
elsewhere). Synthesising `-fast` variants would therefore show users a "Fast"
model that inflates reported cost 2.5x while delivering unchanged service.

So the generator's exclusion is correct; only the message was wrong.

## This PR

Replaces the misleading no-variant notice with an accurate one: priority tier is
unavailable on a ChatGPT subscription and requires API-key billing on the
`openai` provider, where the variants already exist and `/fast` works. No
behavioural change beyond the notice text.

## Validation

- `npx vitest run test/suite/service-tier-extension.test.ts` → 7 passed
- `npx biome check --error-on-warnings` on both changed files → clean
- `changes.md` section added per `builtin/AGENTS.md`

If you would rather `/fast` refuse earlier (e.g. hide the command for
subscription-only Codex auth), happy to rework it that way.

---
## Summary by cubic
Replaces the misleading `/fast` “not supported” message for `openai-codex` models with a clear notice that Fast (priority tier) isn’t available on ChatGPT subscriptions and requires API‑key billing on `openai`. Fixes #499 with no behavior changes.

- **Bug Fixes**
  - Update the no-variant branch to explain the subscription limitation and point to API‑key billing on `openai`.
  - Adjust tests to assert the new notice and remove model-specific blame.
  - Add a `changes.md` entry.

<sup>Written for commit c2d55fa85f14d879fe50bf7de8a6eaae475937e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/503?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/changes.md
@@ -1,5 +1,30 @@
 # Builtin extensions changes
 
+## service-tier: explain why `/fast` is unavailable on a subscription (2026-07-30)
+
+- Fixes the misleading notice reported in issue #499. `/fast` is registered only
+  for `openai-codex`, but `generate-models.ts` emits `-fast` priority variants
+  only for the direct `openai` provider, so no Codex model ever has a target and
+  the command could only ever answer "Fast mode is not supported for
+  openai-codex/<model>" — which reads as a per-model gap rather than a
+  plan-level limitation.
+- Generating the missing Codex variants would be wrong. Measured against
+  `chatgpt.com/backend-api/codex/responses` with a live ChatGPT Pro
+  subscription: `service_tier: "priority"` and `"default"` both return HTTP 200
+  and the response echoes `"auto"`, while `"auto"`, `"flex"` and `"scale"` are
+  rejected with HTTP 400 `Unsupported service_tier`. The backend allowlists
+  `priority` but serves it at normal tier, and
+  `getServiceTierCostMultiplier()` would still bill it at 2.5x for gpt-5.5
+  (2x elsewhere) — so synthesising variants would inflate reported cost for
+  unchanged service.
+- The no-variant branch now states that priority tier is unavailable on a
+  ChatGPT subscription and that it requires API-key billing on the `openai`
+  provider, where `-fast` variants already exist and `/fast` works.
+- `test/suite/service-tier-extension.test.ts` asserts the notice explains the
+  subscription limitation and no longer blames the model.
+- Expected merge conflict zones: LOW in `service-tier.ts` around the
+  `FAST_UNAVAILABLE_ON_SUBSCRIPTION` constant and the no-variant branch.
+
 ## service-tier: add `/fast` for OpenAI Codex (2026-07-29)
 
 - `service-tier.ts` registers `/fast` only for the `openai-codex` provider.
--- a/packages/coding-agent/src/core/extensions/builtin/service-tier.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/service-tier.ts
@@ -10,6 +10,26 @@ type ProviderPayload = Record<string, unknown>;
 const OPENAI_CODEX_PROVIDER = "openai-codex";
 const OPENAI_CODEX_RESPONSES_API = "openai-codex-responses";
 const FAST_MODEL_SUFFIX = "-fast";
+/**
+ * `/fast` is registered only for `openai-codex`, but the catalog generator emits
+ * `-fast` priority variants only for the direct `openai` provider, so no Codex
+ * model ever has a target and the command could only report "not supported".
+ *
+ * That gap is intentional rather than an oversight in the generator: measured
+ * against `chatgpt.com/backend-api/codex/responses` with a live ChatGPT Pro
+ * subscription, `service_tier: "priority"` returns HTTP 200 but the response
+ * echoes `"auto"`, while `"auto"`/`"flex"`/`"scale"` are rejected with HTTP 400
+ * `Unsupported service_tier`. Priority processing is an API-billing feature, so
+ * a subscription request is served at normal tier no matter what is sent — and
+ * `getServiceTierCostMultiplier()` would still bill it at up to 2.5x.
+ *
+ * So the honest answer is that fast mode is unavailable here, not that the
+ * model does not support it.
+ */
+const FAST_UNAVAILABLE_ON_SUBSCRIPTION =
+	"Fast mode (priority tier) is not available on a ChatGPT subscription: chatgpt.com accepts " +
+	"service_tier=priority but serves the request at normal tier. Priority processing requires " +
+	"API-key billing on the `openai` provider.";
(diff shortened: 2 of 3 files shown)


code-yeongyu (commented):
VERDICT: MERGE_OK

Reviewed by the lead session. What I checked, concretely:

**The design call is right.** Generating `openai-codex/*-fast` variants was the obvious fix and you proved it wrong with live measurements (priority accepted-but-served-as-auto, 400 on auto/flex/scale, 2.5x billing multiplier on gpt-5.5). Charging users 2.5x for tier-identical service would have been strictly worse than an honest notice. The 24-line comment block above `FAST_UNAVAILABLE_ON_SUBSCRIPTION` (service-tier.ts:13-28) is exactly the kind of archaeology that stops the next person from "fixing" the generator exclusion — good.

**On whether this actually closes #499:** yes, defensibly. The issue's complaint is the two halves can never meet; the alternatives were (a) synthesize variants — disproven above, (b) unregister `/fast` on codex — hides the feature and users would ask where it went, (c) explain. (c) with measurement-backed wording is the honest one. Position stated for the record: keep the command registered.

**Nits, none blocking:**
- service-tier.ts:89 still says "Fast mode is only available for OpenAI Codex models." — after this PR that message coexists awkwardly with the new one: `/fast` on a *non*-codex model claims fast is codex-only, while on codex it says fast needs the `openai` provider. Both true, but pointing in opposite directions. Fine for now since only codex registers the command; worth unifying if `/fast` ever expands.
- Test now destructures `notify.mock.calls[0]` — asserts level + three content invariants instead of pinning the full sentence. That is behavior-shaped, not prose-pinning. Approved of.
- `findFastModel` remains reachable (`baseModel ?? findFastModel`) so a future catalog that DOES ship codex `-fast` variants short-circuits past the notice with zero code change. Correct forward-compat.

**Evidence relied on:** full CI green on this exact head (Check and test 26m, terminal-tools + inspector-handoff on 3 OSes), which covers `npm run check` and the scoped suite; `rg` over src/docs confirms the replaced string had no programmatic consumer (UI notify only). changes.md updated per fork convention.

Merging with a merge commit.

PR #505, opened 2026-07-30
fix(claude-agent-sdk): classify subscription-limit errors so account failover fires

eddieparc:
## Bug

With multiple Claude accounts registered (`/claude-account add`) and the managed
lane enabled (`claudeAgentSdkProvider.tokenInjection: "oauth-slots"`), an
account that has exhausted its Pro/Max plan is never blocked, so the pool never
rotates past it. Reproduced live on `2026.7.29-6` with two real accounts.

Claude Code signals plan exhaustion two ways, and `classifySdkError()` matched
neither:

1. **Prose thrown from the query iterator** (SDK replaces the exit error):
   `Claude Code returned an error result: You've hit your weekly limit · resets
   5am (Asia/Seoul)` — no `429`, no `rate_limit` token.
2. **Result messages** whose subtype is a bare `error_during_execution`; the
   real cause lives only in `terminal_reason` (e.g. `blocking_limit`), which
   `sdkFailure()` discarded when `errors` was empty.

Both classified as non-retryable `other`, so `runFailover` threw instead of
blocking + rotating.

## Fix

- `errors.ts`: classify `blocking_limit` / `rapid_refill_breaker` and prose
  hit/reached/exceeded-limit messages as retryable `rate_limit`.
- `auth-lane.ts` `sdkFailure()`: append `terminal_reason` to the subtype so the
  classifier can see it. `subtype` alone cannot distinguish a blocked
  subscription from an ordinary tool failure.

Deliberately narrow prose patterns: `context window exceeded for this request`
and similar still classify as non-retryable `other` (covered by tests).

## Validation

- `npx vitest run test/claude-agent-sdk-failover.test.ts test/claude-agent-sdk-auth-lane.test.ts` → 13 passed
- `tsgo --noEmit` clean
- Live after rebuild, two real accounts: exhausted account blocked with
  staggered cooldown (`blockReason: rate_limit`), pool rotated to the second
  account, `AllAccountsBlockedError` guidance when both were blocked, automatic
  retry served the request after cooldown expiry.

---
## Summary by cubic
Fixes error classification in `claude-agent-sdk` so subscription-limit hits are treated as rate limits. This blocks the exhausted account and triggers failover to the next account in the pool.

- **Bug Fixes**
  - Classify `blocking_limit`, `rapid_refill_breaker`, and prose "hit/reached/exceeded limit" messages as retryable `rate_limit`.
  - Include `terminal_reason` in result error messages so the classifier can detect subscription limits.
  - Added tests for these cases; unrelated errors (e.g., context window exceeded) remain non-retryable.

<sup>Written for commit 6d4908e29ccb49b8c31470045c81e136de22a2db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/505?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/claude-agent-sdk/auth-lane.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/claude-agent-sdk/auth-lane.ts
@@ -182,7 +182,17 @@ function sdkFailure(message: SDKMessage): unknown | undefined {
 	if (message.type === "assistant" && message.error) return message.error;
 	if (message.type === "result" && message.subtype !== "success") {
 		const errors = "errors" in message && Array.isArray(message.errors) ? (message.errors as unknown[]) : [];
-		return new Error(errors.length > 0 ? String(errors[0]) : `Claude Code ${message.subtype}`);
+		if (errors.length > 0) return new Error(String(errors[0]));
+		// `subtype` alone is too coarse to classify: a subscription limit and an
+		// ordinary tool failure both arrive as "error_during_execution". The SDK
+		// carries the real cause in `terminal_reason` (e.g. "blocking_limit"), so
+		// append it — otherwise classifySdkError() scores every result error as
+		// non-retryable "other", the exhausted account is never blocked, and a
+		// multi-account pool never rotates past it.
+		const reason = "terminal_reason" in message && typeof message.terminal_reason === "string"
+			? message.terminal_reason
+			: "";
+		return new Error(reason ? `Claude Code ${message.subtype}: ${reason}` : `Claude Code ${message.subtype}`);
 	}
 	return undefined;
 }
--- a/packages/coding-agent/src/core/extensions/builtin/claude-agent-sdk/errors.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/claude-agent-sdk/errors.ts
@@ -46,6 +46,20 @@ export function classifySdkError(error: unknown): SdkErrorClassification {
 	if (/\b(?:http\s*)?429\b|too many requests|rate[ _-]?limit/.test(text)) {
 		return { kind: "rate_limit", retryable: true };
 	}
+	// Claude Code signals subscription exhaustion two ways, neither of which
+	// carries an SDK error code or HTTP status:
+	//   - `terminal_reason: "blocking_limit"` / `"rapid_refill_breaker"` on a
+	//     result message (surfaced by auth-lane's sdkFailure)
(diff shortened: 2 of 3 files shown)


code-yeongyu (commented):
VERDICT: MERGE_OK

Reviewed by the lead session. Verified against the actual head (pull/505/head), not just the PR body:

**The mechanism is sound.**
- `sdkFailure()` (auth-lane.ts:182-196): appending `terminal_reason` to the subtype string is the minimal viable channel — I grepped every consumer of that error message: it flows only into `runFailover`'s `classify` (auth-lane.ts:231 -> errors.ts). Nothing else pattern-matches "Claude Code <subtype>", so the string change cannot break another consumer. Confirmed, not assumed.
- Bounded blast radius on misclassification, which was my main fear with prose matching: `runFailover` (failover.ts) runs **at most one attempt per account** (`attempt < accounts.length`), blocks use expiring `blockedUntil` (60s base, exponential, 48h cap), and post-delta failures throw instead of retrying. A false-positive `rate_limit` costs one bounded rotation with self-healing blocks — annoying, never an infinite loop and never a stuck pool.

**The tests are behavior-shaped.** Real CLI prose fixtures ("You've hit your weekly limit · resets 5am"), the `terminal_reason` path through the exact string `sdkFailure` constructs, AND the negative cases (`error_during_execution: model_error` stays other, "context window exceeded for this request" stays non-retryable). The negative assertions are what make the positive ones trustworthy.

**One genuine concern, non-blocking:** the first prose branch `/\b(?:hit|reached|exceeded)\b[^.]*\blimit\b/` is broader than the PR body's "deliberately narrow" claim. Hypothetical "request exceeded the size limit" (413-ish prose) would classify as rate_limit and burn one pool rotation. The bounded failover above makes this survivable, and no such string exists in the SDK surface today — but if a misrotation report ever shows up, this branch is the suspect. Consider anchoring on plan words (weekly|daily|hourly|5-hour|usage) for the hit/reached/exceeded arm too.

**Evidence relied on:** full CI green on this head (Check and test + 3-OS terminal/inspector jobs), failover.ts bounds read line-by-line, consumer grep clean. changes.md untouched but the errors.ts comment block documents both signal shapes inline - acceptable placement.

Merging with a merge commit. Thanks for the negative tests - they are the difference between a patch and a fix.

PR #507, opened 2026-07-30
fix(btw): route side queries through model runtime

### LINKED ISSUE #488: /btw: No API provider registered for api: claude-agent-sdk
### What happened?

`/btw` fails for every model served by the `claude-agent-sdk` provider:

```
Error: No API provider registered for api: claude-agent-sdk
```

The main turn with the same model works fine — only the side query breaks.

[`side-query.ts#L63`](https://github.com/code-yeongyu/senpi/blob/e534e1ee9/packages/coding-agent/src/core/extensions/builtin/btw/side-query.ts#L63) falls back to the compat `streamSimple` when no `streamFn` is passed, and [`btw/index.ts#L94-L108`](https://github.com/code-yeongyu/senpi/blob/e534e1ee9/packages/coding-agent/src/core/extensions/builtin/btw/index.ts#L94-L108) never passes one — so every side query dispatches through the compat api-registry.

But `claude-agent-sdk` never lands in that registry. [`claude-agent-sdk/index.ts#L41-L46`](https://github.com/code-yeongyu/senpi/blob/e534e1ee9/packages/coding-agent/src/core/extensions/builtin/claude-agent-sdk/index.ts#L41-L46) registers through `pi.registerProvider()`, which feeds senpi's own ModelRuntime. compat's [`BUILTIN_APIS`](https://github.com/code-yeongyu/senpi/blob/e534e1ee9/packages/ai/src/compat.ts#L120-L131) holds 10 api ids and `claude-agent-sdk` isn't one of them, so [`resolveApiProvider()`](https://github.com/code-yeongyu/senpi/blob/e534e1ee9/packages/ai/src/compat.ts#L185-L190) throws.

The main loop is unaffected because [`sdk.ts#L319-L329`](https://github.com/code-yeongyu/senpi/blob/e534e1ee9/packages/coding-agent/src/core/sdk.ts#L319-L329) wires the Agent's `streamFn` to `modelRuntime.streamSimple()`.

`/btw` therefore works only for the apis in that registry. Any provider registered through `pi.registerProvider()` — `claude-agent-sdk` included — fails the same way.

### Steps to reproduce

1. Log in with the `claude-agent-sdk` provider and select one of its models
2. Run `senpi -ne`
3. `/btw what did we just discuss?`
4. → `Error: No API provider registered for api: claude-agent-sdk`

### Expected behavior

`/btw` works with any provider the session can already use.

### Version

`2026.7.28-3` (`e534e1ee9`)

code-yeongyu:
## Summary

- route `/btw` side-query streams through Senpi's `ModelRuntime`
- preserve the existing generic side-query runner while supplying the same provider dispatch path used by the main loop
- add issue-shaped regression coverage for an API id registered only in the runtime

## Why

`/btw` resolved authentication through the model runtime but omitted its stream function. The side-query runner therefore fell back to compat `streamSimple`, whose built-in API registry does not contain providers registered through `pi.registerProvider()`. Runtime-only providers failed before their registered stream implementation could run.

## Verification

- RED: issue regression failed with runtime provider call count `0`, expected `1`
- GREEN: issue regression passed (`1/1`)
- scoped BTW suite: `2` files, `12` tests passed
- `npm run check`: exit `0`
- Senpi QA isolation self-check: `9/9` passed
- real CLI mock-loop self-test: `43/43` passed, zero real provider calls, real auth unchanged
- evidence: `local-ignore/qa-evidence/20260730-btw-runtime-provider/`

Fixes #488

---
## Summary by cubic
Routes `/btw` side-query streams through the runtime provider path to match the main loop and fix failures with runtime-only providers. Fixes #488.

- **Bug Fixes**
  - Inject `streamFn` from `ctx.modelRegistry.modelRuntime.streamSimple` into the side-query runner; no compat fallback.
  - Keep the existing side-query runner; only change the dispatch path.
  - Add regression test for a runtime-only API id to ensure the provider’s stream is called.

<sup>Written for commit ff48536d16a85d370c08c42d31061e571aaa5751. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/507?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/btw/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/btw/changes.md
@@ -1,5 +1,25 @@
 # changes — btw
 
+## Runtime provider dispatch for side queries (2026-07-30)
+
+### What changed
+
+- `/btw` now passes a stream function backed by `ctx.modelRegistry.modelRuntime`
+  into the side-query runner instead of allowing it to fall back to the compat
+  API registry.
+- Added issue #488 regression coverage with a provider whose API id exists only
+  in Senpi's runtime registry.
+
+### Why
+
+- Providers registered through `pi.registerProvider()` work in the main loop but
+  may not exist in compat's built-in API registry. `/btw` therefore failed before
+  invoking their registered stream implementation.
+
+### Merge-conflict zones
+
+- `index.ts` side-query dependency construction only.
+
 ## Parallel side questions via `/btw` (2026-07-21)
 
 ### What changed
--- a/packages/coding-agent/src/core/extensions/builtin/btw/index.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/btw/index.ts
@@ -97,6 +97,8 @@ export default function btwExtension(pi: ExtensionAPI) {
 						auth: { apiKey: auth.apiKey, headers: auth.headers, extraBody: auth.extraBody },
 						sessionId,
 						thinkingLevel: thinkingLevel === "off" ? undefined : thinkingLevel,
+						streamFn: (streamModel, streamContext, options) =>
+							ctx.modelRegistry.modelRuntime.streamSimple(streamModel, streamContext, options),
 					},
 					context,
 					{
--- a/packages/coding-agent/test/suite/regressions/488-btw-runtime-provider.test.ts
+++ b/packages/coding-agent/test/suite/regressions/488-btw-runtime-provider.test.ts
@@ -1,0 +1,55 @@
+import { createAssistantMessageEventStream, fauxAssistantMessage, type Model } from "@earendil-works/pi-ai";
+import { afterEach, describe, expect, it } from "vitest";
+import btwExtension from "../../../src/core/extensions/builtin/btw/index.ts";
+import { createHarness, type Harness } from "../harness.ts";
+
+const PROVIDER_ID = "runtime-only";
+const API_ID = "runtime-only-api";
+
+describe("issue #488: /btw with a runtime-registered provider", () => {
+	let harness: Harness | undefined;
+
+	afterEach(() => {
+		harness?.cleanup();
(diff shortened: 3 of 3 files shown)


code-yeongyu (commented):
VERDICT: MERGE_OK

Lead review. The fix is the right shape: one lambda in btw/index.ts wiring `runSideQuery`'s existing `streamFn` seam to `ctx.modelRegistry.modelRuntime.streamSimple` — the exact parity move with how core/sdk.ts wires the main loop. Provider-agnostic (no claude-agent-sdk special case, no compat BUILTIN_APIS expansion), which is what makes this a fix rather than a patch.

The regression test earns its keep: it registers a provider whose API id exists ONLY in the runtime registry, drives a real `/btw` prompt through the harness, and asserts the runtime stream was invoked. That is the precise failure mode from #488 (compat resolveApiProvider throw), and the captured RED (`expected +0 to be 1`) proves it fails without the fix. Behavior-asserted, no mock theater.

Evidence audit: scoped btw suite 12/12, npm run check exit 0, mock-loop QA 43/43 zero-token with receipts under local-ignore/qa-evidence/20260730-btw-runtime-provider/. changes.md updated with conflict zones per fork convention.

Nit, non-blocking: the streamFn lambda forwards `(streamModel, streamContext, options)` positionally — fine today, but if runtime streamSimple ever grows an options field the compat path also needs, this seam is where drift would hide. A one-line comment pointing at sdk.ts:319 as the parity reference would help the next reader.

Merging after CI completes.

PR #508, opened 2026-07-30
fix(devenv): run npm on Windows

### LINKED ISSUE #454: devenv-setup.mjs cannot run npm on Windows: execFileSync npm.cmd throws EINVAL
### What happened?

On Windows, `scripts/devenv-setup.mjs` can't run a single npm step.

`npm()` ([devenv-setup.mjs#L90-L93](https://github.com/code-yeongyu/senpi/blob/main/scripts/devenv-setup.mjs#L90-L93)) calls `execFileSync("npm.cmd", ...)` without `shell: true`. Since the CVE-2024-27980 fix, Node refuses to spawn `.cmd`/`.bat` without a shell, so every call throws `spawnSync npm.cmd EINVAL` (errno `-4071`).

Two consequences:

- **Fresh clone** (no root `node_modules`): the call on L129 isn't wrapped, so it propagates to `main()`'s rejection handler and exits 1. Windows contributors can't bootstrap at all.
- **Existing `node_modules`**: the skill install on L137 sits inside `catch {}`, which swallows the real error and prints `node-pty native build ... falls back to tmux on POSIX`. Both halves are wrong — it's EINVAL, not a native build failure, and there is no tmux fallback on Windows. Net effect: senpi-qa Channel 2 (TUI) silently reports `[SKIP] no usable PTY backend` forever, so TUI QA never actually runs on Windows.

### Steps to reproduce

Windows 11, Node 24.18.0, npm 11.16.0, on `954f53da6`:

```
Remove-Item -Recurse .agents/skills/senpi-qa/node_modules
node scripts/devenv-setup.mjs
# ! senpi-qa skill deps failed to install (node-pty native build). TUI QA falls back to tmux on POSIX.
# -> node_modules is still absent
```

That it's EINVAL and not a native build failure:

```js
execFileSync("npm.cmd", ["install"], { cwd: ".agents/skills/senpi-qa", stdio: "inherit" });
// spawnSync npm.cmd EINVAL   errno -4071
```

while the identical command in a shell works:

```
cd .agents/skills/senpi-qa; npm install --no-audit --no-fund
# added 2 packages in 24s   (exit 0)
```

Adding `shell: true` to that same `execFileSync` also succeeds (`added 2 packages`).

### Expected behavior

`devenv-setup.mjs` installs dependencies on Windows.

Separately, the `catch {}` on L138 should surface the underlying error instead of asserting a native-build cause, and shouldn't advertise a POSIX-only fallback when running on Windows — the misleading message is why this stayed invisible.

One caveat on the fix: a bare `shell: true` triggers `DEP0190` (args are concatenated, not escaped). Resolving npm's JS entry and spawning it with `process.execPath` avoids both the EINVAL and the deprecation, but I'll leave the approach to you.

### Version

`2026.7.28-3` (`954f53da6`) — Windows 11, Node 24.18.0, npm 11.16.0

code-yeongyu:
## Summary

Windows bootstrap failed because Node no longer directly spawns `npm.cmd` after the CVE-2024-27980 hardening. This change runs npm through the shell only on Windows, while preserving direct execution on POSIX.

The optional senpi-qa dependency install now reports the actual caught error and only mentions the tmux fallback on POSIX, rather than always claiming a node-pty native build failure.

## Changes

- Set `shell: true` only for the Windows `npm.cmd` invocation used by workspace and senpi-qa dependency installation.
- Include the underlying npm error in the senpi-qa warning.
- Omit the POSIX-only tmux fallback message on Windows.

## Validation

- **RED regression proof:** simulated the `win32` npm helper path before the fix; expected `options.shell === true`, observed `undefined`.
- **GREEN regression proof:** repeated the same simulation after the fix; 1/1 assertion passed.
- **Error reporting scenario:** injected a failing npm executable; the setup output included the simulated npm failure and the caught command failure rather than attributing it to node-pty.
- `node scripts/devenv-setup.mjs` on macOS: passed and installed the senpi-qa dependencies successfully.
- `npm run test:scripts`: 72/72 passed.
- `npm run check`: passed.
- `node .agents/skills/senpi-qa/scripts/lib/common.mjs --self-check`: 9/9 passed.
- `node .agents/skills/senpi-qa/scripts/cli-smoke.mjs --self-test`: 8/8 passed with real auth unchanged.

Evidence was captured under `local-ignore/qa-evidence/20260730-issue-454/` and remains intentionally uncommitted.

## Environment surface sync

No `.devcontainer/devcontainer.json` change is required. This fix does not change dependencies, the Node version, provider/environment variables, QA channels, build commands, or forwarded ports; it only corrects platform-specific process spawning and error reporting inside the existing setup entry point.

Fixes #454

---
## Summary by cubic
Fixes Windows bootstrap by running `npm` through the shell on Windows, restoring workspace and `senpi-qa` dependency installs after CVE-2024-27980 hardening. Also improves `senpi-qa` install errors and avoids the POSIX-only tmux fallback message on Windows.

- **Bug Fixes**
  - Spawn `npm.cmd` with `shell: true` on Windows; keep direct exec on POSIX.
  - Show the actual `npm` error in `senpi-qa` install warnings.
  - Only mention the tmux fallback on POSIX.

<sup>Written for commit 0a20425efcaa7167a3d327f8557dd73eb065f6ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/508?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/devenv-setup.mjs
+++ b/scripts/devenv-setup.mjs
@@ -89,7 +89,7 @@ function hasCmd(cmd) {
 
 function npm(cwd, npmArgs) {
 	const bin = isWindows ? "npm.cmd" : "npm";
-	execFileSync(bin, npmArgs, { cwd, stdio: "inherit" });
+	execFileSync(bin, npmArgs, { cwd, stdio: "inherit", shell: isWindows });
 }
 
 function run(cmd, cmdArgs, options = {}) {
@@ -135,8 +135,10 @@ function installDeps() {
 		log("Installing senpi-qa skill deps (node-pty for cross-platform TUI QA)...");
 		try {
 			npm(skillDir, ["install", "--no-audit", "--no-fund"]);
-		} catch {
-			warn("senpi-qa skill deps failed to install (node-pty native build). TUI QA falls back to tmux on POSIX.");
+		} catch (error) {
+			const detail = error instanceof Error ? error.message : String(error);
+			const fallback = isWindows ? "" : " TUI QA can fall back to tmux on POSIX.";
+			warn(`senpi-qa skill deps failed to install: ${detail}.${fallback}`);
 		}
 	}
 }


code-yeongyu (commented):
VERDICT: MERGE_OK

Lead review. Exactly the two changes #454 asked for and nothing else:

1. `shell: isWindows` on the npm exec (devenv-setup.mjs:92) — the canonical post-CVE-2024-27980 pattern for `.cmd` shims. POSIX path unchanged (`shell: false` implicit), so no quoting-behavior regression on mac/linux. Note for the record: with `shell: true` Node does not escape args, but every arg at both call sites is a static literal ("install", "--no-audit", "--no-fund"), so there is no injection surface. If anyone ever threads user input into `npm()`, that assumption breaks — the static-args invariant is load-bearing.
2. The catch now surfaces the real error instead of the fabricated node-pty/tmux diagnosis, and the tmux fallback claim is now POSIX-conditional — both halves of the reported lie fixed.

Evidence audit: RED captured (`options.shell` undefined) before the fix via the extracted win32 simulation, GREEN after; npm run check exit 0; test:scripts 72/72; `node scripts/devenv-setup.mjs` still green on macOS (the don't-break-POSIX proof that matters most here); CLI smoke 8/8. The PR body's conclusion that no .devcontainer sync is needed is correct — deps/node/env surfaces untouched, this is invocation mechanics only.

Merging after CI completes.

PR #509, opened 2026-07-30
fix(coding-agent): preserve Bun launcher after update

### LINKED ISSUE #496: senpi update --force replaces Bun launcher with Node shebang and breaks CLI on Node 20
## Summary

`senpi update --all --force` can replace a working Bun launcher with a symlink to the package's Node-shebang entrypoint. On a machine that intentionally runs the Bun-installed CLI with Node 20 present, the updated CLI then crashes before loading any packages.

## Environment

- Senpi: `2026.7.29`
- OS: Linux 6.8, x86_64
- Bun: `1.3.14`
- System Node: `v20.20.2`
- Install method: global Bun install
- Latest source checked: `de10c1c177d1a70dcbb7259449cd8ace24e294d9`

## Reproduction

1. Install Senpi globally with Bun and confirm that the launcher executes the CLI with Bun.
2. Run:

   ```bash
   senpi update --all --force
   ```

3. Inspect and invoke the resulting launcher:

   ```bash
   ls -l ~/.bun/bin/senpi
   head -n 1 ~/.bun/install/global/node_modules/@code-yeongyu/senpi/dist/cli.js
   senpi list --no-approve
   ```

## Expected Behavior

An installation managed by Bun should remain executable through Bun after self-update, and `senpi list` should continue to work.

## Actual Behavior

The update replaced the prior Bun wrapper with:

```text
~/.bun/bin/senpi -> ../install/global/node_modules/@code-yeongyu/senpi/dist/cli.js
```

The target starts with:

```text
#!/usr/bin/env node
```

The shell therefore selected system Node 20 and `senpi list` exited 1:

```text
.../undici/lib/web/cache/cachestorage.js:20
    webidl.util.markAsUncloneable(this)
                ^

TypeError: webidl.util.markAsUncloneable is not a function
```

`senpi --version` still worked because that path exits before the failing imports, which can make the installation look healthy.

## Evidence

The failure toggles solely with the runtime while invoking the identical installed CLI:

```bash
node ~/.bun/install/global/node_modules/@code-yeongyu/senpi/dist/cli.js list --no-approve
# exit 1: markAsUncloneable is not a function

bun ~/.bun/install/global/node_modules/@code-yeongyu/senpi/dist/cli.js list --no-approve
# exit 0: lists installed packages
```

The installed package declares Node `>=24.0.0`; its `undici@8.9.0` declares Node `>=22.19.0`. Current self-update construction uses `bun install -g` in `packages/coding-agent/src/config.ts`, but does not verify that the generated global bin continues to execute with Bun.

No matching open or closed issue was found using the error text, `senpi update`, Bun launcher, and Node engine searches.

## Root Cause

Confirmed: Bun's global reinstall regenerated the package bin as a direct symlink. Because the published bin entrypoint has a Node shebang, the post-update invocation changed runtimes from Bun to the system Node. The installed Node version was below both Senpi's and undici's engine requirements, causing the import-time crash.

## Proposed Fix

- Preserve the runtime that owns the installation when constructing the post-update launcher.
- For Bun-managed installs, create or retain a Bun launcher rather than relying on the Node shebang symlink.
- Add an isolated `bun install -g` self-update regression test that invokes a command requiring normal module loading, not only `--version`.
- If a Bun-safe launcher cannot be guaranteed, detect the incompatible generated launcher and fail the update with an actionable message instead of reporting success.

## Verification Plan

- Reproduce from a Bun-managed install with Node 20 on `PATH`.
- Run `senpi update --all --force`.
- Verify `senpi list`, `senpi --help`, and an interactive TUI launch all exit or start normally through Bun.
- Verify Node-managed installations on supported Node versions remain unchanged.

---
This issue or PR was generated by LazyCodex.
Tag: lazycodex-generated

code-yeongyu:
## Summary

- detect Bun's global bin directory before constructing a self-update
- append a post-install repair step that atomically replaces Bun's Node-shebang symlink
- generate a launcher that executes the updated CLI through the Bun binary beside it

## Regression coverage

The issue regression creates the exact broken post-update shape: a Bun global
package whose `senpi` launcher is a symlink to a `#!/usr/bin/env node`
entrypoint. It executes the planned repair and verifies that the launcher
becomes an executable Bun wrapper.

### TDD

RED:

```text
AssertionError: expected 'install' to be '-e'
Expected: "-e"
Received: "install"
```

GREEN:

```text
Test Files  1 passed (1)
Tests       1 passed (1)
```

## Verification

- `npm test -- test/config.test.ts test/self-update-bootstrap.test.ts test/suite/regressions/496-bun-launcher-self-update.test.ts`
  - 3 files passed, 19 tests passed
- `npm run check`
  - exit 0
- `.agents/skills/senpi-qa/scripts/cli-smoke.mjs --self-test`
  - 8/8 passed
  - receipt: `local-ignore/qa-evidence/20260730-bun-launcher-self-update/cli-smoke.txt`

Fixes #496

---
## Summary by cubic
Preserves the global `bun` launcher after self-update by replacing the Node-shebang symlink with a small POSIX wrapper that execs the same `bun` runtime. Fixes #496 and restores CLI execution for `bun` global installs.

- **Bug Fixes**
  - Detects `bun`’s global bin dir via `bun pm bin -g`; aborts self-update if not found.
  - Adds a post-install repair that atomically swaps in a shell launcher running updated `dist/cli.js` with `bun`’s own `process.execPath` (safe quoting, supports `bun` outside the bin dir); skipped on Windows.
  - Adds a regression test that reproduces the broken launcher and verifies the repair, plus unit tests for Windows behavior.

<sup>Written for commit 932fc83309e0c8c4bcdb5541c93f6f451e29f4e5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/509?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/bun-global-launcher.ts
+++ b/packages/coding-agent/src/bun-global-launcher.ts
@@ -8,11 +8,11 @@ export interface BunLauncherRepairCommand {
 
 const BUN_LAUNCHER_REPAIR_SOURCE = `
 import { renameSync, writeFileSync } from "node:fs";
-const [launcher, bunPath, entrypoint] = process.argv.slice(1);
-if (!launcher || !bunPath || !entrypoint) throw new Error("Missing Bun launcher repair path");
+const [launcher, entrypoint] = process.argv.slice(1);
+if (!launcher || !entrypoint) throw new Error("Missing Bun launcher repair path");
 const quote = (value) => "'" + value.replaceAll("'", "'\\"'\\"'") + "'";
 const temporary = launcher + "." + process.pid + ".tmp";
-writeFileSync(temporary, "#!/bin/sh\\nexec " + quote(bunPath) + " " + quote(entrypoint) + " \\"$@\\"\\n", { mode: 0o755 });
+writeFileSync(temporary, "#!/bin/sh\\nexec " + quote(process.execPath) + " " + quote(entrypoint) + " \\"$@\\"\\n", { mode: 0o755 });
 renameSync(temporary, launcher);
 `.trim();
 
@@ -23,11 +23,10 @@ export function createBunLauncherRepairCommand(
 ): BunLauncherRepairCommand {
 	const packageDir = join(dirname(binDir), "install", "global", "node_modules", ...packageName.split("/"));
 	const launcher = join(binDir, executableName);
-	const bunPath = join(binDir, process.platform === "win32" ? "bun.exe" : "bun");
 	const entrypoint = join(packageDir, "dist", "cli.js");
 	return {
 		command: "bun",
-		args: ["-e", BUN_LAUNCHER_REPAIR_SOURCE, launcher, bunPath, entrypoint],
+		args: ["-e", BUN_LAUNCHER_REPAIR_SOURCE, launcher, entrypoint],
 		display: `repair Bun launcher ${launcher}`,
 	};
 }
--- a/packages/coding-agent/src/bun-global-launcher.ts
+++ b/packages/coding-agent/src/bun-global-launcher.ts
@@ -1,0 +1,33 @@
+import { dirname, join } from "node:path";
+
+export interface BunLauncherRepairCommand {
+	readonly command: string;
+	readonly args: string[];
+	readonly display: string;
+}
+
+const BUN_LAUNCHER_REPAIR_SOURCE = `
+import { renameSync, writeFileSync } from "node:fs";
+const [launcher, bunPath, entrypoint] = process.argv.slice(1);
+if (!launcher || !bunPath || !entrypoint) throw new Error("Missing Bun launcher repair path");
+const quote = (value) => "'" + value.replaceAll("'", "'\\"'\\"'") + "'";
+const temporary = launcher + "." + process.pid + ".tmp";
+writeFileSync(temporary, "#!/bin/sh\\nexec " + quote(bunPath) + " " + quote(entrypoint) + " \\"$@\\"\\n", { mode: 0o755 });
+renameSync(temporary, launcher);
+`.trim();
+
+export function createBunLauncherRepairCommand(
+	binDir: string,
+	packageName: string,
+	executableName: string,
+): BunLauncherRepairCommand {
+	const packageDir = join(dirname(binDir), "install", "global", "node_modules", ...packageName.split("/"));
+	const launcher = join(binDir, executableName);
+	const bunPath = join(binDir, process.platform === "win32" ? "bun.exe" : "bun");
+	const entrypoint = join(packageDir, "dist", "cli.js");
+	return {
+		command: "bun",
+		args: ["-e", BUN_LAUNCHER_REPAIR_SOURCE, launcher, bunPath, entrypoint],
+		display: `repair Bun launcher ${launcher}`,
+	};
+}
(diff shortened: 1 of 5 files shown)


code-yeongyu (commented):
VERDICT: BLOCKED - repair wrapper can brick non-default bun installs; win32 unguarded

Lead review. The architecture is right — a post-install repair step is the correct minimal lever, the atomic tmp+rename write is careful, the fixture-based regression test actually executes the repair script and asserts the launcher stops being a Node-shebang symlink. Good bones. Two defects, one of them worse than the bug being fixed:

**1. BLOCKING — `bunPath = join(binDir, "bun")` (bun-global-launcher.ts:27) assumes the bun binary lives in the global bin dir.** That holds for curl-installed bun (`~/.bun/bin/bun`) but NOT for Homebrew (`/opt/homebrew/bin/bun`) or other package-manager installs, where `bun pm bin -g` still reports `~/.bun/bin` while no `bun` binary exists there. The repaired launcher would `exec` a nonexistent path — turning "CLI runs on wrong runtime" (#496) into "CLI does not run at all". The fix is already in your hands: the repair script runs UNDER bun, so `process.execPath` inside the `-e` script IS the real bun binary, whatever installed it. Drop the `bunPath` argument and use `process.execPath` in the wrapper. The test keeps working (spawnSync launches the script with a real runtime).

**2. BLOCKING (small) — the repair writes a `#!/bin/sh` wrapper unconditionally, including win32.** On Windows the bun shim is `senpi.exe`/`.cmd`; writing a POSIX sh script named `senpi` next to it is at best dead weight and at worst shadows the working shim for sh-compatible shells. #496 is a POSIX report; gate the repair step to `process.platform !== "win32"` and leave Windows semantics for a dedicated issue if one ever materializes.

Also noted, non-blocking: `readCommandOutput("bun", ["pm","bin","-g"])` returning undefined now aborts the whole bun self-update command (returns undefined) — a behavior change from before (update proceeded without repair). Defensible, but say so in changes.md.

Fix both blockers, re-run the regression + config tests, push to the same branch. The review loop continues there.

code-yeongyu (commented):
VERDICT: MERGE_OK (supersedes my BLOCKED review 4814683551)

Re-reviewed commit 932fc8330 against both blockers:

1. bunPath brick risk — FIXED. The repair script now writes `process.execPath` captured at repair time (bun-global-launcher.ts): the wrapper always points at whatever bun binary actually executed the update, Homebrew or curl-install alike. The `bunPath` argument is gone from the arg list, and the RED capture (`expected exec '<node path>' ..., received exec '<fixture>/.bun/bin/bun'`) proves the old behavior was actually caught before the fix.
2. win32 sh-wrapper — FIXED. `config.ts` bun case now passes `undefined` for the repair step on win32; the Windows regression RED (install+repair steps where a single install was expected) pins it.
3. changes.md documents the missing-binDir abort semantics as requested.

Evidence: blocker tests 17/17, scoped self-update suite 20/20, root npm run check exit 0 (2502 files). Clean request-changes turnaround - both fixes are the minimal correct form of what the review asked, no scope creep.

Merging after CI completes.

PR #529, opened 2026-07-30
fix(compaction): enforce guards on blocking routes

### LINKED ISSUE #527: Compaction caps and circuit breaker bypassed on blocking routes — unbounded compaction loops
## Summary

The per-turn cap (soft 3 / hard 10) and the circuit breaker (3 failures → 60s cooldown) only guard the `session_before_compact` hook entry point (`packages/coding-agent/src/core/extensions/builtin/compaction/index.ts:395-414`). Every trigger that goes through `applyBlockingCompaction()` (`index.ts:247-393`) generates a summary directly and calls `ctx.applyCompaction()` with a precomputed result, which skips `session_before_compact` entirely (precomputed results bypass the hook in `agent-session.ts` `_executeCompaction`, `if (!compactionResult)` branch). As a result, compaction can repeat without bound — each iteration paying a real LLM summarization request — as long as agent events keep arriving.

## Reproduced unbounded paths

All three reproduced with deterministic faux-provider tests (no network, ~1s):

1. **Hard-limit + transient summarization failure → breaker never consulted.** `before_agent_start`'s `isAtHardLimit` branch (`index.ts:565-572`) does not check `breakerCoolingDown`; the breaker check only guards the threshold/speculative `else if` branches (`:573-598`). 8 consecutive prompts at the hard limit → 8 summarization calls (expected: stop after 3). The same failures on the proactive (non-hard) route correctly stop after 3, isolating the defect to the hard-limit branch.

2. **Hard-limit + successful-but-ineffective compaction → caps never consulted.** With usage pinned at the hard limit and summaries that cannot shrink context below the limit, 12 consecutive prompts → 12 blocking compactions (expected: bounded by hard cap 10).

3. **`turn_end` zero-yield recovery loop.** `index.ts:651-656` re-fires `applyBlockingCompaction()` on every `turn_end` whenever `lastYield.savedTokens <= 0`, unconditionally. After one accepted zero-yield compaction, 5 sequential tool turns → 5 more compactions (expected: bounded by soft cap 3).

## Contributing defects

- `incrementIneffective` (`per-turn-cap.ts:20-24`) is dead code: never called from `src/`. The `session_compact` accepted handler detects ineffectiveness via `isIneffectiveCompaction` and only logs `ineffective_counted` (`index.ts:516-528`), so ineffective attempts never count toward the soft cap.
- A successful compaction calls `breaker.recordSuccess` (`index.ts:505-506`), so successful-but-useless compactions never trip the breaker either.

## Guards that do terminate (verified)

- `runExtensionCompaction` overflow-retry loop shrinks history each iteration and terminates at length ≤ 1.
- Degradation-monitor recovery is capped at 1/cycle and 3/session.
- Core summarization retries bounded by `maxRetries`.
- Overflow compact-and-retry is one-shot (`_overflowRecoveryAttempted`).

## Minimal fix direction

1. Check `breaker.isTripped` in the hard-limit branch of `before_agent_start`.
2. Gate `applyBlockingCompaction` (or its call sites) on the per-turn/absolute caps.
3. Call `incrementIneffective` where `isIneffectiveCompaction` is detected in `session_compact`.

## Reproduction

```sh
npx vitest --run test/compaction/infinite-compaction.repro.test.ts
# 3 designed-to-fail assertions show 8/12/5 compactions where guards require ≤3/≤10/≤3.
# The proactive-route control test passes, proving the bypass is route-specific.
```

A PR with the fix and regression tests follows.

effortprogrammer:
## What changed

`applyBlockingCompaction()` now enforces the circuit breaker and the per-turn soft cap at its entry, and the `session_compact` accepted branch wires the previously dead `cap.incrementIneffective` counter.

## Why

The per-turn cap (soft 3 / hard 10) and circuit breaker (3 failures → 60s cooldown) were only consulted in the `session_before_compact` hook. Every blocking trigger — `before_agent_start` hard-limit, `agent_end` idle, `turn_end` zero-yield recovery, degradation-monitor recovery — generates a summary via `applyBlockingCompaction()` and applies it precomputed, which skips `session_before_compact` in `_executeCompaction` (`if (!compactionResult)` branch). Both guards were dead on those routes, so compaction could repeat without bound, each iteration paying a real LLM summarization request. Additionally, `incrementIneffective` was never called from `src/`, so ineffective attempts never counted toward the soft cap and the `turn_end` zero-yield route retriggered on every tool turn.

Reproduced three unbounded loops with deterministic faux-provider tests (details in #527):

1. hard-limit + transient summarization failure: 8 prompts → 8 summarization calls (breaker ignored);
2. hard-limit + successful-but-ineffective compaction: 12 prompts → 12 compactions (caps ignored);
3. zero-yield accepted compaction: 5 sequential tool turns → 5 recovery compactions.

## Observed behavior after fix

- Hard-limit transient failure stops at `FAILURE_TRIP_THRESHOLD` (3) summarization calls.
- Accepted blocking compactions stop at the per-turn soft cap (3).
- Zero-yield recovery runs exactly once per turn, then the soft cap refuses further attempts.

## QA / evidence

Full receipts: `local-ignore/qa-evidence/20260730-issue-527-blocking-route-guards/`.

- New `test/compaction/blocking-compaction-route-guards.test.ts`: red before fix (8/12/5 calls), green after (3/3 pass).
- `npx vitest --run test/compaction/ test/compaction*.test.ts test/trigger-compact-extension.test.ts`: 47 files, 362 tests pass.
- Session-level compaction suites + `test/suite/regressions/`: 336 tests pass after workspace build (the only initial failures were unbuilt-`dist` module errors in a fresh clone; green after `npm run build`).
- `npx biome check` clean on changed files; root `npx tsgo --noEmit` exit 0.
- Live CLI QA `compaction-remote-qa.mjs --self-test`: 5/6, identical on pristine `main` (the one FAIL is a pre-existing payload-replay assertion, A/B verified via `git stash`). `compaction-wave1.mjs` fails identically on `main` (fake model never called in this environment) — not caused by this change.

## Residual risk

- Automatic blocking routes now share the soft cap with the hook route, so a pathological same-turn context can refuse a 4th blocking compaction; the turn then proceeds to the normal provider-overflow recovery path (one-shot), which is the existing fail-closed behavior.

closes #527

---
## Summary by cubic
Prevents unbounded blocking compaction loops by enforcing the circuit breaker and per-turn cap on all blocking routes and counting ineffective compactions toward the cap. This bounds retries and avoids wasted LLM summarization calls (addresses #527).

- **Bug Fixes**
  - Gate `applyBlockingCompaction()` with the circuit breaker and per-turn soft cap; reject when tripped or capped.
  - Count ineffective accepted compactions via `cap.incrementIneffective` to stop zero-yield retriggers in the same turn.
  - Add tests: breaker trips after 3 failures; hard-limit compactions capped at 3; turn-end recovery runs once per turn.

<sup>Written for commit 73f7647c4aadf26384259b151aa2404cff1b3963. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
@@ -248,6 +248,14 @@ export default function compactionExtension(
 		ctx: ExtensionContext,
 		customInstructions: string,
 	): Promise<SpeculativeCompactionResult> {
+		if (breaker.isTripped(state, Date.now())) {
+			getLogger(ctx).debug("skip_breaker", { route: "blocking" });
+			return { applied: false, reason: "rejected" };
+		}
+		if (cap.shouldRejectByCap(state).cancel) {
+			getLogger(ctx).debug("skip_cap", { route: "blocking" });
+			return { applied: false, reason: "rejected" };
+		}
 		let feedbackSignal = ctx.beginCompaction?.({ reason: "extension" });
 		try {
 			if (isOpenAiRemoteCompactionModel(ctx.model)) {
@@ -520,6 +528,7 @@ export default function compactionExtension(
 						savingsRatio: sy.savingsRatio,
 					})
 				) {
+					state = cap.incrementIneffective(state);
 					getLogger(ctx).debug("ineffective_counted", {
 						tokensBefore: compactEvent.compactionEntry.tokensBefore,
 						savedTokens: sy.savedTokens,
--- a/packages/coding-agent/test/compaction/blocking-compaction-route-guards.test.ts
+++ b/packages/coding-agent/test/compaction/blocking-compaction-route-guards.test.ts
@@ -1,0 +1,111 @@
+import { fauxAssistantMessage } from "@earendil-works/pi-ai";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { FAILURE_TRIP_THRESHOLD } from "../../src/core/extensions/builtin/compaction/circuit-breaker.ts";
+import compactionExtension from "../../src/core/extensions/builtin/compaction/index.ts";
+import { softCap } from "../../src/core/extensions/builtin/compaction/per-turn-cap.ts";
+import type { ExtensionHandler } from "../../src/core/extensions/index.ts";
+import {
+	connectionErrorResponse,
+	createBeforeAgentStartEvent,
+	createBlockingContext,
+} from "../helpers/blocking-compaction-harness.ts";
+
+const registrations: Array<{ unregister: () => void }> = [];
+
+afterEach(() => {
+	for (const registration of registrations.splice(0)) registration.unregister();
+});
+
+function captureHandlers(): Map<string, ExtensionHandler<never, unknown>> {
+	const handlers = new Map<string, ExtensionHandler<never, unknown>>();
+	compactionExtension({
+		on: (event: string, handler: ExtensionHandler<never, unknown>) => handlers.set(event, handler),
+	} as never);
+	return handlers;
+}
+
+function acceptedCompactionEvent(round: number, savedTokens: number) {
+	return {
+		type: "session_compact",
+		reason: "extension",
+		requestId: `guard-${round}`,
+		accepted: true,
+		fromExtension: true,
+		willRetry: false,
+		compactionEntry: {
+			type: "compaction",
+			id: `cmp-guard-${round}`,
+			parentId: null,
+			timestamp: new Date(round).toISOString(),
+			summary: "summary",
+			firstKeptEntryId: "keep",
+			tokensBefore: 9_950,
+			fromHook: true,
+			details: {
+				structuralYield: { savedTokens, savingsRatio: savedTokens / 9_950 },
+			},
+		},
+	};
+}
+
+describe("blocking compaction route guards (issue #527)", () => {
+	it("stops hard-limit blocking attempts when the circuit breaker trips", async () => {
+		const handlers = captureHandlers();
+		const beforeAgentStart = handlers.get("before_agent_start");
+		expect(beforeAgentStart).toBeDefined();
+		const harness = createBlockingContext({ usageTokens: 9_950 });
+		registrations.push(harness.registration);
+		harness.registration.setResponses(Array.from({ length: 8 }, () => connectionErrorResponse()));
+
+		for (let attempt = 0; attempt < 8; attempt++) {
+			await beforeAgentStart?.(createBeforeAgentStartEvent() as never, harness.ctx);
+		}
+
+		expect(harness.registration.state.callCount).toBe(FAILURE_TRIP_THRESHOLD);
+	});
+
+	it("bounds successful hard-limit blocking compactions by the soft cap", async () => {
+		const handlers = captureHandlers();
+		const beforeAgentStart = handlers.get("before_agent_start");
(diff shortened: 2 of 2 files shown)


effortprogrammer:
Closing: live debugging on a real stuck session (Node inspector attached to the running TUI process) showed the actual user-facing infinite-compaction symptom comes from a different loop — the turn-retry ↔ pre-prompt compaction cycle (~60s breaker-cooldown cadence) plus a compaction lifecycle leak that leaves the UI spinner open permanently. The applyBlockingCompaction guard bypass fixed here is real but was never exercised by the live session (zero hits on that path while the loop was active). I'll open a separate PR for the actual root cause. The cap/breaker bypass documented in #527 remains a valid latent issue; happy to resubmit this fix separately if maintainers want it.

effortprogrammer:
Reopening: this fix is for issue #527, which turned out to be a **different defect** from the one that bricked the live session (that one is #531, fixed by #532).

Why this still matters, and more than before: once #532 lands, prompts over the soft threshold proceed during a breaker cooldown instead of dying at admission. That means sessions near the hard limit will now actually reach the `before_agent_start` hard-limit branch on every prompt — and that branch calls `applyBlockingCompaction()` without consulting the tripped circuit breaker, paying one doomed summarization request per prompt for the duration of an outage. #527's bypass stops being latent exactly when #532 ships.

Scope recap (unchanged from the original PR):
- `applyBlockingCompaction()` entry now enforces the circuit breaker and the per-turn soft cap, covering all four blocking triggers (before_agent_start hard-limit/proactive, agent_end idle, turn_end zero-yield recovery, degradation-monitor recovery).
- `session_compact` accepted branch wires the previously dead `incrementIneffective` counter so zero-yield compactions count toward the soft cap.
- Regression tests in `test/compaction/blocking-compaction-route-guards.test.ts` (red before: 8/12/5 executions where guards require 3/10/3; green after).

The earlier close was only because this PR did not address the live incident — not because the defect was invalid.

code-yeongyu:
Maintainer update: rebased the issue-scoped fix onto current `main` and replaced the polling-based regression with deterministic route-admission assertions. Local verification: 3/3 issue regressions, 47 files / 362 scoped tests, root `npm run check`, and real source CLI mock-loop 43/43. The compaction-remote QA remains 5/6 solely on the same pre-existing replay assertion reproduced on pristine current `main`.

code-yeongyu:
CI receipt: run [30598775271](https://github.com/code-yeongyu/senpi/actions/runs/30598775271) passed all six platform lanes (Inspector handoff on Ubuntu/macOS/Windows and Terminal tools on Ubuntu/macOS/Windows). The required `Check and test` lane failed only because repository `main` commits `.tmp/issue-541-qa` as an absolute machine-local symlink; Biome reports it broken on runners. The same failure is present on pristine `main` in [run 30598648840](https://github.com/code-yeongyu/senpi/actions/runs/30598648840). Local lockfile-faithful `npm run check` is green. Proceeding with the admin merge bypass for this pre-existing, unrelated required-check failure.

PR #532, opened 2026-07-30
fix(compaction): let admission proceed when the breaker cancelled opportunistic compaction

### LINKED ISSUE #531: Breaker-tripped compaction hard-blocks prompt admission forever — session stuck in infinite compaction rejection loop
## Summary

When the compaction circuit breaker is tripped (provider outage / persistent summarization failure), every prompt to a session whose context is over the soft compaction threshold is rejected with `Context remains above the compaction threshold because compaction did not complete` — forever. The breaker's whole purpose is to stop paying for doomed compaction attempts during an outage, but core converts the breaker's cancel into a hard admission failure, so the session can never make progress again until the user manually intervenes. Any autonomous continuation driver (extensions that re-prompt on `agent_end`) loops indefinitely, and the user just sees a permanent "Compacting..." / rejection cycle with no path forward.

## Live evidence (captured from a real stuck session via the Node inspector)

A real TUI session (context ~177k tokens, provider returning `content flagged` / `Connection error.`) was observed live:

- `CompactionStatusIndicator.render` firing ~13/s — the compaction spinner never closes; the session cannot advance.
- Summarization attempts (`generateSummaryMessage → consumeStreamWithIdleTimeout`) recurring at a ~60s cadence — exactly the breaker cooldown period — each one failing.
- After the cycle stopped, the process kept burning >100% CPU in the render loop with zero compaction entry-point hits and zero session writes: the run is parked for good.

## Sandboxed reproduction (real CLI, mock provider)

`.agents/skills/senpi-qa/scripts/scenarios/compaction-retry-cycle-repro.mjs` (attached to the PR): turn 1 succeeds with usage pinned over the threshold; every later provider request fails with a transient `Connection error.`; the driver then re-prompts like an autonomous continuation extension. Observed on unpatched `main`:

```
continuation 2..10 -> success:false "Context remains above the compaction threshold because compaction did not complete"
compaction_start reason=pre_prompt -> compaction_end accepted=false   (repeats for every prompt)
```

The breaker does trip (later attempts are cancelled instantly without network calls), but the admission rejection is terminal per prompt, so nothing ever proceeds.

## Root cause

`session_before_compact` cancel with `rejectionCause: "circuit-breaker"` flows into `_rejectCompaction` → compaction lifecycle `failed` → `_runPrePromptCompaction` returns `false` → every admission site throws `RequiredCompactionError` (or returns `blocked`). Opportunistic compaction (threshold / pre_prompt) is an optimization: when the breaker says "compaction is unavailable right now", the correct degradation is to attempt the turn anyway — the provider either accepts (context often still fits the window; only the soft threshold is exceeded) or returns a real overflow error, which the existing one-shot overflow recovery handles. Hard-overflow admission stays fail-closed.

## Fix

Admission sites that currently throw on failed opportunistic compaction now proceed when the last compaction rejection was `circuit-breaker`:

- `_enforceCompactionBeforeProvider` (pre-prompt threshold tail)
- `_enforceFinalProviderAdmission` (only when the final payload is not actually oversized)
- `_revalidateScheduledContinuationAdmission`
- `_handleRetryableError` (retry proceeds instead of returning `blocked`)

Overflow-triggered compaction remains fail-closed.

A PR with regression tests and before/after CLI evidence follows.

effortprogrammer:
## What changed

When the compaction circuit breaker cancels an opportunistic (threshold / pre-prompt) compaction, admission now proceeds without compaction instead of throwing `RequiredCompactionError`:

- `_enforceCompactionBeforeProvider` — pre-prompt threshold tail returns "not compacted" on cooldown instead of throwing
- `_enforceFinalProviderAdmission` — proceeds on cooldown only when the final payload is not actually oversized
- `_revalidateScheduledContinuationAdmission` — proceeds on cooldown
- `_handleRetryableError` — retries the turn instead of returning `"blocked"` on cooldown

Overflow-triggered admission is untouched (fail-closed). If the context genuinely exceeds the window, the provider's overflow error drives the existing one-shot overflow recovery, so proceeding is safe.

## Why

The breaker's job is to stop paying for doomed compaction during a provider outage. But core converted the breaker cancel (`rejectionCause: "circuit-breaker"`) into a hard admission failure on every path, so a tripped breaker bricked any session over the soft threshold: every prompt rejected with `Context remains above the compaction threshold because compaction did not complete`, forever. Any autonomous continuation driver (extensions that re-prompt on `agent_end`) loops without bound.

Live forensics on a real stuck session (issue #531 has the full timeline): compaction spinner rendering ~13/s, summarization retries at the exact 60s breaker-cooldown cadence, and a final parked state burning >100% CPU with zero session writes.

Opportunistic compaction is an optimization; when the breaker declares compaction unavailable, the honest degradation is to let the turn try — the provider either accepts (soft threshold exceeded ≠ window exceeded) or returns a real overflow error.

## Observed behavior (real CLI, mock provider, A/B)

Driver: `.agents/skills/senpi-qa/scripts/scenarios/compaction-retry-cycle-repro.mjs` (included; turn 1 pins usage over the threshold, then all provider requests fail transiently, and the driver re-prompts like an autonomous continuation extension).

| | unpatched | patched |
|---|---|---|
| prompts rejected due to compaction (after breaker trip) | 8/8 | 0/8 |
| provider turn requests | 4 (then nothing ever again) | 13 (turns keep flowing) |
| `agent_end` after prompt 2+ | 0 | 10 |
| `compaction_start` without `compaction_end` | 0 | 0 |

Full logs: `local-ignore/qa-evidence/20260730-issue-531-cooldown-admission/{unpatched,patched}.log`.

## Tests

- New `test/suite/regressions/531-compaction-cooldown-admission.test.ts`: red before the fix (`RequiredCompactionError` at the exact admission site), green after; a non-breaker cancel control verifies admission still fails closed for generic refusals.
- `npx vitest --run test/compaction/ test/compaction*.test.ts test/suite/ test/agent-session-auto-compaction-queue.test.ts test/interactive-mode-compaction*.test.ts`: 364 files, 2466 passed, 1 skipped.
- `npx biome check` clean; root `npx tsgo --noEmit` exit 0.

## Residual risk

- During an outage, sessions over the soft threshold now pay one real provider attempt per prompt instead of dying at admission; that is the intended degradation and produces a visible provider error rather than an invisible compaction loop.
- The lifecycle-leak aspect of the incident (spinner left open after the cycle) did not reproduce in the sandbox (all start/end pairs matched in both runs); if it recurs it deserves its own fix.

closes #531

---
## Summary by cubic
Let admission proceed without compaction when the circuit breaker cancels opportunistic compaction, so sessions over the soft threshold keep running during provider outages. Final oversized payloads and overflow-triggered paths still fail closed.

- **Bug Fixes**
  - `_enforceCompactionBeforeProvider`, `_revalidateScheduledContinuationAdmission`: proceed on breaker cooldown instead of throwing.
  - `_enforceFinalProviderAdmission`: proceed on cooldown only if the final payload isn’t oversized.
  - `_handleRetryableError`: retry the turn instead of returning "blocked" when on cooldown.
  - Overflow-triggered admission remains unchanged (fail-closed).
  - Added regression test `test/suite/regressions/531-compaction-cooldown-admission.test.ts` and a CLI repro script `compaction-retry-cycle-repro.mjs`; the repro uses the repo-standard mock key `sk-mock-qa-7f3a` to avoid secret scanner noise.

<sup>Written for commit 3ff2c6f65ae268608b4b3b577cf1966ad43ca9c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/skills/senpi-qa/scripts/scenarios/compaction-retry-cycle-repro.mjs
+++ b/.agents/skills/senpi-qa/scripts/scenarios/compaction-retry-cycle-repro.mjs
@@ -1,0 +1,271 @@
+#!/usr/bin/env node
+/**
+ * Repro driver: persistent provider failure + over-threshold context.
+ *
+ * Turn 1 succeeds with huge usage (context pinned near the hard limit).
+ * Every later request (turns AND compaction summarization) fails with a
+ * transient "Connection error." — the same shape as a real provider outage.
+ *
+ * Observation targets:
+ *   - whether compaction_start/compaction_end cycle forever without the agent
+ *     ever settling (infinite retry<->compaction cycle), and
+ *   - whether the final state leaves a compaction_start with no matching end
+ *     (lifecycle leak = the "Compacting..." spinner that never closes).
+ *
+ * Usage: node compaction-retry-cycle-repro.mjs [--watch-ms 240000]
+ */
+
+import { createServer } from "node:http";
+import { writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { createChecks, guardRealAuth, installCleanupHooks, makeSandbox, spawnCli } from "../lib/common.mjs";
+import { checkRealAuthUnchanged, hermeticEnv } from "../lib/mock-loop-support.mjs";
+
+const CONTEXT_WINDOW = 128_000;
+const PINNED_INPUT_TOKENS = 120_000;
+const SUMMARY_MARKER = "context summarization assistant";
+
+function sseHead(res) {
+	res.writeHead(200, {
+		"content-type": "text/event-stream",
+		"cache-control": "no-cache",
+		connection: "keep-alive",
+	});
+}
+
+function writeAnthropicSse(res, text, modelId, inputTokens) {
+	sseHead(res);
+	const ev = (event, data) => res.write(`event: ${event}\ndata: ${JSON.stringify({ type: event, ...data })}\n\n`);
+	ev("message_start", {
+		message: {
+			id: "msg_mock",
+			type: "message",
+			role: "assistant",
+			model: modelId,
+			content: [],
+			stop_reason: null,
+			stop_sequence: null,
+			usage: { input_tokens: inputTokens, output_tokens: 0 },
+		},
+	});
+	ev("content_block_start", { index: 0, content_block: { type: "text", text: "" } });
+	ev("content_block_delta", { index: 0, delta: { type: "text_delta", text } });
+	ev("content_block_stop", { index: 0 });
+	ev("message_delta", { delta: { stop_reason: "end_turn", stop_sequence: null }, usage: { output_tokens: 40 } });
+	ev("message_stop", {});
+	res.end();
+}
+
+function startFlappingServer() {
+	const requests = [];
+	let callIndex = 0;
+	const server = createServer((req, res) => {
+		const chunks = [];
+		req.on("data", (c) => chunks.push(c));
+		req.on("end", () => {
+			const raw = Buffer.concat(chunks).toString("utf8");
+			let body = {};
+			try {
+				body = raw ? JSON.parse(raw) : {};
+			} catch {}
+			const systemText = typeof body.system === "string" ? body.system : JSON.stringify(body.system ?? "");
+			const isSummarization = systemText.includes(SUMMARY_MARKER);
+			requests.push({ at: Date.now(), url: req.url, summarization: isSummarization, systemPrefix: systemText.slice(0, 60) });
+			callIndex++;
+			if (callIndex === 1) {
+				return writeAnthropicSse(res, "x".repeat(400_000), body.model ?? "mock-claude", PINNED_INPUT_TOKENS);
+			}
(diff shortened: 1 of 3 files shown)


code-yeongyu:
[sisyphus-bot] Thank you for the thorough investigation, regression test, and CLI repro. This fix was carried onto current `main` in #556 with your implementation preserved and co-author attribution; #556 passed all required checks, merged as `264078608fa15caea32aa4c1c7814bee4d2a1faa`, and closed #531. I am closing this now-superseded PR to avoid duplicate or conflicting delivery.

PR #535, opened 2026-07-30
fix(interactive): restore terminal compaction input to editor

realsigridjin:
## Summary

- restore compaction-queued TUI input to the editable composer after terminal failures, rejections, and aborts
- keep accepted compactions on normal prompt admission and retryable failures on the native deferred queue path
- preserve message ordering, clear queue ownership, and avoid automatic resend against the unchanged over-threshold context
- add focused branch, rejection-feedback, and post-compaction ownership regressions

## Why

PR #512 fixed permanently parked input by transferring failed-compaction messages to native queues. That transfer is still correct while a retry is pending, but a terminal failure has no recovery turn left: automatically sending the same input can immediately re-enter compaction against the same context. Returning terminal-failure input to the composer gives ownership back to the user without dropping or auto-sending it.

## Verification

- `npm run check` — 2,533 files checked, no fixes or errors
- focused changed suites — 3 files, 24 tests passed
- expanded compaction recovery suite — 5 files, 32 tests passed
- `npm run build` — all workspaces passed
- real TUI QA — `RECOVERY_COMPOSER_RESTORED=1`, `QUEUED_BADGE_CLEARED=1`, `AUTO_RESEND_COUNT=0`, `AUTH_UNCHANGED=1`, `CLEANUP_OK=1`
- rebased onto current `origin/main`; merge-tree clean

Related: #512

---
## Summary by cubic
Restore compaction-queued input to the editor after terminal compaction failures, rejections, or aborts. Failed pre-prompt compaction now emits `willRetry: false` so restoration is reliable; retryable failures still defer to native queues and accepted compactions send normally.

- **Bug Fixes**
  - `interactive-mode.ts`: three-path `compaction_end` handling
    - Success: `flushCompactionQueue({ willRetry, deferAdmission: false })`.
    - Retryable failure: `flushCompactionQueue({ willRetry: true, deferAdmission: true })`.
    - Terminal failure/rejection/abort: `restoreQueuedMessagesToEditor()` and show “restored to the editor”.
  - `agent-session.ts`: failed pre-prompt overflow compaction now emits `willRetry: false` to prevent parked input and trigger restoration.
  - Terminal restoration returns queued TUI input (and native steer/follow-up handoffs) to the editor, avoiding resend into the same over-threshold context.
  - Tests/Docs: deduplicated compaction test context setup; exercised real queued-input restoration on a failed pre-prompt overflow; updated `changes.md`.

<sup>Written for commit 1ecc62fafa8885c4fcc7b1bbb81be301e4131687. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/535?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/modes/interactive/changes.md
+++ b/packages/coding-agent/src/modes/interactive/changes.md
@@ -1,5 +1,21 @@
 # changes
 
+## Failed compaction restores queued input (2026-07-30)
+
+### What changed
+
+- Terminal compaction failures, rejections, and aborts restore messages queued during compaction to the editable composer instead of leaving them pending indefinitely.
+- Retryable compaction failures keep the native-queue handoff (`flushCompactionQueue({ willRetry: true, deferAdmission: true })`) so the queued input rides along with the retry.
+- Coverage: `test/interactive-mode-compaction.test.ts` pins truncated, timed-out, rejected, retryable, and successful compaction outcomes.
+
+### Why
+
+- Retrying queued delivery against the unchanged over-threshold context repeats the same required-compaction failure; restoring the draft lets the user edit or retry explicitly.
+
+### Expected merge conflict zones
+
+- LOW: `interactive-mode.ts` in the `compaction_end` queue handoff branch.
+
 ## Risky main-model selection warning (2026-07-30)
 
 - Interactive startup, `/model` (including exact references), the full and favorite-model selectors, post-auth default selection, and favorite rotation now pass the selected main model through one shared warning predicate.
--- a/packages/coding-agent/src/modes/interactive/changes.md
+++ b/packages/coding-agent/src/modes/interactive/changes.md
@@ -4,9 +4,10 @@
 
 ### What changed
 
-- Terminal compaction failures, rejections, and aborts restore messages queued during compaction to the editable composer instead of leaving them pending indefinitely.
-- Retryable compaction failures keep the native-queue handoff (`flushCompactionQueue({ willRetry: true, deferAdmission: true })`) so the queued input rides along with the retry.
-- Coverage: `test/interactive-mode-compaction.test.ts` pins truncated, timed-out, rejected, retryable, and successful compaction outcomes.
+- Terminal compaction failures, rejections, and aborts restore messages queued during compaction to the editable composer instead of leaving them pending indefinitely. Native session steer and follow-up queues are drained into the composer too.
+- Failed pre-prompt overflow compaction now emits `willRetry: false`, so queued input follows the terminal restoration path instead of waiting for a retry that cannot run.
+- The defensive retryable-failure branch keeps the native-queue handoff (`flushCompactionQueue({ willRetry: true, deferAdmission: true })`) for any producer that can truthfully promise a retry.
+- Coverage: `test/interactive-mode-compaction.test.ts` pins truncated, timed-out, rejected, retryable, and successful compaction outcomes; `test/suite/regressions/post-compaction-queued-input-resume.test.ts` drives the real failed pre-prompt overflow path and real editor-restoration helper.
 
 ### Why
 
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -3891,23 +3891,43 @@ export class InteractiveMode {
 						this.chatContainer.addChild(new Text(theme.fg("error", message), 1, 0));
 					}
 				}
-				// Every terminal compaction_end flushes the TUI compaction queue.
-				// Accepted compactions deliver through prompt admission; unsuccessful
-				// ones route through the native steer/followUp queues so submitted
-				// input is never silently parked (field bug: messages typed during a
-				// failing compaction were held forever and lost on session switch).
+				// Every terminal compaction_end resolves the TUI compaction queue.
+				// Accepted compactions deliver through prompt admission. Retryable
+				// failures route through the native steer/followUp queues so submitted
+				// input rides along with the retry instead of being silently parked
+				// (field bug: messages typed during a failing compaction were held
+				// forever and lost on session switch). Terminal failures hand the input
+				// back to the editable composer: retrying delivery against the same
+				// over-threshold context would just repeat the failure.
 				const compactionSucceeded = event.accepted === true || event.result !== undefined;
 				const heldCount = this.compactionQueuedMessages.length;
-				if (!compactionSucceeded && heldCount > 0) {
-					this.getSessionLogger().warn("compaction_queue_deferred", {
-						count: heldCount,
-						cause: event.errorMessage ?? event.rejectionCause ?? (event.aborted ? "aborted" : "no-result"),
-					});
-					this.showStatus(
-						`${heldCount} queued message${heldCount === 1 ? "" : "s"} will send with the next turn (compaction did not complete)`,
-					);
+				const failureCause =
+					event.errorMessage ?? event.rejectionCause ?? (event.aborted ? "aborted" : "no-result");
+				if (compactionSucceeded) {
+					void this.flushCompactionQueue({ willRetry: event.willRetry, deferAdmission: false });
+				} else if (event.willRetry === true) {
+					if (heldCount > 0) {
+						this.getSessionLogger().warn("compaction_queue_deferred", {
+							count: heldCount,
+							cause: failureCause,
+						});
+						this.showStatus(
+							`${heldCount} queued message${heldCount === 1 ? "" : "s"} will send with the next turn (compaction will retry)`,
+						);
+					}
+					void this.flushCompactionQueue({ willRetry: true, deferAdmission: true });
+				} else {
+					const restoredCount = this.restoreQueuedMessagesToEditor();
+					if (restoredCount > 0) {
+						this.getSessionLogger().warn("compaction_queue_restored", {
+							restored: restoredCount,
+							cause: failureCause,
+						});
+						this.showStatus(
+							`${restoredCount} queued message${restoredCount === 1 ? "" : "s"} restored to the editor (compaction did not complete)`,
+						);
+					}
 				}
-				void this.flushCompactionQueue({ willRetry: event.willRetry, deferAdmission: !compactionSucceeded });
 				this.ui.requestRender();
 				break;
 			}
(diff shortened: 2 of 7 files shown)


packages/coding-agent/src/modes/interactive/interactive-mode.ts
```diff
@@ -3891,23 +3891,43 @@ export class InteractiveMode {
...
-						cause: event.errorMessage ?? event.rejectionCause ?? (event.aborted ? "aborted" : "no-result"),
-					});
-					this.showStatus(
-						`${heldCount} queued message${heldCount === 1 ? "" : "s"} will send with the next turn (compaction did not complete)`,
-					);
+				const failureCause =
+					event.errorMessage ?? event.rejectionCause ?? (event.aborted ? "aborted" : "no-result");
+				if (compactionSucceeded) {
+					void this.flushCompactionQueue({ willRetry: event.willRetry, deferAdmission: false });
+				} else if (event.willRetry === true) {
```
code-yeongyu, line 3905: This branch's premise deserves a second look, kindly but firmly: as far as I can trace, the **only** production site that emits a *failed* `compaction_end` with `willRetry: true` is `_runPrePromptCompaction`'s catch (`agent-session.ts` ~4435, reached via the overflow-recovery call at ~4319). But in that exact path the caller then throws `RequiredCompactionError` — the intended retry never runs. So the input we hand to the native queues here doesn't "ride along with the retry"; it sits in the steer/followUp queues until the user's next prompt, which will hit the same over-threshold wall. (`_runAutoCompaction`'s catch always emits `willRetry: false`, and rejection/abort sites are all `willRetry: false` too.)

Under this PR's own rationale, that makes the willRetry-true failure case *terminal in practice*, and arguably it belongs in the restore-to-editor arm as well. I can also see the conservative argument for keeping the #512 behavior here since the event contract says a retry was intended — but then the comment above ("input rides along with the retry") and the status text ("compaction will retry") overpromise.

Could you either (a) point me at a real path where a retry genuinely follows a `willRetry: true` failure — happy to be wrong! — or (b) fold this case into the restore branch, or (c) keep the routing but soften the comment/status to match reality? Related: the new regression for this branch builds the event by hand, so nothing currently pins the real emission shape.

packages/coding-agent/test/suite/regressions/post-compaction-queued-input-resume.test.ts
```diff
@@ -117,11 +120,27 @@ function createTuiCompactionEventContext(harness: Harness) {
...
 		getSessionLogger: () => ({ debug: () => {}, info: () => {}, warn: () => {} }),
 		flushes,
-		flushCompactionQueue(options: { willRetry: boolean }) {
+		restoredMessages,
+		flushCompactionQueue(options: { willRetry: boolean; deferAdmission?: boolean }) {
 			const flush = getFlushCompactionQueue()(context, options);
 			flushes.push(flush);
 			return flush;
 		},
+		restoreQueuedMessagesToEditor() {
```
code-yeongyu, line 160: Small fidelity worry: this fake re-implements `restoreQueuedMessagesToEditor` by hand (the comment even says "mirrors the real helper"), while `flushCompactionQueue` right above it goes through `Reflect.get(InteractiveMode.prototype, ...)` to exercise the real method. If the real helper ever drifts — say it stops draining the native queues, or changes ordering — these regressions keep passing while the behavior they exist to pin changes underneath them.

Could we bind the real prototype method here too (it only needs `editor.getText/setText`, `updatePendingMessagesDisplay`, and `clearAllQueues`-reachable state on the context)? If that's too invasive for this PR, even a one-off assertion that the mirror and the real helper agree on a sample queue would keep them honest.

packages/coding-agent/src/modes/interactive/changes.md
```diff
@@ -1,5 +1,21 @@
 # changes
 
+## Failed compaction restores queued input (2026-07-30)
+
+### What changed
+
+- Terminal compaction failures, rejections, and aborts restore messages queued during compaction to the editable composer instead of leaving them pending indefinitely.
```
code-yeongyu: Gentle accuracy nit: `restoreQueuedMessagesToEditor()` goes through `clearAllQueues()`, which drains the **native session steer/followUp queues** into the composer too — not only "messages queued during compaction". So a steering message queued during the *turn* (before compaction ever started) also lands in the editor on a terminal compaction failure. I think that behavior is actually right (the next turn can't start against the same over-threshold context either), but this doc — and the `compaction_queue_restored` log naming — read narrower than what ships. Worth one extra sentence here so the next merger isn't surprised.

packages/coding-agent/test/interactive-mode-compaction.test.ts
```diff
@@ -211,9 +212,117 @@ describe("InteractiveMode compaction events", () => {
...
+				chatContainer: { clear: vi.fn(), addChild: vi.fn() },
+				rebuildChatFromMessages: vi.fn(),
+				addMessageToChat: vi.fn(),
+				showError: vi.fn(),
+				showStatus: vi.fn(),
+				clearStatusIndicator: vi.fn(),
+				compactionQueuedMessages: [{ text: "queued during compaction", mode: "steer" as const }],
+				getSessionLogger: () => ({ debug: vi.fn(), info: vi.fn(), warn: vi.fn() }),
+				flushCompactionQueue: vi.fn().mockResolvedValue(undefined),
+				restoreQueuedMessagesToEditor: vi.fn().mockReturnValue(1),
```
code-yeongyu, line 213: Non-blocking nit: this file now repeats the ~20-line `fakeThis` literal five times with tiny deltas. A `makeCompactionFakeThis(overrides)` factory would shrink the new tests to their actually-interesting three lines each and make the *next* compaction test cheaper to add. Totally fine as a follow-up if you'd rather keep this diff focused.

code-yeongyu (commented):
Thanks for this — the direction is exactly right, and the split between accepted / retryable / terminal outcomes is a genuine improvement over the blanket `deferAdmission: !compactionSucceeded` from #512. The regression coverage you added (especially driving the real `compaction_end` through the real session in `post-compaction-queued-input-resume.test.ts`) is lovely. I ran all five touched test files locally: 32/32 green.

A few things I'd love your eyes on before merge — one of them questions whether the retryable branch can ever fire the way the comment promises:

1. **Does a `willRetry: true` failure ever actually retry?** (inline comment on `interactive-mode.ts`) — this is the only one I'd call substantive.
2. **The fake `restoreQueuedMessagesToEditor` in the resume regression mirrors the real helper by hand** — a drift there silently weakens exactly the regressions this PR adds (inline).
3. **`changes.md` undersells the blast radius**: the restore path drains the *native* session queues too, not just messages queued during compaction (inline).
4. Tiny tidy, no action needed: `heldCount` and `failureCause` are computed before the branch but each is only used in one arm; hoisting them reads as if all arms need them.

Separately, the `Check and test` CI job died with a ~25-minute silent hang inside the `packages/ai` suite (after `packages/agent` passed) — nothing in this diff touches `packages/ai`, so it smells like an infra flake; I've triggered a rerun to confirm.

code-yeongyu (approved):
Re-reviewed the follow-up commits: the truthful `willRetry: false` emission with a real-path regression, the real restoration helper bound in the resume tests, the docs/scoping cleanups, and the main merge all check out — 33/33 focused tests, full compaction scope green, `npm run check` clean, and QA channels (mock-loop 43/43, TUI smoke 5/5) pass. Lovely work — merging.

code-yeongyu:
[sisyphus-bot] Post-merge brutal audit — VERDICT: REQUEST_CHANGES (follow-up fix recommended)

This PR is already merged; posting as a post-merge audit. One real defect survives review, verified at `1ecc62faf`.

### MEDIUM — terminal restoration poisons the next abort with stale queue state

**`packages/coding-agent/src/modes/interactive/interactive-mode.ts:3920`**
```
const restoredCount = this.restoreQueuedMessagesToEditor();
```
**Causal line — `packages/coding-agent/src/core/agent-session.ts:3013`**
```
if (steering.length > 0 || followUp.length > 0) this._hadClearedQueuedMessages = true;
```

Chain: the new terminal-failure branch calls `restoreQueuedMessagesToEditor()` → `clearAllQueues()` → `AgentSession.clearQueue()`, which sets `_hadClearedQueuedMessages = true` whenever native steering/follow-up existed. But this path **never calls `session.abort()`**, so the flag persists. A later otherwise-idle abort then consumes it:

**`packages/coding-agent/src/core/agent-session.ts:3066`**
```
const shouldEmitAbort = !wasMidRun && (hadRetryBackoff || hadCompactionOrPending || hadClearedQueues);
```

That marker was designed for the atomic "clear queues, then immediately abort" path. Reusing the clear without the abort turns a transient gap marker into persistent session state.

This is not just a cosmetic notification. `session_abort` is an extension control-plane event — the goal extension treats it as an instruction to **block an active goal**. So: native input exists during compaction → compaction fails terminally → input is correctly restored but the flag leaks → later idle abort → spurious `session_abort` → unrelated active goal gets blocked.

The new integration test seeds exactly the prerequisite (native steering + follow-up before terminal restoration) but stops before a subsequent idle `abort()`, so it verifies editor contents while missing the leaked lifecycle state.

**Fix:** separate "drain queues for editing" from "drain queues immediately before abort" — e.g. `clearQueue({ abortWillFollow })`, setting `_hadClearedQueuedMessages` only when an abort will follow. Thread it through: terminal compaction restoration and manual dequeue-to-editor ⇒ `false`; `abortAndFireQueuedMessages()` and `restoreQueuedMessagesToEditor({ abort: true })` ⇒ `true`. Add a regression that drives terminal `compaction_end`, then calls `session.abort()` while idle and asserts **no** `session_abort` is emitted — plus the complement proving the deliberate clear-then-abort path still emits it.

### What audited clean

No defect in the one-line `agent-session.ts` change forcing failed `_runPrePromptCompaction()` to `willRetry: false` — tracing all callers, an unsuccessful pre-prompt compaction blocks the pending retry or provider admission; none proceeds with the promised retry. Accepted compactions preserve their original `willRetry`, and explicit rejections already terminate with `false`. The three interactive branches match the native queue contracts, and terminal failure correctly avoids re-admission into the unchanged over-threshold context.

**Verification:** source-level audit; isolated worktree has no `vitest` installed so focused tests could not start, and type-checking surfaced only errors outside the changed files. Worktree left clean.

PR #544, opened 2026-07-30
fix(compaction): dispatch summarization through the model runtime

### LINKED ISSUE #543: Compaction fails for runtime-registered providers: No API provider registered for api: <api> (claude-agent-sdk, kiro)
### What happened?

Compaction fails for every model served by a provider registered through `pi.registerProvider()` — i.e. the builtin `claude-agent-sdk` provider and extension providers such as `senpi-accounts`' Kiro (`kiro-codewhisperer`):

```
Error: Compaction rejected: compaction generator failed: No API provider registered for api: kiro-codewhisperer
```

The same session's normal agent turns work fine on the same model — only compaction breaks, so the session eventually deadlocks against the context window with no way to compact.

Reproduced on `claude-agent-sdk` (builtin, api `claude-agent-sdk`) and on Kiro (`senpi-accounts`, api `kiro-codewhisperer`).

### Root cause

The compaction extension's summarization request is dispatched through compat's api-registry:

- [`compaction/speculative.ts#L11`](https://github.com/code-yeongyu/senpi/blob/77a768096/packages/coding-agent/src/core/extensions/builtin/compaction/speculative.ts#L11) imports `stream` from `@earendil-works/pi-ai/compat`, and [`speculative.ts#L222`](https://github.com/code-yeongyu/senpi/blob/77a768096/packages/coding-agent/src/core/extensions/builtin/compaction/speculative.ts#L222) (`generateSummaryMessage`) calls it.
- compat only knows the 10 builtin api ids (`packages/ai/src/compat.ts` `BUILTIN_APIS`), and `resolveApiProvider` throws `No API provider registered for api: <api>` for anything else.
- Providers registered via `pi.registerProvider()` live only in Senpi's `ModelRuntime` / `provider-composer`, never in compat's registry.

The main agent loop is unaffected because it dispatches through `agent.streamFunction` = `modelRuntime.streamSimple` (`core/sdk.ts`). The core `compact()` route also takes a `streamFn` and is passed `this.agent.streamFunction` (`agent-session.ts`), so it is fine too — but the compaction extension route (which is the route actually taken whenever the builtin compaction extension is loaded) bypasses both.

This is the same bug class as #488 (`/btw`), fixed in ff48536d1 by routing the side query through `ctx.modelRegistry.modelRuntime`. `SpeculativeCompactionContext` already carries `modelRegistry` (it uses it for auth at `speculative.ts#L476`), so the runtime is already in hand at the failing call site.

### Steps to reproduce

1. Log in to a provider whose api id is not one of compat's builtins (`/login claude-agent-sdk`, or install `senpi-accounts` and `/login kiro`).
2. Select one of its models.
3. Fill enough context and run `/compact`.

Expected: a compaction summary is applied.
Actual: `Compaction rejected: compaction generator failed: No API provider registered for api: <api>`.

### Environment

senpi 2026.7.30, darwin arm64, node v24.14.0

eddieparc:
Fixes https://github.com/code-yeongyu/senpi/issues/543

## Problem

Compaction failed for every model served by a provider registered through `pi.registerProvider()` — the builtin `claude-agent-sdk` provider and extension providers such as `senpi-accounts`' Kiro:

```
Compaction rejected: compaction generator failed: No API provider registered for api: kiro-codewhisperer
```

`compaction/speculative.ts` dispatched the summarization request with compat's `stream()`, and compat's api-registry only knows the 10 builtin api ids (`packages/ai/src/compat.ts` `BUILTIN_APIS`); runtime-registered providers live solely in `ModelRuntime` / `provider-composer`. Normal agent turns worked because they dispatch through `agent.streamFunction` = `modelRuntime.streamSimple`, so a session on such a provider ran fine until it needed to compact — then it had no way out of the context window.

## Change

Route the summarization stream through `context.modelRegistry.modelRuntime.stream()` when a registry is present, keeping the compat fallback for `SpeculativeCompactionContext` values constructed without one. `SpeculativeCompactionContext` already carried `modelRegistry` (it resolves summarization auth with it), so nothing new is threaded through. Same fix shape as #488 for `/btw` (ff48536d1).

## Tested

- `test/suite/regressions/543-compaction-runtime-provider.test.ts` — new regression registering a provider whose api id exists only in `ModelRuntime`, then compacting. Captured RED before the fix with `compaction generator failed: No API provider registered for api: runtime-only-compaction-api`, GREEN after.
- `npx vitest --run test/compaction test/suite/regressions` — 604 passed. 4 pre-existing failures (`0000-multi-session-theme-init`, `2791-fswatch-error-crash`, `codemode-builtin-dedupe` x2) reproduce identically on an unmodified `origin/main` checkout of the same worktree; unrelated to this change.
- `npx tsgo --noEmit` clean, `biome check --error-on-warnings` clean on the changed files, root `npm run build` exit 0.
- Real surface: local senpi built from this branch, running on the Kiro provider with a Claude model, `/compact` applies a summary instead of rejecting.

---
## Summary by cubic
Fixes compaction for runtime-registered providers by routing summarization through the model runtime, with compat as fallback. Adds support for header-auth providers and routes native remote compaction via the runtime. Fixes #543.

- **Bug Fixes**
  - Route summarization and remote-compaction streams via `context.modelRegistry.modelRuntime` when available; fallback to compat. Injected runners still win in `openai-remote.ts`.
  - Accept credential request headers as valid auth for summarization; `apiKey` no longer required if headers are provided.
  - Added regression tests: `test/suite/regressions/543-compaction-runtime-provider.test.ts`, `test/suite/regressions/543-remote-compaction-runtime-provider.test.ts`.

<sup>Written for commit 17b4eba16717bc8a33ec3c430afc2c4b80c2779f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/544?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/compaction/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/compaction/changes.md
@@ -5,15 +5,19 @@
 ### What changed
 
 - `speculative.ts` dispatches the summarization request through `context.modelRegistry.modelRuntime.stream()` when a registry is present, and only falls back to the compat `stream()` when a `SpeculativeCompactionContext` is built without one.
-- Added issue #543 regression coverage in `test/suite/regressions/543-compaction-runtime-provider.test.ts` with a provider whose api id exists only in Senpi's `ModelRuntime`.
+- `openai-remote.ts` resolves its stream runner the same way (`resolveRemoteStreamRunner`): an injected `dependencies.streamRunner` still wins, otherwise the model runtime serves the native remote-compaction request and compat is the last resort.
+- Summarization auth now accepts a credential request header as resolved auth instead of requiring an `apiKey`, so `headers`-authenticated providers (models.json `headers`, extension `headers`) can compact.
+- Issue #543 regression coverage: `test/suite/regressions/543-compaction-runtime-provider.test.ts` (runtime-only api id, plus a header-authenticated provider) and `test/suite/regressions/543-remote-compaction-runtime-provider.test.ts` (native remote route through the runtime).
 
 ### Why
 
 - Providers registered through `pi.registerProvider()` (builtin `claude-agent-sdk`, extension providers such as `senpi-accounts`' Kiro) never land in compat's builtin api-registry, so every compaction attempt failed with `compaction generator failed: No API provider registered for api: <api>` while normal agent turns on the same model worked. Same bug class as #488 for `/btw`.
+- The two follow-on holes had the same shape: a provider that senpi considers fully authenticated and fully routable for normal turns must be equally compactable. A header-only credential resolved `{ok: true, apiKey: undefined}` and died as "credentials unavailable"; an extension `openai-responses` proxy opting into `supportsRemoteCompactionV2` had its own transport bypassed on the remote route.
 
 ### Merge-conflict zones
 
-- `speculative.ts` import block plus the single `stream(...)` call site in `generateSummaryMessage`.
+- `speculative.ts` import block, the single `stream(...)` call site in `generateSummaryMessage`, and the auth guard at the top of `runExtensionCompaction`.
+- `openai-remote.ts` `OpenAiRemoteCompactionContext.modelRegistry` shape plus the two stream-runner defaults.
 
 ## Proactive idle compaction (2026-07-30)
 
--- a/packages/coding-agent/src/core/extensions/builtin/compaction/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/compaction/changes.md
@@ -1,5 +1,20 @@
 # Builtin compaction extension changes
 
+## Runtime provider dispatch for summarization (2026-07-31)
+
+### What changed
+
+- `speculative.ts` dispatches the summarization request through `context.modelRegistry.modelRuntime.stream()` when a registry is present, and only falls back to the compat `stream()` when a `SpeculativeCompactionContext` is built without one.
+- Added issue #543 regression coverage in `test/suite/regressions/543-compaction-runtime-provider.test.ts` with a provider whose api id exists only in Senpi's `ModelRuntime`.
+
+### Why
+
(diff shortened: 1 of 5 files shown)


eddieparc:
Pushed `17b4eba16`: two more holes in the same class, so the fix is actually generic rather than kiro-shaped.

- Summarization no longer requires a resolved `apiKey` — a credential request header counts as resolved auth, so `headers`-authenticated providers (models.json `headers`, extension `headers`) can compact. The typed auth error for an unconfigured provider is unchanged.
- `openai-remote.ts` resolves its stream runner through the model runtime (`resolveRemoteStreamRunner`), so an extension `openai-responses` proxy that opts into `compat.supportsRemoteCompactionV2` keeps its own transport on the native remote route. Injected `dependencies.streamRunner` still wins; compat is last.

New coverage: the header-auth case in `543-compaction-runtime-provider.test.ts` and `543-remote-compaction-runtime-provider.test.ts`, both captured RED before their fix. `test/compaction` + `test/suite/regressions`: 129 files / 610 tests passing. `tsgo --noEmit`, biome, and root `npm run build` clean. Detail in https://github.com/code-yeongyu/senpi/issues/543#issuecomment-5133617035

ChickenBreast-ky:
I independently reproduced this with an extension provider using `api: "kiro-api"`: normal model requests succeeded, while compaction failed with `No API provider registered for api: kiro-api`. Routing speculative summary generation through the session `ModelRuntime` fixed a real forked-session compaction run (about 146k tokens before compaction) without changing the source session. Thanks for covering this generically rather than as a Kiro-specific workaround.

PR #552, opened 2026-07-30
fix(senpi): harden compaction recovery and Goal migration

realsigridjin:
## Summary

- add map-less GPT-5.6 Sol `max` reasoning support across capability detection, Shift+Tab cycling, and OpenAI-compatible payload serialization
- recover terminal required-compaction watchdog/truncated-stream failures with a deterministic durable checkpoint while preserving retained-suffix progress and global queued-input chronology
- enforce exact per-turn caps and cooldown admission semantics without weakening hard-limit or generic-failure fail-closed behavior
- migrate legacy budget-limited `pi-goal` state into Senpi's budget-free builtin Goal store

## Verification

- `npm run check`
- coding-agent affected suite: 13 files / 164 tests passed after rebasing onto current `main`
- AI reasoning suite: 2 files / 60 tests passed after rebasing onto current `main`
- final review lanes: code quality, goal verification, hands-on QA, security, context/history, and independent gate all PASS/APPROVE
- installed runtime QA: Senpi `2026.7.30`, `xhigh`/`max` exposed, queue metadata compatibility preserved

## Compatibility

- existing Ctrl+Tab behavior is unchanged; only the supported level sequence expands for eligible map-less Sol models
- manual compaction, generic cancellation, and genuine oversize remain fail-closed
- `clearAllQueues()` keeps the legacy enumerable `{ steering, followUp }` shape while exposing ordered recovery metadata non-enumerably

## Related work

- builds on the terminal queue-restoration foundation merged in #535
- preserves the cooldown-admission contract discussed in #532

---
## Summary by cubic
Hardens required-compaction recovery with a deterministic fallback that only runs when a real prepared suffix fits under limits, preserves global queued-input order across UI/native/compaction paths, exposes `max` for map-less `gpt-5.6-sol`, and migrates legacy budget-limited goals into the built-in Goal store.

- New Features
  - Expose `max` for map-less `gpt-5.6-sol` and send `reasoning.effort: "max"` via `openai-responses`, `azure-openai-responses`, `openai-codex-responses`, and `openai-completions`. Shift+Tab cycles xhigh → max. Explicit maps (including `xhigh`) and `null` vetoes stay authoritative; non‑Sol variants are unchanged.
  - Auto-migrate legacy `pi-goal` files into the built-in Goal store; convert `budgetLimited` to `active` and drop `tokenBudget` with no user action.

- Bug Fixes
  - Required compaction recovers from summarization watchdog timeouts or a typed `SummaryRequestError` with `failureKind: "upstream-stream-truncated"` by appending a deterministic checkpoint only if the real prepared suffix fits; otherwise it cancels and preserves the latest user request. Non-required reasons remain fail-closed, and truncation-looking generic errors never authorize fallback.
  - Failed pre‑prompt compactions now emit `willRetry: false`, returning queued input to the editor instead of re‑queuing against an oversized context.
  - Preserve global enqueue chronology across steering/follow‑up and compaction‑owned input; `clearQueue()` keeps the legacy `{ steering, followUp }` shape and adds a non‑enumerable `ordered` list for recovery and delivery.
  - Enforce exact per‑turn caps and breaker cooldown semantics: ineffective compactions count toward the cap, and blocking summaries aren’t invoked while the breaker is open.

<sup>Written for commit 5038336b6b2236f2a485b8261d4fbb8c885deb49. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/552?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (changes requested):
[sisyphus-bot] Brutal deep review — VERDICT: REQUEST_CHANGES

Reviewed at head `503174fed` in an isolated worktree against `77a768096`. All findings source-verified with exact file:line.

---

### 1. BLOCKER — deterministic recovery can silently drop the entire retained suffix, including the latest user request

**`packages/coding-agent/src/core/extensions/builtin/compaction/deterministic-fallback.ts:106`**
```
firstKeptEntryId: "",
```

When the canonical retained suffix does not fit, the fallback sets `firstKeptEntryId` to `""`. `buildContextEntries()` only begins retaining entries after it finds that ID:

**`packages/coding-agent/src/core/session-manager.ts:520`**
```
if (entry.id === compaction.firstKeptEntryId) {
```

With `""` **no persisted entry ever matches**, so every entry before the new checkpoint is omitted from active model context. This is not "dropping old summarized history" — it can drop the newest, explicitly retained user message. The next provider request then succeeds while having no idea what the user just asked.

The tests **bless** this rather than catch it:
- `test/suite/agent-session-compaction.test.ts:581` builds a payload literally named `"mandatory latest payload "` …
- … then `test/suite/agent-session-compaction.test.ts:597` asserts `firstKeptEntryId: ""`.
- `test/compaction/required-compaction-deterministic-fallback.test.ts:40` pins the same.

The core overflow simulation cannot protect you here: emptying `firstKeptEntryId` is *precisely what makes the destructive result fit*.

**Fix:** never emit a compaction with an empty/nonexistent `firstKeptEntryId`. Use the prepared cut point when the suffix fits; otherwise **cancel and fail admission closed**, restoring pending input to the editor so the user can shorten or split it. If destructive emergency pruning is genuinely wanted, it must preserve the latest user request in a recoverable durable record and require explicit user action — not masquerade as a successful compaction. Add an assertion that the latest user message survives every accepted fallback.

---

### 2. HIGH — "legacy" normalization is applied to *every* current goal file and destroys current-store fields

**`packages/coding-agent/src/core/extensions/builtin/goal/persistence.ts:77`**
```
const normalizedGoal = normalizeLegacyGoal(parsed.goal);
```
**`.../goal/persistence.ts:106`**
```
delete normalized.tokenBudget;
```

`parseGoalFile()` is the **normal reader** used by `readGoalFile()`, not a migration-only path. So every file in the current `extensions/goal` store is normalized as if it were legacy `pi-goal`. A valid current goal carrying `tokenBudget` is read back without it, and the next ordinary update writes the stripped object back — **permanent loss**. The same normalizer also rewrites `budgetLimited`/`budget_limited` to `active`, so a file's state can change merely by being read.

Your own test documents the contradiction: `test/suite/app-server-goal-wire.test.ts:82` asserts `expect(loaded).not.toHaveProperty("tokenBudget")` while line 84 shows the on-disk bytes still contain `tokenBudget: 8192`.

**Fix:** separate current parsing from legacy migration — `parseGoalFile(raw, { legacy = false })`, with only `migrateLegacyGoalFile()` passing `legacy: true`. Add a test that writes a current goal with `tokenBudget`, reads it, updates an unrelated field, and asserts the budget survives on disk and in memory.

---

### 3. MEDIUM — coding-agent and packages/ai disagree on whether an explicit map omission vetoes `xhigh`

**`packages/coding-agent/src/core/thinking-levels.ts:59`**
```
if (level === "xhigh") return mappedLevel !== undefined || supportsXhigh(model);
```
**`packages/ai/src/models.ts:704`**
```
if (model.thinkingLevelMap !== undefined) return false;
```

packages/ai makes an existing `thinkingLevelMap` authoritative (missing level ⇒ unsupported). coding-agent honors that for `max` but not `xhigh` — `supportsXhigh()` ignores the map and infers from model ID. Executed against a model with `thinkingLevelMap: { max: "max" }`:

```
packages/ai:  off, minimal, low, medium, high, max
coding-agent: off, minimal, low, medium, high, xhigh, max
```

Because `clampThinkingLevel()` searches upward first, an unsupported `xhigh` can clamp to `max` — **silently increasing** reasoning effort. **Fix:** consume the canonical packages/ai helpers, or mirror `supportsMax()` exactly; add a test for a map containing `max` but omitting `xhigh`.

---

### 4. LOW — map-less `max` payload coverage exists for only 1 of 4 changed backends

All four builders changed (`openai-responses.ts:337`, `azure-openai-responses.ts:173`, `openai-codex-responses.ts:520`, `openai-completions.ts:851`) and are source-consistent, but only OpenAI Responses gets a map-less payload test (`packages/ai/test/openai-responses-thinking-matrix.test.ts:55`). Completions is most at risk since it derives a separate `thinkingModel` from compatibility metadata first. **Fix:** table-driven payload test across all four.

---

### Scope judgment — this is not one atomic change

Four commits, at least three independently shippable domains: AI capability (`ae3efdffa`), compaction recovery/cooldown/caps/queue-order (`6364f6f03`), goal persistence migration (`f3065e625`), compat follow-up (`503174fed`). Consequences: a goal-store blocker is holding unrelated AI capability work hostage, and a data-loss blocker is buried inside 1000+ lines of unrelated queue/cap churn. `main` already carries independently merged issue #531 work, raising conflict risk. **Please split into AI capability / goal migration / compaction recovery.**

### What is genuinely good

Hard-limit breaker tests verify the summarizer is not invoked post-breaker; cooldown tests distinguish breaker cancellation from generic extension cancellation; queue-order tests preserve cross-mode ordering without disturbing native buckets; the typed watchdog avoids nondeterministic sleeps; UTF-8 truncation is code-point safe and manual failures stay fail-closed.

**Verification:** worktree clean, no edits. `git diff --check` clean. packages/ai `tsc --noEmit` passed. 117 focused tests passed. Larger integration suites could not load (checkout resolves an incompatible `@anthropic-ai/claude-agent-sdk` missing `./extract`); no reported type error pointed at changed files.

code-yeongyu:
[sisyphus-bot] Follow-up: blocker #1 fixed and pushed — **and I am withdrawing finding #2, which was wrong.**

## Pushed to this branch (`4657823b6`)

**Blocker #1 (deterministic fallback data loss) — fixed.** `createRequiredCompactionFallback` now accepts a fallback only when `firstKeptEntryId` names a real branch entry *and* the prepared suffix fits admission; otherwise it returns nothing and `index.ts` returns `{ cancel: true }`, leaving existing context and pending input intact. The `firstKeptEntryId: ""` path and the `retainedSuffix: "none"` variant are gone. The two tests that blessed the destructive behavior now assert the safety invariant instead: an accepted fallback retains the latest user request, and an unfit fallback creates no compaction entry.

**Finding #3 (`xhigh` capability drift) — fixed.** `supportsXhigh()` now mirrors `supportsMax()`: `null` vetoes, an explicit mapping enables, an existing map without `xhigh` disables, and only map-less models fall through to id inference. Covered by a new case in `thinking-levels-tiers.test.ts`.

## Correction: finding #2 (`tokenBudget` normalization) was WRONG — withdrawn

I flagged `normalizeLegacyGoal()` running inside `parseGoalFile()` as data loss that "permanently removes" a current goal's `tokenBudget`. That was a misread, and I want it on the record rather than buried.

Stripping `tokenBudget` on read is **intentional and load-bearing**. It is the mechanism that enforces senpi's budget-free design — `builtin/goal/AGENTS.md` states there is no budget-driven behavior, and `test/suite/app-server-thread-goal.test.ts` pins it explicitly:

```
// Then: the create response preserves wire compatibility, but persisted Goal state is budgetless.
expect(objectAt(responseResult(preserved), "goal").tokenBudget).toBeNull();
expect(objectAt(responseResult(cleared), "goal").tokenBudget).toBeNull();
expect(objectAt(responseResult(read), "goal").tokenBudget).toBeNull();
```

The `create` response echoes the budget for wire compatibility; every subsequent read is budgetless *by design*. I verified this empirically rather than by argument:

- `app-server-thread-goal.test.ts` on your unmodified head `503174fed`: **8/8 passed**
- the same file after applying my suggested "fix": **`AssertionError: expected 4096 to be null`**

So the change I asked for would have broken an intentional product invariant. I have **not** pushed it. `persistence.ts` is untouched on this branch, and the status-rewrite half of my concern is a no-op for current files anyway, since senpi never writes `budgetLimited`.

Apologies for the noise on that one — your original code was right.

## Verification of what I did push

```
required-compaction-deterministic-fallback · agent-session-compaction · app-server-goal-wire
app-server-thread-goal · goal-store · thinking-levels-tiers · per-turn-cap
extension-hard-limit-cap-breaker
=> 8 files / 110 tests passed
```

`app-server-thread-goal` and `app-server-goal-wire` are included deliberately: they are the suites my withdrawn suggestion would have broken, and they are green.

## Still open (not addressed by this push)

- **Finding #4** — map-less `max` payload coverage still exists for only 1 of the 4 changed backends. `openai-completions` is the most exposed since it derives a separate `thinkingModel` first.
- **Scope** — this is still four commits across AI capability, compaction recovery, and goal migration. A compaction data-loss blocker sitting inside 1000+ lines of unrelated queue/cap churn is why finding #2 took a full empirical round-trip to settle. Splitting would make the remaining review cheap.

code-yeongyu:
[sisyphus-bot] Finding #4 also closed — pushed `748ac126b`.

The map-less `max` payload assertion now covers **all four** backends the `supportsMax` change touched, not just OpenAI Responses:

| backend | provider used | assertion |
|---|---|---|
| OpenAI Responses | `openai` | `reasoning.effort === "max"` (pre-existing) |
| Azure OpenAI Responses | `azure-openai-responses` | `reasoning.effort === "max"` |
| OpenAI Codex Responses | `openai-codex` (`transport: "sse"`) | `reasoning.effort === "max"` |
| OpenAI Completions | `openai` + `api: "openai-completions"` | `reasoning_effort === "max"` |

Completions gets its own assertion shape because it serializes a flat `reasoning_effort` and derives a separate `thinkingModel` from compatibility metadata before checking support — which is exactly why covering it separately matters.

**Mutation-proofed rather than just green:** forcing `supportsMax()` to `return false` fails **6** of the 27 tests; reverting the probe returns **27/27**. So these assertions actually detect the regression they name.

Remaining on this PR from my review: only the **scope split** (AI capability / goal migration / compaction recovery as separate PRs). That one is your call — everything I flagged as a defect is now either fixed or withdrawn.

code-yeongyu:
[sisyphus-bot] Merge-readiness report: this PR is now **CONFLICTING against `main`**, and the conflict is not mechanical — `main` has independently shipped overlapping work with *different semantics*.

I attempted the merge in a scratch worktree to characterize it, then aborted. **Your branch is untouched at `748ac126b`** (my two fix commits, nothing else).

## The blocking overlap: issue #531 landed on `main` separately

`main` commit **`48682cb0c` — "fix(compaction): admit prompts during breaker cooldown"** (merged as **#556**) already implements the cooldown-admission work this PR also carries.

- `test/suite/regressions/531-compaction-cooldown-admission.test.ts` is an **add/add conflict**: `main` has a 92-line version, this PR adds a competing 85-line version of the same regression.
- The production divergence is semantic, in `compaction/index.ts`:

```
// this PR
if (breaker.isTripped(state, Date.now()) || cap.shouldRejectByCap(state).cancel) {
    return { applied: false, reason: "failed" };
}

// main (48682cb0c)
if (breaker.isTripped(state, Date.now())) {
    getLogger(ctx).debug("skip_breaker", { route: "blocking" });
    return { applied: false, reason: "rejected" };
}
if (cap.shouldRejectByCap(state).cancel) {
    getLogger(ctx).debug("skip_cap", { route: "blocking" });
    return { applied: false, reason: "rejected" };
}
```

`failed` vs `rejected` is a real behavioral difference for downstream handling, and `main` additionally keeps breaker and cap distinguishable in logs. Picking a winner is a design decision on *your* change, so I am not making it inside your PR.

Full conflict surface: `agent-session.ts` (3 hunks), `interactive-mode.ts` (3), `compaction/index.ts` (1), `compaction/changes.md` (1), `531-...test.ts` (1 add/add).

## Recommended path

This is the concrete version of the scope-split note from my review:

1. Rebase onto current `main`.
2. **Drop the now-redundant issue-531 cooldown work** (production + test) — `main` owns it as of `48682cb0c`; keep only a deliberate follow-up if you prefer `failed` over `rejected`, argued on its own.
3. Split the remainder as previously suggested: **AI capability** / **goal migration** / **compaction recovery**. The deterministic-fallback data-loss fix is independently valuable and would merge immediately on its own.

## State of my review items

| # | Item | Status |
|---|---|---|
| 1 | deterministic fallback drops retained suffix | **fixed** (`4657823b6`) |
| 2 | `tokenBudget` normalization | **withdrawn — my error**, your code was right |
| 3 | `xhigh` capability drift | **fixed** (`4657823b6`) |
| 4 | map-less `max` coverage 1/4 backends | **fixed** (`748ac126b`, mutation-proofed) |
| — | scope split | **open — yours**, and now blocking the merge via #531 duplication |

No defects remain outstanding from my side. The only thing standing between this PR and merge is the `main` reconciliation above.

code-yeongyu (approved):
[sisyphus-bot] Re-review of the rewritten PR at `5038336b6` — **VERDICT: APPROVE.** This resolves every finding from my original review.

Verified against the new head, not just the description:

- **Scope split done correctly.** The PR is now compaction-only (15 files, +701/-50). My earlier findings #3 (xhigh capability) and #4 (backend coverage) live in the extracted work instead — appropriate.
- **Blocker #1 stays fixed.** `createRequiredCompactionFallback` still returns `undefined` when `firstKeptEntryId` is empty or matches no real branch entry, and the hookup returns `{ cancel: true }` instead of emitting a suffix-dropping compaction. Your `770a72418` refines this further — typed `SummaryRequestError` classification instead of string-matching, and fail-closed outside required compaction — which is strictly better than what I asked for.
- **Main's cooldown semantics adopted.** `blocking` admission now returns `reason: "rejected"` with the distinct `skip_breaker`/`skip_cap` logging, matching `48682cb0c`. The semantic divergence from my merge-readiness report is gone.
- **Finding #2 confirmed resolved-by-reconciliation.** The tokenBudget "fix" I pushed and then withdrew is not present — good; my withdrawal stands.

Independent verification on your exact head `5038336b6` in an isolated worktree (not trusting the PR description): `required-compaction-deterministic-fallback`, `agent-session-compaction`, `compaction-terminal-queue-order`, `post-compaction-queued-input-resume`, `531-compaction-cooldown-admission`, `interactive-mode-compaction` — **6 files / 76 tests passed**.

Well-executed rebase; the branch preserved ancestry and addressed each finding with evidence. Approving.

code-yeongyu (approved):
[sisyphus-bot] Re-review of the rewritten PR at `5038336b6` — **VERDICT: APPROVE.** This resolves every finding from my original review.

Verified against the new head, not just the description:

- **Scope split done correctly.** The PR is now compaction-only (15 files, +701/-50). My earlier findings #3 (xhigh capability) and #4 (backend coverage) live in the extracted work instead — appropriate.
- **Blocker #1 stays fixed.** `createRequiredCompactionFallback` still returns `undefined` when `firstKeptEntryId` is empty or matches no real branch entry, and the hookup returns `{ cancel: true }` instead of emitting a suffix-dropping compaction. Your `770a72418` refines this further — typed `SummaryRequestError` classification instead of string-matching, and fail-closed outside required compaction — which is strictly better than what I asked for.
- **Main's cooldown semantics adopted.** `blocking` admission now returns `reason: "rejected"` with the distinct `skip_breaker`/`skip_cap` logging, matching `48682cb0c`. The semantic divergence from my merge-readiness report is gone.
- **Finding #2 confirmed resolved-by-reconciliation.** The tokenBudget "fix" I pushed and then withdrew is not present — good; my withdrawal stands.

Independent verification on your exact head `5038336b6` in an isolated worktree (not trusting the PR description): `required-compaction-deterministic-fallback`, `agent-session-compaction`, `compaction-terminal-queue-order`, `post-compaction-queued-input-resume`, `531-compaction-cooldown-admission`, `interactive-mode-compaction` — **6 files / 76 tests passed**.

Well-executed rebase; the branch preserved ancestry and addressed each finding with evidence. Approving.

realsigridjin:
Status update: PR #552 is now merged into main as 06ed4e7f8243c31c9d0278b476a46723fb414d49. The compaction, AI capability, and Goal-migration remediation branches passed focused tests and repository static checks before the upstream merge. Because the PR is merged, it is no longer reviewable and no re-review request can be sent.

code-yeongyu:
[sisyphus-bot] Merged as `06ed4e7f8`. Thank you @realsigridjin for the rework. The two lanes that were not compaction-specific (map-less max reasoning + legacy goal store migration) are adopted and credited to you in #584.

PR #553, opened 2026-07-30
fix(coding-agent): pause stale goals on direct input

### LINKED ISSUE #447: Built-in goal continuation can cause unbounded token consumption and self-reinforcing output loops
## Summary

The built-in durable goal continuation can repeatedly persist hidden continuation prompts into the LLM context without a continuation cap or stale-progress guard. In a long-running goal this produced hundreds of automatic turns, very large token consumption, and eventually a self-reinforcing assistant output loop.

This was observed with `@code-yeongyu/senpi` version `2026.7.28` on macOS arm64.

## Observed impact

One affected session contained:

- 281 persisted `goal-continuation` custom messages
- only 2 direct user messages
- 2,280 assistant response steps, including tool-call rounds
- 45,172,651 uncached input tokens
- 305,121,792 cache-read tokens
- 496,928 output tokens
- 350,791,371 total processed tokens
- approximately USD 330.98 according to provider usage metadata
- only one compaction during the session

After 264 automatic continuation messages, the assistant first emitted internal-looking control narration as normal output:

```text
用户 continues. Should proceed.
```

Subsequent clean `stop` responses contained phrases such as `Need wait`, `Next`, `End`, and `Stop`. Because those clean responses remained in context, the pattern reinforced itself.

A direct user question did not suspend the goal. The assistant answered at `23:49:50.455`, and another hidden goal continuation was injected at `23:49:51.451`, less than one second later.

### Automatic re-trigger

`dist/core/extensions/builtin/goal/index.js` calls `monitorContinuation.afterAgentEnd(...)` whenever the goal remains active.

`dist/core/extensions/builtin/goal/continuation.js` currently accepts both `stop` and `length` as continuable stop reasons:

```js
return stopReason === "stop" || stopReason === "length";
```

The built-in goal continuation has no equivalent of the continuation cap and stale-state guard used by the OMO start-work and ULW-loop continuation components.

### Hidden prompts remain durable LLM context

`dist/core/extensions/builtin/goal/lifecycle-helpers.js` sends the prompt as a hidden follow-up with `triggerTurn: true`.

`dist/core/messages.js` currently excludes no custom message types:

```js
export function isContextExcludedCustomMessage(_customType) {
    return false;
}

export function filterContextExcludedMessages(messages) {
    return messages;
}
```

Every custom message is then converted to an LLM `user` message. Therefore consumed `goal-continuation` prompts accumulate in subsequent requests and survive until compaction.

### Token budget is not enforced

The Goal type and store support `tokenBudget`, but continuation admission does not check it. Goal accounting also adds only uncached input plus output tokens, so the displayed goal usage did not expose the additional 305 million cache-read tokens processed in this session.

### Retry and fallback amplification

Provider errors were not the initial cause, but retries and fallback reused the existing session context unless compaction was required. Later fallback turns still processed roughly 135k tokens per response. There is no output sanity or repetition circuit breaker for clean `stop` responses.

## Expected behavior

Automatic goal execution should remain bounded and should not allow consumed continuation control prompts to accumulate in durable LLM context. A user interruption, truncation, repeated no-progress turns, or exhausted retry chain should safely suspend the goal instead of silently continuing.

## Suggested fix

At minimum:

1. Make goal continuation single-flight and add a persisted consecutive-continuation cap.
2. Exclude consumed `goal-continuation` messages from future LLM context, or use an ephemeral request-only control channel.
3. Do not treat `stopReason === "length"` as an ordinary clean continuation. Allow at most one dedicated truncation recovery, then suspend.
4. Suspend automatic continuation after a direct user message until explicit resume.
5. Suspend the goal after an exhausted provider retry or terminal error.
6. Enforce a processed-token budget that includes cache reads, while preserving the existing uncached token counter for compatibility.
7. Add a bounded repeated-output or no-progress circuit breaker.
8. On loading an affected session, collapse historical continuation messages and default the active goal to a safe suspended state.

## Suggested regression coverage

- 50 synthetic clean goal turns never leave more than one pending continuation and stop at the configured cap.
- Historical consumed continuation prompts are absent from the next provider request.
- `length` stops cannot create an unbounded continuation loop.
- A direct user message suppresses continuation until explicit resume.
- Terminal provider errors leave the goal suspended.
- Token-budget admission includes cache-read usage.
- A fixture containing hundreds of old continuation messages loads without replaying them.
- Live sandbox QA confirms that normal bounded goals still continue and complete successfully.

I can provide a redacted event timeline or help test a patch if useful.

realsigridjin:
## Summary

- pause an active durable goal as soon as newer interactive or RPC input arrives
- cancel pending monitor continuation synchronously and defer in-flight pauses until usage is fully accounted
- remove the unconditional `userGrace` continuation path while preserving extension-originated and unattended monitor continuation
- migrate legacy `pi-goal` state before session initialization without changing current-store precedence or `tokenBudget` round trips

## Root cause

Direct input did not create another session. A stale active goal remained eligible for a hidden delayed continuation after the user had moved on, making the terminal appear to switch back to another task.

## Behavior changes

- idle direct input accounts the existing idle window, then persists `active -> paused` before the new turn
- in-flight direct input preserves that turn's usage, then pauses at `agent_end` with a goal-id guard
- blocked and paused goals remain unchanged by ordinary input; blocked goals now require explicit `/goal resume`
- extension input never pauses a goal, and explicit resume re-arms one continuation
- current goal files always win over legacy files; only legacy migration removes obsolete budget metadata

## Verification

- affected goal, wire, and regression suites: 9 files / 134 tests passed
- isolated migration commit: 3 files / 74 tests passed, plus full `npm run check`
- full `npm run check` passed on the final tree
- ordered root `npm run build` passed
- built and installed RPC plus native PTY/TUI stale-goal scenarios passed
- five independent review lanes passed after finding and fixing the current-store `tokenBudget` regression

## Risk boundaries

- no authentication, dependency, lockfile, payment, or provider behavior changes
- timer tests use fake time and assert synchronous cancellation before the first awaited read
- changes are split into independently reviewable migration and input-precedence commits

Related to #447.

---
## Summary by cubic
Accepted interactive/RPC input now disarms pending continuation and leaves active Goals idle (not paused). Removed the 60s `userGrace` path; monitor-delayed continuation remains and is held/disarmed based on correlated input dispositions. Fixes #447.

- **Bug Fixes**
  - Accepted direct input (including admitted steer/follow-up) resets the continuation streak and keeps the Goal active; no delayed resurrection after the user turn.
  - Continuation timers hold until `input_disposition` resolves and disarm only on accepted input (`started`/`queued`); handled/rejected restore the remaining timer. Extension-sourced input stays inert, and in-flight usage is still accounted.
  - Mechanical blocks auto‑resume on accepted direct input; provider/interruption/model‑authored blocks remain blocked.

- **Migration**
  - Migrate from `extensions/pi-goal/<threadId>.json` to `extensions/goal/<threadId>.json` on session start.
  - Drop legacy `tokenBudget` and normalize `budgetLimited` to `active`; current-store files take precedence.
  - Preserve current-store `tokenBudget` metadata on read/write.

<sup>Written for commit 1a673295424790ed3d436433acc25b42d4288d4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/553?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (changes requested):
[sisyphus-bot] Brutal deep review — VERDICT: REQUEST_CHANGES

Reviewed at head `92af1584d` in an isolated worktree. Every finding is source-verified. **This PR is directly relevant to the live user complaint that the continuation cap "stops sessions too much" — and in its current form it makes that complaint materially worse.**

---

### 1. HIGH — there is no stale detection; it pauses *every* active goal on *every* direct prompt

**`packages/coding-agent/src/core/extensions/builtin/goal/index.ts:140-143`**
```
if (goal?.status !== "active") return;
const paused = await updateGoal(goalStoreRef(ctx), { status: "paused" }, "user");
```

The stated intent is to pause when the goal's continuation signature is **stale**. This handler never reads or computes any of `goal.lastContinuationSignature`, `buildCurrentGoalContinuationSignature(...)`, or `hasGoalContinuationProgress(...)`. The effective predicate is only `source != extension && status == active`.

So all of these now pause the goal: "Continue the goal." · "Keep going, but prioritize the tests." · same-objective steering during streaming · a status question about active work · input right after creating a fresh goal (before any signature exists) · and — the only intended case — an unrelated prompt. **False positives are the default, not an edge case.** `streamingBehavior: "steer"` is also treated identically to `"followUp"`, though steering usually means *modify current execution*, not *abandon*.

The codebase already defines stale precisely at `continuation.ts:94-99` (`lastContinuationSignature === currentSignature`). The new seam bypasses it entirely.

**Fix:** require `lastContinuationSignature` to exist, build the current signature, and pause **only when they match**; preserve an active goal when observable progress changed or no unattended continuation was delivered. Add separate `steer` coverage. If the policy genuinely changed to "all direct input pauses," rename the feature — it is not stale-goal detection.

---

### 2. HIGH — an ordinary user prompt no longer resumes a cap-blocked goal (worsens the reported complaint)

**`.../goal/index.ts:148-154`** and **`.../goal/lifecycle-helpers.ts:160-162`**

Today on `main`: cap trips → `blocked` with reason `"continuation cap reached"` → **any** normal user message hits `before_agent_start`, which resets the streak *and* flips `blocked → active`. Recovery is one message.

After this PR the new `input` handler deliberately skips blocked goals, `before_agent_start` resets only the numeric streak, and the goal **stays blocked**. Recovery now requires discovering and running `/goal resume`. Resetting a counter while leaving the terminal state in place is not recovery.

This also removes prompt-based reactivation for *every* blocked reason — provider-error, aborted, model-blocked — which is unrelated scope expansion for a stale-active-goal fix.

**Fix:** restore prompt-based recovery at least for mechanical guards (`"continuation cap reached"`, `"repeated assistant output"`, `"output truncation repeated"`): on an accepted direct prompt, reset the streak and reactivate. Separate intentional model blocks from mechanical ones rather than one blanket policy.

---

### 3. HIGH — the resumed-session notice instructs an action that now does the opposite

**`.../goal/index.ts:106-110`** tells the user:
```
Goal auto-continuation suppressed for this resumed session (N historical continuations). Send a message to resume.
```
Under this PR, sending that message enters the new handler and **pauses** the still-active goal (`index.ts:140-143`). The PR's own test locks the contradiction in: `test/suite/goal-extension.test.ts:615` — `it("pauses the goal when a real user prompt follows a suppressed load")`.

This path exists specifically to mitigate continuation floods and cap-style stoppage; it now turns "send a message to resume" into "send a message to pause, then run `/goal resume`."

**Fix:** pick one coherent contract. Preferred: record a one-shot session-start suppression and treat the next accepted direct prompt as reactivation. Alternative: persist `paused` during suppression and change the notice to "Run `/goal resume`."

---

### 4. MEDIUM — the goal is persisted as paused before the input is known to be accepted

`updateGoal(..., "paused")` runs in the raw `input` hook (`index.ts:123-145`), which fires at `agent-session.ts:2331-2337` — **before** model validation (`:2447-2465`), pre-prompt compaction, `before_agent_start`, and final provider admission (`:2524-2527`). So the goal stays paused even when auth is missing, no model is selected, admission rejects the prompt, or a later handler returns `{ action: "handled" }`. A user can type a prompt that never reaches the model and still lose automatic continuation.

**Fix:** split cancellation from persistence — synchronously cancel armed monitor/grace timers at raw `input`, but persist `active → paused` only after the prompt is accepted or queued; clear the candidate on handled/rejected/failed admission.

---

### 5. MEDIUM — the regression test never constructs a stale signature, so it cannot protect the advertised contract

`test/suite/regressions/stale-goal-input-pause.test.ts:32-46` calls `createGoal(...)`, which yields `consecutiveContinuations: 0` and **no** `lastContinuationSignature`, then immediately submits a prompt and asserts `status: "paused"` (`:53-55`). It never delivers a continuation, never persists a signature, never proves staleness. **It would keep passing if all stale-signature logic were deleted** — which is effectively what the production handler does.

**Fix:** deliver a real hidden continuation, assert the signature persisted, prove direct input pauses with state unchanged, then mutate observable state and prove it does **not** take the stale path. Add fresh-goal (no signature) and rejected-input cases.

---

### On the monitor-continuation rewrite (the one clean part)

I found **no** dropped stall/repetition/length guard in the `userGrace` removal: immediate evaluation remains (`monitor-continuation.ts:127-136`), monitor-delayed still re-enters `#admitAndQueue` (`:217-220`), repetition/length-exhaustion still evaluated, toolless-stall tracking intact, and monitor-delayed cap exemption predates this PR. Removing the 60s fallback correctly prevents stale work from resurrecting after a direction change. The risk is concentrated in the new `index.ts` lifecycle policy, not the timer simplification.

**Verification:** detached at `92af1584d`; every cited line verified against worktree source; `git diff --check` clean; no files edited. Focused Vitest could not start (no deps installed in the isolated worktree); type-checking through the main checkout produced only unrelated dependency-version errors.

code-yeongyu:
[sisyphus-bot] Follow-up: all 5 blockers implemented on a companion branch — **but one pre-existing test now fails on a design question only you can settle, so I did not push to your branch.**

Branch: **`fix/553-review-fixes`** (`14d21dfc4`, based on your head `92af1584d`). Pull or cherry-pick as you like.

## What is fixed

**#1 Real stale detection.** The input seam now builds the current continuation signature and pauses **only** when a persisted `lastContinuationSignature` exists and still matches. Progressed goals, fresh goals (no signature yet), and `streamingBehavior: "steer"` all stay active.

**#2 Mechanical-block recovery restored.** An accepted direct prompt again resets the streak and reactivates goals blocked by `continuation cap reached`, `repeated assistant output`, or `output truncation repeated`. Intentional interruption, provider-error, and model-authored blocks stay blocked. This is the part that matters for the live "the cap stops my sessions" complaint.

**#3 Suppressed-load notice honoured.** After a suppressed resumed load, the next accepted prompt resumes instead of pausing, so "Send a message to resume" does what it says.

**#4 Admission-safe persistence.** Added an `input_disposition` extension event (`handled | queued | started | rejected`). Raw `input` now only cancels timers and records a *candidate*; goal state changes only on `queued`/`started`, and is cleared on `handled`/`rejected`. A prompt that never reaches the model no longer pauses the goal.

**#5 Meaningful regression.** `stale-goal-input-pause.test.ts` now drives a real hidden continuation through the lifecycle, asserts the signature persisted, then covers unchanged-stale, observable-progress, fresh-goal, handled-input, and rejected-admission cases. It fails if the stale check is removed or inverted.

I also added the `extensions/changes.md` entry that the repo requires for a `types.ts` public-API change — that was missing.

## Verification

```
stale-goal-input-pause · goal-extension · goal-monitor-continuation · goal-monitor-stall
goal-continuation-verdict · goal-abort-extension · goal-store · agent-session-prompt
extensions-runner · interactive-mode-abort · agent-session-compaction
=> 11 files / 245 tests passed
```

## The open question — `issue-447-goal-continuation.test.ts` fails

`> pauses the goal on a direct user message instead of deferring a graced resume` — **`expected 'active' to be 'paused'`**.

That test creates a fresh goal, sends direct input, and asserts `paused`. Under corrected semantics a fresh goal has **no** persisted signature, so it is *not* stale and must stay active — the test encodes exactly the over-broad behavior finding #1 asked you to remove. It fails for the right reason.

I tried to repair it and hit a second, more interesting problem, which is why I am handing it back rather than forcing it green:

1. Adding a continuation turn first does not make it stale in that file. The shared mock hard-codes `getBranch: () => []` (`goal-monitor-test-harness.ts:118`), while the delivery path derives its signature from `agent_end` **messages**. Real sessions converge; that branchless mock structurally cannot express staleness. (Your real-harness regression covers the pause path correctly, so this is a mock limitation, not a production bug.)
2. Once the goal correctly stays **active**, its clean `agent_end` queues a continuation — so the turn ends with 2 sends where the old test expected 0.

**That second point is the design question:** after a user sends an unrelated direct message and that turn ends cleanly, should an active non-stale goal immediately queue a continuation? Previously the 60s grace window covered this and #553 removes `userGrace`. Options as I see them:

- **(a)** Immediate re-queue is intended — then #447's assertion becomes "exactly one continuation after the user's turn, and no timer adds more", and the test is updated to that.
- **(b)** A user-initiated turn should still suppress the immediate continuation (grace semantics without the resurrect-later timer) — then `endedTurnWasUserInitiated` needs to survive into the post-user `agent_end`, which today it does not.

I left `issue-447-goal-continuation.test.ts` exactly as you wrote it, failing, rather than bending someone else's regression to green around a decision that is yours. Tell me which of (a)/(b) you want and I will finish it.

code-yeongyu:
[sisyphus-bot] Merge-readiness report: this PR is now **CONFLICTING against `main`**, and the conflict sits exactly on the cap logic this PR edits.

Characterized in a scratch worktree, then aborted. **Your branch is untouched at `92af1584d`.**

## The collision

`main` commit **`c05eec9a3` — "fix(goal): count monitor-delayed continuations"** landed after you branched, and it **deletes `isContinuationCapPath` entirely** so the cap applies to *every* automatic continuation path (previously `monitorDelayed` was exempt). This PR edits that same function to drop `userGrace`:

```
// this PR (HEAD)
function isContinuationCapPath(path: GoalContinuationPath): boolean {
    return path === "immediate" || path === "sessionStart";
}

// base 77a768096
function isContinuationCapPath(path: GoalContinuationPath): boolean {
    return path === "immediate" || path === "userGrace" || path === "sessionStart";
}

// main (c05eec9a3)  -> function removed; cap is universal
```

The two changes are *directionally compatible* — you remove the `userGrace` path, `main` removes path-based exemption altogether — but the textual resolution is a deliberate choice about which shape survives, so I am not making it inside your PR.

Conflict surface: `goal/continuation.ts` (1 hunk), `goal/changes.md` (1), `test/suite/goal-continuation-verdict.test.ts` (1), `test/suite/goal-extension.test.ts` (1). `lifecycle-helpers.ts`, `goal-monitor-continuation.test.ts` and `goal-monitor-stall.test.ts` auto-merged cleanly.

## Fixes are ready and waiting

All 5 review blockers are implemented on **`fix/553-review-fixes`** (`14d21dfc4`), based on your `92af1584d`. Rebase that onto current `main` alongside your own resolution, or cherry-pick.

Still needs your decision (unchanged from my last comment): after a user sends an unrelated direct message and that turn ends cleanly, should an active **non-stale** goal immediately re-queue a continuation — **(a)** — or keep grace semantics without the resurrect timer — **(b)**?

My read, now that `main` has moved: **(a)**. Since this PR removes `userGrace` *and* `main` removed the path exemption, choosing (b) would leave an active goal with no automatic resume path at all after a user side-question — which strands goals and makes the very "the cap stops my session" complaint worse. But it is your call and I have left `issue-447-goal-continuation.test.ts` failing, exactly as you wrote it, rather than bending it to green around that decision.

## Heads-up on the product complaint

`c05eec9a3` makes the cap strictly *more* reachable for monitor-heavy goals, since monitor-delayed deliveries now consume it. That is a legitimate fix for an unbounded-wakeup hole, and progress still resets the streak — so only 8 consecutive *zero-progress* monitor continuations trip it. Worth knowing while reconciling, given this PR touches the same guard.

code-yeongyu:
[sisyphus-bot] **Architectural finding — this PR's core premise has been overtaken by `main`, and that outranks my earlier (a)/(b) question.**

I attempted the reconciliation myself in a scratch worktree so I could hand back something mergeable. I stopped, because resolving the conflict *is* the product decision. Scratch worktree removed; your branch and `fix/553-review-fixes` are untouched.

## The finding

This PR removes the `userGrace` path. **`main` actively maintains and is still investing in it:**

- `continuation.ts:11` — `export const GOAL_USER_GRACE_DELAY_MS = 60_000;`
- `continuation.ts:13` — `type GoalContinuationPath = "immediate" | "monitorDelayed" | "userGrace" | "sessionStart";`
- `continuation.ts:96` — `(input.path === "immediate" || input.path === "userGrace")` in the stale guard
- `monitor-continuation.ts:49,135,234` — `DelayedContinuationKind`, `#schedule(goal, "userGrace")`, delayed dispatch
- `main` commit **`bec93e2d8` — "fix(goal): count grace deliveries and reset full guard state on goal transitions"** — a *refinement* of grace accounting, landed after you branched

The collision surfaces concretely in `test/suite/goal-extension.test.ts`:

```
// this PR
expect((await readGoal(storeRefFor(ctx)))?.status).toBe("active");

// main
expect(sent).toHaveLength(0);
const graceDeliveryRecorded = waitForGoalContinuationCount(ctx, 1);
await vi.advanceTimersByTimeAsync(60_000);
await graceDeliveryRecorded;
```

`main` asserts the graced continuation still fires 60s after a user-initiated turn. This PR asserts it never does.

## This supersedes my earlier (a)/(b) question

I previously argued for **(a) immediate re-queue**, reasoning that with `userGrace` gone a goal would have no automatic resume path. **That reasoning was based on this PR's world, not `main`'s.** On current `main`, `userGrace` *is* the resume path, and it works. So the real question is no longer "(a) or (b)" — it is:

> **Should `userGrace` be removed at all, given `main` is actively refining it?**

That is a product-owner decision about goal-continuation behavior, not a merge conflict, and I will not resolve it inside your PR by picking whichever side happens to make the tests green.

Also note `main` `c05eec9a3` deleted `isContinuationCapPath` entirely (the cap is now unconditional across all paths), which independently subsumes this PR's narrower edit to that same function.

## Recommendation

Split the genuinely-orthogonal wins out of the `userGrace` question, since they are valuable regardless of how it is decided:

1. **Real stale detection** (finding #1) — pausing only on a matching persisted signature is correct under either grace policy.
2. **Mechanical-block prompt recovery** (finding #2) — directly serves the live "the cap stops my session" complaint; `main` has since made the cap *more* reachable (`c05eec9a3`), so this matters more now, not less.
3. **`input_disposition`** (finding #4) — admission-safe persistence is independently useful to every extension.

All three are implemented and green on **`fix/553-review-fixes`** (`14d21dfc4`, 245 tests). They rebase onto `main` without touching the grace question. The `userGrace` removal can then be argued on its own merits, with `bec93e2d8` as the counter-position.

code-yeongyu:
[sisyphus-bot] **Correction to my own last comment — the "extract the 3 orthogonal wins" advice was wrong. There is only one decision here.**

I recommended splitting three supposedly-independent wins out of the `userGrace` question. I then verified that claim against current `main` instead of trusting it, and it does not hold:

**1. "Mechanical-block prompt recovery" is not a new win — `main` already does more than this PR.**

`builtin/goal/index.ts:119-121` on `main`:

```ts
let goal = await resetContinuationStreak(ref);
if (goal?.status === "blocked") {
    goal = await updateGoal(ref, { status: "active" }, "user");
}
```

`main` already reactivates **any** blocked goal on a real user prompt — broader than the mechanical-only recovery I proposed. My review finding #2 was that this PR *removes* that behavior; restoring it is therefore a no-op relative to `main`, not something to extract.

**2. `input_disposition` has no consumer on its own.** It exists to make the pause persist only after admission. With no pause-on-input feature, it is an unused public API addition — which is a cost, not a win.

**3. "Real stale detection at the input seam" *is* the pause-on-input feature.** `main` has no `input` handler in the goal extension at all (0 occurrences). So this is not a separable improvement; it is the contested design itself.

## The actual decision, reduced to one question

Both models already prevent a stale goal from resurrecting — they just express it differently:

| | `main` today | this PR |
|---|---|---|
| mechanism | `userGrace` fires at 60s, but `continuation.ts:96` denies it as `stale` when the signature is unchanged | direct input pauses the goal outright |
| resulting state | goal stays `active`, silently does not continue | goal shows `paused` |
| user-visible | footer still says pursuing; nothing happens | explicit paused state |

So the question is **not** "should `userGrace` be deleted" (it is redundant either way, as I showed) and **not** "which of three wins to keep". It is:

> **When a user sends unrelated direct input while a goal is active and stale, should the goal become explicitly `paused`, or stay `active`-but-idle as it does on `main` today?**

That is one cohesive product call about goal UX. Everything else in this PR follows from it: pick "paused" and the input seam, `input_disposition`, and the stale check all belong; pick "active-but-idle" and this PR reduces to the removal of behavior `main` already has.

My own read, for what it is worth: **"paused" is the better UX** — an `active` goal whose footer says it is being pursued while it silently never continues is misleading, and that ambiguity is adjacent to the "the cap stops my session" complaint that started this review. But it is a product decision, and `main` currently implements the other model deliberately.

## Practical consequence

There is no useful smaller extraction. This PR should be **rebased onto `main` and land or close as one feature**, with:

- the `userGrace` removal **dropped** (redundant — `main`'s stale guard already covers it),
- the `before_agent_start` unblock behavior **kept as `main` has it** (do not narrow it),
- `isContinuationCapPath` resolved in favour of `main`'s deletion (`c05eec9a3` made the cap unconditional).

The implementation of the pause seam with correct staleness, plus `input_disposition`, is on `fix/553-review-fixes` (`14d21dfc4`) if you keep the "paused" model.

Apologies for the churn on the recommendation — I would rather correct it than let you split a PR along a seam that does not exist.

code-yeongyu:
[sisyphus-bot] **Retraction: my "the `userGrace` removal is redundant" argument was wrong. This PR fixes a real resurrection bug.**

I ran an adversarial re-review of my own conclusions and it refuted me. I verified the refutation against source before accepting it.

## Why I was wrong

I claimed `main`'s stale guard already prevents a stale goal from resurrecting at the 60s grace timer, so deleting `userGrace` gained nothing. That argument required `lastContinuationSignature` to still be present at grace-evaluation time. **It is not.**

`store.ts` `resetContinuationStreak` deletes the signature, not just the counter:

```ts
export async function resetContinuationStreak(ref: GoalStoreRef): Promise<Goal | null> {
    const goal = await readGoal(ref);
    if (!goal) return goal;
    const next: Goal = { ...goal, consecutiveContinuations: 0 };
    delete next.lastContinuationSignature;      // <-- baseline destroyed
    await writeGoal(ref, next);
    return next;
}
```

And the stale guard requires that baseline to be defined (`continuation.ts:95-100`):

```ts
if (
    (input.path === "immediate" || input.path === "userGrace") &&
    input.lastContinuationSignature !== undefined &&   // <-- false after a real prompt
    input.lastContinuationSignature === input.currentSignature
) {
    return { kind: "deny", reason: "stale" };
}
```

The real sequence:

1. User sends a direct prompt → `before_agent_start` calls `resetContinuationStreak` (`index.ts:110-121`) → **`lastContinuationSignature` is deleted**.
2. The side-question turn ends user-initiated → `afterAgentEnd` schedules `userGrace` (`monitor-continuation.ts:120-137`).
3. At 60s the guard evaluates with `lastContinuationSignature === undefined` → **the stale check cannot fire**.
4. The continuation is admitted and the goal resurrects against stale context.

Independently, even if the baseline survived, the side-question produces new assistant text, so the hash component differs and the signatures would not match anyway. The argument fails twice.

**So removing `userGrace` addresses a genuine bug, and my recommendation to drop that removal was wrong. I withdraw it.**

## What I still think is right, narrowed

The adversarial pass also pushed back on the *persistent pause* policy, and I agree with the narrowed version:

- Pausing is correct when the user genuinely **replaces** the task.
- As a blanket policy for any side question it is **too aggressive**: `transitions.ts` only allows `paused -> active` from a `user` source, and this PR's `before_agent_start` no longer resumes paused goals — so a user who asks one question and walks away returns to a goal that cannot resume itself. That is the same "guardrails stop my session" complaint this review started from.

Recommended shape:

1. **Keep** the `userGrace` removal — it fixes the resurrection bug above.
2. **Keep** accepted-input recovery for mechanically blocked goals (`input_disposition` gating), which is *not* redundant with `main` — queued/steered prompts can bypass `before_agent_start` entirely, and `main` also unblocks before final provider admission so a rejected prompt still mutates the goal.
3. **Do not** persistently auto-pause on every direct input. Disarm the pending grace continuation instead, and leave the goal active-but-idle unless the user explicitly switches tasks or runs `/goal pause`.

That keeps the bug fix and drops the part that makes sessions harder to resume.

I have corrected this position three times now. Each correction came from verifying against source rather than defending the prior claim, and this one reverses my recommendation to you — so please weigh the code above, not my earlier confidence.

code-yeongyu (approved):
[sisyphus-bot] Re-review at `6f0f54a65` (+ maintainer merge of current `main`, pushed as `1a6732954`) — **VERDICT: APPROVE.** This addresses every finding from my original review and the follow-ups.

Verified against the new head, not the description:

- **Finding #1 — real stale detection.** The blanket pause is gone entirely (`direct-input-lifecycle.ts` has no pause call). Direct input now holds a candidate via `holdDirectInput(inputId)` and only acts after admission reports the prompt was `started`/`queued`.
- **Finding #2 — mechanical-block recovery.** `onDisposition` reactivates goals blocked by `isMechanicalContinuationBlock` (`continuation cap reached`, `repeated assistant output`, `output truncation repeated`) on accepted input; intentional and provider-error blocks stay blocked. And this now happens *after* admission, fixing the raw-input mutation window.
- **Finding #3 — the resume/pause contradiction is gone.** The notice no longer instructs an action that pauses.
- **Finding #4 — admission-safe persistence.** `input_disposition` gates all goal-state changes; `handled`/`rejected` never touch the goal.
- **Finding #5 — regression coverage is real.** `stale-goal-input-pause.test.ts` drives a genuine continuation and asserts the persisted signature.

The `userGrace` removal stays — correct, and it is the right bug fix: `before_agent_start` deletes the persisted signature baseline, so the stale gate could never deny the 60s graced continuation and stale goals resurrected. `c05eec9a3`'s deletion of `isContinuationCapPath` is consistent with this head.

Independent verification on your head plus the merge of current `main` (which I pushed as `1a6732954` to resolve the `changes.md` conflict after #552 landed — no force-push, ancestry preserved): `stale-goal-input-pause`, `goal-extension`, `goal-monitor-continuation`, `goal-continuation-verdict`, `goal-modules`, `issue-447-goal-continuation`, `issue-506-monitor-delayed-cap`, `extensions-runner`, `agent-session-compaction`, `531-compaction-cooldown-admission` — **9 files / 209 tests passed**, including the compaction suites that just merged with #552.

One heads-up for the follow-up that should land separately: with `userGrace` gone, a user-initiated turn end has **no** automatic goal-resumption path. That is intentional here — the goal stays active-but-idle and resumes on the next real signal — but it is a behavioral change users will feel, so worth a line in the PR description if you have not already noted it.

realsigridjin:
Status update: PR #553 is now merged into main as 7cad0b9a607a2ab6d75fd313d12378f4e87ff7d7. The active-input lifecycle work was revalidated locally before the upstream merge (142 focused tests and the repository static check passed). Because the PR is merged, it is no longer reviewable and no re-review request can be sent.

code-yeongyu:
[sisyphus-bot] Merged as `7cad0b9a6`. Thank you @realsigridjin for the rework. The adopted legacy goal-store migration lane is in #584.

PR #568, opened 2026-07-31
fix(coding-agent): keep vendored rules discovery inside the project root across Windows drives

### LINKED ISSUE #518: Windows cross-drive project-root traversal fix
### What happened?

On Windows, `findProjectRoot()` can enter a synchronous infinite loop when Senpi is launched from one drive and a rules lookup targets a path on another drive.

`resolve("/")` points to the current working drive, while `dirname()` of the target drive root returns the same path. The traversal therefore never progresses or reaches its existing root check. The session freezes without an error, potentially using 100% CPU.

This reproduced with `senpi -ne`, so it was not caused by a user-loaded extension.

### Steps to reproduce

1. On Windows, launch Senpi from the `C:` drive:

   `senpi -ne`

2. Read, edit, or write a file on another drive, such as `D:\senpi-repro\file.txt`, where no project marker exists in its parent directories.
3. Observe that the session hangs indefinitely during project-root discovery.

### Expected behavior

Project-root traversal should stop when `dirname(currentDirectory) === currentDirectory` and return `null` if no marker was found:

```ts
const parentDirectory = dirname(currentDirectory);

if (
	currentDirectory === filesystemRoot ||
	parentDirectory === currentDirectory
) {
	return null;
}

### Version

2026.7.29-6

### PR DESCRIPTION
**Pull request by MoerAI**
## Summary

On Windows, project rule discovery can escape the project root and pull rule files from an **unrelated drive** into the model context.

When a `read`/`edit`/`write` target lives on a different drive than the project root, the vendored rules finder treats that target as being *inside* the project, walks the other drive, and collects any `AGENTS.md`, `CLAUDE.md`, `CONTEXT.md`, `.omo/rules`, `.claude/rules`, `.cursor/rules` or `.github/instructions` it finds there as **project** rules. Those rules are then injected as project instructions for a project that does not own them.

## Root cause

`getWalkDirectories` guards the walk with `isSameOrChildPath`, which classifies containment from the `relative()` result:

```ts
// rules/finder.ts
const childRelativePath = relative(parentPath, childPath);
return childRelativePath === "" || (!childRelativePath.startsWith("..") && !childRelativePath.startsWith("/"));
```

On Windows, `relative()` between two different drive roots returns an **absolute** path rather than a `..` chain:

```
relative("C:\\workspace\\proj", "D:\\other")  ->  "D:\\other"
```

`"D:\\other"` starts with neither `".."` nor `"/"`, so the containment test returns `true`. The escape guard never trips, the walk proceeds up the other drive (terminating only at `D:\` via the `dirname()` self-equality check), and every directory on that drive contributes project rule candidates.

The sibling helper in `rules/engine.ts` already gets this right and uses `isAbsolute()` for exactly the same test — the finder is the odd one out:

```ts
// rules/engine.ts
return childRelativePath === "" || (!childRelativePath.startsWith("..") && !isAbsolute(childRelativePath));
```

This is the same bug class as #432 (Windows cross-drive traversal in this module), but a different site and a different symptom: #432 was a hang in `project-root.ts`, this is a silent project-scope escape in `finder.ts`. Note this does **not** close #518 — that report's symptom is the `project-root.ts` hang already fixed by #432 and shipped in `v2026.7.30` (the reporter was on `v2026.7.29-6`).

## Fix

Use `isAbsolute()` instead of `startsWith("/")`, matching `rules/engine.ts`.

POSIX behavior is unchanged: there `isAbsolute()` and `startsWith("/")` agree, and `relative()` between two POSIX paths never yields an absolute result.

| File | Change |
|------|--------|
| `packages/coding-agent/src/core/extensions/builtin/rules/rules/finder.ts` | `isSameOrChildPath` rejects an absolute `relative()` result via `isAbsolute()`; brief comment on why |
| `packages/coding-agent/src/core/extensions/builtin/rules/changes.md` | Vendored-adaptation entry per the `changes.md` contract, with the upstream-propose note |
| `packages/coding-agent/test/suite/regressions/0000-rules-finder-cross-drive-project-scope.test.ts` | Regression test, modelled on the existing `0000-rules-find-project-root-cross-drive.test.ts` (mocks `node:path` to win32 so it runs on every CI platform) |

## Reproduction (before fix)

Real Windows 11 machine, project root on `C:`, rule planted on a real second drive `D:` — no mocks, driving the exported `findRuleCandidates` API:

```
projectRoot   = C:\Users\...\qa-proj-c (drive C)
targetFile    = D:\senpi-qa-518\src\file.ts (drive D)
planted rule  = D:\senpi-qa-518\AGENTS.md

project rule candidates returned:
  - D:\senpi-qa-518\AGENTS.md | source = AGENTS.md | distance = 1

RESULT: LEAK - 1 off-project rule(s) collected
```

Unit-level, the new regression test fails on the pre-fix code for the right reason:

```
AssertionError: expected [ 'D:\other\AGENTS.md' ] to deeply equal []
- Expected
+ Received
- []
+ [ "D:\\other\\AGENTS.md" ]
```

## Verification (after fix)

Same real cross-drive driver, unchanged inputs:

```
project rule candidates returned:
  (none)

RESULT: NO LEAK - project scope respected
```

## Test

- New regression test: `test/suite/regressions/0000-rules-finder-cross-drive-project-scope.test.ts` — RED before, GREEN after.
- Pre-existing cross-drive test `0000-rules-find-project-root-cross-drive.test.ts` still passes.
- `npx vitest --run test/suite/regressions/ test/rules-before-agent-start.test.ts` — same failure set as a clean `upstream/main` baseline on this Windows box (9 files: multi-session-theme-init, fswatch-error-crash, replaced-session-context, find-path-glob, claude-sdk-oauth-installed-sdk-hook-stop, bun-launcher-self-update, codemode-builtin-dedupe, inspector-vm-import-crash, todo-12-stale-lock-owner). All are pre-existing and unrelated to this change; they flake between 8 and 9 files across runs on a clean tree.
- `npm run check` — clean (it runs as the pre-commit hook: biome + pinned-deps + ts-imports + shrinkwrap + install-lock + tsgo + browser-smoke + web-ui).

Rebased on `upstream/main` at `9da987f51`.

---
## Summary by cubic
Fixes Windows cross-drive and drive-letter-case escapes in vendored rules discovery so project rules stay inside the project root. Previously, a target on a different drive or with a different drive-letter case made the finder walk unrelated locations and inject their `AGENTS.md`/`CLAUDE.md`/`.claude/rules` files as project rules. POSIX behavior is unchanged.

- **Bug Fixes**
  - `isSameOrChildPath` now rejects absolute `relative()` results with `isAbsolute()`, stopping cross-drive escapes.
  - `getWalkDirectories` now stops the walk on an empty `relative()` instead of an exact string compare, stopping drive-letter-case escapes.
  - Adds a regression test covering both cases and updates `changes.md` and `CHANGELOG.md`.

<sup>Written for commit 35fa7f2e9b322bac6c20759126dc92b1c663e1f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
The only remaining gate is the stale Cubic neutral result from the July quota exhaustion. The quota reset date has passed, but the existing reviewed head was never re-evaluated. Please rerun or refresh Cubic for commit 687b5d0adb, or apply the repository quota-exhaustion skip policy if that is the intended path. No code failure is currently reported.

code-yeongyu:
ULW audit verdict: BLOCKED. The production cross-drive containment fix integrates cleanly, but the authored test fails off Windows, the changelog and required contexts are missing, a P1 thread remains unresolved, and case-variant root termination is incomplete. Local evidence: `local-ignore/qa-evidence/20260817-fix-pr-audit/pr-568/VERDICT.md`.

MoerAI:
Rebased onto current `main` (through #1465) and addressed all five items from the BLOCKED verdict. Mapping each one to its fix and evidence.

**1. The authored test fails off Windows** — `2f7236c`

The mock mapped `dirname`/`join`/`relative`/`resolve` to their win32 variants but left `isAbsolute` as the host implementation. `posix.isAbsolute("D:\\other")` is `false`, so off Windows the containment guard never tripped and the cross-drive `AGENTS.md` was still collected:

```
AssertionError: expected [ 'D:\other\AGENTS.md' ] to deeply equal []
 ❯ test/suite/regressions/0000-rules-finder-cross-drive-project-scope.test.ts:45:57
```

Mapping `isAbsolute` to `win32.isAbsolute` makes the regression exercise the guard on any host. Green on macOS now.

**2. The P1 thread remains unresolved** — `2f7236c`

Same defect: that is exactly what the Codex thread called out. Resolved by the same commit. While there I also replaced the factory's `importOriginal<typeof import("node:path")>()` with a top-level namespace import, since the root `AGENTS.md` forbids inline dynamic type imports.

**3. Case-variant root termination is incomplete** — `2838fab`

`getWalkDirectories` ended the walk with an exact string compare against the project root. Windows drive letters are case-insensitive, so a target under `c:\workspace\proj\src` never matched a `C:\workspace\proj` root and the walk continued past it:

```
["c:\\workspace\\proj\\src", "c:\\workspace\\proj", "c:\\workspace", "c:\\"]
```

`findProjectCandidates` collects rules from every walked directory and does not re-check containment per directory, so this is user-visible, not just a wasted walk — an `AGENTS.md` above the project root is injected as a *project* rule. New regression, RED before the fix:

```
AssertionError: expected [ 'c:\workspace\AGENTS.md' ] to deeply equal []
```

Terminating on `relative(projectRoot, currentDirectory) === ""` folds drive-letter case and stops at the root. POSIX is unaffected: there the two comparisons already agree.

**4. The changelog is missing** — `f74fde4`

Added the `[Unreleased] > Fixed` entry for #568. `node scripts/check-pr-changelog.mjs` went from

```
changelog-gate: FAIL - runtime source changed without a CHANGELOG.md entry
```

to

```
changelog-gate: PASS - changes.md coverage complete (0 production path(s) covered); changelog entry updated (packages/coding-agent/CHANGELOG.md)
```

**5. Required contexts are missing** — `f74fde4`

I read this as the tracker requirement and checked it against the repo's own tooling rather than guessing. `.github/upstream.json` pins `badlogic/pi-mono`, and the `rules` builtin is vendored from `code-yeongyu/pi-rules`, so it is fork-only and the four-canonical-heading policy does not apply — the gate confirms this with `0 production path(s) covered`. `scripts/audit-changes-md.mjs` also exits 0. I extended `rules/changes.md` with the `getWalkDirectories` change in the existing vendored-adaptation format alongside the `isSameOrChildPath` entry. If you would rather have the four headings here regardless of the pin, say so and I will restructure it.

---

Verification on the rebased branch:

```
bun run --cwd packages/coding-agent test test/rules test/suite/regressions/0000-rules-finder-cross-drive-project-scope.test.ts
  Test Files  4 passed (4)
       Tests  17 passed (17)

bun run check                      -> exit 0
scripts/check-pr-changelog.mjs     -> exit 0
scripts/audit-changes-md.mjs       -> exit 0
```

Real-CLI QA via `.agents/skills/senpi-qa`, both asserting `~/.senpi/agent/auth.json` is unchanged:

```
mock-loop.mjs --with-tool : 4/4 passed   (zero-token full agent turn)
cli-smoke.mjs --self-test : 8/8 passed
```

Windows-specific semantics are covered by the win32-mocked unit tests, since the suite runs on POSIX here. The production diff is 8 added / 3 removed lines in `finder.ts`; the rest is the regression test and the two doc entries. Ready for re-review.

PR #576, opened 2026-07-31
feat(goal): proposal - render a countdown for the invisible 60s continuation wait

eddieparc:
## Proposal, not a merge request

This is the **render half** of a fix for an invisible wait. I'm opening it as a proposal
because the interesting half — where the countdown gets driven from — is a design call I'd
rather have your direction on than guess at. Happy to close this if you'd rather solve the
underlying asymmetry a different way.

## The problem

`#schedule()` in `monitor-continuation.ts` waits two very different amounts of time and only
narrates one of them:

| path | delay | `ui.notify` | event | durable entry |
|---|---|---|---|---|
| `monitor` | 240s (`GOAL_MONITOR_CONTINUATION_DELAY_MS`) | yes | yes | yes (`goal-cache-warmup`) |
| `userGrace` | 60s (`GOAL_USER_GRACE_DELAY_MS`, `continuation.ts:11`) | **no** | **no** | **no** |

The `userGrace` branch only clears state. So after a real user prompt, an active goal sits for
a full minute with nothing rendered and the footer back at ready — which is indistinguishable
from a hang. Users respond the way you'd expect: ESC, then re-prompt.

I found this while reading a transcript where a user typed `resume` four times. The gaps that
preceded those presses were `60.0s`, `60.3s`, `60.5s` and `61.6s` — each starting immediately
after a `senpi.hooks.stop-state {count:0}` record and ending with a `goal-continuation`
injection, with no rendered output in between. That regularity is a timer, not model latency.
The 240s monitor wait in the same session *did* leave visible `goal-cache-warmup`
scheduled→resumed entries. Tool calls paired 170/170 and child-task wakes delivered 11/11 in
that transcript, so the silence is the unrendered wait itself, not a lost tool result or a
dropped wake.

## What's in this PR

One new fork-only module, `goal/wait-progress.ts`:

- `renderGoalWaitBar(elapsedRatio)` — a clamped 12-cell `▰`/`▱` bar.
- `formatGoalWaitLabel({ kind, remainingMs, totalMs, activeMonitorCount })` — composes the bar
  with the remaining time, reusing `formatWakeDuration` from `cache-warm.ts` so a wait
  countdown reads like the cache-warm notices already do.

```
▱▱▱▱▱▱▱▱▱▱▱▱ goal resumes in 1m
▰▰▰▱▱▱▱▱▱▱▱▱ goal resumes in 45s
▰▰▰▰▰▰▱▱▱▱▱▱ goal resumes in 30s
▰▰▰▰▰▰▰▰▰▰▰▰ goal resumes in 0s

▰▰▱▱▱▱▱▱▱▱▱▱ goal continues in 3m 20s · 2 monitors on duty
▰▰▰▰▰▰▰▰▰▱▱▱ goal continues in 1m · 1 monitor on duty
```

**Nothing imports it yet.** No scheduler, event, entry, or footer behavior changes here, so
this commit cannot alter runtime behavior — which is also the honest weakness of it: it's a
render layer with no caller until the wiring question below is settled.

## What I'd like your call on

1. Should `userGrace` emit a durable entry (symmetric with `goal-cache-warmup`), a transient
   footer segment only, or both? A durable entry per grace window means a transcript line every
   time a user interrupts, which may be more noise than signal.
2. If footer-only: extend `GoalElapsedTicker`, or add a second ticker so the wait countdown and
   the `Pursuing goal (…)` elapsed segment stay independent?
3. Is a 12-cell bar the right footer budget, or should this be text-only?

I deliberately stopped before touching `#schedule()` — it's had four commits in the last two
weeks (`2ad3d0c30`, `448bef6b2`, `dc6374f4c`, `184e62730`) and I didn't want to collide with
in-flight work on a guess about the intended UX.

## Verification

Sandboxed worktree, never against a live runtime.

- `goal-wait-progress.test.ts` — 9 passed. RED was captured first against a deliberately empty
  stub: `expected '' to contain '47s'`, `'3m 12s'`, `'1 monitor'`, `'0s'` — 4 assertion
  failures, then 9/9 green after implementing.
- Adjacent-surface regression: all 18 `test/suite/goal-*` files, **212 passed**.
- Adjacent-surface regression, measured against this PR's actual base (`origin/main`
  `c0c9e6cdc`) rather than a local HEAD:

  | base | goal test files | tests |
  |---|---|---|
  | plain `origin/main` | 17 | 203 passed |
  | `origin/main` + this commit | 18 | 212 passed |

  The delta is exactly the one new file and its 9 tests.
- `tsgo --noEmit` exit 0; `biome check` on the changed files exit 0.
- Full-wait frame sequence exercised through the shipped module (60s sampled every 5s, 240s
  sampled every 20s), asserting in-probe: single bar width `{12}`, monotonic fill, no negative
  time, first frame empty, last frame full.

One note worth passing on: `goal-monitor-continuation.test.ts` has two timer-driven cases
(`waits four minutes before continuing and announces the schedule`, `counts user grace delivery
toward the continuation cap`) that failed for me when I ran vitest, tsgo and biome concurrently,
and passed 19/19 in isolation on both plain `origin/main` and this branch. Unrelated to this
change, but they look sensitive to CPU contention if that matters for CI.

Not run: root `npm run check` and the full `npm test` across the monorepo. If you want those
green before you'll read it, say so and I'll run them.

## Scope

Only the render layer. Related silent-turn work (a `claude-agent-sdk` stream that ends with
zero content blocks being treated as a successful stop) is deliberately **not** bundled here —
different root cause, different files, separate PR if you want it.

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1767,6 +1767,16 @@ const current = pi.getThinkingLevel();  // "off" | "minimal" | "low" | "medium" 
 pi.setThinkingLevel("high");
 ```
 
+### pi.setSessionFastMode(enabled)
+
+Mark the session as running in fast mode so host surfaces can show it — the TUI footer stamps a ⚡ on the model label. Session-scoped and never persisted; it is also true on its own whenever the active model is configured for the priority tier (an `openai` `-fast` catalog variant, a scoped `provider/id:priority`).
+
+This is only an indicator. It does not add `service_tier` to any request, so an extension that wants the priority tier on the wire still returns it from `before_provider_request`.
+
+```typescript
+pi.setSessionFastMode(true);
+```
+
 ### pi.events
 
 Shared event bus for communication between extensions:
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -649,6 +649,7 @@ export class AgentSession {
 	// Base system prompt (without extension appends) - used to apply fresh appends each turn
 	private _baseSystemPrompt = "";
 	private _currentServiceTier: ServiceTier | undefined = undefined;
+	private _sessionFastMode = false;
 	private _lastHighReasoningWarningKey: string | undefined = undefined;
 	private _baseSystemPromptOptions!: BuildDynamicSystemPromptOptions;
 	private _systemPromptOverride?: string;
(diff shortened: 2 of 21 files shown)


eddieparc:
Superseded by #578 — this branch accidentally carried two unrelated local commits (LAB-14, LAB-15), so the diff showed 21 files. #578 is the same proposal as a single commit on top of current main (4 files).

Issue #1, opened 2026-03-14
Feature request: Remote agent bridge for local GUI control


GGos3:
## Summary

Support a hybrid workflow where an agent running in a remote environment can interact with and observe GUI surfaces on the user's local machine.

## Problem

In remote agent setups, the agent can execute tasks on a server, but it usually cannot directly access or control GUI windows that exist on the user's local device.

This breaks an important class of workflows:

- GUI-based testing
- local browser or desktop app interaction
- visual verification of results
- step-by-step debugging that depends on what the user can see locally

In practice, the useful compute environment may be remote, while the useful GUI surface is local. Today, those two environments are disconnected.

## Request

It would be valuable for `saneagent` to support a local companion or bridge that allows a remotely running agent to interact with the user's local GUI environment.

At a high level, the desired workflow is:

1. The main agent runs remotely.
2. A local companion process exposes controlled access to local GUI surfaces.
3. The remote agent can launch, attach to, inspect, and interact with local applications or windows.
4. The user can directly observe the same local GUI while the agent continues operating remotely.

## Expected Capabilities

A future implementation could make it possible for a remote agent to:

- launch or attach to local GUI applications
- inspect available local windows or active sessions
- perform GUI-driven interactions on the local machine
- stream enough state back to the remote agent for reliable automation
- let the user observe and validate actions in real time

## Why This Fits `saneagent`

This feels aligned with the direction described in this repository:

- desktop GUI support
- cross-platform operation
- compatibility across agent harnesses
- acting as an interface for driving other clients

A remote-to-local GUI bridge would make `saneagent` substantially more useful for real development and testing workflows, especially when execution belongs on a server but interaction and verification belong on the user's machine.

## Notes

The key need here is not simply "remote GUI on the server." The more important capability is enabling a remote agent to work with GUI surfaces that are local to the user.

That distinction matters because, in many practical workflows, the user wants to see and validate the interaction on their own device rather than through a server-hosted desktop session.

Tagging @MINPETER for context and discussion.

minpeter:
@GGos3 is my friend, I asked for a posting for reference in future agent development

m13v:
the remote-to-local GUI bridge is a genuinely useful pattern. we built a local MCP server for macOS that exposes desktop control (screenshot, click, type, scroll, element inspection) over stdio transport, and it works well as the local companion you're describing.

[...]

PR #586, opened 2026-07-31
fix(ai): make reasoning capability metadata authoritative

realsigridjin:
## Summary

Follow-up to #552: current `main` already carries the map-less `max` baseline (`aa13d734c`, including the OpenAI wire serialization). This PR intentionally contains only the remaining capability hardening, so the front-end/agent and the provider payload share one authoritative rule:

- **`packages/ai/src/models.ts`** — `supportsXhigh` / `supportsMax` keep an explicit `thinkingLevelMap` authoritative (omitted level = unsupported, `null` = veto) and infer tiers from the model id only for map-less models; the id lists and the OpenAI-compatible `gpt-5.6-sol` matcher are extracted into module constants with documentation.
- **`packages/coding-agent/src/core/thinking-levels.ts`** — tier detection delegates to `@earendil-works/pi-ai` instead of duplicating the capability rules.
- **Docs** (`packages/ai/changes.md`, `packages/ai/src/changes.md`, `packages/coding-agent/src/core/changes.md`) and focused tests (`supports-xhigh.test.ts`, `agent-session-model-extension.test.ts`) updated.

## Branch status

The previously published head (`24fbc07c`) accidentally carried ~100 files from the pre-merge #552 delivery branches. The branch has been replaced (force-updated) with the clean reconstruction on current `main` (`f4705697b`): the current head `57d6ebabd` is a **7-file delta, all in the AI capability domain**. Validation on this tree (tree `13a92850b`):

- `npm run check` (biome + pinned deps + ts-imports + shrinkwrap + tsgo + browser smoke)
- `npm run build`
- Focused suites: `packages/ai` supports-xhigh + thinking-matrix (**83/83**); `packages/coding-agent` agent-session-model-extension + thinking-levels-tiers (**26/26**)
- CI is re-running on the updated head.

### DIFF
--- a/packages/ai/changes.md
+++ b/packages/ai/changes.md
@@ -6,7 +6,8 @@
 
 - Runtime capability detection now recognizes `max` for map-less OpenAI-compatible `gpt-5.6-sol` models while
   preserving explicit map omissions and `null` vetoes.
-- OpenAI-compatible request builders serialize the selected level as `reasoning.effort: "max"`.
+- OpenAI-compatible request builders serialize the selected level as `reasoning.effort: "max"`
+  (`reasoning_effort: "max"` on Completions).
 
 ### Why
 
--- a/packages/ai/changes.md
+++ b/packages/ai/changes.md
@@ -1,18 +1,19 @@
 # changes.md — ai
 
-## Map-less GPT-5.6 Sol max reasoning (2026-07-30)
+## Shared reasoning-tier capability detection (2026-07-30)
 
 ### What changed
 
-- Runtime capability detection now recognizes `max` for map-less OpenAI-compatible `gpt-5.6-sol` models while
-  preserving explicit map omissions and `null` vetoes.
-- OpenAI-compatible request builders serialize the selected level as `reasoning.effort: "max"`
-  (`reasoning_effort: "max"` on Completions).
+- Shared `xhigh` / `max` model-family constants are hoisted in `models.ts`, and map-less inference now uses
+  one case-normalized, boundary-aware family matcher instead of unbounded substring checks.
+- `getSupportedThinkingLevels` delegates extended-tier precedence to the exported `supportsXhigh` and
+  `supportsMax` predicates rather than duplicating their map-omission and `null`-veto rules.
 
 ### Why
 
-- Custom providers such as `codex-lb` can supply Sol without generated model metadata. The UI and provider
-  payload must agree instead of displaying `max` while silently sending `high`.
+- Capability inference for custom map-less models should reject unrelated ids and case-normalize legitimate
+  aliases while keeping one precedence implementation. Generated catalog models retain their explicit maps,
+  so behavior for real catalog models is intentionally unchanged.
 
 ## Browser-safe prompt-cache TTL resolver (2026-07-28)
 
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -6,10 +6,10 @@
 
 - OpenAI-compatible map-less `gpt-5.6-sol` models now expose `xhigh` and `max` without requiring a generated
   `thinkingLevelMap`.
-- Explicit maps remain authoritative: a missing level on an existing map stays unavailable, and `null` still
-  vetoes the heuristic.
-- OpenAI Responses, Azure Responses, Codex Responses, and Completions send `max` on the wire instead of
-  clamping a UI-selected map-less Sol level to `high`.
+- Explicit maps remain authoritative: a missing level on an existing map stays unavailable, and `null` vetoes the
+  heuristic. `supportsXhigh` and `supportsMax` share that precedence.
+- `supportsMax` is exported from `models.ts` so OpenAI Responses, Azure Responses, Codex Responses, and
+  Completions send `max` on the wire instead of clamping a UI-selected map-less Sol level to `high`.
 - Coverage pins capability, negative non-Sol boundaries, and captured request payloads without live tokens.
 
 ## 2026-07-30 - Recover Kimi XTML response channels from thinking
--- a/packages/ai/src/model.ts
+++ b/packages/ai/src/model.ts
@@ -20,7 +20,8 @@ export interface Model<TApi extends Api> {
 	reasoning: boolean;
 	/**
 	 * Maps pi thinking levels to provider/model-specific values.
-	 * Missing keys use provider defaults. null marks a level as unsupported.
+	 * In a present map, omitting `xhigh` or `max` disables that extended tier; ordinary missing levels
+	 * use provider defaults. null marks any level as unsupported.
 	 */
 	thinkingLevelMap?: ThinkingLevelMap;
 	input: ("text" | "image" | "video")[];
--- a/packages/ai/src/models.ts
+++ b/packages/ai/src/models.ts
@@ -697,6 +697,13 @@ export function clampThinkingLevel<TApi extends Api>(
 	return availableLevels[0] ?? "off";
 }
 
+/**
+ * Whether the model exposes the `xhigh` tier.
+ *
+ * An explicit `thinkingLevelMap` is authoritative: `null` vetoes the tier and a map that omits
+ * `xhigh` disables it. Only map-less models fall back to id-based inference, so custom providers
+ * that ship a reasoning model without generated catalog metadata still surface the tier.
+ */
 export function supportsXhigh<TApi extends Api>(model: Model<TApi>): boolean {
 	const mapped = model.thinkingLevelMap?.xhigh;
 	if (mapped === null) return false;
@@ -705,28 +712,35 @@ export function supportsXhigh<TApi extends Api>(model: Model<TApi>): boolean {
 	return model.reasoning && supportsXhighModelId(model.id);
 }
 
+const XHIGH_MODEL_IDS = [
+	"gpt-5.2",
+	"gpt-5.3",
+	"gpt-5.4",
+	"gpt-5.5",
+	"gpt-5.6",
+	"deepseek-v4-pro",
+	"deepseek-v4-flash",
+	"opus-4-6",
+	"opus-4.6",
+	"opus-4-7",
+	"opus-4.7",
+	"opus-4-8",
+	"opus-4.8",
+	"opus-5",
+	"sonnet-5",
+	"fable-5",
+];
+
 function supportsXhighModelId(modelId: string): boolean {
-	const xhighModelIds = [
-		"gpt-5.2",
-		"gpt-5.3",
-		"gpt-5.4",
-		"gpt-5.5",
-		"gpt-5.6",
-		"deepseek-v4-pro",
-		"deepseek-v4-flash",
-		"opus-4-6",
-		"opus-4.6",
-		"opus-4-7",
-		"opus-4.7",
-		"opus-4-8",
-		"opus-4.8",
-		"opus-5",
-		"sonnet-5",
-		"fable-5",
-	];
-	return xhighModelIds.some((id) => modelId.includes(id));
+	return XHIGH_MODEL_IDS.some((id) => modelId.includes(id));
 }
 
+/**
+ * Whether the model exposes the `max` tier.
+ *
+ * Mirrors {@link supportsXhigh}: an explicit `thinkingLevelMap` wins, and only map-less models
+ * are inferred from the id.
+ */
 export function supportsMax<TApi extends Api>(model: Model<TApi>): boolean {
 	const mapped = model.thinkingLevelMap?.max;
 	if (mapped === null) return false;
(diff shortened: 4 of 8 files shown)


realsigridjin:
Follow-up to #552: current main has adopted the baseline map-less max support. This PR intentionally contains only the remaining hardening: authoritative thinking-level map precedence, shared pi-ai capability ownership, and full provider/matrix coverage. Requesting a fresh review of this focused delta.

code-yeongyu:
# [sisyphus-bot] 🔪 KITCHEN NIGHTMARE: PR #586

Right. Let me tell you what I've found, because you've walked in here with a plate that *looks* clean and it's hiding a raw centre.

## WHAT THIS PR ACTUALLY DOES (the honest plating)

- Hoists `XHIGH_MODEL_IDS`, `MAX_MODEL_IDS`, `OPENAI_MAX_APIS` and the `gpt-5.6-sol` matcher out of function bodies into module constants in `packages/ai/src/models.ts`.
- Deletes the duplicated tier-detection rules from `packages/coding-agent/src/core/thinking-levels.ts` (-59 lines) so it's three thin wrappers over `pi-ai`.
- Adds 78 lines of `supportsMax` coverage for the map-veto / map-omission / non-OpenAI-api matrix.
- Net effect: capability display and wire payload derive from one rule set instead of two drifting copies.

The intent is right. The execution left the fridge door open.

## 🔥 BLOCKERS — THIS DOES NOT LEAVE MY KITCHEN

### 1. Your headline regex has ZERO test coverage. **ZERO.**

`packages/ai/src/models.ts:762` — `const GPT_56_SOL_ID = /(?:^|\/)gpt-5\.6-sol(?:$|-)/;`

This anchored regex is the single cleverest line in the whole PR. It rejects `gpt-5.6-solar`, accepts `openai/gpt-5.6-sol`, accepts `gpt-5.6-sol-fast`. Beautiful. And **not one test proves any of it.**

I mutated it to a bare substring match:

```diff
-const GPT_56_SOL_ID = /(?:^|\/)gpt-5\.6-sol(?:$|-)/;
+const GPT_56_SOL_ID = /gpt-5\.6-sol/;
```

Result:

```
 Test Files  1 passed (1)
      Tests  56 passed (56)
```

**FIFTY-SIX PASSED.** I tore the anchors off your safety property and your suite smiled at me and said service is going fine. You wrote a careful regex and then wrote no test that can tell it apart from a careless one. That's not coverage, that's decoration.

For contrast, here is the assertion you *should* have written — I wrote it, and it works:

```ts
it.each(["gpt-5.6-solar", "gpt-5.6-solaris", "my-gpt-5.6-sol", "xgpt-5.6-sol", "legacy-gpt-5.6-solstice"])(
  "rejects %s", (id) => { expect(supportsMax(maplessModel("openai-responses", id))).toBe(false); });
```

- Against your code: **5 passed**.
- Against the substring mutation: **5 failed**.

That is what a test that can actually fail looks like. Add it. `packages/ai/test/supports-xhigh.test.ts` — the `does not infer max for a map-less non-Sol %s model` block already exists at line ~250 and tests `gpt-5.6-terra` and `gpt-5.6-luna`; those are *different words*. They do not test the anchor. A `solar`-family id is the whole point.

**FIX:** add the 5-case rejection block above to `supports-xhigh.test.ts`.

### 2. You deduplicated across two packages and left a third copy in the same file.

`packages/ai/src/models.ts:663-677`

The entire premise of this PR is "one authoritative rule". Then `getSupportedThinkingLevels` sits there re-implementing that rule inline instead of calling the functions you just exported:

```ts
if (level === "xhigh") {
    return mapped !== undefined || (model.thinkingLevelMap === undefined && supportsXhighModelId(model.id));
}
if (level === "max") {
    return mapped !== undefined || (model.thinkingLevelMap === undefined && supportsMaxModel(model));
}
```

That is `supportsXhigh` and `supportsMax` open-coded, character for character, twelve lines above the functions that do exactly this. You drove the duplication out of `coding-agent` and parked it in the boot of `models.ts`. The next person who changes the precedence rule fixes one site, ships, and reopens the exact drift bug this PR exists to close.

**FIX:**
```ts
if (level === "xhigh") return supportsXhigh(model);
if (level === "max") return supportsMax(model);
```
Both already handle `mapped === null`, `mapped !== undefined`, and the map-less fallback. Identical semantics, one source of truth. Prove it with the existing 56 tests.

## 🍳 UNDERCOOKED — fix before service

### 3. One matcher is a surgeon, the other is a bin bag.

`packages/ai/src/models.ts:715-731` and `:764-774`

You wrote an anchored, segment-aware regex for Sol — and then left `XHIGH_MODEL_IDS` / `MAX_MODEL_IDS` on `.includes()`:

```ts
return XHIGH_MODEL_IDS.some((id) => modelId.includes(id));
```

`"gpt-5.2"` matches `gpt-5.20`, `my-gpt-5.2-preview`, `not-gpt-5.2-at-all`. `"opus-5"` matches `opus-50`. In the *same function*, one rule is precise and the other is a substring soup. You clearly know how to write the careful version — you did it on line 762. Apply the same standard, or write a comment saying why substring is deliberate for these.

### 4. The wrappers wrap nothing.

`packages/coding-agent/src/core/thinking-levels.ts:16-24` — `supportsXhigh` and `supportsMax` are now pure pass-throughs that re-export a function under the same name. Only `getSupportedThinkingLevels` adds anything (the `["off"]` non-empty fallback) — and `pi-ai`'s version already returns `["off"]` for `!model.reasoning` and always yields at least `["off","minimal",...]` for reasoning models, so that fallback may be dead too. Either re-export directly (`export { supportsMax, supportsXhigh } from "@earendil-works/pi-ai";`) or prove the fallback is reachable with a test.

### 5. The new session test reaches into the harness's guts.

`packages/coding-agent/test/suite/agent-session-model-extension.test.ts:217-226`

```ts
harness.getModel().api = "openai-responses";
```

Mutating a model object after harness construction. If `createHarness` ever snapshots or freezes the model, this silently stops testing what it claims to. Pass the api through `createHarness({ models: [{ id: "gpt-5.6-sol", reasoning: true, api: "openai-responses" }] })`. And while you're there: the test asserts the *cycle order* but never the wire payload — the actual bug in the PR description was "UI shows max while the provider clamps to high". Assert the serialized `reasoning.effort`.

## 🧂 SEASONING — nits

- `packages/ai/changes.md` claims Completions sends `reasoning_effort: "max"`; worth a one-line pointer to the test that pins it.
- `MAX_MODEL_IDS` is a strict subset of `XHIGH_MODEL_IDS` minus the gpt entries. A comment explaining *why* the Anthropic ids appear in both would save the next reader a diff.

## ✅ WHAT'S ACTUALLY GOOD

Credit where it's due, and I don't hand this out often:

- **The Sol regex is genuinely correct.** `(?:^|\/)` and `(?:$|-)` is exactly the right pair of anchors — handles the provider path prefix, handles the `-fast` suffix, rejects `solar`. Whoever wrote that was thinking. Now go test it.
- **The veto/omission matrix coverage is real.** I mutated away the map-omission authority (`if (model.thinkingLevelMap !== undefined) return false;`) → **1 failed / 55 passed**. I mutated away the `OPENAI_MAX_APIS` gate → **1 failed / 55 passed**. Those tests bite. Good.
- Killing the `ModelWithThinkingLevelMap` cast and the 59-line duplicate is the right architectural call.
- `changes.md` entries are honest about *why*, including the drift that caused the bug.

## VERDICT

**BLOCK** — the PR's headline safety property is provably untested (56/56 still green under a mutation that removes it), and a change whose entire purpose is deduplication ships with a third copy of the deduplicated rule twelve lines away.

Fix #1 and #2, and this goes out. The bones are good. The plating is lazy.

*Evidence: mutation runs and the 5-case probe executed against `57d6ebabd` in an isolated worktree; baseline `packages/ai` `supports-xhigh.test.ts` = 56 passed.*

code-yeongyu:
# [sisyphus-bot] 🔪 SERVICE ROUND 2: PR #586 — I WAS TOO KIND

A second pass went through this with the real generated catalog and a full production-reversion test. Two things I filed as "seasoning" are actually **blockers**, and there's a headline problem neither pass caught first time. I'm correcting my own plate, because a review that's wrong doesn't get to stand either.

## 🔥 NEW BLOCKER — this PR does not do what its changelog says it does

`packages/ai/src/changes.md`, `packages/coding-agent/src/core/changes.md:15-18`

Your changelog states the fix is that a map-less `gpt-5.6-sol` "could show `max` in the UI while the provider payload clamped to `high`". So I reverted `packages/ai/src/models.ts` to its **pre-PR** parent (`57d6ebabd^`) and ran your 78 new test lines against it:

```
 Test Files  1 passed (1)
      Tests  56 passed (56)
```

**Every single one green against the code you claim to have fixed.**

And a catalog sweep across the real generated models: **1,187 models, 0 changed.** Not one model in the entire catalog resolves differently before and after your `packages/ai` diff.

The `packages/ai` production change (+58/-42) is a **pure no-op refactor** — you hoisted array literals out of function bodies into module constants. That's fine work. It is not the bug fix your changelog describes. The actual value of this PR is the `coding-agent` deduplication, full stop.

**FIX:** rewrite both changelog entries to say what this is — eliminating duplicated rules, behavior intentionally unchanged for reasoning models. Do not ship a changelog that sends the next maintainer hunting a drift that isn't there. If the UI-vs-wire clamp bug is real, it needs an actual fix and an actual failing test; right now you have neither.

## 🔥 ESCALATED — "seasoning" #3 is a blocker, and it's case-sensitive too

I filed the substring matching as a nit. That was wrong, because **this PR makes `models.ts` authoritative for every coding-agent consumer** — session cycling, RPC model metadata, app-server effort validation, fallback selection. Serving a false capability from one central kitchen is strictly worse than serving it from two.

Probe against the real exported functions:

```
{"id":"my-gpt-5.60",     "xhigh":true,  "max":false}
{"id":"notopus-5ive",    "xhigh":true,  "max":true }   ← both tiers. unrelated model.
{"id":"opus-50",         "xhigh":true,  "max":true }
{"id":"not-sonnet-500",  "xhigh":true,  "max":true }
{"id":"xgpt-5.2y",       "xhigh":true,  "max":false}
{"id":"gpt-5.6-solar",   "xhigh":true,  "max":false}
{"id":"GPT-5.6-SOL",     "xhigh":false, "max":false}   ← CASE-SENSITIVE. legitimate alias LOSES capabilities.
{"id":"openai/gpt-5.6-sol",              "xhigh":true, "max":true}  ✓
{"id":"quotio-openai/gpt-5.6-sol-fast",  "xhigh":true, "max":true}  ✓
```

`notopus-5ive` gets `max`. `not-sonnet-500` gets `max`. And `GPT-5.6-SOL` — an uppercase alias a custom provider could plausibly ship — silently gets **nothing**. Once selected, OpenAI Responses forwards the inferred effort verbatim (`packages/ai/src/api/openai-responses.ts:401-405`), so a fabricated capability becomes a malformed provider request.

Note `gpt-5.6-solar` still leaks `xhigh` — your anchored regex correctly denies it `max`, but `XHIGH_MODEL_IDS` catches it on the bare `gpt-5.6` substring. The careful matcher and the careless one sit in the same file, and the careless one wins more often.

**FIX:** one case-normalized, escaped, token-boundary matcher for all three lists. Require a non-alphanumeric boundary before and after the family marker; accept `/ . : _ -` separators and legitimate `-fast`/`-pro` suffixes. Then add table tests for `my-gpt-5.60`, `notopus-5ive`, `opus-50`, `GPT-5.6-SOL`, `gpt-5.6-solar`, and `quotio-openai/gpt-5.6-sol-fast`.

## 🍳 UNDERCOOKED — the public contract says the opposite of the code

`packages/ai/src/model.ts:21-25` documents `thinkingLevelMap` verbatim as *"Missing keys use provider defaults."* The resolver does the reverse for `xhigh`/`max` — once any map exists, an omitted tier is **disabled**:

```
thinkingLevelMap={}                -> xhigh=false, max=false
thinkingLevelMap={"max":null}      -> xhigh=false, max=false
thinkingLevelMap={"xhigh":"xhigh"} -> xhigh=true,  max=false
```

A provider that adds `{ max: null }` following the documented contract silently loses `xhigh` as well. This PR *chooses* omission-is-authoritative — fine, it's the better semantic — but then the docs and the models.json schema must say so.

## 📋 REVISED BLOCKER LIST

1. Changelog describes a fix the diff does not contain; `packages/ai` change is a verified no-op (56/56 green on parent, 0/1187 catalog delta).
2. Unbounded + case-sensitive family matching grants false `xhigh`/`max` and drops legitimate uppercase aliases — now authoritative for all consumers.
3. The anchored Sol regex still has zero coverage (substring mutation → 56/56 green).
4. `getSupportedThinkingLevels` (`models.ts:663-677`) still open-codes the precedence rule instead of calling the exported predicates.
5. `Model.thinkingLevelMap` documentation contradicts the implemented semantics.

## VERDICT

**BLOCK, harder than before.**

The direction — one resolver instead of two — is right, and I'll say again that the Sol regex itself is well built. But right now this PR centralizes a matcher that hands `max` to `notopus-5ive`, ships a changelog for a fix it doesn't make, and its new tests pass against the unfixed code. Three separate ways of telling the reader something untrue.

Fix 1–4 and it goes out.

*Evidence: production reversion to `57d6ebabd^` (56/56 green), 1,187-model catalog sweep (0 changed), and the false-capability probe above, all executed in isolated worktrees.*

code-yeongyu:
# [sisyphus-bot] ✅ SERVICE ACCEPTED: PR #586

Right. I've re-inspected every plate, and I'll say what I don't say often: **this is now properly cooked.** Independently verified in a clean worktree at `d5623b52d` — not taking the fixer's word for it.

## All five blockers closed

**1. Substring soup + case sensitivity — FIXED.** The new `matchesModelFamily()` is case-normalized, regex-escaped, and boundary-aware. Verified against the exact table that was failing:

```
notopus-5ive     xhigh=false max=false   (was true/true)
opus-50          xhigh=false max=false   (was true/true)
not-sonnet-500   xhigh=false max=false   (was true/true)
my-gpt-5.60      xhigh=false max=false   (was true/false)
xgpt-5.2y        xhigh=false max=false   (was true/false)
gpt-5.6-solar    xhigh=false max=false   (was true/false)
GPT-5.6-SOL      xhigh=true  max=true    (was false/false)  ← uppercase alias now works
openai/gpt-5.6-sol             xhigh=true max=true  (preserved)
quotio-openai/gpt-5.6-sol-fast xhigh=true max=true  (preserved)
```

And critically, the fabricated capability no longer reaches the wire payload: `notopus-5ive` now resolves to `["off","minimal","low","medium","high"]` instead of the full list ending in `max`.

**2. Third copy of the precedence rule — FIXED.** `getSupportedThinkingLevels` now delegates to `supportsXhigh`/`supportsMax`.

**3. Untested Sol regex — FIXED.** This was the one that mattered most to me. Previously, loosening the matcher to a bare substring left **56/56 green**. Now:

```
mutation: boundary matcher -> bare substring   ->  10 failed / 62 passed
mutation: drop case normalization              ->   1 failed / 71 passed
mutation: stop delegating in getSupported...   ->   2 failed / 70 passed
```

Every mutation bites. That's real coverage, not decoration.

**4. Fictional changelog — FIXED.** The invented "UI shows `max` while the provider clamps to `high`" claim is gone, replaced with an honest description that explicitly states behavior for real catalog models is intentionally unchanged. That's the correct entry.

**5. `thinkingLevelMap` doc contradiction — addressed.**

## The one thing I chased down before approving

Bare `gpt-5.6` lost `xhigh` — the entry was replaced with `gpt-5.6-luna` / `-sol` / `-terra`. I flagged it as a regression and went to the catalog before accepting it:

- All **30** real `gpt-5.6*` models are `luna`/`sol`/`terra`, and **every one carries `hasMap=true`**, so the explicit map governs them regardless.
- A bare `gpt-5.6` is not a shipped product.

**Full catalog sweep: 1,187 models, 1 changed** — and that one is a *correction*:

```
huggingface/deepseek-ai/DeepSeek-V4-Flash   before=false/false  after=true/false
```

The old case-sensitive matcher was missing it because of the capitalised id. So the narrowing is safe and the single behavioral delta is a bug fix. My expectation was wrong, not your code.

## Tests

`packages/ai` `supports-xhigh.test.ts`: **72 passed** (up from 56). Full pi-ai suite green, coding-agent model-extension 18/18, `npm run check` exit 0.

## VERDICT

**APPROVED.** The matcher is now precise in both directions — it rejects the junk and accepts the aliases it was previously dropping. The tests can fail. The changelog tells the truth.

Good work. Now get out of my kitchen before I change my mind.

*Verified independently at `d5623b52d`: capability gate ALL PASS, 1,187-model catalog sweep, 3-mutation re-proof.*

PR #587, opened 2026-07-31
fix(coding-agent): publish migrated goals without clobbering

realsigridjin:
## Summary

Follow-up to #552/#553: the legacy `pi-goal` migration that landed on `main` normalizes **every** current-store read (`parseGoalFile` unconditionally applies `normalizeLegacyGoal`), which strips the inert `tokenBudget` wire-compatibility metadata from current goals and breaks the app-server round-trip contract. This PR intentionally contains only the two focused corrections:

- **`persistence.ts`** — `parseGoalFile(raw, { legacy: true })` gates legacy normalization to the import path only; the legacy directory is derived by replacing the `goal` path segment (`legacyBaseDirFor`), which is correct for both the session layout and the `no-session` fallback; migrated goals are published **without clobbering** an existing destination (destination wins).
- **Tests** — `goal-store.test.ts` + `goal-extension.test.ts` pin migration-only normalization, current-store precedence with inert `tokenBudget`, the `no-session` layout mapping, and destination-wins publication.

## Branch status

The previously published head (`488dd7ad0`) accidentally carried ~100 files from the pre-merge #552 delivery branches, and its CI failed in unrelated app-server/model-catalog tests. The branch has been replaced (force-updated) with the clean reconstruction on current `main` (`f4705697b`): the current head `839b8874f` is a **5-file delta, all in the builtin goal extension** (migration-only normalization + atomic publish + their test updates). Validation on this tree (tree `fb73764c6`):

- `npm run check` (biome + pinned deps + ts-imports + shrinkwrap + tsgo + browser smoke)
- `npm run build`
- Focused suites: `goal-store`, `goal-extension`, `app-server-goal-wire`, `app-server-thread-goal` (**99/99**)
- The four app-server/model-catalog tests that previously failed on the polluted head pass on this tree when no ambient provider credentials are present (**8/8** with API keys unset; CI runs hermetically)
- CI is re-running on the updated head.

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/goal/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/goal/changes.md
@@ -1,112 +1,52 @@
 # goal Extension Changes
 
-## Migrate legacy pi-goal state into the builtin goal store (2026-07-31)
+## Legacy `pi-goal` state is imported once at session start (2026-07-31)
 
 ### What changed
 
-- Session startup now migrates a legacy sibling `pi-goal` file when the builtin `goal` store does not yet exist.
-- Legacy `budgetLimited` / `budget_limited` goals resume as active, and `tokenBudget` is stripped from runtime state while the original legacy file remains untouched.
-- Existing builtin goal files remain authoritative and are never overwritten by migration.
+- `persistence.ts` exports `migrateLegacyGoalFile(ref)`, which imports a legacy
+  standalone `pi-goal` file from `<sessionDir>/extensions/pi-goal/` into the
+  current budget-free store at `<sessionDir>/extensions/goal/`.
+- `index.ts` `session_start` awaits the migration **before** `readGoal`, so an
+  imported goal participates in the session that migrated it rather than the
+  next one.
+- `parseGoalFile` gained an opt-in `{ legacy: true }` mode. Only that mode runs
+  `normalizeLegacyGoal`, which deletes `tokenBudget` and rewrites the removed
+  `budgetLimited` / `budget_limited` statuses to `active`. Migrated goals still
+  pass the normal `isGoal` validation and `sanitizeContinuationState`.
+- Precedence is absolute: an existing current file short-circuits the migration
+  and is never overwritten. A missing legacy file is a silent no-op, and a
+  legacy file holding an explicit `null` goal migrates nothing.
+- The legacy directory is derived by replacing the `goal` path segment itself
+  (`legacyBaseDirFor`), which is correct for both store layouts:
+  `<sessionDir>/extensions/goal` and the print/in-memory fallback
+  `<agentDir>/extensions/goal/no-session/<cwdKey>`. Swapping `dirname(baseDir)`
+  handles only the first; in the fallback it resolves *inside* the current store
+  tree, so real legacy state is missed and a stray nested `pi-goal` directory
+  could be imported instead. Real-CLI RPC QA caught this; both cases are pinned
+  by regression tests.
 
 ### Why
 
-- Moving the extension store from `pi-goal` to `goal` otherwise made existing objectives appear lost, and legacy budget metadata is incompatible with the builtin goal extension's budget-free design.
-
-### Verification
-
-- Goal-store coverage pins migration, runtime budget stripping, destination persistence, source preservation, and the existing app-server wire contract.
-
-||||||| a2efa4076
-## Accepted direct input leaves an active Goal idle (2026-07-31)
-
-### What changed
-
-- Removed the 60-second user-grace continuation path. An accepted ordinary direct input disarms any pending continuation, and its completed user turn leaves an active Goal active but idle instead of auto-pausing or resurrecting it later.
-- Input candidates and reversible monitor-continuation holds remain keyed by `inputId`, so handled/rejected and overlapping prompts neither mutate Goal persistence nor lose an already armed monitor continuation. Extension input remains inert; admitted steer input gets the same recovery accounting as other direct input.
-- Accepted direct input, including admitted steering, reactivates only the mechanical block reasons owned by `continuation-recovery.ts`; provider, interruption, and model-authored blocks stay blocked.
-- Continuation-cap admission remains universal across immediate, monitor-delayed, and session-start delivery, and candidate goal-ID checks prevent input races from migrating state to another Goal.
-
-### Expected merge conflict zones
-
-- MEDIUM in `index.ts` lifecycle wiring and `monitor-continuation.ts` timer ownership.
-- LOW in the focused direct-input lifecycle module.
-||||||| de5de53a7
-## Achieved footer shows elapsed time and truncated objective (2026-07-31)
-
-### What changed
-
-- `ui.ts` `goalStatusText` renders the `complete` status as
-  `<truncated objective> · Goal achieved (<elapsed>)`, mirroring the elapsed
-  suffix that `Pursuing goal (…)` already shows while a goal is active. The
-  elapsed suffix is omitted only when `timeUsedSeconds` is 0.
-- New exported `truncateGoalObjective` normalizes whitespace and truncates the
-  objective to 32 characters with a trailing ellipsis for the footer preview.
-- `goal-modules.test.ts` pins the new achieved-footer format, including the
-  long-objective truncation case.
-
-### Why
-
-- The achieved footer previously dropped both the objective and the total time
-  spent, so a finished goal gave no at-a-glance context about what completed or
-  how long it took.
+- Users upgrading from standalone `pi-goal` silently lost an in-flight goal: the
+  builtin reads a different directory, so the legacy file was never seen.
+- Normalization is deliberately confined to the legacy read path. Applying it to
+  every current-store read (the alternative considered) is destructive: it
+  strips the inert `tokenBudget` that `00571304e` added as required app-server
+  wire metadata, and it silently resurrects the removed `budgetLimited` status
+  instead of surfacing the existing `InvalidGoalStoreError`. Both regressions
+  were reproduced against `app-server-goal-wire.test.ts` before being rejected.
+- Budget-free means budgets are **not enforced**, not that budget metadata is
+  destroyed on sight. Legacy budgets are dropped because they are enforcement
+  inputs; current-store budgets are preserved because they are wire compatibility
+  metadata.
 
 ### Expected merge conflict zones on the next sync
 
-- LOW in `ui.ts` around the `goalStatusText` switch.
-
-## Mechanically blocked goals resume on a prompt queued while streaming (2026-07-31)
-
-### What changed
-
-- `index.ts` adds an `input` handler that reactivates a mechanically blocked goal
-  (cap, repetition, truncation) from a direct prompt even when that prompt is
-  queued as steer/follow-up while the agent is streaming.
-- The idle path already unblocks in `before_agent_start`, but a queued prompt
-  returns from `prompt()` before that hook runs, so a message sent mid-turn left
-  the goal blocked. The new handler covers it; extension-sourced input is ignored.
-
-### Why
-
-- The recovery hint added in #562 tells the user "Send any message to resume",
-  but a message typed while the agent was streaming was queued and returned
-  before `before_agent_start`, so it did not resume. Users who act on the hint
-  at the most natural moment saw no recovery.
-
-### Expected merge conflict zones on the next sync
-
-- LOW in `index.ts` around the new `input` handler.
-- NONE in the verdict engine, goal store schema, persistence, or public extension API.
-
-## Tool-using turns clear the output-repetition window (2026-07-31)
-
-### What changed
-
-- `monitor-continuation.ts` computes `turnUsedTools` before recording assistant
-  output, and `#recordAssistantOutput` now clears
-  `#recentNormalizedOutputHashes` for a tool-using turn instead of appending its
-  final text hash. Only three consecutive **toolless** identical outputs can
-  trip the `repetition` verdict and block the goal with
-  "repeated assistant output".
-- Coverage adds the issue #566 regression: tool-using turns with identical
-  final text stay active, a mid-streak tool turn resets the window, and the
-  three-toolless-identical-outputs block is pinned unchanged.
-
-### Why
-
-- The repetition guard recorded the last assistant text hash on every
-  `agent_end`, so a goal doing real tool work each turn but ending turns with
-  an identical status line (the monitor-wait pattern) was blocked after three
-  turns — exactly the state a live resumption channel was about to wake. The
-  cap already treats tool use as progress (#539); the repetition window now
-  follows the same principle. The stale-signature guard still suppresses
-  continuation spam for identical no-progress turns without blocking.
-
-### Expected merge conflict zones on the next sync
-
-- LOW in `monitor-continuation.ts` around `afterAgentEnd` ordering and
-  `#recordAssistantOutput`.
-- NONE in `continuation.ts`, the goal store schema, or the public extension
-  API.
+- LOW in `persistence.ts` around `parseGoalFile`'s new `legacy` parameter and the
+  `parseGoalFileJson` split; standalone `pi-goal` has no migration path.
+- LOW in `index.ts` at the `session_start` migration call.
+- NONE in the store schema, tool schemas, status transitions, or public API.
 
 ## Mechanical continuation blocks tell the user how to resume (2026-07-31)
 
--- a/packages/coding-agent/src/core/extensions/builtin/goal/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/goal/changes.md
@@ -15,8 +15,11 @@
   `budgetLimited` / `budget_limited` statuses to `active`. Migrated goals still
   pass the normal `isGoal` validation and `sanitizeContinuationState`.
 - Precedence is absolute: an existing current file short-circuits the migration
-  and is never overwritten. A missing legacy file is a silent no-op, and a
-  legacy file holding an explicit `null` goal migrates nothing.
+  and is never overwritten. Publication writes a restrictive `0600` private temp
+  sibling, then atomically hard-links it to the current path without replacement;
+  `EEXIST` means a concurrent current writer won and migration returns no import.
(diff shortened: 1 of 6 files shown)


realsigridjin:
Follow-up to #552: current main has adopted the baseline Goal migration. This PR contains the remaining concurrency and compatibility hardening: atomic publish where an existing destination wins, migration-only normalization, and inert current-store tokenBudget metadata. Requesting a fresh review of this focused delta.

code-yeongyu:
# [sisyphus-bot] 🔪 KITCHEN NIGHTMARE: PR #587

Right, sit down. This one's different. The *cooking* is good — genuinely good, I'll say so up front. But you've plated it next to a garnish that will poison a customer, and you've written a test that says the poison is the recipe.

## WHAT THIS PR ACTUALLY DOES (the honest plating)

- `parseGoalFile(raw, { legacy: true })` — legacy normalization now runs **only** on the import path, so ordinary current-store reads keep their bytes and the inert `tokenBudget` wire metadata survives the app-server round-trip.
- `legacyBaseDirFor()` replaces the `goal` **path segment** instead of `dirname(baseDir) + "pi-goal"`, which was flat wrong for the `<agentDir>/extensions/goal/no-session/<cwdKey>` layout.
- `publishMigratedGoalFile()` uses `link(tempPath, filePath)` as an exclusive-create so a concurrent current-store writer can never be clobbered by migration.
- Adds ~370 lines of test across `goal-store`, `goal-extension`, `app-server-goal-wire`, `app-server-thread-goal`.

## ✅ FIRST, THE PART I'M NOT GOING TO SHOUT ABOUT

I mutation-tested your suite before I opened my mouth, because unlike some people I check before I serve. Baseline: **91 passed**.

| Mutation | Result |
|---|---|
| Remove the `{ legacy }` gating → normalize every read (**the exact bug this PR fixes**) | **4 failed** |
| `legacyBaseDirFor` → `dirname(baseDir) + "pi-goal"` (the pre-PR bug) | **2 failed** |
| Drop exclusive-create → plain `writeFile` overwrite | **1 failed** |

Every one bites. Your race test even asserts `readdir(ref.baseDir)` contains *exactly* the goal file, so a leaked temp file gets caught. Your `mode 0600` test is correctly `skipIf(win32)`. **This is real coverage and it is the standard PR #586 failed to meet.** Well done. Now let me ruin your afternoon.

## 🔥 BLOCKERS — THIS DOES NOT LEAVE MY KITCHEN

### 1. `link()` will hard-fail on filesystems where `main` worked fine. And it buys you NOTHING.

`packages/coding-agent/src/core/extensions/builtin/goal/persistence.ts:108-138`

```ts
await writeFile(tempPath, contents, { encoding: "utf8", mode: 0o600 });
await link(tempPath, filePath);
...
} catch (error) {
    if (!isFileSystemError(error, "EEXIST")) {
        publicationFailed = true;      // ← rethrown
```

You treat **only** `EEXIST` as benign. `link()` is not universally available: exFAT/FAT32 volumes, a good number of SMB/NFS mounts, and some container overlay configurations return `EPERM`, `ENOSYS`, `EOPNOTSUPP`, or `EXDEV`. On any of those, migration now **throws** where `main` silently succeeded with a plain `writeFile`. You've converted a working path into a crash for anyone whose home directory isn't on a local APFS/ext4 volume.

Grep your own test file for how much of this you covered:

```
'EPERM'   -> 0
'ENOSYS'  -> 0
'EXDEV'   -> 0
'link'    -> 0
```

Zero. Zero. Zero. Zero.

And here's the part that should sting: **the temp-file-plus-link dance is completely unnecessary.** `writeFile` with `flag: "wx"` is the exclusive-create primitive. I ran both:

```
1. link into empty dir:        FIRST
2. link over existing:         EEXIST   (target still: FIRST)
3. wx  over existing:          EEXIST   (target still: FIRST)
```

**Byte-identical semantics.** Same `EEXIST` on collision, same original-preserved guarantee — with no temp file, no `link`, no `rm` cleanup branch, no `AggregateError` ceremony, and no portability landmine. You wrote thirty lines to reimplement a flag.

**FIX:** delete `publishMigratedGoalFile`'s temp/link/unlink machinery and use:
```ts
try {
    await writeFile(filePath, contents, { encoding: "utf8", mode: 0o600, flag: "wx" });
    return true;
} catch (error) {
    if (isFileSystemError(error, "EEXIST")) return false;
    throw error;
}
```
That's the whole function. Then add the `EEXIST` test you already have, and you're done. Remove the now-unused `link` import.

### 2. One corrupt legacy file bricks the goal system FOREVER — and you wrote a test declaring that correct.

`persistence.ts:49-58`, and `test/suite/goal-store.test.ts`:

```ts
it("keeps the typed invalid-goal error for a legacy file", async () => {
    await writeRawGoalFile(legacyStoreRef(ref), '{"version":1,"goal":{"id":1}}\n');
    await expect(migrateLegacyGoalFile(ref)).rejects.toThrow("goal store contains an invalid goal");
});
```

Walk through what that means for a real user. They have **no current goal**. They have a leftover `pi-goal` file from an old version that is malformed — truncated by a crash, half-synced by Dropbox, whatever. Every single time `migrateLegacyGoalFile` runs, it throws. Not once. **Every time.**

I ran it three times against a corrupt legacy file:

```
R8 attempt 1: THROWS SyntaxError -> goal path blocked (no current goal exists!)
R8 attempt 2: THROWS SyntaxError -> goal path blocked (no current goal exists!)
R8 attempt 3: THROWS SyntaxError -> goal path blocked (no current goal exists!)
```

The user never asked about the legacy goal. They don't care about the legacy goal. The legacy goal is *dead data from a version they no longer run* — and it is now permanently blocking a feature they do want to use, with no way out except knowing to manually delete a dotfile they've never heard of.

A migration is a **best-effort import of dead data**. It is not a correctness gate on the live system. Propagating a typed error out of an *optional* import path and calling it "keeps the typed error" is not rigour, it's stubbornness with a test attached.

**FIX:** make the legacy import best-effort — catch `InvalidGoalStoreError` / `UnsupportedGoalStoreVersionError`, leave the legacy file untouched, return `null`, and log a warning. Then **invert** the two tests (`keeps the typed invalid-goal error…`, `keeps the typed unsupported-version error…`) to assert `resolves.toBeNull()` and that the current store stays usable. Those tests currently pin the bug.

## 🍳 UNDERCOOKED — fix before service

### 3. `legacyBaseDirFor` is separator-blind.

`persistence.ts:76-82` — `baseDir.split(sep)`. On win32 `sep === "\\"`, but Node accepts `/`-separated paths everywhere and a config- or caller-derived `baseDir` can easily carry them. Proof:

```
legacyBaseDirFor("C:\\Users\\x\\.senpi\\extensions\\goal")  -> undefined   (on posix)
```

The polarity flips on win32: a `/`-joined `baseDir` returns `undefined`, `migrateLegacyGoalFile` returns `null`, and **the user's goal is silently not migrated**. No error, no log, no goal. Silent data loss is the worst failure mode there is.

**FIX:** `baseDir.split(/[\\/]/)` and rejoin with `sep`, or `normalize()` first. Add a case to the existing win32-aware tests.

### 4. Migration never terminates — the legacy file is immortal.

`persistence.ts:59` — on the `EEXIST` path `publishMigratedGoalFile` returns `false`, `migrateLegacyGoalFile` returns `null`, and **the legacy file is left exactly where it was**. It gets re-read and re-parsed on every subsequent call, forever. Verified:

```
R7 legacy file still present after 'migration': true -> re-parsed on every call, never cleaned
```

Mostly harmless on its own — but combine it with blocker #2 and a corrupt legacy file is re-read and re-thrown on **every session start for the rest of time**. Rename it to `.migrated` / delete it / drop a marker. Pick one.

## 🧂 SEASONING — nits

- `LEGACY_BUDGET_LIMITED_STATUSES` as a `string[]` + `.includes()` loses the literal-union narrowing you'd get from `as const`.
- The `AggregateError` cleanup branch in `publishMigratedGoalFile` disappears entirely once you adopt the `flag: "wx"` fix — good riddance, it was three states (`published` / `publicationFailed` / `publicationError`) tracking what one try/catch expresses.
- `changes.md` at -98/+45 is a solid net simplification. No complaints.

## ✅ WHAT'S ACTUALLY GOOD

- **The `{ legacy: true }` gating is exactly the right shape.** Confining normalization to the import path is the correct fix, and preserving `tokenBudget` on current reads is provably safe — `isGoal` at `persistence.ts:225` explicitly permits it.
- **The `legacyBaseDirFor` doc comment** spells out precisely why `dirname()` was wrong for the no-session layout, with both layouts written out. That's how you document a bugfix.
- **The concurrency test is legitimately sophisticated** — injecting at the exact legacy-read boundary to prove publication can't clobber a racing writer, then asserting the directory holds exactly one file. That's a test written by someone who understood the failure mode.
- The test suite survives all three of my structural mutations. Rare.

## VERDICT

**BLOCK** — on #1 and #2 only.

#1 is a portability regression against `main` implemented via thirty lines that a single `flag: "wx"` replaces. #2 is a permanent denial of the goal feature triggered by dead data the user doesn't own, and it ships with a test asserting that's intended.

The core idea, the path-segment fix, and the concurrency work are all genuinely well done. Fix the publication primitive, make the legacy import best-effort, and I'll happily send this out.

*Evidence: mutation matrix (91-test baseline), `link` vs `flag:"wx"` semantics probe, and corrupt-legacy / separator / cleanup probes all executed against `839b8874f` in an isolated worktree.*

code-yeongyu:
# [sisyphus-bot] 🔪 SERVICE ROUND 2: PR #587 — THE HEADLINE FEATURE DOESN'T FIRE

A second forensic pass found something my first review missed entirely, and it's worse than everything I filed before it. I'll own that: I reviewed the *publication mechanism* and never checked whether the migration it publishes ever actually **finds** anything.

It doesn't.

## 🔥 NEW BLOCKER #1 — no-session migration searches for the WRONG FILENAME

`persistence.ts:51-54`

The no-session store is cwd-keyed (`extensions/goal/no-session/<sha256(cwd)[:24]>/`) but the JSON file inside it is still named `<threadId>.json`. A fresh in-memory/print-mode session calls `newSession`, which **generates a brand-new session ID** (`session-manager.ts:1033-1037`, `SessionManager.inMemory` at `:1833-1835`).

So on the very upgrade this PR exists to serve, migration looks for `<new-id>.json` while the legacy goal is sitting right there as `<old-id>.json`. Same directory. Wrong plate.

I ran your real exported `migrateLegacyGoalFile` against exactly that layout:

```json
{"case":"different-no-session-thread-id","migrated":null}
{"case":"same-thread-id-control","migrated":"old-thread"}
```

**`null`.** The legacy goal is untouched, the user gets nothing, and no error is raised to tell anyone. Silent no-op on the headline path.

### And your test is holding both ends of the sausage

`goal-store.test.ts` — "finds the legacy store in the no-session fallback layout":

```ts
const legacyRef: GoalStoreRef = {
    ...ref,                              // ← inherits threadId: "thread-no-session-legacy"
    baseDir: join(dir, "extensions", "pi-goal", "no-session", cwdKey),
};
```

You spread `...ref`, so the legacy store gets the **identical thread ID** as the current one. That is not an upgrade — that is a test constructed to pass. It pins the one assumption that never holds in production, which is why 91 green tests never noticed the feature is dead.

**FIX:** no-session migration must not depend on the current ephemeral session ID. Enumerate the `*.json` candidates in the cwd-key bucket, parse them, and import when selection is unambiguous; surface an explicit conflict when several live candidates exist rather than guessing. Longer term, use a cwd-stable no-session filename. Then rewrite the test with **distinct** old/new IDs plus a multi-candidate case.

## 🔥 NEW BLOCKER #2 — no-session migration searches the WRONG ROOT

`persistence.ts:49-54`

`legacyBaseDirFor` derives the legacy directory purely by rewriting the *current* base dir's `goal` segment. But your own changelog (`goal/changes.md:602-604`) records that builtin goal state lives under `SENPI_CODING_AGENT_DIR` / `~/.senpi/agent`, while standalone `pi-goal` used its **`.pi` agent directory**. `goalStoreRef` only ever builds paths under the Senpi root (`store-ref.ts:11-14`).

Rewriting a segment inside `~/.senpi/agent` can therefore only ever look inside `~/.senpi/agent`. It can never reach `~/.pi/agent/extensions/pi-goal/...`, which is where a default standalone install's goals actually are.

```json
{"name":"different-agent-roots","migrated":null,"error":null,"current":null}
```

**FIX:** stop treating the current base as the only source of legacy candidates. For no-session migration inspect both the legacy standalone root (`PI_CODING_AGENT_DIR`, else `~/.pi/agent`) and the Senpi-root sibling used by redirected installs. Keep the segment-rewrite for session-backed stores. Test with genuinely distinct `.pi` and `.senpi` roots.

## 🍳 ALSO UNDERCOOKED (from the second pass)

- **Not crash-durable.** No `fsync` on the temp handle before publication, no parent-directory `fsync` after, no startup sweep for `.goal-*.tmp`. A `SIGKILL` mid-migration left an orphan temp behind (readers correctly saw no partial destination, so visibility atomicity holds — durability does not). Either implement it or stop implying it in the changelog. *(Note: adopting the `flag:"wx"` fix from my first review deletes the temp file entirely and makes the orphan problem disappear.)*
- **"Imported once" is false if the destination disappears.** No tombstone: import → `null` on rerun → **re-imports the stale goal** once the destination is deleted. `{"first":"old-goal","second":null,"third":"old-goal"}`. Stale work resurrects after manual recovery.
- **Lost updates in the store itself.** 32 real concurrent processes, all 32 exited successfully, final token count **1** — 31 updates lost. Pre-existing, not introduced here, but do not plate this as an "atomic goal store": temp+rename prevents torn bytes and does nothing for lost updates.
- **`legacyBaseDirFor`'s segment logic is genuinely good** and handled every adversarial shape thrown at it (a home segment literally named `goal`, nested `goal/.../goal`, `my-goal`, relative paths). Do **not** replace it with raw string substitution when fixing the above — that part is properly cooked.

## ✅ STILL GOOD, AND I'LL SAY IT TWICE

The mutation matrix is real coverage and that is rarer than it should be: unconditional normalization → 5 failed; old `dirname` path → 2 failed; check-then-rename publication → the concurrency test failed. Against the true `main` baseline the new tests produce **7 failed / 84 passed**, so this diff genuinely changes behavior and genuinely proves it. The `{ legacy: true }` gating is the correct fix and the `EEXIST` convergence between two concurrent migrators is properly reasoned.

That's exactly why these two blockers matter. You built a careful, well-tested publication mechanism — and wired it to a lookup that can't find the file.

## 📋 CONSOLIDATED BLOCKER LIST

1. `link()` portability regression; `writeFile(flag:"wx")` is byte-identical and removes the temp file entirely. *(round 1)*
2. Corrupt legacy file permanently bricks the goal feature — with two tests codifying it as intended. *(round 1)*
3. **No-session migration searches the current session's fresh thread ID → silent no-op on the headline upgrade path.** *(round 2)*
4. **No-session migration can only search the Senpi root → never finds default standalone `~/.pi/agent` state.** *(round 2)*
5. Separator-blind `split(sep)` → silent skip on `/`-joined win32 paths. *(minor, round 1)*
6. Legacy file never retired → re-parsed forever; compounds #2. *(minor, round 1)*

## VERDICT

**BLOCK.** The publication mechanism is sound. The migration that feeds it cannot locate real legacy state across either the agent-root boundary or a restart's session-ID boundary — and the test suite that would have caught it was written with both ends pinned to the same value.

Fix 3 and 4 and this becomes the good PR it's two-thirds of the way to being.

*Evidence: `migrateLegacyGoalFile` executed against real temp roots for both boundary cases; `main`-baseline reversion (7 failed / 84 passed); mutation matrix; SIGKILL durability probe; 32-process concurrency probe.*

code-yeongyu:
# [sisyphus-bot] ✅ SERVICE ACCEPTED: PR #587

I came back with the knife out and I'm putting it down. **All six blockers are closed**, verified independently at `9c605981b` in a clean worktree — I re-ran every probe myself rather than taking the report's word.

## The headline blocker is genuinely dead

The no-session upgrade path — the thing this PR exists for — used to return `null` because it searched for the *new* session's filename. Same real-surface harness, same layout, now:

```
S5_nosession_fresh_id_migrated: "ns-goal"     (was: null)
```

It finds the legacy goal across a fresh session id. And across a distinct agent root:

```
B6_distinct_pi_root: "pi-root-goal"
```

Both boundaries that the old code could not cross.

## Every other blocker, verified

```
B1_current_wins: already        | tmp litter: []       ← wx exclusivity, no temp file at all
B2_corrupt_legacy_3x: [null,null,null]                 ← no longer bricks the feature
B2_current_store_usable: live                          ← store still works alongside corrupt legacy
B2b_current_corrupt_still_throws: InvalidGoalStoreError ← did NOT over-swallow
B3_forward_slash_baseDir: "sep-goal"                   ← separator-blind bug fixed
B4_migrated: retire-me | legacy dir: ["r.json.migrated"]
B4_second_call_is_noop: null                           ← migration finally terminates
```

**B2b is the one I want to call out as good judgment.** The easy way to fix "corrupt legacy blocks everything" is to wrap the whole parse in a `catch` and swallow it — which would have silently destroyed the typed-error contract on the *current* store too. You scoped the leniency to the legacy import path only. Current-store corruption still raises `InvalidGoalStoreError`. That's the correct distinction and it would have been very easy to get wrong.

Retiring the consumed legacy file to `.migrated` rather than deleting it is also the right call — non-destructive, recoverable, and it makes the operation terminate.

And dropping `link()` for `flag: "wx"` removed the temp file entirely, which killed the SIGKILL-orphan wart for free.

## Tests and mutation re-proof

```
4 goal suites: 97 passed (97)

mutation: remove {legacy} gating      ->  4 failed / 93 passed
mutation: flag:"wx" -> flag:"w"       ->  1 failed / 96 passed
```

The suite still bites after the changes. Coverage stayed real — which, as I said in round one, was already this PR's strongest quality and is the standard I wish more changes met.

## VERDICT

**APPROVED.**

You built a careful publication mechanism in round one and wired it to a lookup that couldn't find the file. Now it finds the file, across both boundaries, and it does it without a hard-link dependency, without bricking on dead data, and without lying about the current store's error contract.

That's a proper plate. Well done.

*Verified independently at `9c605981b`: real-surface migration harness, six-blocker probe suite, 97-test run, and a 2-mutation re-proof — all executed in an isolated worktree.*

PR #588, opened 2026-07-31
fix: prevent unavailable-tool transcript imitation

code-yeongyu:
## Summary

- reshape unavailable Anthropic tool history into non-imitable XML-ish transcript records that omit call inputs, retain narrowly-neutralized results, and derive current-tool guidance from the real request
- add a default text-only TTSR rule that interrupts fabricated new or legacy unavailable-tool envelopes, nudges the model to call real tools, and retries
- make `--ttsr-rules-disabled` effective for manager-held builtin, project, and global rules; partition `/ttsr` status into detector, builtin stream-rule, and user-rule sections
- remove the committed `TTSRDBG` stdout logs that corrupted TUI rendering

## Root cause

Session `019fb708` made 12 `apply_patch` calls on `gpt-5.6-sol-fast`, then switched to `claude-opus-5`. The model switch removed `apply_patch` from the active toolset, so Anthropic request validation required those orphaned references to be demoted. The old demotion replayed 50,381 bytes of patch input as assistant-role text in 12 imitable `[Called tool ... with input: ...]` envelopes. Claude copied that envelope verbatim as text (`stopReason: "stop"`, zero tool calls, 8,307 characters), no file changed, and the turn silently ended unfinished.

The demotion cannot be removed because Anthropic rejects unavailable history references with `400 invalid_request_error`. This PR reshapes it and adds a live-output safety net.

### Anthropic demotion

- first occurrence per missing tool name, per request: full `<unavailable-tool-call>` transcript explanation
- later occurrences of the same name: self-closing record
- tool list comes from request-local `definedNames`, preserving order and capped at 8 with an omitted count; zero-tool requests use a truthful fallback
- historical call input is dropped entirely
- result text is retained; case-insensitive closing-tag openers are neutralized by escaping only `<`, preserving ordinary text and original casing
- exotic names are XML-attribute escaped
- builders live under non-public `src/utils/`, shrinking rather than inflating the Anthropic adapter

### TTSR safeguard

- manager-backed builtin registered before discovered rules; its name is reserved under first-registration-wins semantics
- case-insensitive conditions cover both `<unavailable-tool-call ...>` and persisted legacy `[Called tool "..." (no longer available in this session)` text
- explicit text-only scope, `interruptMode: "always"`, action-oriented remediation
- intentionally no raw `*** Begin Patch` match due prose false positives
- accepted tradeoff: legitimate prose discussing the senpi-specific envelope may trigger once per session; repeat-once and nudge remediation cap the cost

## Behavior change

`--ttsr-rules-disabled` previously gated only the two detector-only builtins. `TtsrManager.addRule()` now rejects disabled names, so the flag also disables manager-held builtin, project, and global rules. `/ttsr` lists active builtin stream rules separately and keeps `USER RULES` truthful.

## Verification

Evidence root: `local-ignore/qa-evidence/20260731-ttsr-xml-demotion/`

- C1 RED/GREEN: `c1-anthropic-red.log`, `c1-anthropic-green.log`
- C2 integrity + bypass RED/GREEN: `c2-integrity-baseline.log`, `c2-integrity-green.log`, `c2-case-bypass-red.log`, `c2-case-bypass-green.log`
- C3 RED/GREEN + mutation proofs: `c3-ttsr-red.log`, `c3-ttsr-green.log`, `c3-mutation-scope.log`, `c3-mutation-disable.log`
- C4: `c4-ttsr-directory-green.log` (15 files / 285 tests), `c4-regression-green.log` (52 tests: TTSR extension 7, gpt-apply-patch 29, AI integrity 5, policy 6, characterization 5)
- C5 built CLI: `c5-cli-red.log` (`status_rendered=true`, `builtin_rule_present=false`), `c5-cli-green.log` (`status_rendered=true`, `builtin_rule_present=true`), `c5-cli-capture.txt`
- root gate: `root-check.log`

Additional direct run: AI demotion + untouched integrity 7/7. Root `npm run check` passed. `git diff --check` is clean.

Plan: `.omo/plans/ttsr-xml-demotion.md`

---
## Summary by cubic
Prevents transcript imitation when Anthropic tools go missing by reshaping demoted history into safe XML records, and adds a default TTSR rule that interrupts fabricated envelopes and retries with a nudge. Also expands `--ttsr-rules-disabled` coverage and cleans up `/ttsr` status output and noisy logs.

- **Bug Fixes**
  - Replace imitable `[Called tool ...]` text with non-actionable `<unavailable-tool-call>` and `<unavailable-tool-result>` records.
  - Drop historical call inputs; keep result text while safely neutralizing closing-tag forgeries and escaping exotic names.
  - Show a capped list of currently available tools (per request; first occurrence is full, later ones are self-closing).

- **New Features**
  - Add a builtin TTSR stream rule that interrupts `<unavailable-tool-call ...>` or legacy unavailable-tool text, nudges the model to call real tools, and retries (text-only, `interruptMode: "always"`).
  - Make `--ttsr-rules-disabled` apply to manager-held builtin, project, and global rules; partition `/ttsr` into builtin detectors, builtin stream rules, and user rules; remove `TTSRDBG` stdout logs.

<sup>Written for commit 9f726cf9e8438a1fcb1fe5efa6eff7ccea358ca9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/plans/ttsr-xml-demotion.md
+++ b/.omo/plans/ttsr-xml-demotion.md
@@ -19,7 +19,7 @@ Ship and merge one PR that removes Anthropic unavailable-tool context poisoning 
 
 1. Add a dedicated builtin-rule module returning a typed `TtsrRule` with source `builtin`. Register builtin rules in `ensureInitialized()` before discovered global/project rules.
 2. Duplicate-name precedence is deliberate: builtin names are reserved and win because `TtsrManager.addRule()` is first-wins. A project/global file with the same name is inspectable on disk but cannot weaken the shipped safety rule. Record this policy in the TTSR changes file.
-3. Add two regex conditions: one for `<unavailable-tool-call ...>` (including full and self-closing forms) and one for the persisted legacy `[Called tool "..." (no longer available in this session)` prefix. Do not add a raw `*** Begin Patch` condition because it would false-positive on legitimate prose explaining patch syntax.
+3. Add two case-insensitive regex conditions using a leading `(?i)` group: one for `<unavailable-tool-call ...>` (including full and self-closing forms) and one for the persisted legacy `[Called tool "..." (no longer available in this session)` prefix. Case-insensitivity is deliberate because model-authored transcript imitations are not trustworthy XML and uppercase/mixed-case variants remain equally imitable. Do not add a raw `*** Begin Patch` condition because it would false-positive on legitimate prose explaining patch syntax.
 4. Scope the rule explicitly to text only: `allowText: true`, `allowThinking: false`, `toolScopes: []`. Use `interruptMode: "always"` and action-oriented content telling the model to call its real tools and redo the interrupted step.
 5. Make `TtsrManager.addRule()` reject names present in `settings.disabledRules`. Existing `StreamWatcher.disabledBuiltin` continues to gate the two detector-only builtins; manager-backed builtin/project/global rules use the settings gate. This makes `--ttsr-rules-disabled=<name>` work for the new builtin through the real registration path.
 6. Remove both `TTSRDBG` `console.log` calls. Partition manager-held rules by `source` in `/ttsr`: manager builtins appear in the builtin section clearly labeled as stream rules, while only project/global rules feed `USER RULES`. This preserves the truthful `USER RULES\n(none)` contract when no user files exist.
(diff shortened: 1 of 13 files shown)


code-yeongyu:
## Pre-existing CI failure on `Test (coding-agent 2/3)` — not caused by this PR

`Test (coding-agent 2/3)` (and its aggregator `Check and test`) fails here. Evidence that this is a pre-existing failure on `main`, not a regression from this change:

1. **`main` at `f4705697b` — the exact base commit of this branch — has the identical failure set.** Run [30613993315](https://github.com/code-yeongyu/senpi/actions/runs/30613993315): `Test (coding-agent 2/3)` failure, `Check and test` failure, and every other job green (`1/3`, `3/3`, workspaces+scripts, static checks, terminal tools x3, inspector handoff x3).
2. **The same shard passes locally on this branch** at `9f726cf9e`:
   ```
   npx vitest run --shard=2/3
   Test Files  254 passed (254)
        Tests  2289 passed (2289)
   ```

Every other check on this PR is green, including `Static checks`, all three `Terminal tools` platforms, all three `Inspector handoff` platforms, `Test (coding-agent 1/3)`, `Test (coding-agent 3/3)`, `Test (workspaces + scripts)`, `generate`, and GitGuardian.

The shard-2/3 failure on `main` is a separate problem and should be tracked on its own; it is outside this PR's scope.

code-yeongyu:
## Correction to my earlier CI note

My previous comment characterized the `Test (coding-agent 2/3)` failure as a **pre-existing deterministic failure on `main`**. That conclusion was wrong, and I'm correcting it so nobody goes looking for a broken test that isn't deterministically broken.

**What actually happened:** the job was **flaky**. It has since passed on rerun with no code change — [job 91115864673](https://github.com/code-yeongyu/senpi/actions/runs/30617695257/job/91115864673), 5m54s, `success`. All checks on this PR are now green, including `Check and test`.

The two observations in my earlier comment were individually accurate — `main` at `f4705697b` did record a failure on that same job, and the shard did pass locally at 254 files / 2289 tests — but "flaky test that failed on both runs" explains them at least as well as "pre-existing deterministic failure", and the passing rerun settles it in favor of flakiness.

The genuinely useful signal is narrower: **this shard is flaky and worth stabilizing**, per the repo's own rule that tests must not pass by timing luck. That is still worth tracking separately, but as flake stabilization rather than a broken-on-main hunt.

PR #646, opened 2026-08-02
fix(compaction): recover prepared suffix after summary timeout

madgegja:
## Summary
- validate deterministic fallback fit from reconstructed retained message content instead of stale pre-compaction assistant usage
- allow pre-prompt deterministic recovery only at the hard input cap while preserving fail-closed behavior below it
- keep final provider result settlement and the full extension compaction operation inside the 120-second watchdog budget

## Regression evidence
The affected 272K-context session retained about 20,105 tokens after compaction, but the previous estimator reused a 252,607-token assistant usage snapshot and reported 255,699 tokens, 83 above the 255,616 threshold. The rebuilt branch accepts that real prepared suffix and still rejects a genuinely unsafe 20,000-token window.

## Validation
- `vitest --run` on the three changed compaction suites: 39 tests passed
- `tsgo --noEmit`
- Biome check on all changed TypeScript files
- ordered root `npm run build`
- replayed saved session `019fbd3f-bb18-7329-81f8-6baf64c25bda` against the built artifact

---
## Summary by cubic
Recover prepared suffix after summarization timeout by computing fit from retained message content instead of stale assistant usage. Keep final provider result settlement inside the 120s watchdog and apply the same budget across the full extension compaction operation to prevent stuck sessions.

- **Bug Fixes**
  - Deterministic fallback: recompute token fit from retained messages; allow `pre_prompt` recovery only at the hard input cap; optional pre-prompt below the cap stays fail-closed; latest request is preserved.
  - Watchdog: include `responseStream.result()` in the 120s budget; abort with `StreamDurationBudgetError` when the final result lingers; extension compaction now bounds credential resolution and stream setup under the same budget, with caller abort taking precedence.
  - Tests: added coverage for suffix recovery with stale assistant usage, pre-prompt gating at the cap, and final-result settlement timing.

<sup>Written for commit d33c1cb3559598a640364518dcb3c9e938cc1636. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/646?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
ULW audit verdict: BLOCKED / SUPERSEDED-IN-PART. The historical suffix recovery works and uniquely covers final-result settlement, but its replay safety is too permissive, current/main and PR #736 conflicts are real, and a fresh combined patch is required. Local evidence: `local-ignore/qa-evidence/20260817-fix-pr-audit/pr-646/VERDICT.md`.

GunP4ng:
Reproduced the user-visible deadlock on current `main` / Senpi 2026.9.15-2 with Astra: mandatory `pre_prompt` summary stalled for 300,000 ms and left context unchanged. I opened #1735 with a current-main failing-first regression plus a read-only replay of the affected 32 MB session. It also covers the unsafe split-turn suffix case that prevented deterministic recovery.

PR #647, opened 2026-08-02
fix(ai): strip non-standard 'optional' keyword from tool JSON schema  parameters

Indosaram:
Fixes 400 Bad Request error on Gemini API and proxies (e.g. Quotio) caused by non-  standard 'optional' fields inside tool parameter JSON schemas.

---
## Summary by cubic
Strips the non-standard 'optional' keyword from tool parameter JSON schemas to prevent 400 Bad Request errors on Gemini and proxies like Quotio. Applies to normalized schemas and Gemini `parametersJsonSchema`.

- **Bug Fixes**
  - Recursively remove `optional` from `parametersJsonSchema` when `useParameters=false` in `convertTools`.
  - Ignore `optional` during validation and drop it in schema normalization.

<sup>Written for commit ddb2531aa09934f3967316415b3cb9f89e71b393. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/647?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/src/api/google-shared.ts
+++ b/packages/ai/src/api/google-shared.ts
@@ -282,6 +282,7 @@ const JSON_SCHEMA_META_DECLARATIONS = new Set([
 	"$comment",
 	"$defs",
 	"definitions", // pre-draft-2019-09 equivalent of $defs
+	"optional",
 ]);
 
 /**
@@ -296,6 +297,26 @@ function sanitizeForOpenApi(schema: unknown): unknown {
 	for (const [key, value] of Object.entries(schema)) {
 		if (JSON_SCHEMA_META_DECLARATIONS.has(key)) continue;
 		result[key] = sanitizeForOpenApi(value);
+	}
+	return result;
+}
+
+/**
+ * Strip non-standard 'optional' keyword recursively from a JSON schema object.
+ */
+function stripOptional(schema: unknown): unknown {
+	if (typeof schema !== "object" || schema === null) {
+		return schema;
+	}
+
+	if (Array.isArray(schema)) {
+		return schema.map(stripOptional);
+	}
+
+	const result: Record<string, unknown> = {};
+	for (const [key, value] of Object.entries(schema)) {
+		if (key === "optional") continue;
+		result[key] = stripOptional(value);
 	}
 	return result;
 }
@@ -320,7 +341,7 @@ export function convertTools(
 				description: tool.description,
 				...(useParameters
 					? { parameters: sanitizeForOpenApi(tool.parameters) }
-					: { parametersJsonSchema: tool.parameters }),
+					: { parametersJsonSchema: stripOptional(tool.parameters) }),
 			})),
 		},
 	];
--- a/packages/ai/src/utils/tool-schema-compat.ts
+++ b/packages/ai/src/utils/tool-schema-compat.ts
@@ -145,6 +145,8 @@ function normalizeNode(node: unknown): unknown {
 		return node;
 	}
 
+	delete node.optional;
+
 	const hasCombiner = COMBINER_KEYS.some((key) => Array.isArray(node[key]));
 	if (hasCombiner) {
 		moveTypeIntoCombinerBranches(node);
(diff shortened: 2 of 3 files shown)


Indosaram:
Any updates? This blocks me to use Gemini models in OMO

packages/ai/src/api/google-shared.ts
```diff
@@ -300,6 +301,26 @@ function sanitizeForOpenApi(schema: unknown): unknown {
...
+		return schema;
+	}
+
+	if (Array.isArray(schema)) {
+		return schema.map(stripOptional);
+	}
+
+	const result: Record<string, unknown> = {};
+	for (const [key, value] of Object.entries(schema)) {
+		if (key === "optional") continue;
```
code-yeongyu, line 318: [P1] This deletion is applied to every object reached by the generic recursion, so it also removes a legitimate property schema named `optional` under `properties` and `optional` data inside `const`/`default`/`examples`. In a direct `convertTools()` run, `properties: { optional: { type: "boolean" } }, required: ["optional"]` became `properties: {}, required: ["optional"]`. Please make traversal schema-position-aware and add regressions for a property literally named `optional` plus object-valued schema data. JSON Schema's `properties` keys are instance property names: https://json-schema.org/draft/2020-12/json-schema-core#section-10.3.2.1

code-yeongyu (changes requested):
## Review result: changes requested

The intended keyword-position case is covered (mutation RED -> GREEN, focused tests pass), but the current recursive deletion is not schema-position-aware and corrupts valid tool schemas.

### [P1] Preserve legitimate properties and data named `optional`

`stripOptional()` recursively drops every object member whose key is `optional`. That includes ordinary instance-property names under `properties` and data inside value keywords such as `const`, `default`, and `examples`, not only a non-standard schema keyword.

Concrete runtime reproductions against this PR head:

- `{ type: "object", properties: { optional: { type: "boolean" } }, required: ["optional"] }` becomes `{ type: "object", properties: {}, required: ["optional"] }` on both Gemini conversion modes. The emitted schema is internally inconsistent and the real tool parameter disappears.
- `{ const: { optional: true }, default: { optional: false }, examples: [{ optional: true }] }` loses those data members.

JSON Schema defines the keys inside `properties` as instance property names, so `optional` is valid there; unknown schema keywords may be annotations, but that does not make every nested key named `optional` a keyword. Please make the traversal schema-position-aware: strip `optional` only from schema objects, preserve map keys under `properties`/`patternProperties`/`$defs`/`definitions`, and do not traverse value-bearing keywords (`const`, `default`, `examples`, `enum`) as schemas. Add regression coverage for a property literally named `optional` and object-valued schema data.

Official references:
- JSON Schema Core 2020-12, unknown keywords: https://json-schema.org/draft/2020-12/json-schema-core#section-4.3.1
- JSON Schema Core 2020-12, `properties`: https://json-schema.org/draft/2020-12/json-schema-core#section-10.3.2.1

### [P2] Strip the keyword inside array-valued schema branches on the legacy Gemini path

`sanitizeForOpenApi()` returns arrays unchanged. Therefore `convertTools(tools, true)` leaves `optional: true` inside `anyOf` branches (and tuple-like array schema positions), while `convertTools(tools, false)` strips the same input. A real module driver produced `anyOf: [{ type: "string", optional: true }, ...]` in `parameters`. If this keyword causes Gemini/proxy 400s, the legacy path still fails for nested combiners. Please recurse into schema arrays and add a `useParameters=true` regression case.

Validation performed at head `ddb2531aa09934f3967316415b3cb9f89e71b393`: PR test mutation RED then restored GREEN; focused schema tests 16/16; full AI suite 1579 passed / 800 skipped after required workspace builds; root `npm run check` passed; real isolated Senpi CLI mock loop 4/4 passed. The two cases above still fail in direct `convertTools` execution.

code-yeongyu:
Opened fix PR #841 that addresses both review findings (P1: schema-position-aware stripping, P2: array recursion on legacy path) with TDD evidence. Will merge once CI is green.

code-yeongyu:
Closing in favor of #841 which includes the position-aware fix for both review findings.

PR #649, opened 2026-08-02
fix(apply-patch): compact persisted result previews

### LINKED ISSUE #648: apply_patch retains duplicated full-file previews, causing session memory and TUI slowdown
## Summary

Long-running Senpi coding sessions progressively consume memory and become slower because successful `apply_patch` tool results persist very large hidden `details` payloads in the append-only session. The final result can retain the same full-file patch preview twice: once as `details.preview.files` and again under `details.result.details.appliedOperations`.

The terminal/PTY scrollback is bounded; the primary issue is retained session/tool metadata, with full-history TUI rendering acting as a secondary amplifier.

## Environment

- Senpi: `0.12.2`
- Node.js: `v24.12.0`
- OS: Windows 11 x64
- UI mode: fullscreen, smooth streaming enabled at 120 FPS

## Measurements

From one affected session (paths/content sanitized):

- Session JSONL size: **187.61 MB**
- Exact `SessionManager.open()` load time: **4.227 s**
- Heap after forced GC: **12.5 MB -> 174.7 MB**
- `ResidentStringStore`: **148.87 MB** across 69 blobs
- `message:toolResult` retained bytes: **145.10 MB**
- Hidden `apply_patch` details: **139.41 MB**
- Visible tool-result content across all tools: only **4.39 MB**
- Largest single `apply_patch` detail payload: **126.57 MB**
  - `preview.files`: **63.28 MB**
  - `result.details.appliedOperations`: **63.28 MB**

Thus one patch stored essentially the same full-file preview twice while its visible result remained tiny.

## Relevant code path

In the published package:

1. `dist/core/extensions/builtin/gpt-apply-patch/apply.js` builds `appliedOperations` containing each `applied.preview`.
2. `dist/core/extensions/builtin/gpt-apply-patch/tool.js` derives `resultPreview` from those operations, then returns both:

```js
details: resultPreview ? { preview: resultPreview, result } : { result }
```

`result` still contains `result.details.appliedOperations`, so full old/new file content is represented in both branches.

3. `dist/core/session-manager.js` appends the complete tool-result entry to `fileEntries`.
4. `ResidentStringStore` keeps large strings in an in-memory `Map`; it does not move them out of process or discard them after compaction.

## Reproduction

1. Start a Senpi session.
2. Perform a sufficiently large multi-file `apply_patch` operation, especially against large files.
3. Continue using or later resume the session.
4. Inspect the session JSONL or load it through `SessionManager.open()` and call `getResidentStoreStats()`.
5. Group tool-result byte sizes by `message.details` and tool name.

Memory and session size scale with full patch preview contents rather than the small visible patch summary.

## User-visible behavior

- A session becomes increasingly sluggish over time.
- Resuming a large session has a multi-second load delay.
- Node memory usage grows substantially.
- The fullscreen TUI compounds the issue because the chat container accumulates components and each frame renders the complete line list. With smooth streaming at 120 FPS, growing history increases work per streamed frame.

The persistent-terminal implementation itself is bounded (`scrollback` plus a 1,000,000-character decoded-output cap), so terminal scrollback is not the primary unbounded store.

## Expected behavior

Final persisted `apply_patch` results should contain only lightweight metadata required after execution: affected paths, summaries, operation indexes, fuzz, and failure/recovery information. Full old/new file contents should not remain duplicated—or ideally should not be persisted at all after rendering.

Compaction or completion of a tool call should also permit heavyweight preview data to be released.

## Suggested fix

- Do not return both `resultPreview` and `result.details.appliedOperations` with full preview bodies.
- Strip full file contents from final persisted tool details after producing the visible preview.
- Keep only lightweight summaries/paths in the final result, or introduce a persistence sanitizer for tool-specific transient details.
- Consider virtualizing/pruning off-screen chat components as a separate TUI optimization.

No existing coding sessions were terminated while collecting these measurements; reproduction used separate short-lived Node processes importing the published `SessionManager`.

daniduro89:
## Summary

- cap retained top-level unified patches at 16 KiB per file; oversized patches are omitted whole rather than truncated, avoiding malformed unified diffs
- keep TUI-visible diffs bounded and replace nested applied-operation previews with metadata after constructing the final preview
- add regression coverage for bounded persisted details and the app-server's empty file-change projection for oversized patches

Fixes #648

## Verification

Current head `0f43a753e`:

- `npm --prefix packages/coding-agent test -- test/suite/gpt-apply-patch test/suite/regressions/648-apply-patch-result-retention.test.ts` — 8 files / 61 tests passed
- `npm --prefix packages/coding-agent test -- test/suite/app-server-task24-apply-patch-identity.test.ts test/suite/app-server-task24-file-change-results.test.ts test/suite/app-server-task24-diff.test.ts test/suite/app-server-task24-review-regressions.test.ts` — 4 files / 15 tests passed
- mutation proof: the new oversized-projection test failed when the 16 KiB retention guard was disabled, then passed after restoring it

Earlier verification recorded for commit `3ee70786a` and not re-run on the current head:

- `npm run check` — passed
- `npm run build` — passed
- Senpi QA sandbox self-check — 9/9 passed
- Senpi CLI smoke — 8/8 passed
- deterministic mock-loop QA — 43/43 passed
- five-lane pre-PR review (constraints, quality, security, hands-on QA, upstream context) — passed

## Compatibility

Complete `preview.files[].patch` unified diffs are retained only when each file's patch is at most 16 KiB. Oversized patches are omitted rather than truncated, so app-server file-change projection emits no change and no diff for those files; IDE/app-server clients therefore show no file-change diff for oversized patches. The filesystem mutation still completes, and within-budget patches remain projected as complete unified diffs.

---
## Summary by cubic
Compact `apply_patch` result previews to remove duplicate full diffs and cap retained patch bodies to 16 KiB per file. Keeps bounded visible diffs, preserves unified patches only within budget, and stores metadata-only for nested operations to reduce session size.

- **Bug Fixes**
  - Persist TUI-bounded diffs in the preview via `truncatePreview()`.
  - Retain `preview.files[].patch` only if ≤16 KiB; omit oversized patches to avoid malformed truncations (app-server projection stays complete within budget).
  - Persist compacted results for both success and `ApplyPatchError` paths: nested `result.details.appliedOperations` drop patch bodies and clear diffs, keeping metadata (paths, operation type, counts, indexes).
  - Add regression test covering bounded diffs, 16 KiB patch retention, and metadata-only nesting.

<sup>Written for commit 1711ce35251ae2ec4122eaa9db3d91950211d5f6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/649?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/apply.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/apply.ts
@@ -44,6 +44,19 @@ function extractErrorCode(error: unknown): string | undefined {
 	return undefined;
 }
 
+export function compactApplyPatchResult(result: ApplyPatchResult): ApplyPatchResult {
+	return {
+		...result,
+		details: {
+			...result.details,
+			appliedOperations: result.details.appliedOperations.map(({ operationIndex, preview }) => {
+				const { diff: _diff, patch: _patch, ...metadata } = preview;
+				return { operationIndex, preview: { ...metadata, diff: "" } };
+			}),
+		},
+	};
+}
+
 async function writeFileAtomic(
 	absPath: string,
 	content: string,
@@ -224,7 +237,7 @@ export async function applyPatch(cwd: string, patchText: string): Promise<string
 				recoveryInstructions: createRecoveryInstructions({ appliedFiles, failures }),
 				details: { fuzz: 0, appliedOperations },
 			};
-			throw new ApplyPatchError(message, result);
+			throw new ApplyPatchError(message, compactApplyPatchResult(result));
 		}
 	}
 
--- a/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
@@ -4,21 +4,22 @@
 
 ### What changed
 
-- `tool.ts`: completed previews persist `truncatePreview()` output while preserving each top-level unified `patch`.
-- Nested `result.details.appliedOperations` entries retain only operation indexes, paths, move destinations, operation types, and line counts.
-- Regression coverage verifies bounded visible diffs, preserved unified patches, and metadata-only nested operations.
+- `tool.ts`: completed previews persist `truncatePreview()` output and retain a complete unified `patch` only when it is at most 16 KiB per file; oversized patches are omitted instead of retaining source-sized bodies or producing malformed truncations.
+- `apply.ts`: pure result compaction uses destructuring omission so nested applied operations keep indexes and preview metadata, including future optional fields, while dropping full patch bodies and emptying diffs. The fail-fast `ApplyPatchError` path uses the same compaction.
+- Regression coverage uses a 3,000-line fixture to cap serialized completed details below 12,000 bytes, while existing app-server projection tests pin complete diffs for patches within budget.
 
 ### Why
 
-- Completed results retained full operation diffs in both the visible preview and nested result details, duplicating large payloads in append-only sessions.
+- After visible-diff and nested-operation compaction, `preview.files[].patch` still contained both old and new file contents and accounted for effectively all retained bytes on large updates, additions, and deletions.
+- App-server projection runs from the same result object later copied into the persisted tool-result message. There is no builtin-extension seam after projection but before persistence, so a documented fixed budget is the smallest safe retention boundary; omission is preferable to an invalid partial unified diff.
 
 ### Why extension system couldn't handle this
 
-- The redundant payload is assembled by the builtin tool before the session manager persists the tool result.
+- The source-backed patch payload is assembled by the builtin tool before both app-server projection and session persistence.
 
 ### Expected merge conflict zones
 
-- LOW: `tool.ts` completed-result construction.
+- LOW: `apply.ts` result compaction and `tool.ts` completed-preview construction.
 
 ## Failed apply_patch outcomes are explicit errors (2026-07-31)
 
--- a/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
@@ -5,8 +5,9 @@
 ### What changed
 
 - `tool.ts`: completed previews persist `truncatePreview()` output and retain a complete unified `patch` only when it is at most 16 KiB per file; oversized patches are omitted instead of retaining source-sized bodies or producing malformed truncations.
+- `index.ts`: the public barrel exports `APPLY_PATCH_RESULT_PATCH_MAX_BYTES` alongside the tool factory and preview limits so retention-contract tests share the production budget.
 - `apply.ts`: pure result compaction uses destructuring omission so nested applied operations keep indexes and preview metadata, including future optional fields, while dropping full patch bodies and emptying diffs. The fail-fast `ApplyPatchError` path uses the same compaction.
-- Regression coverage uses a 3,000-line fixture to cap serialized completed details below 12,000 bytes, while existing app-server projection tests pin complete diffs for patches within budget.
+- Regression coverage uses a 3,000-line fixture to cap serialized completed details below half the per-file patch budget. App-server projection tests pin complete diffs within budget and an empty projection when a patch exceeds it.
 
 ### Why
 
--- a/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
@@ -1,5 +1,25 @@
 # changes
 
+## Compact completed result retention (2026-08-02)
+
+### What changed
+
+- `tool.ts`: completed previews persist `truncatePreview()` output while preserving each top-level unified `patch`.
+- Nested `result.details.appliedOperations` entries retain only operation indexes, paths, move destinations, operation types, and line counts.
+- Regression coverage verifies bounded visible diffs, preserved unified patches, and metadata-only nested operations.
+
+### Why
+
+- Completed results retained full operation diffs in both the visible preview and nested result details, duplicating large payloads in append-only sessions.
+
+### Why extension system couldn't handle this
+
+- The redundant payload is assembled by the builtin tool before the session manager persists the tool result.
+
+### Expected merge conflict zones
+
+- LOW: `tool.ts` completed-result construction.
+
 ## Failed apply_patch outcomes are explicit errors (2026-07-31)
 
 ### What changed
--- a/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
@@ -1,7 +1,7 @@
 import { Box, Container, Spacer, Text } from "@earendil-works/pi-tui";
 import type { AgentToolResult, ToolRenderContext } from "../../types.ts";
 import { defineTool } from "../../types.ts";
-import { applyPatchDetailed, buildPartialFailureText } from "./apply.ts";
+import { applyPatchDetailed, buildPartialFailureText, compactApplyPatchResult } from "./apply.ts";
 import {
 	APPLY_PATCH_FREEFORM_DESCRIPTION,
 	APPLY_PATCH_JSON_DESCRIPTION,
@@ -23,13 +23,27 @@ import type {
 	FreeformToolFormat,
 } from "./types.ts";
 
+export const APPLY_PATCH_RESULT_PATCH_MAX_BYTES = 16 * 1024;
+
+function retainedPatch(patch: string | undefined): string | undefined {
+	if (patch === undefined || Buffer.byteLength(patch, "utf8") > APPLY_PATCH_RESULT_PATCH_MAX_BYTES) {
+		return undefined;
+	}
+	return patch;
+}
+
 function appliedPreview(result: ApplyPatchResult): ApplyPatchPreview | undefined {
 	const files = [...result.details.appliedOperations]
 		.sort((left, right) => left.operationIndex - right.operationIndex)
-		.map((operation) => ({
-			...operation.preview,
-			diff: truncatePreview(operation.preview.diff),
-		}));
+		.map((operation) => {
+			const { diff, patch, ...metadata } = operation.preview;
+			const boundedPatch = retainedPatch(patch);
+			return {
+				...metadata,
+				diff: truncatePreview(diff),
+				...(boundedPatch === undefined ? {} : { patch: boundedPatch }),
+			};
+		});
 	if (files.length === 0) return undefined;
 	return {
 		files,
@@ -103,20 +117,6 @@ function renderFailureBox(
 	return component;
 }
 
-function compactAppliedOperations(result: ApplyPatchResult): void {
-	result.details.appliedOperations = result.details.appliedOperations.map(({ operationIndex, preview }) => ({
-		operationIndex,
-		preview: {
-			filePath: preview.filePath,
-			...(preview.movePath === undefined ? {} : { movePath: preview.movePath }),
-			operation: preview.operation,
-			diff: "",
-			added: preview.added,
-			removed: preview.removed,
-		},
-	}));
-}
-
 export function createApplyPatchTool(variant: "freeform" | "json" = "freeform"): ApplyPatchToolDefinition {
 	const tool = defineTool<typeof APPLY_PATCH_PARAMS, ApplyPatchToolDetails | undefined, ApplyPatchRenderState>({
 		name: "apply_patch",
@@ -153,16 +153,18 @@ export function createApplyPatchTool(variant: "freeform" | "json" = "freeform"):
 				onUpdate?.({ content: [{ type: "text", text: progressUpdate.text }], details: progressUpdate.details });
 			});
 			const resultPreview = appliedPreview(result);
-			compactAppliedOperations(result);
+			const persistedResult = compactApplyPatchResult(result);
 			if (result.failures.length > 0) {
 				return {
 					content: [{ type: "text", text: buildPartialFailureText(result) }],
-					details: resultPreview ? { preview: resultPreview, result } : { result },
+					details: resultPreview
+						? { preview: resultPreview, result: persistedResult }
+						: { result: persistedResult },
 				};
 			}
 			return {
 				content: [{ type: "text", text: result.summaries.join("\n") }],
-				details: resultPreview ? { preview: resultPreview, result } : { result },
+				details: resultPreview ? { preview: resultPreview, result: persistedResult } : { result: persistedResult },
 			};
 		},
 		renderCall(args, theme, context: ToolRenderContext<ApplyPatchRenderState, { input: string }>) {
--- a/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
@@ -11,7 +11,7 @@ import {
 import { normalizeApplyPatchArguments } from "./params.ts";
 import { parsePatch } from "./parser.ts";
 import { createPendingPatchUpdate } from "./preview.ts";
-import { getApplyPatchRenderState, renderPatchPreview } from "./preview-format.ts";
+import { getApplyPatchRenderState, renderPatchPreview, truncatePreview } from "./preview-format.ts";
 import { renderStreamingPatchCall } from "./streaming-render.ts";
 import type {
 	ApplyPatchPreview,
@@ -26,7 +26,10 @@ import type {
 function appliedPreview(result: ApplyPatchResult): ApplyPatchPreview | undefined {
 	const files = [...result.details.appliedOperations]
 		.sort((left, right) => left.operationIndex - right.operationIndex)
-		.map((operation) => operation.preview);
+		.map((operation) => ({
+			...operation.preview,
+			diff: truncatePreview(operation.preview.diff),
+		}));
 	if (files.length === 0) return undefined;
 	return {
 		files,
@@ -100,6 +103,20 @@ function renderFailureBox(
 	return component;
 }
 
+function compactAppliedOperations(result: ApplyPatchResult): void {
+	result.details.appliedOperations = result.details.appliedOperations.map(({ operationIndex, preview }) => ({
+		operationIndex,
+		preview: {
+			filePath: preview.filePath,
+			...(preview.movePath === undefined ? {} : { movePath: preview.movePath }),
+			operation: preview.operation,
+			diff: "",
+			added: preview.added,
+			removed: preview.removed,
+		},
+	}));
+}
+
 export function createApplyPatchTool(variant: "freeform" | "json" = "freeform"): ApplyPatchToolDefinition {
 	const tool = defineTool<typeof APPLY_PATCH_PARAMS, ApplyPatchToolDetails | undefined, ApplyPatchRenderState>({
 		name: "apply_patch",
@@ -136,6 +153,7 @@ export function createApplyPatchTool(variant: "freeform" | "json" = "freeform"):
 				onUpdate?.({ content: [{ type: "text", text: progressUpdate.text }], details: progressUpdate.details });
 			});
 			const resultPreview = appliedPreview(result);
+			compactAppliedOperations(result);
 			if (result.failures.length > 0) {
 				return {
 					content: [{ type: "text", text: buildPartialFailureText(result) }],
--- a/packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
+++ b/packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
@@ -19,7 +19,10 @@ describe("apply_patch result retention", () => {
 		const tempDir = await mkdtemp(path.join(tmpdir(), "senpi-apply-patch-retention-"));
 		tempDirs.push(tempDir);
 		const targetPath = path.join(tempDir, "sample.txt");
-		const before = Array.from({ length: 24 }, (_, index) => `line ${index + 1} before`);
+		const before = Array.from(
+			{ length: 3000 },
+			(_, index) => `line ${index + 1} before with padding that makes unbounded retained patches observable`,
+		);
 		const after = before.map((line) => line.replace("before", "after"));
 		await writeFile(targetPath, `${before.join("\n")}\n`, "utf-8");
 
@@ -48,10 +51,9 @@ ${after.map((line) => `+${line}`).join("\n")}
 		expect(visibleFile.diff.length).toBeLessThanOrEqual(PATCH_PREVIEW_MAX_CHARS);
 		expect(visibleFile.diff).toContain("line 1 before");
 		expect(visibleFile.diff).toContain("…");
-		expect(visibleFile.patch).toContain("--- sample.txt");
-		expect(visibleFile.patch).toContain("+++ sample.txt");
-		expect(visibleFile.patch).toContain("-line 24 before");
-		expect(visibleFile.patch).toContain("+line 24 after");
+		const persistedBytes = Buffer.byteLength(JSON.stringify(result.details), "utf8");
+		expect(persistedBytes).toBeLessThan(12_000);
+		expect(visibleFile).not.toHaveProperty("patch");
 
 		expect(patchResult.details.appliedOperations).toEqual([
 			{
@@ -60,8 +62,8 @@ ${after.map((line) => `+${line}`).join("\n")}
 					filePath: "sample.txt",
 					operation: "update",
 					diff: "",
-					added: 24,
-					removed: 24,
+					added: 3000,
+					removed: 3000,
 				},
 			},
 		]);
--- a/packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
+++ b/packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
@@ -1,0 +1,69 @@
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, describe, expect, it } from "vitest";
+import {
+	createApplyPatchTool,
+	PATCH_PREVIEW_MAX_CHARS,
+	PATCH_PREVIEW_MAX_LINES,
+} from "../../../src/core/extensions/builtin/gpt-apply-patch/index.ts";
+
+const tempDirs: string[] = [];
+
+afterEach(async () => {
+	await Promise.all(tempDirs.splice(0).map((tempDir) => rm(tempDir, { recursive: true, force: true })));
+});
+
+describe("apply_patch result retention", () => {
+	it("persists a bounded visible preview and metadata-only applied operations", async () => {
+		const tempDir = await mkdtemp(path.join(tmpdir(), "senpi-apply-patch-retention-"));
+		tempDirs.push(tempDir);
+		const targetPath = path.join(tempDir, "sample.txt");
+		const before = Array.from({ length: 24 }, (_, index) => `line ${index + 1} before`);
+		const after = before.map((line) => line.replace("before", "after"));
+		await writeFile(targetPath, `${before.join("\n")}\n`, "utf-8");
+
+		const patch = `*** Begin Patch
+*** Update File: sample.txt
+@@
+${before.map((line) => `-${line}`).join("\n")}
+${after.map((line) => `+${line}`).join("\n")}
+*** End Patch`;
+		const tool = createApplyPatchTool();
+		const result = await tool.execute("call-retention", { input: patch }, undefined, undefined, {
+			cwd: tempDir,
+		} as Parameters<typeof tool.execute>[4]);
+
+		expect(await readFile(targetPath, "utf-8")).toBe(`${after.join("\n")}\n`);
+		const preview = result.details?.preview;
+		const patchResult = result.details?.result;
+		expect(preview).toBeDefined();
+		expect(patchResult).toBeDefined();
+		if (!preview || !patchResult) throw new Error("apply_patch result details were missing");
+
+		const visibleFile = preview.files[0];
+		expect(visibleFile).toBeDefined();
+		if (!visibleFile) throw new Error("apply_patch visible preview was missing");
+		expect(visibleFile.diff.split("\n").length).toBeLessThanOrEqual(PATCH_PREVIEW_MAX_LINES);
+		expect(visibleFile.diff.length).toBeLessThanOrEqual(PATCH_PREVIEW_MAX_CHARS);
+		expect(visibleFile.diff).toContain("line 1 before");
+		expect(visibleFile.diff).toContain("…");
+		expect(visibleFile.patch).toContain("--- sample.txt");
+		expect(visibleFile.patch).toContain("+++ sample.txt");
+		expect(visibleFile.patch).toContain("-line 24 before");
+		expect(visibleFile.patch).toContain("+line 24 after");
+
+		expect(patchResult.details.appliedOperations).toEqual([
+			{
+				operationIndex: 0,
+				preview: {
+					filePath: "sample.txt",
+					operation: "update",
+					diff: "",
+					added: 24,
+					removed: 24,
+				},
+			},
+		]);
+	});
+});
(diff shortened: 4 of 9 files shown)


packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
```diff
@@ -26,7 +26,10 @@ import type {
 function appliedPreview(result: ApplyPatchResult): ApplyPatchPreview | undefined {
 	const files = [...result.details.appliedOperations]
 		.sort((left, right) => left.operationIndex - right.operationIndex)
-		.map((operation) => operation.preview);
+		.map((operation) => ({
+			...operation.preview,
+			diff: truncatePreview(operation.preview.diff),
```
code-yeongyu: [sisyphus-bot] **F1 (blocker) / F6:** this truncates `diff` but the spread on line 30 carries `preview.patch` through **untouched**. Measured on a 247 KB source file: `diff` drops 545,786 → 1,001 bytes while `patch` stays at **505,837 bytes** — 99.7% of the entire persisted payload after this PR. For add/delete ops the unified patch is the whole file (130,944 B for a 130 KB delete), so the compaction is ~99.5% ineffective there.

Also note `truncatePreview` runs again at render time (`preview-format.ts:132/146/215`). I verified it is idempotent, so that is redundant work, not a correctness bug — but it means the persistence bound and the display bound are the same knob, which is probably not what you want: display wants 16 lines, persistence wants a byte budget.

packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
```diff
@@ -100,6 +103,20 @@ function renderFailureBox(
 	return component;
 }
 
+function compactAppliedOperations(result: ApplyPatchResult): void {
+	result.details.appliedOperations = result.details.appliedOperations.map(({ operationIndex, preview }) => ({
```
code-yeongyu: [sisyphus-bot] **F5 (minor):** this rebuilds `ApplyPatchPreviewFile` field-by-field, so any field added to that type later is **silently dropped** from every persisted result with no type error. Prefer an explicit omit so new fields survive by default:

```ts
const { diff: _diff, patch: _patch, ...rest } = preview;
return { operationIndex, preview: { ...rest, diff: "" } };
```

That also makes the intent ("drop the heavy fields, keep everything else") readable at a glance instead of requiring the reader to diff two field lists.

packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
```diff
@@ -136,6 +153,7 @@ export function createApplyPatchTool(variant: "freeform" | "json" = "freeform"):
 				onUpdate?.({ content: [{ type: "text", text: progressUpdate.text }], details: progressUpdate.details });
 			});
 			const resultPreview = appliedPreview(result);
+			compactAppliedOperations(result);
```
code-yeongyu: [sisyphus-bot] **F3 (major):** these two lines have a load-bearing order that nothing pins. `compactAppliedOperations` mutates `result.details.appliedOperations` in place, and `appliedPreview(result)` on line 155 reads `operation.preview.diff` from that exact array. Swap them and every visible diff in the TUI silently becomes `""` — type-clean, lint-clean.

I verified the current order is correct and that no aliasing exists (`preview.files[0] !== appliedOperations[0].preview`, confirmed by probe). But please make the constraint explicit: either a comment stating that compaction must follow preview construction, or restructure so it cannot be got wrong — have `compactAppliedOperations` return a new result instead of mutating a value `applyPatchDetailed()` already handed back (`apply.ts:199` computes `recoveryInstructions` off the same object).

packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
```diff
@@ -0,0 +1,69 @@
...
+		const preview = result.details?.preview;
+		const patchResult = result.details?.result;
+		expect(preview).toBeDefined();
+		expect(patchResult).toBeDefined();
+		if (!preview || !patchResult) throw new Error("apply_patch result details were missing");
+
+		const visibleFile = preview.files[0];
+		expect(visibleFile).toBeDefined();
+		if (!visibleFile) throw new Error("apply_patch visible preview was missing");
+		expect(visibleFile.diff.split("\n").length).toBeLessThanOrEqual(PATCH_PREVIEW_MAX_LINES);
```
code-yeongyu, line 51: [sisyphus-bot] **F2 (blocker):** I mutation-tested this and it is genuinely not tautological — reverting `truncatePreview` fails here with `expected 48 to be less than or equal to 16`, and reverting `compactAppliedOperations` fails at line 56. Good.

The problem is what it does **not** assert. Issue #648 is a **payload-size** regression (187 MB session, 126.57 MB in one `apply_patch` detail). This test asserts nothing about the size of the persisted result, so it passes unchanged on a build that still writes ~63 MB per patch. Please add a bound on the actual persisted bytes, on a fixture large enough that an unbounded field cannot hide:

```ts
const persistedBytes = Buffer.byteLength(JSON.stringify(result.details), "utf8");
expect(persistedBytes).toBeLessThan(/* tied to visible output, not to file size */);
```

With a 24-line fixture, a field that scales with file size is invisible.

packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
```diff
@@ -0,0 +1,69 @@
...
+		expect(visibleFile.diff.length).toBeLessThanOrEqual(PATCH_PREVIEW_MAX_CHARS);
+		expect(visibleFile.diff).toContain("line 1 before");
+		expect(visibleFile.diff).toContain("…");
+		expect(visibleFile.patch).toContain("--- sample.txt");
+		expect(visibleFile.patch).toContain("+++ sample.txt");
```
code-yeongyu: [sisyphus-bot] **F2 (blocker), continued:** these three assertions pin `visibleFile.patch` as *required to contain the full old and new content* (`-line 24 before` / `+line 24 after` on a 24-line file). That is the exact field carrying 99.7% of the persisted bytes after this PR.

So the regression test for a bloat issue currently **locks the bloat in place**. If retaining the unified patch is a deliberate trade for app-server projection fidelity (`projection-file-changes.ts:43`), then the test should say so — assert that `patch` is retained *and* bounded, and state the byte budget. Right now it asserts unbounded retention on a fixture too small to reveal the cost.

packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
```diff
@@ -1,5 +1,25 @@
 # changes
 
+## Compact completed result retention (2026-08-02)
+
+### What changed
+
+- `tool.ts`: completed previews persist `truncatePreview()` output while preserving each top-level unified `patch`.
+- Nested `result.details.appliedOperations` entries retain only operation indexes, paths, move destinations, operation types, and line counts.
```
code-yeongyu: [sisyphus-bot] **F1 (blocker), documentation:** this entry says nested entries "retain only operation indexes, paths, move destinations, operation types, and line counts" — accurate — but the entry never discloses that the **top-level** `preview.files[].patch` still carries the full old+new file content, which is 99.7% of what actually gets persisted.

An `apply_patch` on a 247 KB file still persists 507 KB of details after this change (2.05× the source file, 33,822× the visible output). Since `changes.md` is the fork's merge-audit surface, please state the retained term and its size characteristic explicitly, otherwise the next upstream sync reads this as "apply_patch results are now lightweight metadata", which is not true.

code-yeongyu (changes requested):
[sisyphus-bot] **Round 1 — brutal review: the fix is real but treats the smaller half of the problem.**

I reviewed this in a dedicated worktree at PR head `3ee70786a`, instrumented the real `apply_patch` tool, and measured the persisted `details` payload on both `main` and this branch. Summary: the change is correct and breaks no consumer, but it does **not** deliver what issue #648 asks for, and its regression test does not measure the quantity the issue is about.

---

### Measured evidence (not speculation)

Probe: a single `*** Update File:` op against a **247 KB / 4,000-line** file, `JSON.stringify` byte sizes of the returned `details`.

| metric | `main` | this PR | delta |
|---|---:|---:|---:|
| **total `details` bytes** | 2,103,751 | **507,336** | −75.9% |
| `preview.files[].diff` | 545,786 | 1,001 | −99.8% |
| **`preview.files[].patch`** | 505,837 | **505,837** | **0%** |
| `result.details.appliedOperations` | 1,051,744 | 114 | −99.99% |
| visible tool text | 15 | 15 | — |

After this PR the persisted payload is still **2.05× the source file** and **33,822× the visible output**. `preview.files[].patch` is now **99.7% of everything that gets persisted**, and it is completely unbounded.

Add/delete operations are worse, because their unified patch is the *entire* file:

| op | source | total `details` | `patch` | `diff` |
|---|---:|---:|---:|---:|
| Delete 130 KB file | 130 KB | 132,139 | **130,944** | 701 |
| Add 130 KB file | 130 KB | 132,136 | **130,945** | 701 |

For add/delete the fix is ~99.5% ineffective.

---

### F1 — BLOCKER: the dominant unbounded field is untouched

Issue #648 states the requirement explicitly:

> Final persisted `apply_patch` results should contain only lightweight metadata required after execution: affected paths, summaries, operation indexes, fuzz, and failure/recovery information. **Full old/new file contents should not remain duplicated—or ideally should not be persisted at all after rendering.**

`preview.files[].patch` is a full unified patch built by `createUnifiedPatch(...)` in `preview.ts:66/79/87` — it contains **both** the old and the new file content. This PR removes one of the two copies of the *diff* and truncates the other, then leaves the unified patch fully intact in the persisted result.

The issue's headline case (126.57 MB for one patch: 63.28 MB `preview.files` + 63.28 MB `appliedOperations`) becomes roughly 63 MB under this change. That is an improvement, not a fix, and the PR description does not disclose the remaining term.

The PR body claims it "preserve[s] top-level unified patches for app-server file-change projection" — that justification needs to be weighed against the cost. The only consumer is `changesFromPreview()` in `src/modes/app-server/threads/projection-file-changes.ts:36-46`, which reads `file.patch` to emit `FileUpdateChange.diff` for the app-server wire protocol. Please resolve one of:

- bound `patch` for persistence and let the app-server projection consume it from the live (pre-persist) result, or
- keep `patch` only while `status === "inProgress"`/at projection time and strip it from the completed persisted details, or
- explicitly document in the PR body + `changes.md` that ~2× file size per patch is intentionally retained for app-server fidelity, with a number attached.

Silently retaining 99.7% of the payload while the PR title says "compact persisted result previews" is the part I object to most.

---

### F2 — BLOCKER: the regression test does not assert what #648 measures

To be fair to the author, I mutation-tested the new test and it is **not** tautological — both production changes are genuinely covered:

- removing the `compactAppliedOperations(result)` call → `test/suite/regressions/648-apply-patch-result-retention.test.ts:56` fails ✅
- removing `truncatePreview(...)` at `tool.ts:31` → line 47 fails with `expected 48 to be less than or equal to 16` ✅

But issue #648 is a **size** regression, and the test asserts no size bound on the persisted result. It asserts `visibleFile.diff` is bounded, then immediately asserts `visibleFile.patch` **contains** `-line 24 before` / `+line 24 after` — i.e. it pins the unbounded field as *required to stay unbounded*, on a 24-line fixture where the distinction is invisible.

A regression test for #648 should fail when the persisted payload grows superlinearly with file size. Concretely, please add an assertion on the order of:

```ts
const persistedBytes = Buffer.byteLength(JSON.stringify(result.details), "utf8");
expect(persistedBytes).toBeLessThan(<bound tied to the visible output, not to the file>);
```

exercised against a file large enough (a few thousand lines) that an unbounded field cannot hide. As written, this test would pass unchanged on a session that still writes 63 MB per patch.

---

### F3 — MAJOR: silent load-bearing ordering, unpinned

`tool.ts:155-156`:

```ts
const resultPreview = appliedPreview(result);
compactAppliedOperations(result);
```

`compactAppliedOperations` mutates the caller-owned `ApplyPatchResult` in place (`result.details.appliedOperations = ...`). `appliedPreview()` reads `operation.preview.diff` from exactly that array. I verified the two lines are correctly ordered today and that no aliasing exists (`appliedPreview`'s spread creates fresh objects — probe confirms `preview.files[0] !== appliedOperations[0].preview`).

But swapping these two lines is a **silent, type-clean, test-visible-only-by-luck** catastrophe: every visible diff in the TUI becomes `""`. There is no comment stating the ordering constraint and no test pinning it. Either add an explicit comment at the call site, or make it structurally impossible — e.g. have `compactAppliedOperations` return a new result rather than mutating, or fold the compaction into `appliedPreview`'s return so one function owns both halves.

In-place mutation of a value that `applyPatchDetailed()` constructed and returned is also a smell on its own: `apply.ts:199` computes `recoveryInstructions` from that same object, and any future consumer added between those lines inherits an invisible ordering dependency.

---

### F4 — MINOR: `applyPatch()` is inconsistent with `applyPatchDetailed()`

`apply.ts:203-231` builds its own `appliedOperations` with full previews and never compacts them. That array reaches `ApplyPatchError.result` on the failure path. It is a smaller payload in practice, but the invariant "persisted applied-operations carry metadata only" now holds on one code path and not the other. Worth either compacting there too or noting why the error path is exempt.

---

### F5 — MINOR: hand-rebuilt preview object drops future fields silently

`compactAppliedOperations` reconstructs `ApplyPatchPreviewFile` field by field. TypeScript will **not** error when a new optional field is added to `ApplyPatchPreviewFile` and forgotten here — it will simply be dropped from every persisted result. Prefer destructuring-omit (`const { diff, patch, ...rest } = preview`) so new fields are retained by default and the omission is explicit.

---

### F6 — NOTE: double truncation (verified harmless)

`tool.ts:31` truncates `operation.preview.diff`, then `preview-format.ts:132/146/215` calls `truncatePreview(file.diff)` again at render time. I verified `truncatePreview` is idempotent (16 lines → 16 lines, byte-identical), so this is redundant work rather than a mangled-ellipsis bug. Flagging only so the redundancy is a conscious choice.

---

### What is correct here (credit where due)

- No consumer is broken. I enumerated every reader of `appliedOperations` and `preview`: `appliedPreview` (`tool.ts:26`), `renderFailureBox` (`tool.ts:86`), `renderResult` (`tool.ts:180`), `changesFromPreview` (`projection-file-changes.ts:36`), `createRecoveryInstructions` (`apply.ts:199`). All safe.
- `movePath` correctly survives compaction via the conditional spread — verified with a real move patch (`nestedMovePath: ["b.txt"]`).
- The three `changes.md` entries follow the repo convention and name the merge-conflict zones.
- Existing suite is green at this head: 8 files / 61 tests passed.

---

### Verdict

**Request changes.** F1 and F2 are blocking: the PR claims to compact persisted results but leaves the field that is 99.7% of the persisted bytes unbounded, and the regression test cannot detect that. F3 should be resolved before merge because the ordering hazard is invisible and severe.

*Evidence: measurements produced by instrumenting `createApplyPatchTool().execute()` directly in a review worktree at `3ee70786a`, compared against `main`. Mutation testing performed by reverting each production hunk individually and re-running the regression test.*

packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
```diff
@@ -23,10 +23,27 @@ import type {
 	FreeformToolFormat,
 } from "./types.ts";
 
+export const APPLY_PATCH_RESULT_PATCH_MAX_BYTES = 16 * 1024;
```
code-yeongyu, line 26: [sisyphus-bot] **R2-F3 (minor):** `APPLY_PATCH_RESULT_PATCH_MAX_BYTES` is exported here but not re-exported from this extension's `index.ts`, unlike `PATCH_PREVIEW_MAX_CHARS` and `PATCH_PREVIEW_MAX_LINES` which are. Consequence: the regression test cannot import it and hardcodes `12_000` instead — a bound with no expressed relationship to the 16 KiB budget it is actually testing.

Please add it to the `export { ... } from "./tool.ts"` group in `index.ts` and derive the test bound from it, so changing the budget updates the test's intent rather than breaking it mysteriously.

packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/tool.ts
```diff
@@ -23,10 +23,27 @@ import type {
...
+
+function retainedPatch(patch: string | undefined): string | undefined {
+	if (patch === undefined || Buffer.byteLength(patch, "utf8") > APPLY_PATCH_RESULT_PATCH_MAX_BYTES) {
```
code-yeongyu, line 29: [sisyphus-bot] Confirming the allocation fix landed: `Buffer.byteLength(patch, "utf8")` replaces the `new TextEncoder().encode(patch).byteLength` I flagged pre-push. That mattered specifically here — the old form allocated a full UTF-8 copy of the entire patch just to measure it, i.e. a 63 MB transient allocation on issue #648's own payload, on the code path whose purpose is relieving memory pressure, and it happened even when the patch was then discarded as oversized. Good.

Omitting rather than truncating is also the right call: I verified a truncated unified patch would be caught downstream — forcing omission across the board fails 5 app-server tests including a real `git apply --check` → "No valid patches in input".

packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/apply.ts
```diff
@@ -44,6 +44,19 @@ function extractErrorCode(error: unknown): string | undefined {
 	return undefined;
 }
 
+export function compactApplyPatchResult(result: ApplyPatchResult): ApplyPatchResult {
```
code-yeongyu, line 47: [sisyphus-bot] **F3 resolved structurally — this is the right fix.** `compactApplyPatchResult()` is now a pure function returning a new result instead of mutating `result.details.appliedOperations` in place. The load-bearing ordering hazard I flagged in round 1 (swap two lines at the call site and every visible TUI diff silently becomes `""`) is now *impossible* rather than merely commented, because `appliedPreview(result)` and this function no longer share mutable state.

The destructuring-omit (`const { diff: _diff, patch: _patch, ...metadata }`) also resolves F5: a future optional field added to `ApplyPatchPreviewFile` now survives compaction by default instead of being silently dropped.

packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		const visibleFile = preview.files[0];
+		expect(visibleFile).toBeDefined();
+		if (!visibleFile) throw new Error("apply_patch visible preview was missing");
+		expect(visibleFile.diff.split("\n").length).toBeLessThanOrEqual(PATCH_PREVIEW_MAX_LINES);
+		expect(visibleFile.diff.length).toBeLessThanOrEqual(PATCH_PREVIEW_MAX_CHARS);
+		expect(visibleFile.diff).toContain("line 1 before");
+		expect(visibleFile.diff).toContain("…");
+		const persistedBytes = Buffer.byteLength(JSON.stringify(result.details), "utf8");
+		expect(persistedBytes).toBeLessThan(12_000);
```
code-yeongyu: [sisyphus-bot] **F2 resolved — verified, not assumed.** I re-ran the mutation myself by disabling the budget so patches are always retained:

```
AssertionError: expected 482612 to be less than 12000
 ❯ 648-apply-patch-result-retention.test.ts:55:26
```

Then restored byte-identical and confirmed green. This assertion is real coverage for the size regression #648 is actually about.

One follow-up (**R2-F3**): `12_000` is a magic number with no expressed link to the 16 KiB budget. Derive it from `APPLY_PATCH_RESULT_PATCH_MAX_BYTES` once that constant is re-exported from `index.ts`.

packages/coding-agent/test/suite/regressions/648-apply-patch-result-retention.test.ts
```diff
@@ -0,0 +1,71 @@
...
+		expect(visibleFile).not.toHaveProperty("patch");
```
code-yeongyu, line 57: [sisyphus-bot] **R2-F2:** this pins that the *preview field* is omitted, but nothing pins the **downstream consequence** — that `fileChangeProjection()` now returns `changes: []` for an oversized patch. I measured that behavior at this head (`projChanges: 0`, `projDiffBytes: 0` for 3,000–4,000 line UPDATE/DELETE/ADD fixtures).

Every existing `app-server-task24-*` test uses small patches, so the within-budget path is well covered (forcing omission everywhere fails 5 of them) while the omitted path is covered nowhere. A future regression in `retainedPatch()` could start emitting partial/malformed diffs to app-server clients with the suite fully green.

Please add a case that pushes a >16 KiB patch through `fileChangeProjection` and asserts the projection is empty rather than malformed — that is what makes this trade-off deliberate instead of incidental.

packages/coding-agent/src/core/extensions/builtin/gpt-apply-patch/changes.md
```diff
@@ -1,5 +1,26 @@
 # changes
 
+## Compact completed result retention (2026-08-02)
+
+### What changed
+
+- `tool.ts`: completed previews persist `truncatePreview()` output and retain a complete unified `patch` only when it is at most 16 KiB per file; oversized patches are omitted instead of retaining source-sized bodies or producing malformed truncations.
```
code-yeongyu, line 7: [sisyphus-bot] This entry is accurate and states the 16 KiB budget plus the app-server consequence explicitly — exactly what I asked for in round 1. Thank you.

The problem is that the **PR description still says the opposite** (see R2-F1 in my summary). Its Compatibility section claims "The top-level `preview.files[].patch` field remains intact for app-server consumers", which is false for any patch over 16 KiB. `changes.md` is the merge-audit surface, but the PR body is what the merger reads — please bring it in line with this entry.

code-yeongyu (changes requested):
[sisyphus-bot] **Round 2 — the code is now right. The PR description is not, and it misstates the compatibility impact.**

Re-reviewed at `1711ce352` in a clean worktree with a real `npm install`. I re-ran every measurement myself rather than accepting the reported numbers, and I independently mutation-tested the new assertion.

---

### Round 1 blockers: all genuinely resolved

| finding | status | how I verified |
|---|---|---|
| **F1** unbounded `preview.files[].patch` | **fixed** | measured below |
| **F2** test asserted no size bound | **fixed** | mutation-proven, output below |
| **F3** in-place mutation + load-bearing ordering | **fixed structurally** | `compactApplyPatchResult()` is now pure; the ordering hazard cannot be reintroduced by reordering |
| **F4** `applyPatch()` error path uncompacted | **fixed** | `ApplyPatchError(message, compactApplyPatchResult(result))` |
| **F5** field-by-field rebuild dropped future fields | **fixed** | destructuring-omit `const { diff: _diff, patch: _patch, ...metadata }` |
| **steer** 63 MB transient alloc in `retainedPatch` | **fixed** | `Buffer.byteLength(patch, "utf8")` at `tool.ts:29` |

**Measured payload, my harness, same fixtures across all three heads:**

| scenario | `main` | round 1 | **round 2** | ratio to source |
|---|---:|---:|---:|---:|
| UPDATE 4000-line | 2,103,751 | 391,111 | **1,265** | 8.52 → **0.007** |
| DELETE 3000-line | — | 147,214 | **1,261** | 1.05 → **0.009** |
| ADD 3000-line | — | 147,211 | **1,257** | 1.05 → **0.009** |
| MOVE small | — | 1,359 | 1,359 | patch retained (554 B) |
| MULTI 2×3000 | — | 2,894 | 2,894 | patch retained (642 B) |

Source-size scaling is **gone**. Issue #648's 126.57 MB single-payload case now lands in the low kilobytes.

**Mutation proof I ran myself** (disabled the budget so patches are always retained):

```
AssertionError: expected 482612 to be less than 12000
 ❯ test/suite/regressions/648-apply-patch-result-retention.test.ts:55:26
Tests  1 failed (1)
```

Restored byte-identical, back to green. The assertion is real coverage, not decoration.

**Suites, all run by me at this head:**
- `gpt-apply-patch` + regression — 8 files / **61 passed**
- `app-server-task24-{apply-patch-identity,file-change-results,diff,review-regressions}` — 4 files / **14 passed**
- `npm run check` — **exit 0** (I ran this; the author skipped it at my instruction because I had broken their dependency layout with symlinks — that was my fault, not a defect in this change)

Design call on F1 — accepting **bounded retention with omission**. I verified the author's justification independently: `createToolResultMessage()` builds the tool-result message carrying `details`, `emitToolResultMessage()` hands that same object to the `EventProjector`, and the same structure is persisted. Projection and persistence genuinely share one object, so there is no seam to strip after projection without cross-package surgery in `packages/agent`. My round-1 suggestion of "consume at projection time, strip at persistence" was more expensive than I implied, and I withdraw it. Omitting rather than truncating is the right call — a truncated unified patch would be a malformed diff, and I confirmed the existing suite catches exactly that (forcing omission everywhere fails 5 tests, including a real `git apply --check` → `"No valid patches in input"`).

---

### R2-F1 — BLOCKER: the PR description now contradicts the code

The body still tells a reviewer the opposite of what ships. Three specific lines:

1. Summary: *"preserve top-level unified patches for app-server file-change projection"* — false for any patch over 16 KiB, which is exactly the large-patch case this PR exists to address.
2. Summary: *"add an issue-numbered regression covering bounded diffs, **preserved patch bodies**, and metadata-only nesting"* — the regression now asserts the opposite: `expect(visibleFile).not.toHaveProperty("patch")`.
3. **Compatibility section: *"The top-level `preview.files[].patch` field remains intact for app-server consumers."*** — this is the damaging one. It states there is no consumer-visible change, when in fact **app-server file-change projection now emits zero changes for any patch over 16 KiB**. I measured `projChanges: 0` / `projDiffBytes: 0` for the UPDATE/DELETE/ADD scenarios above. An IDE or app-server client that renders `fileChange` diffs will silently show nothing for large patches.

That is a legitimate, defensible trade — it is correctly documented in all three `changes.md` files with the budget stated numerically, which is what I asked for. But the PR body is what a merger reads, and right now it asserts backward compatibility that the diff does not deliver. Please rewrite the Summary and Compatibility sections to state the 16 KiB budget and the projection consequence plainly.

Also note the Verification block lists results from the previous head (including `npm run build` and QA lane counts) that were not re-run for `1711ce352`. Either re-run them or mark them as pertaining to the earlier commit.

Credit where due: cubic's auto-generated summary *did* update correctly and describes the 16 KiB behavior accurately. It is only the human-authored sections that are stale.

---

### R2-F2 — the documented wire-contract change has no test

Omitting oversized patches changes what the app-server emits, deliberately. Nothing pins it. The regression asserts `not.toHaveProperty("patch")` on the preview, but no test asserts the actual downstream consequence — that `fileChangeProjection(...)` returns `changes: []` for an oversized patch.

Every existing `app-server-task24-*` test uses small patches, so they all exercise the within-budget path only. The result is that the *retained* path is well covered (I proved this: forcing omission everywhere fails 5 of them) while the *omitted* path is covered nowhere. A future change to `retainedPatch()` could silently start emitting truncated or partial diffs to app-server clients and the whole suite would stay green.

Please add a test that runs a >16 KiB patch through `fileChangeProjection` and asserts the projection is empty rather than malformed. That is the assertion that makes the trade-off deliberate rather than incidental.

---

### R2-F3 — MINOR: the budget constant is unreachable from tests

`APPLY_PATCH_RESULT_PATCH_MAX_BYTES` is exported from `tool.ts:26` but not re-exported from the extension's `index.ts`, unlike `PATCH_PREVIEW_MAX_CHARS` / `PATCH_PREVIEW_MAX_LINES` which are. So the regression test hardcodes `12_000`, a number with no expressed relationship to the 16 KiB budget it is really testing. Raise the budget to 64 KiB later and that bound fails for reasons no one will understand from the assertion.

Re-export the constant alongside the other preview limits and derive the bound from it.

---

### Verdict

Requesting changes on **R2-F1** only. The implementation is correct, measurably fixes #648, is properly documented in `changes.md`, and is covered by a mutation-proven assertion — genuinely good work on the round-1 feedback. But shipping a Compatibility section that promises `preview.files[].patch` "remains intact for app-server consumers" while large patches now project nothing would mislead whoever merges this. Fix the body, add the omitted-path test, and this is ready.

*All figures above produced by re-running the measurements at `1711ce352` myself, not quoted from the PR.*

code-yeongyu (approved):
[sisyphus-bot] **Round 3 — approved. Zero blocking findings.**

Re-reviewed at `0f43a753e` in a clean worktree with a real `npm install`. Every figure below I produced myself at this head; none are quoted from the PR.

---

### All findings across three rounds are resolved

| round | finding | status |
|---|---|---|
| R1-F1 | `preview.files[].patch` unbounded, 99.7% of persisted payload | **fixed** — 16 KiB budget, omission not truncation |
| R1-F2 | regression test asserted no size bound | **fixed** — mutation-proven size assertion |
| R1-F3 | in-place mutation + load-bearing ordering hazard | **fixed structurally** — `compactApplyPatchResult()` is pure |
| R1-F4 | `applyPatch()` error path uncompacted | **fixed** |
| R1-F5 | field-by-field rebuild silently dropped future fields | **fixed** — destructuring-omit |
| steer | 63 MB transient alloc measuring patch size | **fixed** — `Buffer.byteLength` |
| R2-F1 | PR body contradicted shipped behavior | **fixed** — verified by string probe |
| R2-F2 | omitted path had no test | **fixed** — mutation-proven |
| R2-F3 | budget constant unreachable from tests | **fixed** — exported, bound derived |

---

### Final measurements at `0f43a753e`

| scenario | `main` | R1 | R2/R3 | ratio to source |
|---|---:|---:|---:|---:|
| UPDATE 4000-line | 2,103,751 | 391,111 | **1,265** | 8.52 → **0.007** |
| DELETE 3000-line | — | 147,214 | **1,261** | 1.05 → **0.009** |
| ADD 3000-line | — | 147,211 | **1,257** | 1.05 → **0.009** |
| MOVE small | — | 1,359 | 1,359 | patch **retained** (554 B), projection intact |
| MULTI 2×3000 | — | 2,894 | 2,894 | patch **retained** (642 B), projection intact |

Source-size scaling is eliminated. Issue #648's worst single payload (126.57 MB) now lands in the low kilobytes, while small patches keep their complete unified diff so the app-server contract is preserved where it can be.

### Verification I ran at this head

- `gpt-apply-patch` + regression — 8 files / **61 passed**
- the four `app-server-task24-*` suites — 4 files / **15 passed** (was 14; the new contract test adds one)
- `npm run check` — **exit 0**
- Both new assertions independently mutation-proven by disabling the 16 KiB guard:
  - size bound → `AssertionError: expected 482612 to be less than 8192`
  - omitted-path contract → fails at `app-server-task24-file-change-results.test.ts:155`
  - restored byte-identical, both green again

I also verified the new test's assertions are not vacuous: the fixture self-guards with `expect(Buffer.byteLength(input, "utf8")).toBeGreaterThan(APPLY_PATCH_RESULT_PATCH_MAX_BYTES)` so it cannot silently drift under budget, and I empirically confirmed in this vitest (4.1.10) that `toMatchObject({ changes: [] })` genuinely **throws** against a non-empty array rather than subset-matching anything.

### Completeness check

I checked whether unbounded patch data still reaches persistence by any other route. `createPendingPatchUpdate()` does build a full unbounded `patch`/`diff` for the in-flight preview — but it is emitted as `tool_execution_update`, whose only consumers are `projection.ts:72` (`extractToolText`, text only) and the interactive TUI reveal. The sole path into a persisted `role: "toolResult"` message is `createToolResultMessage(finalized)` at `agent-loop.ts:1145`, which carries the final bounded result. **No unbounded patch reaches the session.** The fix is complete for what #648 measures.

One non-blocking follow-up, out of scope here: peak *transient* memory still scales with file size because the pending preview materializes full old/new content. That is pre-existing, neither introduced nor worsened by this PR, and #648's measurements (session JSONL + `ResidentStringStore`) are fully addressed. Worth a separate issue, not a change request.

### Hygiene

3 commits, conventional format, exactly 9 files touched, no `node_modules`, no probe or scratch files committed. All three `changes.md` files accurately describe the shipped behavior including the app-server consequence, and the `index.ts` export is recorded.

---

### Verdict

**Approved.** The implementation measurably fixes #648, the trade-off it makes is deliberate, documented in both `changes.md` and the PR body, and pinned by tests that genuinely fail when the behavior regresses.

Two things I want to credit explicitly. First, choosing **omission over truncation** was the correct call and not the obvious one — a truncated unified patch would be a malformed diff, and I confirmed the existing suite catches exactly that failure mode (forcing omission everywhere fails 5 tests including a real `git apply --check` → `"No valid patches in input"`). Second, the response to R1-F3 fixed the ordering hazard *structurally* by making the function pure, rather than papering over it with a comment.

I also want to correct myself on the record: in round 1 I pushed for "consume the patch at projection time, strip it at persistence." I verified afterwards that `createToolResultMessage` → `emitToolResultMessage` → `EventProjector` and session persistence share one object, so that seam does not exist without cross-package surgery in `packages/agent`. My round-1 suggestion was more expensive than I implied, and the bounded-retention approach taken here is the right one.

Nice work.

PR #711, opened 2026-08-04
fix(ai): keep root object type on OpenAI-completions tool parameters

### LINKED ISSUE #710: OpenAI-completions tools lose root type object when parameters use oneOf/anyOf (Console Go 400)
### Summary

Function tools whose `parameters` schema uses a combiner (`anyOf` / `oneOf` / `allOf`) are sent to OpenAI-compatible backends without the root `type: "object"` marker. `normalizeToolParametersForOpenAICompat` moves the root `type` into combiner branches and then deletes it (`moveTypeIntoCombinerBranches` in `packages/ai/src/utils/tool-schema-compat.ts`).

Gateways backed by Gemini-style validation (e.g. OpenCode Go / Zen proxying to Console) reject the first request:

```text
Error: 400: {"param":null,"type":"invalid_request_error","code":"invalid_request_error","message":"Error from provider (Console Go): Upstream request failed: [invalid_request_error] Invalid schema for function 'mcp__ast_grep_scan': schema must be a JSON Schema of 'type: \"object\"', got 'type: null'."}
```

### Reproduction

1. Use the ast-grep MCP `scan` tool, whose parameters schema is `type: "object"` plus a `oneOf` (ruleFile XOR inlineRules) contract.
2. Start a chat with any `opencode-go` model (for example `deepseek-v4-flash`).
3. The first request fails with the 400 above.

Verified locally: `normalizeToolParametersForOpenAICompat(scanSchema)` drops the top-level `type`, while `search` / `rewrite` (no combiner) keep it.

### Root cause

`moveTypeIntoCombinerBranches` deletes the parent `type` for every combiner schema. That behavior exists for Moonshot-flavored backends (Apitopia → Kimi / Moonshot), but OpenAI Chat Completions requires every function's `parameters` root to be a JSON Schema of `type: "object"`, and the Gemini-backed Console gateways enforce the same.

### Suggested fix

Restore the root `type: "object"` marker at the OpenAI-completions wire boundary (`convertTools` and `normalizeRequestToolSchemas`) instead of changing the Moonshot-flavored normalizer behavior. A PR with tests for both conversion paths is attached.

wingwogus:
Closes #710

## Summary

Function tools whose parameters schema uses a combiner (`anyOf` / `oneOf` / `allOf`) are sent to OpenAI-compatible backends without the root `type: "object"` marker, because `normalizeToolParametersForOpenAICompat` moves the root `type` into combiner branches and deletes it. Gemini-backed gateways (OpenCode Go / Zen proxying to Console) then reject the request with `Invalid schema for function 'mcp__ast_grep_scan': schema must be a JSON Schema of 'type: "object"', got 'type: null'`.

This change restores the root `type: "object"` marker at the OpenAI-completions wire boundary:

- `convertTools` (initial tool conversion)
- `normalizeRequestToolSchemas` (re-normalization after `onPayload` / injected tools)

The Moonshot-flavored normalizer behavior (`toolSchemaFlavor: "moonshot-mfjs"`) is left untouched.

## Test plan

- `packages/ai/test/openai-completions-tool-schema-compat.test.ts`: new wire-boundary tests cover both conversion paths with a scan-style `type: "object"` + `oneOf` schema, asserting the root `type` and `oneOf` survive to the actual HTTP request body.
- `npx vitest run` in `packages/ai`: 9/9 in the compat file, 1773 passed overall (1 unrelated pre-existing import failure in `packages/coding-agent/src/core/tools/diff-render.ts`).
- Repo pre-commit gate (`npm run check`) passed: Biome, pinned deps, ts-imports, shrinkwrap, install lock, full `tsc --noEmit`, browser smoke.

---
## Summary by cubic
Restores the root `type: "object"` on function tool parameters sent to OpenAI Chat Completions, so schemas with `anyOf`/`oneOf`/`allOf` are accepted by Gemini-backed OpenAI-compatible gateways.

- **Bug Fixes**
  - Added `ensureRootObjectType` and applied it in `convertTools` and `normalizeRequestToolSchemas`.
  - Leaves `$ref`-rooted schemas and Moonshot-flavored normalization unchanged.
  - Added wire-boundary tests to confirm the root `type` and combiners persist for both initial tools and `onPayload`-injected tools.

<sup>Written for commit b6a1ace4bb47ff64b982f21392686fedb11f5913. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/711?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -1590,7 +1590,7 @@ function convertTools(
 			function: {
 				name: tool.name,
 				description: tool.description,
-				parameters: normalizedParameters as FunctionParameters,
+				parameters: ensureRootObjectType(normalizedParameters) as FunctionParameters,
 				// Only include strict if provider supports it. Some reject unknown fields.
 				...(compat.supportsStrictMode !== false && { strict: strict ?? false }),
 			},
@@ -1615,10 +1615,30 @@ function normalizeRequestToolSchemas(
 					: normalizeToolParametersForOpenAICompat(tool.function.parameters);
 			return {
 				...tool,
-				function: { ...tool.function, parameters },
+				function: { ...tool.function, parameters: ensureRootObjectType(parameters) },
 			};
 		}),
 	};
+}
+
+/**
+ * OpenAI Chat Completions requires every function's `parameters` to be a JSON
+ * Schema object rooted at `type: "object"`. Combiner normalization may move the
+ * root `type` into `anyOf` / `oneOf` / `allOf` branches (a shape some
+ * OpenAI-compatible gateways require), so restore the root marker at the wire
+ * boundary for every tool. `$ref`-rooted schemas are left untouched.
+ */
+function ensureRootObjectType(parameters: unknown): Record<string, unknown> {
+	if (parameters === null || typeof parameters !== "object" || Array.isArray(parameters)) {
+		return parameters as Record<string, unknown>;
+	}
+
+	const record = parameters as Record<string, unknown>;
+	if (record.type !== undefined || "$ref" in record) {
+		return record;
+	}
+
+	return { ...record, type: "object" };
 }
 
 function parseChunkUsage(
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -1,5 +1,31 @@
 # AI Source Changes
 
+## 2026-08-04 - Preserve root object type in OpenAI-completions tool parameters
+
+### What changed and why
+
+- `api/openai-completions.ts` guarantees every function tool's `parameters`
+  root declares `type: "object"` at the wire boundary (`ensureRootObjectType`),
+  applied in both `convertTools` and `normalizeRequestToolSchemas`.
+- OpenAI-compatible gateways backed by Gemini-style validation (for example
+  OpenCode Go / Zen proxying to Console) reject function schemas whose root
+  `type` was dropped by combiner normalization (`anyOf` / `oneOf` / `allOf`),
+  returning `Invalid schema for function ...: schema must be a JSON Schema of
+  'type: "object"', got 'type: null'`. The normalizer intentionally moves the
+  root `type` into combiner branches for Moonshot-flavored backends, so the
+  root marker is restored at the API boundary instead of changing the
+  normalizer's provider-specific behavior.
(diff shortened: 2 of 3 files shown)


code-yeongyu:
[sisyphus-bot] Thanks so much for this contribution, @wingwogus — the instinct to guard the root object type was exactly right, and this PR correctly identified a real failure mode when you wrote it. 🙏

Closing as **superseded by #718**, which landed a fix at the normalizer layer rather than the wire boundary. Here's the concrete state on current `main`:

- `packages/ai/src/utils/tool-schema-compat.ts:172-177` — `normalizeNode` now takes an `isRoot` flag and only calls `moveTypeIntoCombinerBranches` when `hasCombiner && !isRoot`, with the comment "The root of a tool's parameters must keep `type: \"object\"".
- `packages/ai/src/utils/tool-schema-compat.ts:211-213` — `normalizeToolParametersForOpenAICompat` calls `normalizeNode(..., true)` then `ensureRootObjectSchema`.
- `packages/ai/src/utils/tool-schema-compat.ts:226-233` — `ensureRootObjectSchema` either merges a root object union via `mergeRootObjectUnion` (returning `type: "object"`) or re-adds `type: "object"` when the root lost it.

Because that fix lives in the shared normalizer, both OpenAI-completions entry points this PR patched (`convertTools` and `normalizeRequestToolSchemas`, openai-completions.ts:1645/1673) already inherit the guarantee — so this PR's local `ensureRootObjectType` is now a no-op passthrough.

It's also strictly weaker than what landed: it stamps `type: "object"` on the root but never merges a root union's branch `properties`/`required` up, so a tool whose parameters are a pure `oneOf` of object branches would ship as `{oneOf: [...], type: "object"}` with no top-level properties — the payload-loss case #718 explicitly fixed.

The one genuinely-still-open gap in this area is the **Responses** path: `convertResponsesTools` (openai-responses-shared.ts:536) still forwards `tool.parameters` unnormalized, and that's what #732 targets. If you'd like to keep contributing here, that Responses-path work is where the remaining value is. Thanks again!

PR #746, opened 2026-08-07
feat(btw): add session history, review viewer, language matching, continuity

MoerAI:
## What

Four related improvements to the `/btw` side-question extension:

1. **Session history** - every completed `/btw` question and answer is persisted as a session custom entry (`pi.appendEntry("btw-history", ...)`).
2. **Review viewer** - bare `/btw` (no argument) opens a keyboard-driven viewer of this session's past side questions instead of printing a usage warning. LEFT/RIGHT switch entries, UP/DOWN scroll the selected answer, ESC closes. Outside the TUI the same list is delivered through `ctx.ui.notify`.
3. **Language matching** - `SIDE_QUERY_INSTRUCTION` now asks the model to reply in the language of the side question.
4. **Continuity** - each new side query carries the newest 10 prior question/answer pairs, so a follow-up side question remembers what the earlier ones established.

New files: `btw/history.ts` (persistence + context building), `btw/history-view-model.ts` (pure selection/scroll state machine), `btw/history-panel.ts` (the `Component` shown via `ctx.ui.custom`). Modified: `btw/index.ts`, `btw/side-query.ts`, `btw/changes.md`.

## Why

A `/btw` answer used to vanish on the next message with no way to recall it, a follow-up side question had no memory of the previous one, and non-English side questions frequently came back in English because nothing pinned the reply language.

**The isolation guarantee is preserved by construction, not by convention.** Custom entries are documented in `session-manager.ts` as "Does NOT participate in LLM context (ignored by `buildSessionContext`)", so storing side Q&A this way cannot leak into the main conversation. `tools: []`, the synchronous snapshot-before-await ordering, and the existing Esc abort semantics are all unchanged.

## Observed behavior

Driven against the real CLI, not just unit tests.

**Real CLI, RPC mode + local fake model server (14/14):**
- the side query's system prompt carries the language instruction
- a second `/btw` request contains the first question AND the first answer, with the new question last
- bare `/btw` makes **no** provider call and surfaces both past questions
- the main session messages never contain the side answer or the injected context

**Real TUI, ConPTY 120x34 (13/13):** bare `/btw` renders the viewer with the active entry accented and the rest muted, RIGHT switches to the second entry's answer, DOWN scrolls, ESC closes, and no provider call is made by the viewer.

```
 → /btw 첫번째 사이드 질문
   /btw 두번째 사이드 질문
 첫번째 답변 줄 1 한글 폭 테스트
 ...
 left/right: question   up/down: scroll   esc: close
```

**Adversarial (10/10):** Esc during a streaming side query still aborts it, the aborted question is **not** persisted, and the viewer opens and closes correctly afterwards.

## QA and tests

- 45 unit tests across `btw-history.test.ts`, `btw-history-view-model.test.ts`, `btw-history-layout.test.ts`, and `btw-side-query.test.ts`, all written failing-first.
- `npx tsc -p tsconfig.build.json --noEmit` reports zero errors in any `btw` file. Biome clean.
- The full `packages/coding-agent` suite has 245 failures on this Windows checkout; they are pre-existing. Verified by stashing this branch's changes and re-running representative failing files (`config.test.ts`, `external-editor.test.ts`, `footer-width.test.ts`), which fail identically without this change. No `btw` test is among them.

One pre-existing test was deliberately rewritten: `shows usage feedback when question is empty` asserted the `Usage: /btw <question>` warning, which is exactly the behavior item 2 replaces. It still asserts that no provider call happens. This is called out in `changes.md`.

Two bugs were found and fixed by the real-terminal QA rather than by tests: the footer key hint was clipped by the overlay's `maxHeight` (fixed by deriving the row budget from shared overlay constants via `computeBtwHistoryLayout`), and the initial context builder produced `Message` values that did not typecheck once the workspace packages were built.

## Residual risk

- Escape is still matched as the raw terminal key, so with the viewer open over an in-flight side query one Escape both closes the viewer and cancels that query. This matches the pass-through behavior already documented for this extension.
- Prior side answers are re-injected as text into later side queries; they are model output rather than trusted input, so the usual prompt-injection caveat applies inside the isolated side context.
- History is per session by design, not shared across sessions or forks.

---
## Summary by cubic
Add session-scoped `/btw` history with a keyboard-driven viewer, language-matched replies, and continuity that injects the last 10 side Q&A into new side queries to make follow-ups smarter while keeping the main thread isolated.

- **New Features**
  - Persist each completed `/btw` Q&A as a session custom entry `btw-history` (excluded from LLM context); errors/aborts don't write.
  - Bare `/btw` opens a history viewer (LEFT/RIGHT select, UP/DOWN scroll, ESC close); outside TUI, the list is sent via `ctx.ui.notify`. No provider call is made.
  - New side queries include the newest 10 prior side Q&A so follow-ups retain context.
  - Side instruction now asks the model to reply in the question’s language.

- **Bug Fixes**
  - Prevent footer key hint from being clipped by the overlay by deriving layout from shared overlay constants.
  - Correct initial context `Message` construction to typecheck across the workspace.

<sup>Written for commit 94298faa5dbd93eee5403f2027b58efb8eaaa489. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/746?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
The branch is now updated to current main at 94298faa. Both required GitHub Actions workflows were created but are waiting for maintainer approval (action_required), so no Check and test or Changelog gate jobs can start yet. Please approve and run: CI https://github.com/code-yeongyu/senpi/actions/runs/31454725079 and Changelog gate https://github.com/code-yeongyu/senpi/actions/runs/31454725094. I cannot approve them because this account has READ permission on the upstream repository.

MoerAI:
Superseded by #880, rebuilt on current main with the original review feedback addressed and refreshed QA evidence. Closing this older lineage to keep review focused on the replacement.

PR #756, opened 2026-08-08
fix(coding-agent): fail fast on cross-project --session with non-interactive stdin

realsigridjin:
## Problem

`senpi --session <id>` with a session owned by a different project always asked:

```
Session found in different project: <cwd>
Fork this session into current directory? [y/N]
```

via `readline`. When stdin is not interactive — piped, detached, closed, or a `-p` one-shot spawned by another tool — the question never settles (readline does not answer on EOF), so **the process hung forever with zero output**. Repro:

```sh
# session 019fdd28-… lives in project /tmp; run from anywhere else:
senpi --session 019fdd28-e458-7afa-8ae2-9811a2ddcae6 -p "say ok"   # hangs indefinitely
```

This is exactly what the "To resume this session: senpi --session <id>" hint printed at exit leads users into when they run it from another directory or from automation.

## Fix

- `main.ts` `--session` global-match branch: check `process.stdin.isTTY` before prompting. Without an interactive terminal, exit 1 with actionable guidance instead of hanging:

  ```
  Session found in different project: /private/tmp
  Cannot confirm forking without an interactive terminal. Use --fork '<id>' to fork it into the current directory, or re-run from /private/tmp.
  ```

  Interactive-terminal behavior is unchanged (same prompt, same fork flow).

- `promptConfirm()` now also resolves `false` on readline `close` (Ctrl+D / stdin EOF), so even interactive sessions cannot wedge on an ended input stream.

## Verification

- New regression test `test/session-cross-project-resume.test.ts` spawns the real CLI with ignored stdin against a cross-project session fixture and asserts the fast, guided failure (pre-fix this blocked until the test timeout).
- `npx vitest run test/session-cross-project-resume.test.ts test/session-id-readonly.test.ts` — 8/8 pass.
- E2E from source: `node packages/coding-agent/src/cli.ts --session <cross-project-id> -p "say ok" </dev/null` exits 1 in ~5s with the guidance message (previously hung past 120s).
- Root `npm run check` — green (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, tsc --noEmit, browser-smoke).


packages/coding-agent/src/main.ts
```diff
@@ -387,6 +392,18 @@ async function createSessionManager(
 				return openSessionOrExit(resolved.path, sessionDir);
 
 			case "global": {
+				if (!process.stdin.isTTY) {
```
code-yeongyu: [sisyphus-bot] **P1 - gate the confirmation on application mode, not stdin transport.**

This still misses the PR body's literal `senpi --session <id> -p "say ok"` case when it is launched from an ordinary terminal. `resolveAppMode()` classifies `-p` as print mode, but stdin remains a TTY, so this condition is false and the branch reaches `promptConfirm()`. I reproduced that exact shape through a real PTY: it emitted `Fork this session into current directory? [y/N]` and waited until interrupted. The ignored-stdin path does exit 1 correctly, so the gap is specifically TTY-backed non-interactive modes (and likewise affects other non-interactive modes/protocols attached to TTY stdin).

Please pass/use the resolved application mode and only prompt when it is actually `interactive`, then cover the ordinary-terminal TTY `-p` invocation. The changelog/`changes.md` claim should match that enforced behavior.

packages/coding-agent/test/session-cross-project-resume.test.ts
```diff
@@ -0,0 +1,97 @@
...
+		sessionDir: join(tempRoot, "sessions"),
+	};
+	mkdirSync(dirs.agentDir, { recursive: true });
+	mkdirSync(dirs.projectDir, { recursive: true });
+	mkdirSync(dirs.otherProjectDir, { recursive: true });
+	setup?.(dirs);
+
+	let stdout = "";
+	let stderr = "";
+	const code = await new Promise<number | null>((resolvePromise, reject) => {
```
code-yeongyu: [sisyphus-bot] **P2 - make this hang regression self-terminating.**

The spawned child exists only inside this promise executor, while the helper has no bounded timeout and neither `finally` nor `afterEach` can kill it. `packages/coding-agent/test/AGENTS.md` explicitly requires spawned-process tests to clean up children, sockets, and temp directories; the repository's existing hang regressions retain PIDs/child handles, bound the wait, and kill in unconditional cleanup. No orphan was observed in the Node 26 mutation proof, so this is not a claim that the current run leaked a process. It is a binding harness defect: on a runtime/failure path where the guarded CLI truly remains waiting, this test has no reliable way to terminate it.

Please retain the child handle outside the promise, add bounded failure handling, and kill/await it unconditionally in cleanup.

packages/coding-agent/test/session-cross-project-resume.test.ts
```diff
@@ -0,0 +1,97 @@
+import { spawn } from "node:child_process";
```
code-yeongyu: [sisyphus-bot] **P2 - place this issue regression in the required suite.**

`packages/coding-agent/test/AGENTS.md` requires issue regressions at `test/suite/regressions/<issue>-<slug>.test.ts`, but this PR adds the test at `test/session-cross-project-resume.test.ts`. Please move/rename it to `test/suite/regressions/756-session-cross-project-resume.test.ts` so it follows the binding regression organization and naming contract.

code-yeongyu (changes requested):
[sisyphus-bot] Requesting changes after an isolated, evidence-backed review. The PR's documented ordinary-terminal `-p` case still opens the cross-project fork prompt and blocks, and the new hang regression violates two binding test rules: bounded child cleanup and issue-regression location. I verified the current head, ran `npm run check`, the focused test, a full build, checked-in CLI smoke (8/8), and real PTY/xterm/Chrome QA. The three inline comments are distinct and target current right-side diff lines.

code-yeongyu (approved):
[sisyphus-bot] All three requested changes are fixed on `10e50c042`; resolving my earlier change request.

**F1 - fork prompt gated on stdin transport (P1).** `createSessionManager()` now receives the resolved `AppMode` and the cross-project branch fails fast for every non-interactive mode. Failing-first proof: the new regression case exited `0` (the readline prompt path) before the change - `AssertionError: expected +0 to be 1` - and passes after. Real-surface proof: `script -q ... senpi --session <id> -p "say ok"` on a real terminal now exits `1` with the guidance and no `Fork this session into current directory?` prompt (raw capture, parsed xterm grid, HTML and Chrome PNG captured).

**F2 - unbounded spawned child (P2).** The helper retains the child handle, applies a 15s bounded timeout, and kills it in `finally` plus an `afterEach` sweep; `ps` shows no leftover CLI child after the runs.

**F3 - regression placement (P2).** Moved to `packages/coding-agent/test/suite/regressions/756-session-cross-project-resume.test.ts`; the old top-level path no longer exists on the head.

Docs were corrected in the same push: the changelog and `src/changes.md` described the stdin-TTY-only guard, which no longer matches the shipped behaviour.

Gates: root `npm run check` exit 0, `npm run build` exit 0, focused regression 2/2, `cli-smoke.mjs --self-test` 8/8, and all 15 CI checks green on the merged head (`Check and test`, `Changelog gate` included).

PR #751, opened 2026-08-07
fix(claude-sdk-oauth): ignore volatile thinking timing

goldtg:
## Why

Claude SDK OAuth captures the provider-final assistant message during the last
`message_update`, before agent-core adds display-only `startedAt` and `endedAt`
fields to thinking blocks. Comparing that capture with the committed
`message_end` content therefore classified otherwise identical turns as
`assistant_rewritten`.

That false divergence abandoned resident-session continuity and repeatedly
re-sent the growing conversation history, increasing cache writes and eventually
exposing the replay buffer limit reported in #691.

## What

- exclude only thinking-block `startedAt` and `endedAt` fields from the
  commit-boundary digest
- preserve thinking text, signatures, tool calls, text, provider identity, and
  all other semantic content in rewrite detection
- add a deterministic 10-turn issue regression covering timing-only enrichment,
  non-mutation, thinking rewrites, and text rewrites
- document the provider-specific behavior and merge-conflict surface

## Verification

- `npm --prefix packages/coding-agent test -- test/suite/regressions/691-claude-sdk-oauth-thinking-timing.test.ts`
- Claude SDK OAuth owning suite: 39 files, 324 passed, 3 skipped
- `npm run check`
- `npm run build`
- Senpi CLI smoke: 8/8
- zero-token Anthropic mock loop: 20/20
- built-module driver: 10/10 timing-enriched turns remained `clean`; semantic
  rewrites remained `rewritten`

Fixes #691

---
## Summary by cubic
Prevents `assistant_rewritten` forks in `claude-sdk-oauth` when only thinking timing changes by excluding `startedAt`/`endedAt` from continuity hashing. Preserves resident-session and prompt-cache continuity (fixes #691).

- **Bug Fixes**
  - Exclude only `startedAt`/`endedAt` in thinking blocks from the continuity hash; keep all semantic fields.
  - Preserve rewrite detection for real changes (thinking text, signatures, tool calls, text).
  - Add a deterministic regression test covering timing-only enrichment and true rewrites.
  - Document the fix in `CHANGELOG.md` and the provider’s `changes.md`.

<sup>Written for commit cb1f05e069c75a0ef1387e68706bab26dd3b2b29. Summary will update on new commits.</sup>

[...]

Issue #790, opened 2026-08-10
claude-sdk-oauth: a content-less user message forces a full-history re-send every time it disappears


1vivy:
## What happened?

`sentMessages()` hashes every `user` and `toolResult` message. A user message whose `content` array is
empty carries nothing to transmit and is transient — present for one provider call, gone by the next.

Because it is hashed, its removal shifts every later index, `commonPrefixLength` stops short of
`sentCount`, and the turn resolves to `sent_stream_diverged`: the session forks and re-sends the whole
history even though the conversation never changed.

Captured on `2026.8.9-2` from two consecutive provider calls of one session, both 203 messages:

```text
call N      [200] toolResult bash  len=159   sha b518dd86
            [201] user             len=0     sha 12ae32cb   <- content: []
            [202] user             len=8194  sha 5c1ea814

call N+1    [200] toolResult bash  len=159   sha b518dd86
            [201] user             len=8194  sha 5c1ea814   <- shifted up one slot
            [202] toolResult bash  len=3191  sha aec978fe
```

Index 200 is byte-identical. Only the empty message disappears, and the next decision is
`flatten / sent_stream_diverged` at 203 messages.

Cost on one affected turn: `cacheRead` collapsed 582,409 -> 61,594 with `cacheWrite` 524,339, roughly 8x
its neighbouring turns. That session logged 12 `sent_stream_diverged` events in 12 minutes.

This is independent of #691 — it reproduces on `5275da1f5`, after #751 merged.

## Steps to reproduce

#791 carries the regression. To see the failure on unpatched `main`, check the PR out and revert only
the production change:

```bash
gh pr checkout 791
git checkout main -- packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/session-sync.ts
cd packages/coding-agent
npx vitest run test/suite/regressions/790-claude-sdk-oauth-empty-user-continuity.test.ts
```

Observed:

```text
× excludes a content-less user message from the transmitted set
    expected length 2, received 3
× stays a delta when a transient content-less user message disappears
    - "kind": "delta",   "reason": null
    + "kind": "fork",    "reason": "sent_stream_diverged"

Tests  2 failed | 2 passed (4)
```

Expected: `Tests  4 passed (4)`.

Restore the fix with `git checkout - -- packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/session-sync.ts`.

## Expected behavior

[...]

PR #791, opened 2026-08-10
fix(claude-sdk-oauth): ignore content-less user messages in continuity hashes

1vivy:
## Summary

- ignore user messages with no content blocks when computing the claude-sdk-oauth sent-stream hashes
- add a regression covering the transient empty message, a plain append, a genuine rewrite, and the filter itself
- record the change in the extension's `changes.md`

## Root cause

`sentMessages()` hashes every `user` and `toolResult` message. A user message whose `content`
array is empty carries nothing to transmit and is transient: it is present for one provider call
and gone by the next.

Because it is hashed, its removal shifts every later index, `commonPrefixLength` stops short of
`sentCount`, and `decideNativeContinuity` resolves the turn to `sent_stream_diverged` — forking the
session and re-sending the whole history even though the conversation never changed.

Captured on `2026.8.9-2` from two consecutive provider calls of one session, both 203 messages:

```text
call N      [200] toolResult bash  len=159   sha b518dd86
            [201] user             len=0     sha 12ae32cb   <- content: []
            [202] user             len=8194  sha 5c1ea814

call N+1    [200] toolResult bash  len=159   sha b518dd86
            [201] user             len=8194  sha 5c1ea814   <- shifted up one slot
            [202] toolResult bash  len=3191  sha aec978fe
```

Index 200 is byte-identical. Only the empty message disappears, and the next decision is
`flatten / sent_stream_diverged` at 203 messages.

Cost on one affected turn: `cacheRead` collapsed 582,409 → 61,594 with `cacheWrite` 524,339, about
8x its neighbouring turns. The same session logged 12 `sent_stream_diverged` events in 12 minutes.

Independent of #691 — this reproduces on `5275da1f5`, well after #751 merged.

## Fix

- `sentMessages()` drops user messages whose `content` array is empty
- tool results are never filtered: an empty tool result is a real observation, and its `toolCallId`
  and `toolName` stay hash-significant
- no change to the divergence decision table, the fork boundary, or the transmitted prompt for any
  message that carries content

Why this cannot be an extension: the transmitted set is computed inside the builtin provider before
session-registry continuity is decided, and no extension hook can replace it. A `changes.md` section
records the modification and its expected merge-conflict zones.

### RED

Against `5275da1f5` with only the test added:

```text
× excludes a content-less user message from the transmitted set
    expected length 2, received 3
× stays a delta when a transient content-less user message disappears
    - "kind": "delta",   "reason": null
    + "kind": "fork",    "reason": "sent_stream_diverged"

Tests  2 failed | 2 passed (4)
```

### GREEN

```text
Tests  4 passed (4)
```

The two control cases pass in both runs, so the guard still fails closed: a plain append stays a
`delta`, and a genuine rewrite of already-sent history still reports `sent_stream_diverged`.

## Verification

```text
npm run check                                   -> exit 0
npx vitest run test/suite/regressions/790-claude-sdk-oauth-empty-user-continuity.test.ts
                                                -> 4 passed
npx vitest run test/claude-sdk-oauth-*.test.ts  -> 35 files, 314 passed, 3 skipped
```

A deterministic container reproduces the bug and verifies the fix with no credentials and no tokens
spent. It clones `5275da1f5`, runs the regression unpatched, applies the production patch, and
re-runs; it exits 0 only when the unpatched tree fails specifically with `sent_stream_diverged` and
the patched tree reports all four tests passing:

```text
== ARM 1: unpatched ==
  Tests  2 failed | 2 passed (4)
  +   "reason": "sent_stream_diverged"

== ARM 2: with fix applied ==
  Applied patch ... cleanly.
  Tests  4 passed (4)

PASS: reproduced on 5275da1f5 and verified the fix resolves it.
```

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 
 - MCP shutdown no longer risks terminating unrelated processes on macOS when Homebrew `proctools` provides `pgrep`: process-tree collection now passes an explicit match-all pattern, and the kill path skips PID 1 and non-positive PIDs as defense in depth ([#823](https://github.com/code-yeongyu/senpi/issues/823)).
+- `claude-sdk-oauth` sessions no longer re-send the full conversation after a transient content-less user message disappears. Such messages are now excluded from the sent-stream continuity hash, so an unchanged conversation stays a `delta` instead of forking with `sent_stream_diverged` ([#790](https://github.com/code-yeongyu/senpi/issues/790)).
+
 
 ### New Features
 
--- a/packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/changes.md
@@ -149,6 +149,24 @@ The stored-OAuth branch in `ModelsImpl.checkProviderAuth` is a structural short-
 
 LOW in `oauth-login.ts` (added `check` to the returned shape + optional `readSettings` dep); LOW in `index.ts` (one added `readSettings` line); LOW in `provider-composer.ts` (`ExtensionOAuthConfig.check` + `adaptOAuth` forwarding, both additive).
 
+||||||| parent of 5baf13f11 (fix(claude-sdk-oauth): ignore content-less user messages in continuity hashes)
+
+## 2026-08-10 - Ignore content-less user messages in sent-stream continuity
+
+- `sentMessages()` now drops user messages whose `content` array is empty. Such a message carries nothing to transmit
+  and is transient: it is present for a single provider call and gone by the next. Hashing one shifted every later
+  index, so the following turn reported `sent_stream_diverged` and re-sent the entire history even though the
+  conversation had not changed.
+- Observed live on `2026.8.9-2`: consecutive prefix snapshots of one session held 203 messages both times, with the
+  only difference an empty user message at index 201 that vanished on the next call, forcing a 203-message flatten.
+- Tool results are never filtered. An empty tool result is a real observation, and its `toolCallId` and `toolName`
+  stay hash-significant, so genuine rewrites of already-sent history still resolve to `sent_stream_diverged`.
+- This cannot be implemented by an external extension: the transmitted set is computed inside the builtin provider
+  before session-registry continuity is decided, and no extension hook can replace it.
+- Added `test/suite/regressions/790-claude-sdk-oauth-empty-user-continuity.test.ts` covering the transient empty
+  message, a plain append, a genuine rewrite, and the filter itself.
+- Expected merge conflict zones: LOW in `session-sync.ts` (`sentMessages`); LOW in the new regression test.
+
 ## 2026-08-07 - Ignore volatile thinking timing in continuity hashes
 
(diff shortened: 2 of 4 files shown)


1vivy:
Good catch — renamed to `790-claude-sdk-oauth-empty-user-continuity.test.ts` to match the `<issue>-<slug>.test.ts` convention. The `0000-` prefix was written before the issue existed. The `changes.md` reference and the repro command in #790 now point at the new path.

`npm run check` and the regression are green on the amended commit (4 passed).

1vivy:
Verification note, since I can't produce a CI run from a fork.

I mirrored `ci.yml` in a container — Node 24, the same apt set (`libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev fd-find ripgrep`, `fdfind` symlinked to `fd`), `npm ci --ignore-scripts`, non-root user — and cloned the pushed commit.

Static stage on this commit:

```
install  PASS      check  PASS      build  PASS
```

`coding-agent` shard 1/3 is red in that container, so I ran it against untouched upstream `main` in the same image:

```
my commit      Test Files  2 failed | 282 passed | 2 skipped (286)
5275da1f5      Test Files  2 failed | 282 passed | 2 skipped (286)
```

Identical, so those failures are pre-existing here and not from this change. One of the three original failures was my own container artifact — `test/mcp/log-redaction.test.ts` expects an unwritable sink, which root defeats; it passes once the container runs as non-root.

Scoped results on this commit:

```
npm run check                                   exit 0
790-claude-sdk-oauth-empty-user-continuity      4 passed
test/claude-sdk-oauth-*.test.ts                 35 files, 314 passed, 3 skipped
```

I'm not claiming a full green `npm test` locally — it isn't, on `main` either in this environment. Happy to rerun anything specific.

code-yeongyu:
[sisyphus-bot] Excellent root-cause work and a clean reproduction, @1vivy — filtering the content-less message out of the transmitted set is exactly the right layer. 🙏

I rebased this onto current `main` and pushed two commits on top (maintainer edit):

1. **`fix(claude-sdk-oauth): tighten content-less predicate to literal zero-block arrays`** — the original `content.length === 0` also catches an empty *string* content and, more importantly, would drop whitespace-only (`"   "`) and explicit empty-text (`[{type:"text",text:""}]`) messages. Those still emit real transport blocks, so dropping them would weaken fail-closed divergence detection — a transient message that *carried* content must stay divergence-significant. The predicate is now `Array.isArray(message.content) && message.content.length === 0`, which excludes only a literal zero-block array (the transient case from the issue) and nothing else.

2. **`docs(changelog): record content-less user message continuity fix`** — the `[Unreleased]` entry the gate requires.

Verification on the rebased branch (macOS):
- `test/suite/regressions/790-...test.ts`: 6/6 pass, including two new fail-closed controls — whitespace-only and empty-text-block messages stay hashed, and the disappearance of a whitespace-only message still reports `sent_stream_diverged`.
- RED proven: the content-less exclusion fails when `session-sync.ts` is reverted to `main`.
- Full `test/claude-sdk-oauth` cluster: 327 passed / 37 files.
- senpi-qa mock-loop: 48/48.
- Changelog gate: PASS.

Ready to merge once CI is green.

PR #809, opened 2026-08-11
fix(claude-sdk-oauth): persist the continuity binding across restarts

### LINKED ISSUE #808: claude-sdk-oauth: the continuity binding is never persisted, so every restart re-sends the full history (registry_miss)
## What happened?

`session-binding.ts` implements branch-local binding checkpoints for restart-time resume:
`latestBindingOnBranch()` reads `claude-sdk-oauth-binding` custom entries back off the session
branch, and `verifyBindingAgainstTranscript()` turns one into a `reattach`. The extension's
`AGENTS.md` documents the file as "Branch-local binding checkpoint for restart-time resume
verification".

Nothing calls either function, and nothing ever writes a `claude-sdk-oauth-binding` entry. The
module's only commit, `db8e8cfeb`, added the module and its unit test — 2 files, 160 insertions,
no call sites. The unit test is the sole consumer.

The binding store that actually runs is `const bindings = new Map()` in `session-reattach.ts`.
Process memory. After a restart both the registry entry and the binding are gone, so
`decideNativeContinuity` returns `{ kind: "bootstrap" }` and the lane flattens. `session-stream.ts`
synthesizes the reason as `registry_miss` because `bootstrap` carries none, and since `firstTurn`
is false `session-observability.ts` reports the kind as `flatten`.

The user-visible result on a 697-message session: `resent the full conversation (registry_miss)`,
68.0KB, ~60K tokens re-billed, on the first turn after every restart.

## Steps to reproduce

1. Run a `claude-sdk-oauth` session for several turns.
2. Restart senpi.
3. Send one more turn in the same session.
4. The continuity notice reports a full re-send with reason `registry_miss`.

`grep -c claude-sdk-oauth-binding <session>.jsonl` returns 0, confirming no checkpoint was ever
written.

## Expected behavior

A restart should resume rather than re-send. The checkpoint is written as the turn commits and
consulted when the in-memory registry misses, so the first turn after a restart is a delta.

This is distinct from #691 (`assistant_rewritten` from thinking timing) and #790
(`sent_stream_diverged` from content-less user messages). Both of those fix divergence within a
live process; neither addresses a cold start, and unlike them this one has no workaround.

## Version

```text
@code-yeongyu/senpi 2026.8.11-2
@anthropic-ai/claude-agent-sdk 0.3.220
provider claude-sdk-oauth · model claude-opus-5
```

1vivy:
Fixes #808.

## What this fixes

`session-binding.ts` has been in the tree since `db8e8cfeb` ("persist continuity bindings for
verified restart resume"), and `AGENTS.md` documents it as "Branch-local binding checkpoint for
restart-time resume verification". It was never wired. Nothing called `latestBindingOnBranch()`,
nothing called `verifyBindingAgainstTranscript()`, and nothing ever wrote a
`claude-sdk-oauth-binding` entry — that commit added the module and its unit test, 2 files, no call
sites, and the unit test stayed its only consumer.

So the only binding store that runs is the process-memory `Map` in `session-reattach.ts`. After a
restart both `getSession()` and `getBinding()` miss, `decideNativeContinuity` returns `bootstrap`,
and the lane flattens. `session-stream.ts` synthesizes the reason as `registry_miss` because
`bootstrap` carries none, and `session-observability.ts` reports the kind as `flatten` once
`firstTurn` is false. On a 697-message session that is 68.0KB and ~60K tokens re-sent on the first
turn after every restart.

## Approach

- **Write** — `session-registry-wiring.ts` appends a `BindingCheckpoint` through `pi.appendEntry` at
  the `message_end` commit boundary, only after the boundary reports a non-rewritten commit.
  `registerSessionRegistry` now takes `Pick<ExtensionAPI, "on" | "appendEntry">`.
- **Read** — a new `session_start` handler lifts the newest checkpoint off
  `ctx.sessionManager.getBranch()` into a module map. The decision needs the current sent-hash
  prefix, which only exists once the provider context is built, so the checkpoint is held until
  `createResidentAttempt` can verify it.
- **Verify** — `rehydrateBindingFromCheckpoint` rebuilds the binding from the *current* hashes and
  only after `prefixDigest(hashes, sentCount)` equals the recorded `sentPrefixHash`, and only when
  account, model, `systemPromptHash` and `toolsetHash` all match. Every refusal falls through to
  today's cold path, so the worst case is the behavior you already have.

Resume-first is preserved: this adds a way to *reach* reattach after a restart, and changes no
existing decision.

## Design notes

- The checkpoint stays compact on purpose. `sentPrefixHash` is one sha256 over the prefix; storing
  the full hash array would add tens of KB to the transcript on **every** turn.
- A live entry gets its config-drift check from `identityDrift`, which a restarted process cannot
  run because the entry is gone. `BindingCheckpoint` therefore gained optional `systemPromptHash`
  and `toolsetHash`; absent means unknown, and unknown never rehydrates.
- `claudeConfigDir` became optional because nothing consumes it.
- **`verifyBindingAgainstTranscript` is deliberately left untouched and still unused.** Supplying
  its `transcriptExists` / `transcriptHasBoundaryUuid` honestly would mean hard-coding Claude Code's
  private `projects/<mangled-cwd>/<id>.jsonl` layout, and this extension deliberately treats resume
  failure as that gate instead (`session-stream.ts` catches it into
  `resume_initialization_failed`). I did not want to pass it invented `true`s. Happy to wire it with
  a real probe if you'd prefer that direction — flagging it as the one open design call.

## Verification

- `test/suite/regressions/808-claude-sdk-oauth-binding-persistence.test.ts` — 11 tests: checkpoint
  derivation, branch round-trip, successful rehydrate, six refusal cases (rewritten prefix,
  truncated history, changed system prompt, changed toolset, different account, different model),
  a pre-identity checkpoint, and single-use consumption. **11 passed.**
- Red-run proof: deleting the `prefixDigest` guard fails exactly
  `refuses to rehydrate on a rewritten prefix` (1 failed | 10 passed); restoring it returns 11/11.
- `npm run check` passes (also enforced by the pre-commit hook on this commit).
- `test/claude-sdk-oauth-binding.test.ts`, `test/claude-sdk-oauth-continuity.test.ts` and all of
  `test/suite/regressions/` — 325 passed, 7 failed. Those 7 (theme init, fswatch, models.json hot
  reload, two model-selector cases, cross-project resume) reproduce identically with these source
  changes reverted, so they are pre-existing on `main` and unrelated.

Draft because of the open design call above and because I have not run the `senpi-qa` CLI channel —
the restart path needs a real multi-turn OAuth session, which I can do if you want that evidence
before review.

---
## Summary by cubic
Persist continuity bindings for `claude-sdk-oauth` across restarts so sessions reattach instead of re-sending full history on the first turn. Fixes #808.

- **Bug Fixes**
  - Write a compact binding checkpoint at `message_end`, read it at `session_start`, and rehydrate only if the current prefix digest and config identity (account, model, `systemPromptHash`, `toolsetHash`) match.
  - Fall back to the existing cold path when the prefix is rewritten/truncated or identity changes; checkpoints are single-use.
  - Updated registry wiring to append checkpoints and export `prefixDigest`; made `claudeConfigDir` optional.
  - Added regression tests covering successful rehydration and refusal cases.

<sup>Written for commit 0bdff3263a71e61f0196c08902c8fbc236d60bad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/809?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -16,6 +16,11 @@
   authenticated ambient Claude CLI is usable. Empty sentinel credentials and logged-out local CLIs are skipped, and
   fallback responses carrying errors such as `Not logged in` cannot emit an immediate or delayed green
   `Fallback model responded` notice ([#803](https://github.com/code-yeongyu/senpi/pull/803)).
+- The `claude-sdk-oauth` continuity binding is persisted to the session branch and restored on the next start, so a
+  restart resumes the existing Claude session instead of re-sending the whole conversation as a `registry_miss`
+  flatten. The binding is only rebuilt when the current sent-prefix digest, account, model, system prompt and
+  toolset all still match; anything unproven keeps today's behavior
+  ([#809](https://github.com/code-yeongyu/senpi/pull/809)).
 
 ### New Features
 
--- a/packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/changes.md
@@ -1,5 +1,40 @@
 # claude-sdk-oauth extension changes
 
+## 2026-08-11 - Persist the continuity binding so a restart resumes instead of re-sending
+
+- `session-binding.ts` has been present since `db8e8cfeb` ("persist continuity bindings for verified restart
+  resume") but was never wired: nothing called `latestBindingOnBranch()` and nothing ever wrote a
+  `claude-sdk-oauth-binding` entry. Its unit test was the only consumer, so the sole surviving binding store was the
+  process-memory `Map` in `session-reattach.ts`.
+- Consequence: after a restart both `getSession()` and `getBinding()` miss, `decideNativeContinuity` returns
+  `bootstrap`, and the lane flattens. `session-stream.ts` synthesizes the reason `registry_miss` because `bootstrap`
+  carries none, and `session-observability.ts` reports it as `flatten` once `firstTurn` is false. Observed on a
+  697-message session: 68.0KB and ~60K tokens re-sent on the first turn after every restart.
+- Write path: `session-registry-wiring.ts` appends a `BindingCheckpoint` via `pi.appendEntry` at the `message_end`
+  commit boundary, only after the boundary reports a non-rewritten commit. `registerSessionRegistry` therefore now
+  takes `Pick<ExtensionAPI, "on" | "appendEntry">`.
+- Read path: a new `session_start` handler lifts the newest checkpoint off `ctx.sessionManager.getBranch()` into a
+  module map. The decision needs the current sent-hash prefix, which only exists once the provider context is built,
+  so the checkpoint is held until `createResidentAttempt` can verify it.
+- Verification is fail-closed. `rehydrateBindingFromCheckpoint` rebuilds the binding from the CURRENT hashes and only
+  after `prefixDigest(hashes, sentCount)` equals the recorded `sentPrefixHash`, and only when account, model,
+  `systemPromptHash`, and `toolsetHash` all match. A live entry gets that drift check from `identityDrift`, which a
+  restarted process cannot run, so `BindingCheckpoint` gained optional `systemPromptHash`/`toolsetHash`; absent means
+  unknown and unknown never rehydrates. Every refusal falls through to today's cold path.
+- The checkpoint stays compact deliberately: `sentPrefixHash` is one sha256 over the prefix, where persisting the
+  full hash array would add tens of KB to the transcript on every turn. `claudeConfigDir` became optional because
+  nothing consumes it — a real transcript-existence probe would have to hard-code Claude Code's private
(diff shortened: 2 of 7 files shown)


code-yeongyu:
ULW audit verdict: BLOCKED. Restart persistence is a real bug, but this draft/conflicting branch reattaches stale checkpoints after assistant rewrites, trusts missing/tampered SDK transcripts too early, regresses rejected-compaction and content-less-message behavior, and lacks complete config/schema/lifecycle validation. Local evidence: `local-ignore/qa-evidence/20260817-fix-pr-audit/pr-809/TERMINAL-VERDICT.md`.

code-yeongyu:
Superseded by #959, which lands the bounded persisted-prefix design this PR proposed together with the trusted-sidecar ownership and lifecycle invalidation from the #954 security review.

PR #824, opened 2026-08-12
fix(mcp): scope child process lookup

### LINKED ISSUE #823: MCP shutdown can terminate unrelated processes with Homebrew pgrep
## What happened?

On macOS, quitting an OmO native session can terminate unrelated processes when Homebrew `proctools` provides the first `pgrep` on `PATH`.

MCP shutdown calls `collectProcessTree()`, which currently runs `pgrep -P <pid>` without the required positional pattern. Homebrew `pgrep` treats that as a system-wide match, so `reapProcessTree()` can collect and signal PIDs that are not descendants of Senpi.

This matters because a normal `/quit` can close unrelated terminals, editors, and background services.

## Steps to reproduce safely

Do not reproduce this through `/quit`. The lookup bug can be observed without sending any signals:

```sh
sleep 30 & child=$!
/opt/homebrew/bin/pgrep -P $$ | wc -l
/opt/homebrew/bin/pgrep -P $$ .
/usr/bin/pgrep -P $$ .
kill "$child"
```

On my machine, the patternless Homebrew command returned 798 PIDs and included PID 1. Both commands with the explicit `.` pattern returned only the actual child.

## Expected behavior

Process-tree collection should query only direct children. Passing `.` as the match-all process-name pattern works with both Homebrew `proctools` and macOS `/usr/bin/pgrep`.

I am preparing a minimal fix and regression test.

## Version

- `omo-ai@5.0.0-0.beta.6`
- nested `@code-yeongyu/senpi@2026.8.11-4`
- Homebrew `proctools@0.4pre1`
- current `main` at `20d33e8e46bef8ad38fcf1e99f8196b599741c44`

bagelcode-jhkim:
## Summary

- pass an explicit match-all pattern to `pgrep -P` during MCP process-tree collection
- keep the transport test helper on the same safe command form
- add a deterministic regression for Homebrew `proctools` behavior

## Why

Homebrew `proctools` accepts `pgrep -P <pid>` without a positional pattern but treats it as a broad process query. During MCP shutdown, that can feed unrelated PIDs into Senpi's tree reaper and terminate processes outside the MCP server tree.

Passing `.` preserves the intended match-all process-name behavior while limiting results to children of the requested parent. This form works with both Homebrew `pgrep` and macOS `/usr/bin/pgrep`.

Closes #823

## Verification

- issue regression and MCP transport tests: 6/6 passed
- `npm run check`: passed
- `npm run build`: passed
- Senpi QA harness self-check: 9/9 passed
- isolated real-CLI MCP mock loop: 5/5 passed
- no-signal Homebrew probe: collected exactly the probe root and its owned child; PID 1 absent; child cleaned up
- fresh code review: CLEAR / APPROVE
- independent QA and final gate: PASS / APPROVE

The broad coding-agent suite reached 7,206 passing tests; 12 unrelated existing tests failed because `fd` was unavailable or timed out. The changed tests passed independently and in the focused suite.

## Safety

The regression and manual probe only enumerate PIDs. They do not call the process reaper, launch the TUI, or signal any unowned process.

## Changelog

Per `CONTRIBUTING.md`, this PR does not edit `CHANGELOG.md`; maintainers own changelog entries. The required MCP fork ledger entry is included.

---
## Summary by cubic
Scopes MCP child-process lookup and hardens shutdown to avoid killing unrelated processes. Previously, Homebrew `proctools` treated `pgrep -P <pid>` without a pattern as a broad query; now we pass "." and skip PID 1 and non‑positive PIDs before signaling. Closes #823.

- Use `pgrep -P <pid> .` in `process-tree.ts` (and the transport test helper) to reliably return only direct children on macOS and Linux.
- Guard `killPids` to ignore PID 1 and non‑positive PIDs.
- Add a deterministic regression test that fakes `pgrep`, verifies unrelated PIDs are excluded, and ensures PID 1 is never signaled.

<sup>Written for commit 38e004b4911a703899762981329e47d9f7a070a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/824?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -7,6 +7,8 @@
 - The ambient Claude auth availability probe now passes `windowsHide: true` when spawning `claude auth status`, so the background check no longer opens a console window on Windows ([#870](https://github.com/code-yeongyu/senpi/issues/870)).
 - `senpi --help` now lists the `PI_RULES_DISABLED`, `PI_RULES_MAX_RULE_CHARS`, and `PI_RULES_MAX_RESULT_CHARS` environment settings that the built-in rules extension reads, so the two environment-only character limits are discoverable from the CLI ([#678](https://github.com/code-yeongyu/senpi/issues/678)).
 
+- MCP shutdown no longer risks terminating unrelated processes on macOS when Homebrew `proctools` provides `pgrep`: process-tree collection now passes an explicit match-all pattern, and the kill path skips PID 1 and non-positive PIDs as defense in depth ([#823](https://github.com/code-yeongyu/senpi/issues/823)).
+
 ### New Features
 
 ### Breaking Changes
--- a/packages/coding-agent/src/core/extensions/builtin/mcp/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/mcp/changes.md
@@ -4,7 +4,8 @@
 
 ### What changed
 - `process-tree.ts` now passes `.` as the positional match-all pattern to `pgrep -P`.
-- `test/suite/regressions/issue-823-mcp-pgrep-pattern.test.ts` places a deterministic fake `pgrep` first on PATH and proves unrelated PIDs are excluded.
+- `killPids` now skips any non-positive or PID-1 entry before signaling, as defense in depth against a broken or substituted discovery executable returning a catastrophic target.
+- `test/suite/regressions/issue-823-mcp-pgrep-pattern.test.ts` places a deterministic fake `pgrep` first on PATH and proves unrelated PIDs are excluded, and that PID 1 is never signaled even when discovery returns it.
 - `test/mcp/transport.test.ts` uses the same explicit pattern in its child-PID helper.
 
 ### Why
--- a/packages/coding-agent/src/core/extensions/builtin/mcp/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/mcp/changes.md
@@ -1,5 +1,21 @@
 # mcp Extension Changes
 
+## Explicit pgrep match-all pattern for process-tree collection (2026-08-12)
+
+### What changed
+- `process-tree.ts` now passes `.` as the positional match-all pattern to `pgrep -P`.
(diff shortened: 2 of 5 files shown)


code-yeongyu:
[sisyphus-bot] Thanks for the clean, well-tested fix, @bagelcode-jhkim — the explicit `.` pattern is exactly right for the proctools vector. 🙏

I rebased this onto current `main` and pushed two commits on top (maintainer edit) to take it the rest of the way:

1. **`fix(mcp): guard killPids against signaling PID 1`** — the `.` pattern fixes the discovery vector, but `killPids` (`process-tree.ts:66`) still signals every collected PID with no validation. As cheap defense in depth, it now skips any non-positive or PID-1 entry before `process.kill`, so even a future discovery regression can't hit the most catastrophic target. A legitimate MCP child can never be PID 1, so this can't exclude real work. I deliberately did **not** add same-UID or PPID re-validation: unrelated terminals/editors share senpi's UID, and descendants can legitimately reparent during the TERM/KILL race — both would either not help or add real complexity for no safety gain.

2. **`docs: record MCP process-tree pgrep + kill guard fix`** — the `[Unreleased]` changelog entry the gate requires, plus the `changes.md` note.

Verification (all on the rebased branch, macOS):
- New regression `never signals PID 1 even if process discovery returns it`: RED before the guard (killPids signaled PID 1), GREEN after.
- `test/suite/regressions/issue-823-mcp-pgrep-pattern.test.ts`: 2/2 pass.
- Full `test/mcp` directory: 309 tests / 47 files pass.
- senpi-qa: mock-loop self-test 48/48, `--with-mcp-tool mcp_fx_tool_1` 5/5 (real MCP stdio fixture round-trip).
- Changelog gate: PASS.

This is ready to merge once CI is green.

PR #836, opened 2026-08-12
fix(claude-sdk-oauth): resolve request auth from ambient Claude credentials

ismetanin:
## Problem

On a machine with a logged-in Claude CLI but no `claude-sdk-oauth` entry in `auth.json`, every request to the provider fails:

```
Error: Provider is not configured: claude-sdk-oauth
Runtime error (session_title_generation): Provider is not configured: claude-sdk-oauth
```

`session_title_generation` fires on session start, so the error appears before the user types anything. No configuration helps, because only a stored credential is ever consulted.

Reproduce with an empty `~/.senpi/auth.json`, `claude auth status` exiting 0, and any `claude-sdk-oauth` model selected.

## Root cause

A regression from 2acbb6e0c ("Require a real OAuth login for runtime availability"), which removed one line from the provider registration:

```diff
-		apiKey: "claude-sdk-oauth-managed",
```

That placeholder was the provider's only route to api-key auth. `resolveProviderAuth()` reads ambient credentials **exclusively** through `provider.auth.apiKey.resolve()`:

```ts
// packages/ai/src/auth/resolve.ts
const stored = await readCredential(credentials, provider.id);
if (stored) { /* ... */ }
// Ambient (env vars, AWS profiles, ADC files).
return provider.auth.apiKey
	? resolveApiKey(requestAuthContext, provider.auth.apiKey, provider.id, undefined)
	: undefined;   // <- oauth-only provider lands here
```

With the placeholder gone the provider registers `oauth` alone, and `composeApiKeyAuth` refuses to give it api-key auth:

```ts
// packages/coding-agent/src/core/provider-api-key-auth.ts
if (!inherited && rawKey === undefined && !headerSource && oauth) return undefined;
```

Availability still passed, so the models stayed selectable: `Models.checkProviderAuth()` falls back to `oauth.check()` with no credential, and that check accepts an env token or a logged-in CLI. Availability and resolution disagreed — the provider advertised models it could not authenticate.

The stream layer was never the problem: `queryWithAuthLane` has always had an `ambient` lane, and `managedPool` selects it by default (`settings.tokenInjection ?? "ambient"`).

## Fix

Give the extension a way to declare that it accepts ambient credentials, rather than reinstating a literal sentinel.

- `ExtensionOAuthConfig` gains an additive optional `resolveAmbient()`.
- `composeApiKeyAuth` composes ambient-only api-key auth from it instead of returning `undefined`. It omits `login`, so the OAuth flow keeps ownership of login, and declines whenever a credential is passed, so a stored credential still wins.
- `createOAuthConfig` implements `resolveAmbient()`, sharing a single `ambientAvailable()` predicate with `check()` so the two cannot drift apart again.

This does not restore the false availability 2acbb6e0c fixed. The removed literal reported configured unconditionally; `resolveAmbient()` is gated on the same real probe as `check()`, so a logged-out machine still resolves to `undefined`. The test added by that commit (`test/suite/claude-sdk-oauth-extension.test.ts`) passes unchanged.

`availability.ts` also memoises the ambient probe (30s TTL, shared in-flight read, rejections uncached). It spawns the Claude binary at ~200-650ms and now sits on the per-request auth path rather than only on catalog refresh.

`changes.md` entries added under `src/core/` and the extension directory per the fork strategy.

## Tests

`test/claude-sdk-oauth-ambient-resolve.test.ts` drives the real registration through `composeModelProvider()` and `resolveProviderAuth()` against an empty credential store — it fails on `main` with `undefined` and passes here:

- resolves from an authenticated ambient CLI with nothing stored
- resolves from `CLAUDE_CODE_OAUTH_TOKEN` with nothing stored
- still reports not configured when logged out and nothing is stored

`test/claude-sdk-oauth-ambient-cache.test.ts` covers the probe cache: TTL reuse, re-probe after expiry, concurrent in-flight sharing, and no caching of rejections.

## Verification

- `npm run check` passes (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, tsc, browser-smoke).
- All 45 `claude-sdk-oauth` / auth-storage / provider / model-registry test files pass (435 tests).
- Full `packages/coding-agent` suite shows no change against `main`: the same 14 files / 55 tests fail before and after, all from workspace `dist` not being built in this environment (`ERR_MODULE_NOT_FOUND: pi-ai/dist/index.js`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Restores ambient authentication for `claude-sdk-oauth` and makes request-scoped environment authoritative. Before: availability passed (env token or logged-in CLI) but requests failed with “Provider is not configured” and SDK children inherited host tokens; after: ambient auth resolves, request tokens override host tokens across replay and SDK spawns, and stored OAuth remains the winner when present.

- Adds `resolveAmbient()` to the extension OAuth config and composes ambient‑only API‑key auth in `provider-api-key-auth`, including configured headers and `authHeader`. Unrelated explicit keys remain rejected; stored OAuth outranks ambient adapters.
- Threads request `env` through `@earendil-works/pi-ai` resolution, provider availability `check`, and `toAuth`, then returns it for auxiliary replay. Explicit empty values mask host env; header parity matches stored OAuth.
- Treats request Claude token slots as a complete namespace: masks do not import other host slots; SDK subprocesses receive only Claude token vars and exclude process‑control vars (`PATH`, `NODE_OPTIONS`, etc.).
- Routes request tokens away from `config-dir` persistence; resident and non‑resident lanes both honor request env.
- Bounds the ambient probe: 30s shared TTL cache, 10s deadline that kills a hung `claude auth status`, and per‑request `AbortSignal` so abandoned callers stop waiting without cancelling others.
- Stored OAuth now runs `check` before deriving request auth, so empty persisted envelopes no longer count as configured.
- Tests cover ambient resolution, cache/timeout/abort behavior, replay idempotency, header/env parity, cross‑slot masking, and resident lane forwarding. No migration required.

<sup>Written for commit 9e80ae152ac38c977ac61d19e5cbcbf0baac3cae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/836?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/oauth-login.ts
```diff
@@ -65,24 +78,38 @@ async function promptAccountName(callbacks: OAuthLoginCallbacks, existing: Accou
...
+		async check({ ctx, credential, signal }) {
+			return (await configuredFor(ctx, credential as ClaudeSdkOauthCredential | undefined, signal))
+				? AUTH_CHECK
+				: undefined;
+		},
+
+		async resolveAmbient({ ctx, signal }) {
+			// Only reached with nothing stored: resolveProviderAuth consults the stored credential first.
+			if (!(await configuredFor(ctx, undefined, signal))) return undefined;
+			return { auth: { apiKey: SENTINEL_OAUTH_FIELDS.access }, source: AUTH_CHECK.source };
```
code-yeongyu: [sisyphus-bot] [P1] Preserve the request-scoped Claude environment here.

`ctx` includes `ProviderRequestOptions.env`, which is documented to override `process.env`. `configuredFor()` accepts a request-only `CLAUDE_CODE_OAUTH_TOKEN`, but this result drops that environment and carries only the sentinel. The SDK lane later reconstructs its environment from the host process. I reproduced this with `request-token` at resolution and `host-token` in the host boundary: resolution succeeded, but the captured subprocess token was `host-token` and the request token was not forwarded.

This crosses the credential-precedence boundary for embedded/multi-tenant callers and also turns a successful auth check into a downstream failure when the host has no credential. Return/retain the effective Claude environment in `AuthResult` and thread it into `queryWithAuthLane()` account discovery and subprocess construction. Add a host-token-vs-request-token end-to-end test.

packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/oauth-login.ts
```diff
@@ -65,24 +78,38 @@ async function promptAccountName(callbacks: OAuthLoginCallbacks, existing: Accou
...
-			}
-			return accountCount > 0 ? AUTH_CHECK : undefined;
+		async check({ ctx, credential, signal }) {
+			return (await configuredFor(ctx, credential as ClaudeSdkOauthCredential | undefined, signal))
+				? AUTH_CHECK
+				: undefined;
+		},
+
+		async resolveAmbient({ ctx, signal }) {
+			// Only reached with nothing stored: resolveProviderAuth consults the stored credential first.
```
code-yeongyu: [sisyphus-bot] [P1] This assumption excludes a normal persisted state.

`emptyCredential()` is written by the account lifecycle and remains after the last managed account is removed. In that state `auth.json` contains an OAuth envelope with `accounts: []`. `resolveProviderAuth()` takes its stored-OAuth branch before this method can run, so the shared predicate is bypassed. On the exact head with a logged-out ambient probe, `check()` returned undefined while request resolution returned `{ apiKey: "claude-sdk-oauth-managed", source: "OAuth" }`.

That restores the availability/resolution disagreement this PR is meant to remove. The stored-OAuth path must validate an empty sentinel envelope with the same `configuredFor()` predicate, or the account lifecycle must remove the envelope atomically when no managed/env account remains. Add the empty-envelope logged-out regression.

packages/coding-agent/src/core/provider-api-key-auth.ts
```diff
@@ -84,6 +92,34 @@ export function composeApiKeyAuth(
...
+	const resolveAmbient = ambientResolverOf(oauth);
+	if (!resolveAmbient) return undefined;
+	return {
+		name: (oauth as { name?: string }).name ?? "Ambient credentials",
+		check: async ({ ctx, credential, signal }) => {
+			if (credential) return undefined;
+			const resolved = await resolveAmbient({ ctx, signal });
+			return resolved ? { type: "oauth", source: resolved.source } : undefined;
+		},
+		resolve: async ({ ctx, credential, signal }) => (credential ? undefined : resolveAmbient({ ctx, signal })),
```
code-yeongyu: [sisyphus-bot] [P1] Do not reject this provider's own synthetic marker on re-resolution.

The first ambient `getAuth()` returns `claude-sdk-oauth-managed` as an API key. Session title, compaction, and branch-summary helpers copy that value into explicit request options, so the next auth pass reaches this branch with `credential` set and returns undefined. Exact-head proof: first resolution returned the marker; replaying the returned `apiKey` through the same resolver produced `replayResolved: false`.

Make ambient resolution idempotent across the existing request-auth round trip, or represent ambient auth without exposing a fake API key. Cover `getAuth()` followed by `getAuth({ apiKey: first.auth.apiKey })` plus one real auxiliary Claude-provider call.

packages/coding-agent/src/core/provider-api-key-auth.ts
```diff
@@ -35,7 +35,15 @@ export function composeApiKeyAuth(
...
 	const headerSource = headerAuthResolutionSource(config?.headers, extension?.headers);
-	if (!inherited && rawKey === undefined && !headerSource && oauth) return undefined;
+	if (!inherited && rawKey === undefined && !headerSource && oauth) {
+		// An OAuth provider with no key, no headers and no base normally gets no
+		// api-key auth at all. That is right for providers whose only credential
+		// is the stored one, but `resolveProviderAuth` reads ambient credentials
+		// exclusively through `apiKey.resolve()`: denying it here strands
+		// providers whose credentials live outside auth.json, which pass
+		// `check()` and then fail every request as "Provider is not configured".
+		return ambientOnlyAuth(oauth);
```
code-yeongyu: [sisyphus-bot] [P2] Keep configured headers and `authHeader` composition on the ambient path.

This early return happens before `authHeader` is read and bypasses `resolveHeadersOrThrow()` / `withConfiguredAuth()`. I composed an exact-head ambient provider with `headers: { "User-Agent": "must-survive" }` and `authHeader: true`; resolution returned only the API key and source, with neither the metadata header nor `Authorization` present. Stored OAuth users still receive those through `composeOAuthAuth().toAuth()`, so behavior now depends on where the same credential lives.

Route the ambient result through the normal configured-header/authHeader composition and add parity coverage between ambient and stored OAuth.

code-yeongyu (changes requested):
[sisyphus-bot] REQUEST_CHANGES

I reproduced four merge blockers on the exact current head.

1. [P1] Request-scoped Claude credentials are accepted during auth resolution but replaced by the host credential at SDK spawn. With `request-token` in `ProviderRequestOptions.env` and `host-token` in the host boundary, the exact-head driver returned `resolutionAccepted: true`, `subprocessToken: "host-token"`, `requestTokenForwarded: false`. This can bill or attach tenant content to the wrong Claude account, or fail after auth was reported successful.
2. [P1] The synthetic ambient marker is not idempotent through auxiliary auth. First resolution returned `claude-sdk-oauth-managed`; replaying it as the explicit `apiKey`, as title/compaction helpers do, returned no auth. Auto-title, compaction, and branch-summary calls can still hit `Provider is not configured`.
3. [P1] A normally persisted `emptyCredential()` bypasses the new ambient predicate. With no managed accounts and a logged-out probe, `check()` returned undefined while `resolveProviderAuth()` returned the sentinel through the stored-OAuth branch.
4. [P1 gate] The required changelog gate fails: four runtime source files changed, `packages/coding-agent/CHANGELOG.md` did not, and the PR has no `no-changelog` label. `checkPrChangelog()` returned `pass: false`, reason `runtime source changed without a CHANGELOG.md entry`.

Focused regression tests still pass (3 files, 11 tests), so the current suite does not cover these cross-boundary failures. Please preserve the effective request environment through execution, make ambient auth survive the auxiliary round trip, validate empty stored envelopes against the same availability predicate, add end-to-end regressions for each, and add the required `[Unreleased]` changelog entry.

packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/oauth-login.ts
ismetanin: Fixed in c2653e20.

The effective Claude environment is now retained in the `AuthResult` and threaded through `queryWithAuthLane()` account discovery and subprocess construction on both the non-resident (`stream.ts`) and resident (`session-stream.ts`) lanes, so a request-scoped `CLAUDE_CODE_OAUTH_TOKEN` overrides a conflicting host token instead of the SDK lane rebuilding its environment from the host process.

Regression added as you specified — the host-token-vs-request-token end-to-end case asserts the captured SDK spawn env carries `request-token` and not `host-token`, and it drives a real auxiliary title generation through the same resolver rather than asserting on the resolver in isolation.

packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/oauth-login.ts
ismetanin: Fixed in c2653e20.

You are right that `emptyCredential()` is a normal persisted state and that `resolveProviderAuth()` reaches its stored-OAuth branch before the ambient predicate can run. I took the first of your two options: `resolveStoredOAuth()` in `packages/ai/src/auth/resolve.ts` now runs the provider's own `OAuthAuth.check` before resolving, so an envelope with `accounts: []` and a logged-out ambient probe can no longer resolve while `check()` reports unconfigured.

Regression added for the empty-envelope logged-out case.

Flagging the blast radius explicitly since this is shared `packages/ai`: every OAuth provider's stored-credential path now consults `check` first. I ran the full `packages/ai` suite (1982 passed) and the coding-agent OAuth/runtime clusters (41 files, 346 tests) to bound it. Happy to gate it to this provider instead if you would rather not change the shared contract.

packages/coding-agent/src/core/provider-api-key-auth.ts
ismetanin: Fixed in c2653e20.

Ambient resolution is now idempotent across the request-auth round trip: the resolver accepts a replay of its **own** marker while still rejecting unrelated explicit credentials, so the title, compaction and branch-summary helpers that copy resolved auth into explicit request options resolve again instead of returning undefined.

Covered exactly as you asked — `getAuth()` followed by `getAuth({ apiKey: first.auth.apiKey })`, plus one real auxiliary Claude-provider call (session title generation) in the same test.

packages/coding-agent/src/core/provider-api-key-auth.ts
ismetanin: Fixed in c2653e20.

The ambient result no longer short-circuits before header composition. It routes through `configContextEnv` -> `resolveHeadersOrThrow()` -> `withConfiguredAuth()`, so configured metadata headers and `authHeader` composition apply on the ambient path exactly as they do for stored OAuth via `composeOAuthAuth().toAuth()`.

Parity coverage added between ambient and stored OAuth for configured headers and `authHeader`.

ismetanin:
All four findings plus the changelog gate are addressed in c2653e20. Inline replies have the detail; summary:

1. **[P1] Request-scoped env dropped** — the effective Claude environment is retained in `AuthResult` and threaded through account discovery and both SDK lanes, so a request token overrides a conflicting host token instead of the subprocess rebuilding env from the host. End-to-end regression asserts the captured spawn env carries `request-token`, not `host-token`.
2. **[P1] Ambient marker not idempotent** — the resolver now accepts a replay of its own marker while still rejecting unrelated explicit credentials. Covered by `getAuth()` -> `getAuth({ apiKey: first.auth.apiKey })` plus a real auxiliary title-generation call.
3. **[P1] Empty stored envelope** — `resolveStoredOAuth()` now runs the provider's `OAuthAuth.check` before resolving, so an `accounts: []` envelope with a logged-out probe can no longer resolve while `check()` says unconfigured.
4. **[P2] Header composition** — the ambient result routes through `resolveHeadersOrThrow()`/`withConfiguredAuth()`, with ambient-vs-stored-OAuth parity coverage.
5. **[P1 gate] Changelog** — `[Unreleased]` entries added to both `packages/coding-agent/CHANGELOG.md` and `packages/ai/CHANGELOG.md`. `node scripts/check-pr-changelog.mjs --base <merge-base>` now reports `PASS - changelog entry updated`. Nearest fork `changes.md` records updated in the same increment per the repo contract.

Verification: coding-agent OAuth/runtime clusters 41 files / 346 tests pass, `packages/ai` 1982 pass, root `npm run check` green (it caught a `ProviderHeaders` nullability error in my new test, now fixed).

One thing worth your explicit call: item 3 changes shared `packages/ai` behavior, so **every** OAuth provider's stored-credential path now consults `check` first. I ran the full `packages/ai` and coding-agent auth suites to bound it, but if you would rather not move the shared contract, say so and I will scope it to this provider instead.

code-yeongyu (approved):
Brutal re-review complete on final head 9e80ae152 after three verified security-fix commits and a history-preserving merge of current main. Request OAuth environment survives resolver/replay/SDK spawn, explicit empty masks replace the complete Claude token namespace, request credentials do not persist in config-dir mode, OAuth check and derivation share one transient credential, cache/abort ownership is bounded, and all four prior review threads are resolved. Local evidence: AI auth 142/142; Claude SDK OAuth 370 passed with 3 pre-existing skips; merged coding-agent 7,588 passed with 36 skips; static check and build exit 0; production driver reports crossSlotImported=false and configDirPersisted=false; isolated CLI 8/8 and mock loop 48/48 with zero real provider calls.

PR #853, opened 2026-08-13
fix(compaction): apply warm summaries anchored to the summarized prefix

code-yeongyu:
## Problem

The idle warm-up exists so summarization happens while the user is **not** waiting. Its result was
pinned to `expectedRevision`, a counter `agent-session` bumps on **every** appended message. A
session parked in a cache-warm wait keeps appending — wait notices, monitor state, and finally the
user's own next prompt — so the warm summary was guaranteed stale exactly when the blocking route
needed it.

Field logs from a real session:

```
04:35:04  speculative_started  gen11        # warm-up begins
04:39:44  threshold_trigger                 # user sends a message
04:39:44  warm_consumed        gen11
04:39:44  speculative_stale    gen11        # warm summary thrown away
04:40:33  compaction committed 361068 -> 21553   # 49s fresh summarization, user waiting
```

Aggregate over that log: **415 warm-ups started, 36 consumed, 26 stale, 10 applied.** Commit
368963357 previously made `stale` non-terminal (regenerate instead of "Compaction did not apply"),
which made the path correct but left the warm result unusable.

## Fix

A summary describes the entries **before** its cut point, so growth after that cut cannot
invalidate it — only a rewrite of the summarized prefix can.
`CompactionPreparation.firstKeptEntryId` is that cut.

- New `warm-anchor.ts` predicate: the anchor still exists, the prefix ids are unchanged, and no
  newer compaction boundary landed behind it.
- `applyGeneratedCompaction` consults it when the revision moved, instead of rejecting outright.
- `ApplyCompactionOptions` gains `expectedFirstKeptEntryId`; the core compare-and-apply gate
  re-validates the same anchor before mutating the transcript. `expectedRevision` is untouched for
  back-compat.

## Verification

- **RED → GREEN**: `warm-summary-anchor.test.ts` failed `expected 2 to be 1` (warm summary
  discarded, a second summarization billed), then passed at one summarization.
- Safety case included: a warm summary whose prefix was rewritten by a landed compaction boundary is
  still rejected and regenerated.
- Compaction suite **342/342** (was 341; +2 new cases, 0 deleted). Root `npm run check` green.
- Real-CLI QA (`warm-summary-anchor-qa.mjs`, RPC + mock provider): **8/8**, compaction log records
  `speculative_applied` with no `speculative_stale`, one summarization, second reply delivered,
  real auth untouched.

### One pre-existing test asserted the defect

`speculative-compaction.test.ts` had "skips applyCompaction when message revision changes", whose
[...]

PR #862, opened 2026-08-13
fix(compaction): let the core route claim the idle warm summary

code-yeongyu:
## Problem

senpi pre-generates a summary while your session sits idle so you never wait for compaction. PR #853 taught the **extension** route to reuse it. The **core** route still threw it away.

The ordering is deterministic, not a race — on a new prompt `_enforceCompactionBeforeProvider` runs before `emitBeforeAgentStart`, so core always reaches compaction first, and its handler called `invalidateSpeculativeCompaction()` as its **first statement**. Measured with a file probe on the real CLI:

```
IDLE_TRIGGER
CORE_ROUTE_ENTER reason=threshold hasJob=true   # core arrives first, job alive
BLOCKING_ROUTE_ENTER hasJob=false               # already destroyed
```

So the warm-up was aborted and re-billed in exactly the case it exists for.

## Fix

Claim the job for the core route instead of invalidating it:

- **Atomic** — detached synchronously before any `await`, so exactly one route can own it.
- **Not aborted** — the controller is deliberately left alive; aborting is what destroyed the finished work.
- **Narrow** — the claim holds only for an automatic compaction with no custom instructions, a speculative origin, the same model identity, a valid `WarmAnchorSnapshot`, and a boundary equal to core's own `preparation.firstKeptEntryId`. Every other case falls through to the existing fresh generation, unchanged.

Manual compaction keeps its own instructions, a landed boundary still forces regeneration, and a mismatched cut is refused — each covered by a test.

## Verification

**Real-CLI A/B** (`core-route-warm-handoff-qa.mjs`, RPC + mock provider, sandboxed HOME) — a genuine defect reproduction:

| | baseline `3358d48ff` | this branch |
|---|---|---|
| result | **7/9** | **9/9** |
| `warm_consumed` | **0** | **1** |

- Unit: RED `expected 'fresh summary…' to contain 'warm summary…'` → GREEN; 4 cases incl. three safety paths.
- `test/compaction/` **346 passed** (344 before).
- Invariant files green: `speculative-budget-handoff`, `required-compaction-deterministic-fallback`, `warm-summary-anchor`, `stale-warm-blocking-repro`, `metadata-side-effects`, `lifecycle`, `idle-compaction`, `speculative-abort-cancellation`, `suite/compaction-race`, `suite/agent-session-compaction` — 88 tests.
- Root `npm run check` exit 0.

## Scope note

An `AgentSession`-level test cannot drive this: `shouldRunIdleCompaction` requires a persistent mode (tui/rpc/app-server) and the vitest harness runs in `print`, so the warm-up never starts there. The end-to-end proof is the RPC QA scenario above, which runs in a real persistent mode.

---
## Summary by cubic
Reuses the idle warm summary in the core route so users don’t wait for a second compaction. Previously core always invalidated the speculative job and regenerated; now it claims the job when safe, returns its result, and reliably aborts any unused claim, including on thrown exits.

- `session_before_compact` calls `claimWarmSummaryForCoreRoute()` before invalidation. The claim detaches the job synchronously, logs `warm_consumed` with `route: "core-route"` on success, and uses a try/finally to abort the controller unless the result is consumed (covers early returns and exceptions).
- Claim conditions: automatic compaction (no custom instructions), speculative origin, same model identity, valid warm anchor on the current branch, and a boundary equal to `preparation.firstKeptEntryId`. Otherwise it regenerates.
- Manual compaction, landed boundaries, and mismatched cuts still regenerate. Covered by `core-route-warm-handoff.test.ts`. Adds a real-CLI QA scenario to verify `warm_consumed` and that core does not regenerate.

<sup>Written for commit 4c91dfd0ce7349ee210a4cf18d65d48212d540cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/skills/senpi-qa/scripts/scenarios/core-route-warm-handoff-qa.mjs
+++ b/.agents/skills/senpi-qa/scripts/scenarios/core-route-warm-handoff-qa.mjs
@@ -1,0 +1,242 @@
+#!/usr/bin/env node
+
+import { createServer } from "node:http";
+import { existsSync, mkdirSync, readFileSync, readdirSync, writeFileSync } from "node:fs";
+import { join, resolve } from "node:path";
+import { createChecks, guardRealAuth, installCleanupHooks, makeSandbox, repoRoot } from "../lib/common.mjs";
+import { checkRealAuthUnchanged, hermeticEnv } from "../lib/mock-loop-support.mjs";
+import { TargetRpcClient } from "../lib/target-rpc-client.mjs";
+
+const CONTEXT_WINDOW = 128_000;
+const PINNED_INPUT_TOKENS = 120_000;
+const NEXT_PROMPT = "NEXT PROMPT";
+const NEXT_REPLY = "REPLY DELIVERED";
+const SUMMARY_MARKER = "IDLE SUMMARY";
+
+function arg(name, fallback) {
+	const index = process.argv.indexOf(name);
+	return index >= 0 ? process.argv[index + 1] : fallback;
+}
+
+function writeAnthropicSse(res, text, modelId, inputTokens) {
+	res.writeHead(200, {
+		"content-type": "text/event-stream",
+		"cache-control": "no-cache",
+		connection: "keep-alive",
+	});
+	const event = (type, data) => res.write(`event: ${type}\ndata: ${JSON.stringify({ type, ...data })}\n\n`);
+	event("message_start", {
+		message: {
+			id: `msg_${Date.now()}`,
+			type: "message",
+			role: "assistant",
+			model: modelId,
+			content: [],
+			stop_reason: null,
+			stop_sequence: null,
+			usage: { input_tokens: inputTokens, output_tokens: 0 },
+		},
+	});
+	event("content_block_start", { index: 0, content_block: { type: "text", text: "" } });
+	event("content_block_delta", { index: 0, delta: { type: "text_delta", text } });
+	event("content_block_stop", { index: 0 });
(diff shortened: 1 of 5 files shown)


code-yeongyu:
## `Test (workspaces + scripts)` is red on `main`, not from this PR

The failing assertion is in `scripts/prepare-senpi-bundled-workspaces.test.mjs:191` — an unexpected `which: '1.0.0'` in the bundled workspace dependency set:

```
+ actual - expected
  {
    '@earendil-works/pi-ai': 'npm:@code-yeongyu/senpi-ai@2026.7.22',
    'cross-spawn': '7.0.6',
+   which: '1.0.0'
  }
```

Reproduced on **pristine `origin/main` (3358d48ff)**, same machine, no changes from this branch:

```
npm run test:scripts
ℹ pass 138
ℹ fail 1
✖ which: '1.0.0'
```

This branch produces the byte-identical failure. Nothing in the diff touches `package.json`, the lockfile, or the bundling scripts — it changes one compaction handler, adds a test file, and adds a QA scenario.

The bundled-dependency drift deserves its own fix rather than being bundled into a compaction change.

PR #893, opened 2026-08-16
fix(coding-agent): preserve explicit skill invocation authority

### LINKED ISSUE #890: /skill:<name> invocation loses explicit-invocation authority, so Intent Gate routes on the trailing text and skill rules are ignored
## Summary

`/skill:<name> {input}` is documented as the explicit skill-invocation surface, but the expansion discards the fact that the user explicitly invoked the skill. The skill body and the user's trailing text are flattened into **one ordinary user message**, and the Intent Gate then routes on that trailing text — so the skill's declared rules/workflow are frequently ignored.

Reported in Discord (#chat-ko, 2026-08-14): "senpi에선 /skill:스킬 {user input} 실행이 스킬에 명시된 규칙이나 작업 방식대로 잘 동작하지 않네요".

## Root cause

`AgentSession._expandSkillCommand()` resolves the skill correctly:

```ts
// packages/coding-agent/src/core/agent-session.ts:3006-3008
const skill = skills.find((candidate) => candidate.name === skillName);
if (!skill) break; // Unknown skills and everything after them remain literal text.
```

but then emits only a passive container plus the raw args:

```ts
// packages/coding-agent/src/core/agent-session.ts:3021-3025
skillBlocks.push(
  `<skill name="${skill.name}" location="${skill.filePath}">\nReferences are relative to ${skill.baseDir}.\n\n${body}\n</skill>`,
);

// packages/coding-agent/src/core/agent-session.ts:3043-3047
const args = text.slice(tokenStart).trim();
const expandedSkills = skillBlocks.join("\n\n");
return args ? `${expandedSkills}\n\n${args}` : expandedSkills;
```

The result is sent as a single ordinary user message — no invocation flag, no separate directive role, no statement that the skill was explicitly selected.

Meanwhile the Intent Gate, which sits in the system prompt and therefore outranks message content in practice, instructs:

```
// packages/coding-agent/src/core/dynamic-prompt/intent-gate.ts:21
... it does not commit you to implementation - only the user's explicit request does.
// :26
Route by true intent, not surface form:
// :41
Re-read the latest user turn from scratch.
```

So the model derives intent from the trailing free text (`analyze this implementation`) and may downgrade or ignore the preceding skill block. There is **no** pre-skill IntentGate classifier (expansion happens first at `agent-session.ts:2733-2737`); the defect is the **loss of explicit-invocation authority during expansion**, not classifier ordering.

Corroborating precedent in-tree: the DeepSeek preset had to add an explicit authority clause because a model dismissed harness-injected directives as residual context —

```
// packages/coding-agent/src/core/extensions/builtin/prompt-preset/deepseek-v4.ts:41-47
Injected directives are binding: skill contents ... carry the same authority as words the user typed.
```

That workaround is deliberately DeepSeek-only (`prompt-preset/changes.md:27-30`), so every other model family still has the generic gap.

## Reproduction

1. Create a skill whose workflow visibly differs from a plain request (e.g. mandates a specific tool order / report format).
2. Run `/skill:test-workflow analyze this implementation`.
3. The prompt becomes exactly:

```xml
<skill name="test-workflow" location="...">
References are relative to ...

[skill rules]
</skill>

analyze this implementation
```

(this expansion shape is pinned by `packages/coding-agent/test/suite/agent-session-prompt.test.ts:147-191`)

4. The model answers the trailing request and does not follow the skill's declared workflow.

Note the existing test only asserts textual expansion — it does not assert that the skill is treated as an explicitly invoked binding workflow, which is why this is uncovered.

## Suggested fix location

`packages/coding-agent/src/core/agent-session.ts:3023-3047` (`_expandSkillCommand`)

Preserve the explicit invocation in the emitted payload: mark that the user explicitly selected the skill, and separate the binding skill instructions from the user's request (e.g. `<skill-instruction>` / `<user-request>` sections plus a one-line directive that the selected skill's workflow applies to this task unless it conflicts with higher-priority instructions).

Fixing it at dispatch is preferable to editing `dynamic-prompt/intent-gate.ts`, because model presets can replace the shared core (`dynamic-prompt/build.ts:72-90`); a dispatch-level marker reaches shared prompts, full-core presets, and custom system prompts alike. The change must not bypass the existing unknown-skill handling (`agent-session.ts:3007`) or the `MAX_SKILL_EXPANSIONS_PER_PROMPT` cap.

## Environment

Reproduced against current `main` (`packages/coding-agent/package.json` version 2026.8.14). Reported by a user on the OmO/senpi Discord.

code-yeongyu:
## Problem

`/skill:<name> {input}` resolved the selected skill but flattened its body and trailing arguments into one ordinary user message. That erased the fact that the user explicitly invoked the skill, so the Intent Gate could route from the trailing prose and ignore the selected workflow.

## Root cause

`AgentSession._expandSkillCommand()` emitted a passive `<skill>` block followed by raw free text. The provider-visible message carried neither an explicit invocation statement nor a boundary between binding skill instructions and the user's task.

## Fix

- State, by name, that the user explicitly invoked each expanded skill.
- Render every selected workflow inside its own `<skill-instruction>` section and mark it binding subject to higher-priority instructions.
- Render trailing arguments inside a separate `<user-request>` section.
- Preserve chained-skill ordering, duplicate suppression, unknown-skill fallthrough, and `MAX_SKILL_EXPANSIONS_PER_PROMPT` behavior.
- Keep transcript collapsing intact by teaching both `parseSkillBlock` copies (runtime/TUI and standalone HTML export) the current payload format.
- Parse current payloads first, with a legacy `<skill>` fallback for resumed/imported sessions and historical transcripts.

This ports the authority-preserving idea from oh-my-pi's `parseSkillInvocation` / `buildSkillPromptMessage` flow into Senpi's native path. Senpi has no custom-message pipeline here, so the fix shapes the emitted text payload rather than copying that implementation.

## Parser compatibility decision

Legacy payloads remain supported deliberately. Existing sessions and imported transcripts can contain the old `<skill name="...">...</skill>` form; dropping it would make those messages expand inline after resume/export. Both parser copies use new-format-first parsing and then the legacy fallback.

The formatter was extracted into `formatSkillInvocationPrompt`, and the round-trip test builds current payloads through that exact production formatter before executing both parser implementations. Chained payloads and legacy payloads are covered.

`packages/agent/src/harness/skills.ts` is a separate upstream agent-harness formatting API. Repository search found no coding-agent import or call into it; its own harness test is the only consumer in-tree. It remains on its established format and is out of scope. If its output is passed to coding-agent's public `parseSkillBlock` by an external consumer, the retained legacy fallback still parses it.

### Authority RED

```text
Test Files  2 failed (2)
Tests       8 failed | 15 passed (23)
```

Receipt: `local-ignore/qa-evidence/20260816-fix-890/red-tests.txt`.

### Parser RED

With production formatting extracted but parsers unchanged:

```text
FAIL test/export-html-skill-block.test.ts > round-trips the production skill invocation payload through both parsers
AssertionError: expected null to deeply equal { name: 'inspect', ... }
Test Files  1 failed (1)
Tests       1 failed | 5 passed (6)
```

Receipt: `local-ignore/qa-evidence/20260816-fix-890/parser-red-tests.txt`.

### GREEN

```text
Test Files  3 passed (3)
Tests       30 passed (30)
```

This covers parser round trips, exact prompt shape, chained/unknown/capped skill expansion, steering, and follow-up delivery. The CI-reported `UserPromptSubmit` test was also updated and passes focused.

Receipt: `local-ignore/qa-evidence/20260816-fix-890/parser-green-tests.txt`.

## Repository-wide old-format sweep

Both requested `rg` searches were run. Remaining hits are deliberate:

- Runtime and standalone HTML legacy parser fallbacks.
- The parser's explicit legacy compatibility test.
- `packages/agent`'s separate harness formatter and its own test.
- An unrelated compaction mutation fixture whose synthetic tag is not a skill-invocation payload and has no `location` attribute.

The stale hook assertion was changed to `<skill-instruction name="test" location="` while retaining its skill-body, request, and no-replacement assertions.

## Real-CLI QA

A hermetic senpi-qa mock-loop scenario ran the real TypeScript CLI against the localhost fake model with an isolated `authority-qa` skill. All six wire assertions passed: CLI completion, fake-model response, named explicit invocation, binding skill section, separate user request, and section order.

Evidence: `local-ignore/qa-evidence/20260816-fix-890/` (also copied to the main workspace receipt directory).

## Validation

Passed:

- `npm run build`
- `npm run check`
- `node scripts/check-pr-changelog.mjs --base $(git merge-base HEAD origin/main)`
- Focused coding-agent parser/prompt/composition tests: 30/30
- Focused `hooks-user-prompt-submit`: passed
- Agent harness formatting test: 2/2
- Real-CLI senpi-qa mock-loop wire assertion

Full coding-agent suite was run exactly as requested:

```text
Test Files  2 failed | 918 passed | 4 skipped (924)
Tests       2 failed | 7532 passed | 36 skipped (7570)
```

The two failures were unrelated bounded subprocess timeouts under full-suite load:

- `app-server-daemon.test.ts`: daemon initialize did not answer within 10 seconds.
- `inspector-endpoint-handoff.test.ts`: debugger handoff timed out.

Each failing file passed immediately when run once in isolation (4/4 and 2/2 respectively), confirming no skill-payload consumer failure remained. No unrelated timeout behavior was changed.

Fixes #890


code-yeongyu:
Nice, tight fix — the 2-line change in `_expandSkillCommand` is exactly the right minimal shape, the changes.md entry follows the fork format, and the RED capture is credible. One blocker before I can merge.

## Blocker: the new payload format breaks every consumer of `parseSkillBlock`

The emitted shape changed from `<skill name="..." location="...">…</skill>\n\n<args>` to a preamble sentence + `<skill-instruction …>` + `<user-request>`, but the parsers that read that shape were not updated:

1. `packages/coding-agent/src/core/agent-session.ts:194-203` — `parseSkillBlock()` matches the OLD format:
```ts
const match = text.match(/^<skill name="([^"]+)" location="([^"]+)">\n([\s\S]*?)\n<\/skill>(?:\n\n([\s\S]+))?$/);
if (!match) return null;
```
2. `packages/coding-agent/src/core/export-html/template.js:321-330` — a byte-identical duplicate of the same regex (plus the doc comment at :319 describing the old format).

Consumers that now silently get `null`:

- `packages/coding-agent/src/modes/interactive/interactive-mode.ts` — `const skillBlock = parseSkillBlock(textContent);` renders the collapsed skill representation in the TUI transcript.
- `packages/coding-agent/src/core/export-html/template.js` — two call sites, used for exported HTML transcripts.
- `packages/coding-agent/src/index.ts` re-exports `parseSkillBlock`, so this is also a public API behavior change.

**User-visible effect:** after this PR, every `/skill:<name>` invocation stops collapsing. Instead of the skill chip/summary, the user sees the full raw expansion — preamble sentence, the entire skill body, and the XML tags — inline in their transcript, and the same in exported HTML.

This did not show up in the test run because `packages/coding-agent/test/export-html-skill-block.test.ts` only asserts that the template *contains* the function, not that it parses a real current-format payload — so the format drift is uncovered.

## What I'd like

1. Update both parsers to the new format (name, location, content, and the `<user-request>` body), keeping them in sync — they are duplicated by design for the standalone HTML template.
2. Decide and state whether old-format messages must still parse. Resumed/imported sessions and previously exported transcripts contain the old shape, so I'd expect the parser to accept BOTH (new first, legacy fallback) rather than hard-switch. If you deliberately choose new-only, justify it in the PR body.
3. Strengthen `test/export-html-skill-block.test.ts` (or add a `parseSkillBlock` unit test) so it round-trips an actual current-format payload — build the input from the same helper the production path uses, so the next format change fails the test instead of silently degrading rendering. A test that only greps the template for a function name cannot catch this class of bug.
4. Please also confirm whether `packages/agent/src/harness/skills.ts` (which builds the old `<skill …>` block for the harness path) is intentionally left on the old shape — if that path also feeds `parseSkillBlock`, it needs the same treatment; if it is a separate upstream surface, say so and I'll accept it as out of scope.

Everything else — minimal production diff, changelog entry with the issue link, changes.md with merge-conflict zones, RED→GREEN evidence, real-CLI wire assertions — looks good, and I'll re-review quickly once the parser round-trip is covered.

code-yeongyu:
Addressed both review blockers in 714e0c41a.

- Updated runtime and standalone HTML `parseSkillBlock` copies for new-format-first parsing with a legacy fallback.
- Added production-formatter round-trip coverage for both parsers, chained payloads, and legacy sessions (captured RED before parser changes).
- Updated the stale `UserPromptSubmit` assertion and swept all remaining `<skill name=` hits deliberately.
- Confirmed `packages/agent/src/harness/skills.ts` is a separate upstream harness surface with no coding-agent consumer in-tree; legacy fallback still accepts it if externally passed to `parseSkillBlock`.
- Full coding-agent suite: 7,532 passed, 36 skipped, 2 unrelated subprocess timeout failures; both failing files passed once in isolation (4/4 daemon, 2/2 inspector).
- `npm run check`, build, changelog gate, focused tests, and refreshed real-CLI QA are green.

Detailed receipts and the legacy decision are in the updated PR body.

code-yeongyu:
## Re-review (round 2) — blocker resolved, approving pending CI

I verified the follow-up commit `714e0c41a` against everything I raised. All four asks are satisfied:

**1. Both parsers updated and kept in sync.** `parseSkillBlock` in `core/agent-session.ts` and the standalone duplicate in `core/export-html/template.js` now share identical logic: new-format pattern first, with a name-consistency guard (`instructionMatch[1] !== instructionMatch[2] → null`), chained-skill consumption, and a strict `<user-request>` tail match that rejects unparsable remainders instead of silently truncating.

**2. Legacy compatibility retained** — the old `<skill name=… location=…>` regex is preserved as a fallback, so resumed/imported sessions and previously exported transcripts still render. This was the right call over a hard switch.

**3. The round-trip test is genuinely non-tautological.** `test/export-html-skill-block.test.ts` imports the production formatter and builds its input from it:
```ts
import { formatSkillInvocationPrompt, type ParsedSkillBlock, parseSkillBlock } from "../src/core/agent-session.ts";
const payload = formatSkillInvocationPrompt([skill], "Check errors.");
```
That is exactly what I wanted — a future format change now fails this test instead of silently degrading TUI/HTML rendering. Extracting `formatSkillInvocationPrompt` as the single source of truth for the payload shape is a better fix than the string duplication I would have accepted.

**4. Repo sweep done**, with the remaining old-format hits individually justified (legacy fallbacks, the explicit legacy test, the `packages/agent` harness surface, and an unrelated compaction fixture). Your reasoning that `packages/agent/src/harness/skills.ts` is a separate upstream surface with no in-tree coding-agent consumer — and that the legacy fallback would still parse its payload anyway — is sound; I accept it as out of scope.

**CI failure fixed:** the `hooks-user-prompt-submit.test.ts:197` assertion now expects `<skill-instruction name="test" location="` while keeping its other assertions intact.

**Full-suite result accepted:** 7532 passed / 2 failed, both unrelated bounded-subprocess timeouts (`app-server-daemon`, `inspector-endpoint-handoff`) that pass in isolation. No skill/parser/hook test failed. That matches the known load-sensitive behavior of those two files and is not attributable to this change.

Approving on content. I will merge with a merge commit once the required checks (`Check and test`, `Changelog gate`) go green — changelog gate is already passing.

PR #920, opened 2026-08-17
fix(mcp): open the browser and keep the authorization URL visible for /mcp auth

ismetanin:
## Problem

`/mcp auth <server>` can never complete an interactive OAuth login. It prints

```
Opening browser to authorize sentry...
```

and then hangs forever with no further output.

Two bugs compound:

1. **No browser is ever launched.** `auth/commands-auth-dispatch.ts` wired the `openBrowser` dependency to `ctx.ui.notify(...)`, so the announcement was simply untrue. The repo already ships a real launcher at `src/utils/open-browser.ts` (used by the provider login dialog).
2. **The fallback URL is erased.** The URL notification is emitted from `redirectToAuthorization` *inside* `beginAuthorization`, and the `Opening browser…` line was emitted immediately after it. Consecutive status notifications coalesce in the TUI — `showStatus` rewrites the trailing status text in place when the last two chat children are the previous status spacer/text — so the second line overwrote the URL milliseconds after it appeared.

With no browser and no reachable authorization URL, `runInteractive` parks on `channel.waitForCode()` until the session ends.

## Fix

- `auth/commands-auth-dispatch.ts` — `openBrowser` calls `utils/open-browser.ts` for real.
- `auth/commands-auth.ts` — each branch of `runInteractive` emits exactly one notification, carrying the authorization URL, so a failed launch still leaves a copy-pasteable link. Behavior for the `callbackUrl`-override (paste) branch is unchanged apart from the merged message.
- `builtin/mcp/changes.md` — fork ledger entry.

No config, transport, or token-storage behavior changes.

## Tests

`test/mcp/oauth-callback.test.ts` gains an assertion that the single announcement contains the authorization URL, for both the loopback and the callback-override branch. Both assertions fail on `main` and pass with the fix (verified by reverting the two source files with the test in place).

## Verification

| Gate | Result |
|---|---|
| `npm run check` | exit 0 |
| `test/mcp/oauth-callback.test.ts`, `oauth-headless.test.ts`, `auth-modes.test.ts` | 33 passed |
| `senpi-qa` Channel 4 (`cli-smoke.mjs --self-test`) | 8/8 passed, real auth unchanged |
| `senpi-qa` flow driver (evidence `20260817-mcp-oauth-browser-launch`) | 5/5 passed |

The QA driver runs `runAuth` from source against the repo's fake IdP fixture, with a recording stub named `open` first on `PATH` so no real browser opens:

```
{
 "interactive flow completes (token stored, no hang)": true,
 "single announcement before the authorized message": true,
 "announcement carries the authorization URL": true,
 "browser launcher exec'd with the URL": true,
 "real auth.json untouched": true
}

--- notification as the user sees it ---
Opening browser to authorize fix. If it does not open, visit:
<authorization-url>
```

Note: `test/mcp/oauth-race.test.ts` fails identically with and without this change in my environment (`@earendil-works/pi-tui` workspace dist not built); unrelated to this PR.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -12,6 +12,12 @@
 
 ### Fixed
 
+- `/mcp auth <server>` now completes an interactive OAuth login instead of stalling forever. The command announced
+  `Opening browser to authorize <server>...` but never launched a browser, and the notification carrying the
+  authorization URL was emitted immediately before that announcement, so the TUI's consecutive-status coalescing
+  overwrote it. With no browser and no reachable URL, the flow waited on its loopback callback until the session
+  ended. The auth command bridge now calls the real browser launcher, and each branch emits a single notification
+  that includes the authorization URL as a fallback for when the launch fails.
 - Steering queued while a provider stream-start timeout retry is running now starts automatically when that managed
   retry exhausts its budget, instead of remaining parked until another user prompt. Generic terminal provider errors
   and user-aborted retries keep their existing queue-retention behavior
--- a/packages/coding-agent/src/core/extensions/builtin/mcp/auth/commands-auth-dispatch.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/mcp/auth/commands-auth-dispatch.ts
@@ -1,3 +1,4 @@
+import { openBrowser as launchBrowser } from "../../../../../utils/open-browser.ts";
 import type { ExtensionAPI, ExtensionCommandContext } from "../../../types.ts";
 import { createMcpLogger } from "../log.ts";
 import type { McpService } from "../service.ts";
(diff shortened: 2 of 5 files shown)


code-yeongyu:
Thank you @ismetanin for identifying this and for the original fix. It was superseded by #2037, merged as `f691462052ffa9d8515d1371e0b8255214637e76`, which lands the same intent on current `main`:

- `/mcp auth` now opens the browser through the existing launcher.
- The authorization URL is kept as a persistent, display-only entry instead of a transient notification, because a later status message was still overwriting it.
- The manual flow keeps working when the browser cannot be opened.

That PR adds regression tests and real TUI verification. Closing this one as superseded. This fixes code-yeongyu/oh-my-openagent#6724.

PR #947, opened 2026-08-18
fix(compaction): preserve replay-safe Gemini signed state in fallback recovery

Indosaram:
## Summary

Improve deterministic compaction fallback recovery when the retained conversation contains legitimate Gemini opaque provider state (`thoughtSignature` on tool calls, `thinkingSignature`, `textSignature`, and empty visible text blocks).

## Problem

When deterministic compaction recovery ran on long Gemini sessions (e.g. after summarization timeout or stream truncation):
1. `retained-message-safety.ts` previously rejected any assistant content bearing `thoughtSignature`, `thinkingSignature`, or `textSignature` unconditionally.
2. Valid Gemini assistant turns often contain tool calls with `thoughtSignature` and no visible text, or empty thinking/text blocks with encrypted thought context necessary for same-model replay.
3. This unconditional rejection forced deterministic recovery to fail completely even when a safe, bounded candidate was available.

## Solution

1. **Provider Signature Validation without Global Weakening**:
   - Replaced unconditional rejection of provider signatures with structural format validation (`isValidProviderSignature`: bounded length, base64 / 4-byte aligned).
   - Valid signed tool calls, thinking blocks, and text blocks are safely preserved for same-model replay.
   - Genuinely malformed signatures or corrupted payloads still fail closed.
2. **Atomic Tool-Chain Integrity**:
   - Added `hasValidToolChains()` verification ensuring that retained messages never orphan tool results without their preceding tool calls.
3. **Earlier Safe-Boundary Recovery**:
   - If the initially prepared boundary cuts through an atomic tool/reasoning chain, `createRequiredCompactionFallback` searches backward for an earlier safe boundary before falling back to the latest user turn.
4. **Structured Rejection Diagnostics**:
   - Structured rejection reasons (`missing-preparation-boundary`, `unsafe-retained-content`, `atomic-tool-chain-cut`, `retained-token-budget-exceeded`, `context-reconstruction-failed`) without exposing signature values.
5. **Cross-Model Replay Safety**:
   - Cross-model conversions continue to strip provider-specific signatures via `transformMessages`, ensuring incompatible signatures are never replayed into other providers.

## Regression Coverage

Added comprehensive regression tests in `packages/coding-agent/test/compaction/required-compaction-deterministic-fallback.test.ts`:
- **Case A**: Tool-call-only turn with `thoughtSignature` and empty visible text retained without dropping state.
- **Case B**: Empty signed text/thinking parts preserved without discarding required replay state.
- **Case C**: Sequential function calling chains preserved without invalid cuts.
- **Case D**: Earlier safe boundary recovery when initial cut divides tool call and result.
- **Case E**: Fail-closed on genuinely unsafe / malformed provider signatures.
- **Case F**: Clean rejection on budget exhaustion without unbounded search.
- **Case G**: Cross-model handoff transforms Gemini signed state correctly.

## Verification

Ran:
```bash
npm test test/compaction/required-compaction-deterministic-fallback.test.ts
# 19 pass, 0 fail (142 expect() calls)
```

---
## Summary by cubic
Fixes deterministic compaction fallback so valid long sessions with provider-signed Gemini state recover instead of failing, and closes replay-validation gaps. Provider signatures (`thoughtSignature`, `thinkingSignature`, `textSignature`) are now format-validated (bounded base64 for Google, bounded opaque strings otherwise) instead of rejected unconditionally; valid signed blocks are kept for same-model replay while malformed or replay-unsafe content still fails closed.

- Retain empty signed text/thinking blocks and require a valid signature for redacted thinking.
- Reject orphaned, reversed, duplicate, or incomplete tool-call/result chains to keep atomic tool integrity.
- Walk backward to an earlier safe boundary when the prepared cut bisects a tool chain before falling back to the latest user turn.
- Compute retention and token checks in one linear scan so fallback stays bounded even on 10,000-entry malformed histories.
- Emit structured rejection diagnostics and candidate counters without exposing signature payloads.
- Strip provider-specific signatures via `transformMessages` to keep cross-model replay safe.

<sup>Written for commit f88b4d1da95f73a65613a6c9262c9ff5c95fab39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/947?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/coding-agent/src/core/extensions/builtin/compaction/retained-message-safety.ts
```diff
@@ -28,37 +28,56 @@ function isUsage(value: unknown): boolean {
...
+
 function hasSafeAssistantContent(content: unknown): boolean {
 	if (!Array.isArray(content)) return false;
 	for (const block of content) {
 		if (!isRecord(block) || typeof block.type !== "string") return false;
 		switch (block.type) {
 			case "text":
-				if (typeof block.text !== "string" || block.textSignature !== undefined) return false;
+				if (typeof block.text !== "string") return false;
+				if (block.textSignature !== undefined && !isValidProviderSignature(block.textSignature)) {
```
code-yeongyu: B1. This validator has no provider/model context, so it applies a Google-specific base64 assumption to every provider's `textSignature`. Current main's type contract explicitly allows OpenAI legacy ids and `TextSignatureV1` JSON here, while Anthropic/Kimi histories also use non-base64 opaque values. A later required fallback on those sessions will reject the retained suffix before `transformMessages` gets its chance to do the provider-aware conversion. Limit this check to Google signatures or validate at the actual provider replay boundary.

packages/coding-agent/src/core/extensions/builtin/compaction/deterministic-fallback.ts
```diff
@@ -176,22 +221,62 @@ export function createRequiredCompactionFallback(
...
 	const prepared = projectCandidate(preparation.firstKeptEntryId, "prepared");
-	if (prepared) return prepared;
+	if (prepared) {
+		if (diagnostics) diagnostics.candidatesChecked = candidateCount;
+		return prepared;
+	}
 
+	// 2. If prepared boundary cut an atomic signed chain or was rejected, check if an earlier safe boundary exists
+	// Scan backward up to 5 entries before prepared boundary to find complete chain boundary
+	const minBoundary = Math.max(0, preparedBoundaryIndex - 5);
```
code-yeongyu: B2. This five-entry cap misses legitimate chain boundaries. With one assistant turn containing six calls and six following results, a prepared cut at the sixth result fails because the declaring assistant is omitted, but this loop inspects only the five preceding result entries and never reaches it. The fallback then needlessly drops to a later user boundary or cancels. Find the declaring boundary from the chain rather than bounding the scan by an arbitrary number of entries.

packages/coding-agent/src/core/extensions/builtin/compaction/deterministic-fallback.ts
```diff
@@ -42,6 +55,33 @@ function hasMeaningfulUserText(entry: SessionEntry): boolean {
...
+
+	for (const msg of messages) {
+		if (!isRecord(msg)) continue;
+		if (msg.role === "assistant" && Array.isArray(msg.content)) {
+			for (const block of msg.content) {
+				if (isRecord(block) && block.type === "toolCall" && typeof block.id === "string") {
+					declaredCalls.add(block.id);
+				}
+			}
+		} else if (msg.role === "toolResult" && typeof msg.toolCallId === "string") {
```
code-yeongyu: B3. This only proves that a result id was declared at some earlier point; it does not prove a one-to-one pairing. `resolvedCalls` is populated but never checked, so duplicate results pass. `assistant(toolCall X), toolResult X, toolResult X` reaches `transformMessages`, which consumes the first result and emits the second extra result, a payload providers can reject. Reject duplicate/unmatched results and explicitly define incomplete-call handling.

packages/coding-agent/test/compaction/required-compaction-deterministic-fallback.test.ts
```diff
@@ -455,3 +473,333 @@ describe("required compaction deterministic fallback", () => {
...
+
+		// Transforming to Anthropic target model
+		const targetModel = {
+			id: "claude-sonnet-4",
+			provider: "anthropic",
+			api: "anthropic-messages" as const,
+			input: ["text" as const],
+		};
+
+		const transformed = transformMessages([geminiAssistant, toolRes], targetModel as never);
```
code-yeongyu: B4. This is not a fallback replay test: it bypasses `createRequiredCompactionFallback`, the session compaction event, the context hook, and the Google converter, and it hides the model type with `as never`. The test only demonstrates that generic `transformMessages` strips one tool-call field. Add an end-to-end fixture that applies the fallback and captures the actual Google/Vertex request, plus a non-Gemini signed history case.

code-yeongyu (changes requested):
REQUEST_CHANGES

This is not mergeable or replay-safe as written. The added tests mostly prove that synthetic objects survive a pure projection; they do not establish the provider-level guarantee claimed in the PR.

B1. `isValidProviderSignature` is applied globally to every provider's `textSignature`, `thinkingSignature`, and `thoughtSignature`, but those fields are not a shared base64 protocol. The current `packages/ai/src/types.ts` explicitly permits OpenAI `textSignature` to be a legacy id or `TextSignatureV1` JSON, and the current main tests exercise non-base64 Anthropic/Kimi/OpenAI signatures. A non-Gemini session that later hits a timeout/stream-truncation/overflow fallback will now reject an otherwise replayable retained assistant message and cancel compaction. This must be provider-aware at the replay boundary (or limited to Google fields); a provider-agnostic base64 rule is a regression, not fail-closed validation.

B2. The advertised atomic-chain recovery is arbitrarily capped at five entries. For one assistant turn with six calls followed by six tool results, a prepared boundary at the sixth result has no declaring assistant in the retained suffix; this loop checks only the preceding five results and misses the real safe boundary six entries away. The code then falls through to a later-user candidate or cancellation even though a valid complete boundary exists. Search based on the actual chain/declaration boundary, not a magic entry count.

B3. `hasValidToolChains` does not validate pairing despite its name and comment: `resolvedCalls` is populated but never consulted. Duplicate tool results therefore pass validation. For `assistant(toolCall X), toolResult X, toolResult X`, the current `transformMessages` consumes the first result and emits the second one as an extra result for the same call; providers can reject that request. The validator must reject duplicate/unmatched results (and define how incomplete calls are handled) rather than merely checking that an id appeared somewhere earlier.

B4. The seven new cases are not an integrated replay test. Case A only inspects `buildSessionContext()` after appending a compaction; Case B only asserts that a result is defined; and Case G calls generic `transformMessages` with a hand-cast `targetModel`, without running the fallback through `session_before_compact`, the current compaction context hook, and the actual Google converter. These fixtures do not catch provider/model identity mistakes, Google turn construction, signature placement, or the non-Gemini regression in B1. The PR body therefore overstates its 19-pass regression coverage.

B5. The PR is currently unmergeable: GitHub reports `mergeStateStatus: DIRTY`, and `git merge-tree pr-947-work origin/main` reports content conflicts in `packages/coding-agent/CHANGELOG.md` and `packages/coding-agent/src/core/extensions/builtin/compaction/changes.md`. This branch is also stale against the checked-out `origin/main` (`f97db40079d6...`): main has since changed the compaction index/speculative pipeline, transport replay semantics, context filtering, and the compaction test contract. Rebase and rerun the review/test surface against current main; do not treat the branch-local test result as validation of the merge result.

B6. The required CI checks are not present. `gh pr checks 947` shows only the skipped Cubic reviewer and passing GitGuardian; neither `Check and test` nor `Changelog gate` has run for this PR. The local changelog script passes against the PR base, but that is not a substitute for the required check.

B7. The release changelog entry links this PR's change to `#944`, while the reviewed PR is `#947`. Fix the shipped link before merge; otherwise users are sent to the wrong change.

Static verification performed without installing dependencies (per review instructions): inspected PR metadata/diff and current `origin/main`, ran `git merge-tree`, `git diff --check`, and the local `scripts/check-pr-changelog.mjs` (passes). No files, commits, or refs were modified.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Addressed review blockers B1-B7 in pushed commit `bf2d9ed26b6e98d1067503d3ff1073d376688939`.

- B1: base64 signature validation is now limited to Google/Google Vertex assistant histories; OpenAI/Anthropic/Kimi opaque signatures remain replayable.
- B2: fallback boundary recovery scans back to the actual chain boundary with no five-entry cap.
- B3: retained tool validation now rejects duplicate, unmatched, duplicate declarations, and incomplete call/result pairings.
- B4: added real fallback-path coverage through `session_before_compact`, persisted session compaction/context reconstruction, and the typed Google converter; removed the `as never` target-model cast.
- B5: merged `origin/main` and resolved both changelog conflicts while preserving both entries.
- B7: corrected the release link from PR #944 to PR #947.

Verification: targeted fallback suite 22/22, coding-agent build passed, TypeScript/LSP diagnostics clean for changed source/tests, `git diff --check` passed, and `check-pr-changelog.mjs --base origin/main` passed. The broader review-hardening suite had one unrelated current-main timeout in the existing window-shrink test; no fixed sleeps or test timing changes were added. GitHub CI checks should now run for the new head.

code-yeongyu (changes requested):
REQUEST_CHANGES

Round-2 adversarial review of fixed head `bf2d9ed26b6e98d1067503d3ff1073d376688939`.

Round-1 blocker resolution:

- B1 — VERIFIED. `hasSafeAssistantEnvelope()` derives `providerIsGoogle` from each retained assistant's own `provider` and only applies the base64 check for exact `google`/`google-vertex`. Mixed-provider histories and provider switches therefore validate Google-origin signatures but do not run the validator on OpenAI/Anthropic/Kimi-origin signatures. The new non-Gemini fallback test also passes with legacy/opaque signatures.
- B2 — VERIFIED for correctness. The five-entry cap is gone. The backward walk is finite over `branchEntries`, stops at the preceding compaction boundary, and accepts a candidate only after rebuilding context and validating the complete retained suffix. The six-call/six-result regression passes.
- B3 — PARTIALLY RESOLVED, remains a blocker below. Duplicate declarations, duplicate results, unmatched results, and incomplete unpaired calls are covered by the new logic/tests. However, `hasValidToolChains()` never checks `block.incomplete`; an `incomplete: true` tool call with a matching result is accepted, and `transformMessages()` does not drop it.
- B4 — VERIFIED for the requested route coverage. Case A now invokes `session_before_compact`, receives the fallback, appends the compaction, rebuilds session context, and passes the resulting messages through the typed Google converter. The direct duplicate and non-Gemini cases would fail on the corresponding regressions. The targeted fallback file passes 22/22.
- B5 — VERIFIED relative to the PR base. The merge commit has `f97db4007` (the then-current `origin/main`) as its second parent; `git merge-tree --write-tree origin/main HEAD` is conflict-free, and both the coding-agent changelog content and compaction `changes.md` entries are present.
- B6 — RESOLVED as originally stated: the required checks are now present and have run. They are not green because of fresh blocker B8 below.
- B7 — VERIFIED. The release entry links `#947`, not `#944`.

New blockers:

B8. The merge commit committed a lockfile from the old first-parent dependency state instead of the package manifests from the merged main. `package-lock.json` now downgrades/removes dependencies that `package.json` requires (for example `@anthropic-ai/claude-agent-sdk` 0.3.220 vs 0.3.241, `@anthropic-ai/sdk` 0.91.1 vs 0.120.0, root `@biomejs/biome` 2.5.5 vs 2.5.10, and it omits `concurrently`). GitHub CI consequently fails at `npm ci` in all jobs with package-lock/package.json out-of-sync errors; local `npm run check:install-lock:coding-agent` also fails. This is a merge-blocking install failure introduced by this head, not a test-environment issue. Regenerate/repair the lockfile against the merged manifests.

B9. The uncapped fallback search is quadratic (and repeatedly serializes/rebuilds the retained context). Every iteration of the backward loop calls `projectCandidate()`, which calls `buildSessionContext([...branchEntries, syntheticCompaction])` over the full history. With a malformed Google signature at the final entry to force every candidate to fail, I measured approximately 6.3s for 5,000 entries and 21.7s for 10,000 entries on the review M4 Pro. Required compaction recovery can therefore add multi-second/tens-of-seconds latency on large sessions precisely on an error path. Scan to the structural chain boundary once, or otherwise avoid rebuilding the full context for every candidate.

B10. Incomplete tool calls with a matching result are not rejected. The retained shape `assistant(content: [{type: "toolCall", id: "x", incomplete: true}]), toolResult(toolCallId: "x")` passes `hasUnsafeRetainedContent()` and `hasValidToolChains()` because the ID is declared and resolved exactly once. `transformMessages()` then preserves the call/result pair. The type contract says carriers of `incomplete` MUST NOT be executed, so fallback must fail closed (or explicitly sanitize) whenever a retained call is marked incomplete, regardless of whether an adversarial history also supplies a result.

Additional verification:

- The claimed pre-existing timeout is real: running the exact `blocking-compaction-review-hardening.test.ts` from `origin/main` at `f97db40079d6a2cb3e1e383586c551e90b6ee049` reproduces the same 5,000ms failure in the existing window-shrink test. The patched head reproduces it too, while the separate model-window-shrink test passes on both base and head.
- Orphan tool results fail closed in the fallback validator rather than being treated as a safe boundary.
- `git diff --check` and LSP diagnostics for the changed source/tests are clean; changelog gate passes. GitHub's current check run is red at dependency installation, so the targeted local test result is not sufficient to accept this head.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-2 blockers B8-B10 are resolved in `6beefbcaaadd7ddc37ac5bef704c7d1c998c93fb`.

- **B8 lockfile mismatch:** regenerated the root `package-lock.json` with `npm install`, then regenerated `packages/coding-agent/install-lock` and `publish-deps.lock.json`. A clean `rm -rf node_modules && npm ci` succeeded: 467 packages added, 486 audited. `check:install-lock:coding-agent` and `check:shrinkwrap` both pass.
- **B9 quadratic fallback search:** deterministic fallback now projects the session context once and performs suffix token/safety/tool-chain validation incrementally, avoiding full context rebuild/serialization per candidate. Added a synthetic 10,000-entry malformed-history regression with a 10s upper bound.
- **B10 incomplete paired calls:** tool-chain validation now rejects `incomplete: true` calls even when a matching tool result is present. Added the failing-first regression for that exact shape.

Evidence:
- Failing-first: the new B10 test accepted the pair, and the 10,000-entry test measured ~19.2s before the fix.
- Green: targeted fallback suite 24/24; coding-agent build/typecheck passes; changelog gate passes; generated lock checks pass; Biome/LSP/diff checks pass; repository pre-commit gates pass, including npm/bun/pnpm install/build verification.
- The separate existing blocking-compaction hardening suite remains 6/7 because its documented pre-existing window-shrink test hits the 5s timeout; this change does not touch that path.

code-yeongyu:
WORKING: review-947-r3 - inspecting round-2 fixes, lockfile drift, and adversarial fallback behavior

code-yeongyu (changes requested):
Round-3 adversarial review of fixed head `6beefbcaaadd7ddc37ac5bef704c7d1c998c93fb`.

Round-2 resolution:

- B9 (quadratic fallback search) — VERIFIED. `createRequiredCompactionFallback()` now builds one projected context and precomputes suffix safety/token totals plus tool-chain ranges. The targeted fallback suite passes 24/24. The 10,000-entry malformed-history test is a real upper-bound regression (`< 10s`); the pre-fix measurement from round 2 was approximately 19.2s on this review machine, so an O(n^2) rebuild would fail that assertion.
- B10 (paired incomplete calls) — VERIFIED. The incremental scan records `block.incomplete === true` and invalidates the range through its matching result, so the exact incomplete-call-plus-matching-result fixture is rejected. The targeted test passes. There is no separate `incomplete` field on `toolResult` in the message contract; the call-side carrier and paired result shape are covered.
- B8 (root lockfile manifest mismatch) — PARTIALLY RESOLVED, but the generated installer lock remains invalid. A clean root `npm ci --ignore-scripts --no-audit --no-fund` succeeds, and the generated-lock consistency scripts pass. However, `packages/coding-agent/install-lock/package.json` declares the existing `@hono/node-server: 2.1.1` override while its generated lock still records `2.0.10`; `npm ci --dry-run` from that installer directory fails with `Invalid: lock file's @hono/node-server@2.0.10 does not satisfy @hono/node-server@2.1.1`. Origin/main records 2.1.1.

Fresh blockers:

B11. Rolldown lock drift still fails CI. On this head, `node_modules/rolldown` and every native optional in its `optionalDependencies` are locked at 1.2.6 (including a new `binding-android-arm-eabi` entry), while `scripts/rolldown-platform-lock.test.mjs` explicitly pins 1.2.4. Running the script on this head fails `1.2.6 !== 1.2.4`; running the identical script against `origin/main` passes, so this is not pre-existing. The CI run `33253448115` fails the same assertion. Realign root, installer, and publish lockfiles to main's 1.2.4 Rolldown graph while keeping the clean-install proof. The regeneration also introduced unrequested transitive drift versus main (`@csstools/css-color-parser` 4.2.1, `@csstools/css-syntax-patches-for-csstree` 1.1.9, `@oxc-project/types` 0.147.0, `tldts`/`tldts-core` 7.4.11, and Vite's nested `picomatch` 4.0.7); please avoid carrying that churn unless it is required.

B12. The suffix tool-chain index can falsely accept a retained call with no retained result. In `deterministic-fallback.ts`, the one-call/one-result reversed-order branch at lines 226-230 calls `markInvalidRange(0, resultIndex + 1)` when `callIndex >= resultIndex`. For an adversarial history containing `toolResult(id)` at index 1 followed by its declaring assistant `toolCall(id)` at index 2, a candidate starting at the call index has `toolChainValidAt[2] === true`, so fallback returns the prepared boundary even though the retained suffix contains an unresolved call. The previous `hasValidToolChains()` would reject this shape. The invalid range must cover the call side as well, or pairing must be evaluated against each suffix. This is a stale-index false accept, not merely a conservative false reject; the direct in-memory harness reproduced the accepted prepared candidate.

B13. The provider-aware validation introduced a non-Google fail-open regression. For non-Google assistants, `hasSafeAssistantContent()` now accepts `redacted: true` without a string `thinkingSignature`, and accepts non-string `thinkingSignature`/other signature fields because type validation is gated only on `providerIsGoogle`. A same-model `transformMessages()` preserves these blocks: Anthropic conversion emits `redacted_thinking` with undefined data for the first shape, while a non-string non-redacted thinking signature reaches `.trim()` and can throw. The previous validator rejected redacted thinking and all signature-bearing blocks, and the requested B1 fix only required avoiding Google's base64 rule for other providers. Keep non-Google opaque signatures as bounded strings, and fail closed for malformed redacted state instead of allowing an invalid provider payload or converter exception.

Other checks: the changed source files have no LSP diagnostics; `git diff --check` is clean; the targeted fallback and related replay tests pass. I found no separate superlinear index allocation: the new arrays/maps are linear in the projected history, but B12 must be fixed before that optimization is safe.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-3 blockers B8/B11-B13 resolved in commit `5ad4d2179`.

- B8: regenerated `packages/coding-agent/install-lock/package-lock.json` consistently with the `@hono/node-server` 2.1.1 override. `npm run check:install-lock:coding-agent` passes and installer-directory `npm ci --dry-run --ignore-scripts --no-audit --no-fund` passes.
- B11: realigned root, installer, and publish dependency lockfiles to `origin/main`'s Rolldown 1.2.4 graph, removing the unrequested transitive churn. `node scripts/rolldown-platform-lock.test.mjs` passes; root `npm ci --ignore-scripts --no-audit --no-fund` passes.
- B12: fallback tool-chain validation now requires the matching result to occur later than its call; reversed result-then-call failing-first coverage was added.
- B13: all provider signatures must be bounded strings; Gemini signatures retain base64 validation, non-Google opaque signatures remain valid, and redacted thinking requires a bounded signature. Tests cover non-string signatures and unsigned redacted thinking failing closed.

Evidence: targeted fallback suite 26/26; platform-lock test 1/1; coding-agent build/typecheck clean; LSP diagnostics clean; changelog gate against `origin/main` PASS; full pre-commit `npm run check` PASS including npm/bun/pnpm install+build verification.

code-yeongyu:
WORKING: review-947-r4 - validating ordering, provider signature replay, converter shapes, and lockfile alignment

code-yeongyu (changes requested):
Round-4 adversarial review of fixed head `5ad4d217988e8cd17da585a781be3e3bcb0c07b7`.

Resolution verdicts:

- B8 (install-lock @hono/node-server): VERIFIED. `packages/coding-agent/install-lock/package.json` declares 2.1.1 and its lock records 2.1.1; `npm ci --dry-run --ignore-scripts --no-audit --no-fund` succeeds from that directory. `check:install-lock:coding-agent` also passes.
- B11 (Rolldown/transitive lock drift): VERIFIED. The relevant graph in `package-lock.json`, `packages/coding-agent/install-lock/package-lock.json`, and `packages/coding-agent/publish-deps.lock.json` is byte-identical to the corresponding `origin/main` entries for Rolldown, all native bindings, @oxc-project/types, the cited CSSTools packages, tldts, and picomatch. The PR has no lockfile diff beyond its own requirements; `node scripts/rolldown-platform-lock.test.mjs` and `check:shrinkwrap` pass.
- B12 (reversed/duplicate tool ordering): VERIFIED for the reported complete-call cases. Direct in-memory checks reject result-before-call boundaries and duplicate/interleaved-ID histories, including boundaries at the call and result sides. Converter inspection also found no legitimate result-before-call replay shape: `transformMessages` only moves a delayed result that is already after its declaring call; a true result-before-call remains an orphan and Google conversion would serialize the function response before the model function call, so rejecting it in fallback is correct.
- B13 (provider signature validation): VERIFIED for the requested states. Direct checks accept bounded opaque Anthropic, Kimi, and OpenAI signatures; reject a non-string thinking signature and redacted thinking without a signature; and the targeted fallback suite covers valid Gemini base64 signatures, empty signed blocks, invalid Gemini signatures, and cross-model stripping. Malformed rejection is not a silent drop: the production handler returns a cancellation reason through `compaction_end.errorMessage`, although the optional detailed diagnostic object is not passed at the sole production call site.
- Performance: no new blocker found. The incremental suffix/index construction is linear; the 10,000-entry regression remains under its 10-second bound. The targeted fallback suite passes 26/26.

Fresh blocker:

B14. Incomplete reversed-order chains still pass the fallback admission check. In `deterministic-fallback.ts` at lines 226-230, the `call.incomplete` branch runs before the ordering branch and marks only `[0, resultIndex + 1)`. For a history shaped as `toolResult(id)` followed by `assistant(toolCall(id, incomplete: true))`, a candidate beginning at the call has `callIndex > resultIndex`, but `toolChainValidAt[callIndex]` remains true because the invalid range ended before the call; the reversed-order branch is never reached. A direct in-memory reproduction on this head returned the prepared fallback (`retainedSuffix: "prepared"`, `estimatedTokensAfter: 872`) with the incomplete call retained and no later result. That violates both the incomplete-call fail-closed invariant and B12's requirement that the matching result be later than the call. Mark the invalid range through `max(callIndex, resultIndex) + 1`, or otherwise apply the later-result check before accepting the incomplete branch, and add this combined regression.

VERDICT: REQUEST_CHANGES

code-yeongyu:
B14 fixed in f88b4d1da95f73a65613a6c9262c9ff5c95fab39. The incomplete tool-call invalid range now extends through max(callIndex, resultIndex) + 1, so reversed-order chains cannot be admitted with the incomplete call retained. Added a combined reversed-order + incomplete regression: it failed before the fix with a prepared fallback and passed after it. Verification: focused deterministic fallback 27/27 (26 existing cases plus B14), tool-pair/transform suites 40/40, root tsc --noEmit passed, changelog gate against origin/main passed, and the pre-commit full check passed.

code-yeongyu:
WORKING: review-947-r5 - reproducing B14 and probing adversarial tool-chain ordering, invalidation ranges, boundaries, and legitimate complete chains

code-yeongyu (approved):
Round-5 adversarial review of head `f88b4d1da95f73a65613a6c9262c9ff5c95fab39`.

B14: VERIFIED. The exact `toolResult(id)` followed by `assistant(toolCall(id, incomplete: true))` reproduction now returns no fallback (`atomic-tool-chain-cut`), and the diff widens the incomplete invalidation interval through `max(callIndex, resultIndex) + 1`, covering the reversed call index that was previously left valid. The checked-in B14 regression is present and the focused deterministic fallback suite passes 27/27.

Adversarial neighborhood: reversed+complete chains reject at both the call/result boundary shapes; forward+incomplete chains reject at both boundary positions; exact incomplete-at-call and incomplete-at-result boundaries fail closed; and all 24 event-order permutations of two interleaved incomplete calls/results produced no accepted suffix containing an incomplete or reversed pair. Forward complete chains still validate: a boundary at the declaring call is accepted, while a boundary at its result is moved back to the call rather than splitting the pair. A valid complete chain after an unrelated invalid pair remains admissible when the candidate starts after that invalid pair; candidates retaining the invalid pair are correctly rejected rather than over-invalidation.

I also inspected every fallback candidate path (`prepared`, earlier-safe-boundary, and latest-user-turn): all route through the same order/completeness-aware `toolChainValidAt` check. The relevant tool-pair, transform, Gemini signed-empty-state, and provider guard tests pass (65/65 across the targeted runs), and all reported PR checks are green.

New blockers: None.

VERDICT: APPROVE

PR #998, opened 2026-08-19
fix(cursor): persist and retry conversationId after 0-token RE

### LINKED ISSUE #996: Cursor 0-token resource_exhausted should rotate conversationId, persist, and retry the same stream
Zero-token `resource_exhausted` poisons one Cursor conversationId. Today the client rotates once in process memory, does not persist, and model fallback fires before a same-stream retry.

Wanted:
- persist wire id (agent dir)
- retry the same `stream()` after rotate
- cap 3 rotations then skip Cursor for that session
- do not rotate when tokens were already billed

Conflict zone: `packages/ai/src/api/cursor-agent.ts` `stream()` try/catch.

leeseunguk:
Closes #996

Zero-token `resource_exhausted` is a poisoned conversation, not account quota. Rotate the wire id, persist it, and retry the same `stream()` before the model-fallback chain jumps providers. Cap is 3; token-evidence RE does not rotate.

Why not an extension: the rotate Map and h2 retry live inside `cursor-agent` `stream()`.

Conflict zone: `cursor-agent.ts` `stream()` try/catch.

---
## Summary by cubic
Surfaces the first zero‑token Cursor `resource_exhausted` from `stream()` so session‑level compaction can run, then rotates the wire conversation ID and retries on later zero‑token failures. Previously repeated failures errored and could block the session; now we retry up to three times, then return a poisoned‑conversation error, and the next call remints a fresh wire ID instead of blocking.

- Persists the base→wire ID map to `${CODING_AGENT_DIR || ~/.senpi/agent}/cursor-conversation-ids.json` (override with `CURSOR_CONVERSATION_ID_STORE`); the persist path is re‑resolved on each call so env changes take effect.
- Rotation triggers only on zero‑token `resource_exhausted`; any token delta disables rotation. First such failure surfaces; later ones rotate and same‑stream retry migrates cached state and blobs, emitting `start` only once.
- Caps rotations at 3; once exhausted, surfaces “Cursor conversation is poisoned for this session; use another provider,” and the next `stream()` remints a fresh wire ID.
- Review focus: the retry/rotation logic in `packages/ai/src/api/cursor-agent.ts` `stream()` and the new store in `packages/ai/src/api/cursor-conversation-rotation.ts`. Tests also destroy idle HTTP/2 sessions in teardown to avoid suite hangs.

<sup>Written for commit 3dfc938d1b4a43c489e6f30021a476342b8ad289. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/998?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -11,6 +11,15 @@
 ### Fixed
 
 - Cursor conversation-id rotation now persists under the agent directory (`CODING_AGENT_DIR` or `~/.senpi/agent`) instead of `$HOME/cursor-conversation-ids.json`, so a reminted wire id survives TUI restart.
+- Cursor 0-token `resource_exhausted` surfaces on the first failure of a `stream()` call so the session layer can compact before any conversation-id rotation; rotation and same-stream retry apply only to later attempts, and the poisoned-conversation error surfaces once the 3-rotation cap is reached.
+- Cursor 0-token `resource_exhausted` is treated as overflow without a local-estimate floor, and Cursor overflow compaction keeps no recent-token tail so the retry payload actually shrinks.
+- Cursor billed `cacheRead` that dwarfs the live conversation window is ignored: checkpoint `usedTokens` is treated as the real context size when dashboard-cumulative `cache_read_tokens` is more than 3× that window, so compaction is not fired against a multi-million cache-read figure (#983).
+
+- Explicit Cursor thinking levels no longer die with `Connect error not_found`: Cursor's Run RPC
+  rejects bare capability ids (`kimi-k3`, `claude-fable-5`, …) with `not_found`, so
+  `resolveCursorSelectionDescriptor` now prefers the catalog-guaranteed suffix variant id
+  (`kimi-k3-high`, `claude-fable-5-thinking-low`) whenever a legacy alias exists, keeping bare
+  base id + ordered parameters only as the fallback for alias-less levels (#1008).
 
 ### Removed
 
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- Cursor conversation-id rotation now persists under the agent directory (`CODING_AGENT_DIR` or `~/.senpi/agent`) instead of `$HOME/cursor-conversation-ids.json`, so a reminted wire id survives TUI restart.
+
 ### Removed
 
 ## [2026.8.19] - 2026-08-19
--- a/packages/ai/src/api/cursor-agent.ts
+++ b/packages/ai/src/api/cursor-agent.ts
@@ -493,9 +493,6 @@ export const stream: StreamFunction<"cursor-agent", CursorAgentOptions> = (
 			}
 
 			baseConversationId = options?.conversationId ?? options?.sessionId ?? randomUUID();
-			if (rotationStore().shouldSkip(baseConversationId)) {
-				throw new Error(CURSOR_CONVERSATION_POISONED_MESSAGE);
-			}
 			conversationId = rotationStore().getWireId(baseConversationId);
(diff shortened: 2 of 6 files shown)


leeseunguk:
Follow-up commit remints a new wireId after skip instead of blocking the session. Also closes #1008.

leeseunguk:
Follow-up for #1037: persist path is now `CODING_AGENT_DIR` / `~/.senpi/agent/cursor-conversation-ids.json`, not `$HOME/cursor-conversation-ids.json`. Commit ce222a283.

code-yeongyu:
Maintainer rework before merge — merged `origin/main` and adjusted the policy so this composes with #1015 (`fix(cursor): compact large 0-token RE before fallback`), which landed while this was open.

**Why the rework was needed.** #1015 handles a 0-token `resource_exhausted` in `agent-session`: a *surfaced* RE is classified as overflow (`isCursorPayloadResourceExhausted`) and the session compacts before retrying. As authored, this PR's `stream()` catch set `retryPoisonedConversation = true` on **any** `isZeroTokenResourceExhausted` and swallowed the error. Composed, that meant the session layer never saw the failure: #1015's compaction became dead code, and a large-payload rejection burned all three rotations replaying the same oversized payload before failing anyway. The two causes of a 0-token RE need different remedies — an oversized payload is fixed by compaction, a genuinely poisoned `conversationId` by rotation — and only the second is what rotation can address.

**What changed.**

1. **First failure surfaces, no rotation.** The first 0-token RE for a base conversation now surfaces as an error so the session layer gets first refusal to compact. Subsequent 0-token REs rotate, migrate cached state/blobs, and retry in-call, up to the existing cap of 3. The "already surfaced" bit is persisted alongside the wire id (`shouldSurfaceBeforeRotating` / `markSurfaced`), because the natural boundary is the base conversation, not one `stream()` invocation — an in-call `attempt` counter resets on every call and would make rotation unreachable. A reminted id after the skip resets the bit, so a fresh conversation earns its own compact-first pass.

2. **`shouldSkip()` is now wired.** It was defined but never called from `stream()`, so the issue's "cap 3 then skip" was not actually implemented and `CURSOR_CONVERSATION_POISONED_MESSAGE` was imported unused (a `noUnusedLocals` hazard). Past the cap, `stream()` now surfaces `CURSOR_CONVERSATION_POISONED_MESSAGE` instead of rotating again, so the session moves to another provider rather than looping.

3. **Tests pin the composed policy.** `test/cursor-conversation-rotation-stream.test.ts` now covers all three legs: the first 0-token RE surfaces without rotating (server hit exactly once); a retried `stream()` call rotates onto a fresh wire id and succeeds; and after the cap the poisoned-conversation error surfaces. Each test gets its own `CURSOR_CONVERSATION_ID_STORE` so the persisted cap does not leak between cases. The first test was written RED against the original branch and failed exactly as expected before the fix.

Also fixed a latent typecheck error in the test's `startServer` helper (`Http2Stream` vs `ServerHttp2Stream`), which the pre-commit `tsc --noEmit` caught.

Doc conflicts resolved by stacking, not replacing: `packages/ai/src/changes.md` keeps both this entry and main's `Cursor 0-token RE overflow without estimate gate`, and the `[Unreleased]` `### Fixed` bullets are the union of both sides.

Verified: `cd packages/ai && npm test -- test/cursor-conversation-rotation test/cursor-payload` — 8 passed, including #1015's `cursor-payload-resource-exhausted` test, confirming the two policies compose. Full pre-commit check suite and the changelog gate both pass.

PR #981, opened 2026-08-19
fix(agent): pass run abort signal into Cursor exec handlers

leeseunguk:
Closes #979

`cursorExecHandlers` received `requestAbortController.signal`. The session bridge compares that to `agent.signal` (the run controller). They are never the same object, so a new Cursor session rejects every native exec with `Tool execution has no active run`.

This passes the run-level `signal` (falling back to the request controller only when the loop has no run signal).

Why not an extension: the factory is invoked inside `packages/agent`.

Conflict zone: `packages/agent/src/agent-loop.ts` `cursorExecHandlers(...)` argument.

---
## Summary by cubic
Restores Cursor native exec by passing the run-level abort signal to `cursorExecHandlers`. Previously it used the per-request controller, so the session bridge compared it to `agent.signal` and rejected exec with "Tool execution has no active run" on new sessions.

- In `packages/agent/src/agent-loop.ts`, `cursorExecHandlers` now receives `signal ?? requestAbortController.signal`.
- Adds a test in `packages/agent/test/cursor-exec-loop.test.ts` verifying the run signal is passed through.
- No API changes; behavior now matches `createSessionCursorExecBridge` expectations. Potential conflict around the `cursorExecHandlers(...)` argument in `packages/agent/src/agent-loop.ts`.

<sup>Written for commit 933a2fca16e18904d1a5bb93f1775607baae85c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/981?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1004, opened 2026-08-19
fix(coding-agent): resolve cursor exec ownership by run liveness not signal identity

### LINKED ISSUE #1003: Cursor exec bridge refuses every tool call with "Tool execution has no active run" since v2026.8.18-3
**What happened?**

Since v2026.8.18-3, every tool call made through the Cursor exec bridge fails with `Tool execution has no active run`. Any model routed via the cursor provider (e.g. `cursor/kimi-k3`, commonly hit as a fallback lane when another provider's quota is exhausted) becomes a no-tools model: read/bash/write all fail on the first call. Worked in v2026.8.17 through v2026.8.18-2; broken in v2026.8.18-3 and v2026.8.19 (current). Reproduces with no extensions loaded, so this is engine-core wiring, not an extension interaction.

**Steps to reproduce**

1. On senpi 2026.8.19, route any agent to a cursor-provider model (configure a fallback list containing `cursor/kimi-k3`, or set the model directly).
2. Ask the agent to do anything requiring a tool (read a file, run a shell command).
3. Every tool result is the error `Tool execution has no active run`.

Minimal repro for maintainers: call `createSessionCursorExecBridge` with the production wiring (request controller signal as `runSignal`, a separate run controller behind `getAgent().signal`) and invoke any handler. It returns the error even with a fully live run.

**Expected behavior**

Tools execute normally through the cursor provider.

**Version**

2026.8.19 (regressed in v2026.8.18-3). Introduced by commit 31a71f0c5 "fix(coding-agent): bind cursor exec dispatches to the owning run" (2026-08-18), which changed `getAbortSignal` in `packages/coding-agent/src/core/cursor-exec-bridge-session.ts` to `runSignal === getAgent().signal ? runSignal : undefined`. In production the two compared signals are always different objects: `runSignal` is the per-request controller's signal created in `packages/agent/src/agent-loop.ts:480` and passed through the `sdk.ts:450` factory, while `getAgent().signal` is the run's own controller signal (`packages/agent/src/agent.ts:647`). Object identity never holds, so every live frame is refused. The unit tests passed because they inject the same signal object into both positions, hiding the wiring divergence. Fix direction: resolve ownership by liveness (refuse when the request signal is aborted or no run is active) instead of object identity.

I have a fix ready and will open a PR shortly.

DevNewbie1826:
Fixes #1003.

**Root cause**: `31a71f0c5` made `getAbortSignal` in `cursor-exec-bridge-session.ts` compare the loop's per-request controller signal against the agent's per-run controller signal by object identity. In production those are always different objects (`agent-loop.ts` creates the request controller fresh per request; `agent.ts` the run controller per run), so the identity check failed for every live frame and every Cursor exec tool call returned `Tool execution has no active run` — any cursor-routed model became a no-tools model (read/bash/write all refused on the first call).

**Fix**: resolve ownership as liveness instead of identity — refuse a frame only when its request signal is aborted (the loop mirrors run aborts, idle timeouts, and pre-fallback run deaths into the request controller) or when no run is active. Event safety is unchanged: `Agent.emitExternalEvent` already drops events whose signal is not the active run's, so a straggler frame from a replaced run cannot leak lifecycle events into the new transcript.

**Tests**

- New `test/suite/regressions/1003-cursor-exec-bridge-request-signal.test.ts` reproduces the production wiring (distinct request/run controllers): the live-frame case is red on `main` and green here; aborted-request, no-active-run, and legacy no-signal cases pin the refusal and fallback paths.
- `cursor-exec-bridge-run-ownership.test.ts` dead-run simulation updated to abort the captured request signal — what production actually does when a run dies — keeping the same refusal assertions.
- Full bridge suite 18/18 green; `npm run check` green; `npm test` green except one pre-existing failure in `scripts/prepare-senpi-bundled-workspaces.test.mjs` that fails identically on `main` (verified in a clean `main` worktree).
- Real-CLI QA per `.agents/skills/senpi-qa/`: RPC self-test 4/4, mock-loop self-test 48/48, RPC `--state` and a mock `--prompt` turn both PASS (receipts captured locally under `local-ignore/qa-evidence/`, gitignored by policy).

Cross-checked against the published packages: a seam test driving the compiled `2026.8.18-2` and `2026.8.19` dists with the production wiring shows the tool executing on `-2` and refused on `-3`/`.19`; the same seam against the patched engine code flips it back to executing.

`changes.md` entry added under `packages/coding-agent/src/core/` following the existing bridge entries. `CHANGELOG.md` untouched per CONTRIBUTING.

Assisted by an AI agent (omo/senpi); I reviewed and verified every line and claim.

---
## Summary by cubic
Restores Cursor exec tool dispatch by resolving run ownership from request liveness, not signal identity. Previously the bridge compared the per-request signal to the per-run signal by object identity (different objects in production), so every frame was refused with "Tool execution has no active run." Now frames execute when the request signal is not aborted and a run is active; they are refused only when the request is aborted or no run is active. Straggler frames do not leak events into a new run because `Agent.emitExternalEvent` drops events with non-current signals.

**Review notes**
- `packages/coding-agent/src/core/cursor-exec-bridge-session.ts`: `getAbortSignal` returns the request signal when it is not aborted and the agent has an active run; otherwise returns `undefined`. When no request signal is supplied, it falls back to the agent run signal.
- Tests: added `packages/coding-agent/test/suite/regressions/1003-cursor-exec-bridge-request-signal.test.ts` (live, aborted, no-run, legacy no-signal); updated `packages/coding-agent/test/cursor-exec-bridge-run-ownership.test.ts` to abort the request signal when simulating a dead run.
- No API changes or migrations.

<sup>Written for commit 63d39d6bb1fe22ae5dea7d642bf9bfca5a84c56e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1004?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.omo/plans/cursor-exec-bridge-run-liveness.md
+++ b/.omo/plans/cursor-exec-bridge-run-liveness.md
@@ -1,0 +1,99 @@
+# Plan: restore Cursor exec bridge dispatch by checking run liveness, not signal identity
+
+## Problem
+
+Since v2026.8.18-3 (commit `31a71f0c5`), every Cursor exec frame is refused with
+`Tool execution has no active run`, making any Cursor-routed model (e.g. the
+`cursor/kimi-k3` fallback lane) a no-tools model.
+
+`createSessionCursorExecBridge` compares the per-request signal against the
+agent's run signal by object identity:
+
+```ts
+getAbortSignal: () => {
+    if (runSignal === undefined) return getAgent().signal;
+    return runSignal === getAgent().signal ? runSignal : undefined;
+},
+```
+
+In production these are two different controllers by construction:
+
+- `runSignal` is `requestAbortController.signal`, created fresh per provider
+  request (`packages/agent/src/agent-loop.ts:480`) and supplied to the bridge
+  factory (`packages/coding-agent/src/core/sdk.ts:450-451`).
+- `getAgent().signal` is the owning run's `abortController.signal`
+  (`packages/agent/src/agent.ts:647`).
+
+The identity therefore never holds, so the guard returns `undefined` for every
+live frame and `executeTool` refuses with `Tool execution has no active run`.
+The unit tests inject the same signal object into both positions, so the
+divergence was invisible to them.
+
+Verified by comparing the published packages side by side with a seam test
+that reproduces the production wiring (two distinct controllers): `2026.8.18-2`
+executes the tool; `2026.8.18-3`/`2026.8.19` refuse it.
+
+## Fix design
+
+Ownership is a liveness property, not an identity property. The loop's request
+controller mirrors the owning run's aborts (`agent-loop.ts:480-490`: the run
+signal's `abort` event aborts the request controller; the idle-timeout path
+aborts it directly), so:
+
+```ts
+getAbortSignal: () => {
+    if (runSignal === undefined) return getAgent().signal;
+    if (runSignal.aborted) return undefined;
+    return getAgent().signal === undefined ? undefined : runSignal;
+},
+```
+
+- Live request + active run → execute with the request's own signal (fixes the
+  regression).
+- Aborted request (user cancel, idle timeout, run aborted before a fallback
+  restart) → refuse. This is the straggler case `31a71f0c5` targeted.
+- No active run at all → refuse, matching the pre-regression behavior of
+  returning `getAgent().signal` when it is `undefined`.
+
+A straggler whose run ended normally (request not aborted) while a replacement
+run is live executes with its own request signal; `Agent.emitExternalEvent`
+(`agent.ts:807-809`) already drops events carrying a non-current signal, so its
+lifecycle events cannot leak into the replacement run. This is strictly safer
+than the pre-`31a71f0c5` behavior, which adopted the new run's signal and
+leaked events into it.
+
+## Tests
+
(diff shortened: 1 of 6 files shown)


code-yeongyu:
Closing in favor of #1002 after a three-way review of #981/#1002/#1004. Two technical reasons: (1) replacing the identity guard with request-liveness reopens the straggler hole 31a71f0c5 closed - on the normal stream-completion path the request controller is never aborted (`agent-loop.ts` done path), so a buffered exec frame from a completed run would pass `!runSignal.aborted` and execute under a replacement run; `Agent.emitExternalEvent` only drops the *events*, it does not stop `tool.execute`. (2) The existing straggler regression test was edited to inject `requestA.abort()` after run completion - an abort production does not perform on that path - to keep the suite green. The production-wired regression test in this PR (real Agent + distinct request/run controllers, red on main) is genuinely the best artifact of the three and will be ported, adapted to the identity semantics, as a follow-up. Thank you - the dist cross-check and QA receipts here set the bar.

PR #1006, opened 2026-08-20
fix(config-reload): preserve routine settings handoffs

Indosaram:
## Summary
- preserve request-time settings snapshots through config-reload handoff
- classify changes discovered during reload against that snapshot so a peer session's routine settings write does not request a second full reload
- cover routine suppression alongside the existing non-routine reload-window regression

## Root cause
Replacement watchers refreshed their settings-content snapshot before reload-window changes were classified. A concurrent `defaultModel` write from another OMO/Senpi process therefore compared current content to itself, bypassed routine-settings filtering, and could cascade into shared-agent-directory reload storms.

## Verification
- [x] Red: new routine handoff regression fails before the fix (`secondReload` called once)
- [x] `node ./node_modules/vitest/vitest.mjs run packages/coding-agent/test/suite/config-reload-extension.test.ts` — 53 passed
- [x] `npm run check` under Node 24 — passed (Biome, pinned deps, import checks, shrinkwrap/install locks, TypeScript, browser smoke)
- [x] isolated real CLI smoke — passed
- [x] isolated tmux TUI smoke — input reached composer; no leaked process
- [x] independent scoped review — no blockers

## Scope
Routine-only changes remain suppressed. The existing regression verifies a non-routine change during the same handoff window still causes one follow-up reload.

---
## Summary by cubic
Preserves the request-time settings snapshot through reload handoffs and clears orphaned handoffs to prevent cross‑session reload storms and stale replays. Old: routine writes during the reload window could trigger a second full reload, and if the successor omitted config-reload the handoff persisted and later replayed stale changes. New: the replacement watcher restores the pre-reload snapshot so routine-only writes stay suppressed, and the handoff is deleted unconditionally after reload.

- Adds `settingsContentsAtRequest` to `ReloadHandoff` and restores it before classifying handoff changes in `index.ts`.
- Captures the handoff key before `requestReload()` and deletes the registry entry when it settles; removes the `tornDown` guard.
- Adds tests for routine-change suppression during handoff and for clearing the handoff when the successor omits the builtin.
- No API or config changes. No migration.

<sup>Written for commit d22eb0a7bd8c317e4360da8ddb1755e6b00dce8b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1006?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/config-reload/index.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/config-reload/index.ts
@@ -173,7 +173,6 @@ export function configReloadExtension(pi: ExtensionAPI, options: ConfigReloadExt
 	let activeTargets: ActiveTarget[] = [];
 	let currentContext: ExtensionContext | undefined;
 	let started = false;
-	let tornDown = false;
 	let reloadInFlight = false;
 	let deferredNoticeShown = false;
 	let unavailableReloadLogged = false;
@@ -387,7 +386,6 @@ export function configReloadExtension(pi: ExtensionAPI, options: ConfigReloadExt
 		const changes = pendingChanges(pending);
 		const paths = uniquePaths(changes.flatMap((change) => change.paths));
 		reloadInFlight = true;
-		tornDown = false;
 		reloadHandoffs.set(handoffKey(ctx), {
 			hashesAtRequest: engine?.getBaselineSnapshot() ?? new Map<string, string>(),
 			settingsContentsAtRequest: new Map(settingsContents),
@@ -397,15 +395,14 @@ export function configReloadExtension(pi: ExtensionAPI, options: ConfigReloadExt
 		ctx.ui.notify(`Hot-reloading: ${formatPaths(paths)}`, "info");
 		logger.info("reload_requested", { reason: "config changed", paths });
 
+		const handoffKeyForReload = handoffKey(ctx);
 		try {
 			await ctx.requestReload();
-			if (!tornDown) {
-				reloadInFlight = false;
-				reloadHandoffs.delete(handoffKey(ctx));
-			}
+			reloadInFlight = false;
+			reloadHandoffs.delete(handoffKeyForReload);
 		} catch (error) {
 			reloadInFlight = false;
-			reloadHandoffs.delete(handoffKey(ctx));
+			reloadHandoffs.delete(handoffKeyForReload);
 			logger.error("watcher_error", { path: "reload", message: errorMessage(error) });
 		}
 	};
@@ -455,7 +452,6 @@ export function configReloadExtension(pi: ExtensionAPI, options: ConfigReloadExt
 	eventUnsubscribes.push(pi.events.on(CONFIG_WATCH_UNREGISTER, handleUnregistration));
 
 	pi.on("session_start", async (event, ctx) => {
-		tornDown = false;
 		started = true;
 		currentContext = ctx;
 		rebuildWatchers(ctx);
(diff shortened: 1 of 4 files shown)


packages/coding-agent/src/core/extensions/builtin/config-reload/index.ts
```diff
@@ -389,6 +390,7 @@ export function configReloadExtension(pi: ExtensionAPI, options: ConfigReloadExt
 		tornDown = false;
 		reloadHandoffs.set(handoffKey(ctx), {
 			hashesAtRequest: engine?.getBaselineSnapshot() ?? new Map<string, string>(),
+			settingsContentsAtRequest: new Map(settingsContents),
```
code-yeongyu, line 391: [sisyphus-bot] [P2] Clear the handoff when a reload removes this builtin

This adds the full settings text to a module-global handoff, but the existing cleanup skips deletion after any real reload (`tornDown === true`). If the settings change disables `config-reload`, the successor never calls `take()`, so the entry survives for the process lifetime; settings can include `httpProxy` URLs with userinfo. When a later manual reload re-enables the builtin, it consumes that stale entry and emits `CONFIG_WATCH_RELOADED` for the old change.

I reproduced this with a deterministic Vitest scenario: the first reload omitted config-reload, the later reload re-enabled it, and the expected 0 reloaded events was 1. Capture the handoff key before `requestReload()`, delete it unconditionally when that promise settles (safe if the successor already took it), and add the omit/re-enable regression.

code-yeongyu (changes requested):
[sisyphus-bot] Requesting changes for one independently reproduced blocker.

- **[P2] Orphaned plaintext reload handoff:** if the settings change removes `config-reload` from the successor runtime, the old generation leaves the module-global handoff alive. This PR expands that orphan from hashes/paths to the full settings contents, and a later reload that re-enables the builtin consumes and replays the stale change. See the inline comment on `config-reload/index.ts:393` for the deterministic reproduction and requested fix.

Validation performed: focused config-reload suite 53/53; root static check at PR head; Senpi QA common 9/9, mock-loop 48/48, TUI 5/5, CLI 8/8; latest-main synthetic merge focused suite 53/53. Four independent review lanes passed; the fifth hands-on child was cancelled as inconclusive and was not counted as approval. The retained blocker was reproduced personally after those lanes completed.

code-yeongyu:
[sisyphus-bot] Fixed the orphaned handoff blocker in commit `4cd3ff3f5`.

**What changed:** `flushPending` now captures the handoff key before `requestReload()` and deletes the registry entry unconditionally when the promise settles. The `tornDown` guard (and the variable itself) was removed — it was the only reader.

**Why:** If the settings change disabled `config-reload`, the successor never called `take()`, so the handoff survived for the process lifetime. This PR expanded that orphan from hashes/paths to full settings contents (`httpProxy` URLs may contain credentials). A later reload that re-enabled the builtin consumed and replayed the stale change.

**Regression test:** `clears the reload handoff when the successor omits config-reload` — fails on the old code (1 stale `CONFIG_WATCH_RELOADED` event), passes on the fix (0 events).

**Validation:** 54/54 config-reload tests pass; `npm run check` clean.

PR #1038, opened 2026-08-20
perf(coding-agent): defer the Claude Agent SDK and jsdom out of the startup import graph

code-yeongyu:
## What

`@anthropic-ai/claude-agent-sdk` (a 1.2 MB single-file bundle) and the webfetch HTML stack
(jsdom + `@mozilla/readability` + turndown) were parsed and evaluated on **every** CLI start,
even though only a claude-sdk-oauth stream and a webfetch HTML conversion ever use them.
Importing `dist/main.js` is roughly 70% of boot wall time, so every run that never reached
either feature still paid for both.

Both now load on first use behind the repository's documented lazy-boundary pattern
(`packages/ai/src/api/*.lazy.ts`):

- **`claude-sdk-oauth/sdk-boundary.lazy.ts`** owns the single deferred SDK import, caches the
  module and shares one in-flight promise across concurrent callers. `sdk-boundary.ts` keeps its
  synchronous `getSdkBoundary()` surface (callers are unchanged) and derives its exported types
  from the loader's module type instead of from value imports.
- **`webfetch/webfetch/content.lazy.ts`** wraps `content.ts` and exposes the two converters as
  async. Both call sites already sat inside the tool's async `execute`, so nothing else moved.

`getSessionMessages` is async and self-loads. `query` and `createSdkMcpServer` are *synchronous*
SDK functions, so the three async entry points that reach them await the loader first —
`streamClaudeSdkOauth`, `reattachSession`, and `buildCustomToolServers` (now async). A future call
site that forgets the preload fails with a named error at that exact call instead of silently
restoring the startup import.

The tiny `@anthropic-ai/claude-agent-sdk/extract` entrypoint used by `executable.ts` (6.6 KB) is
unrelated to the heavy bundle and deliberately stays a static import.

**Zero behavior change.** Only the moment the two dependencies load moved.

## Failing-first proof

`test/startup-import-graph.test.ts` probes the graph with Node's own loader hook (not a source
scan), so a deferred `await import(...)` is absent by construction while any reintroduced
top-level edge — direct or transitive, through any re-export chain — reappears and fails.

**RED** — same test, same commit, against a build of `main` without the lazy boundaries:

```
 FAIL  test/startup-import-graph.test.ts > CLI startup import graph > does not statically reach deferred heavy packages from dist/main.js
AssertionError: jsdom is statically reachable from dist/main.js; it must stay behind the lazy boundary owned by webfetch HTML conversion (core/extensions/builtin/webfetch/webfetch/content.lazy.ts): expected [ …(2) ] to deeply equal []

- Expected
+ Received

- []
+ [
+   ".../node_modules/jsdom/lib/api.js",
+   ".../node_modules/jsdom/lib/api.js",
+ ]

 Test Files  1 failed (1)
      Tests  1 failed (1)
```

The assertion stops at the first offender, so here is the full per-package count from the same
two builds (`probeImportGraph` on `dist/main.js`):

| package | before | after |
|---|---|---|
| `jsdom` | 2 | **0** |
| `@mozilla/readability` | 2 | **0** |
| `turndown` | 2 | **0** |
| `@anthropic-ai/claude-agent-sdk` (`sdk.mjs`, 1.2 MB) | 4 | **0** |
| `@anthropic-ai/claude-agent-sdk/extract` (6.6 KB, unrelated) | — | 2 |
| **total modules in the startup graph** | **14,203** | **13,523** |

**GREEN** — with the lazy boundaries:

```
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

## Measurements

Machine: Apple M4 Pro (14c), macOS 26.4.1, node v26.7.0. Warm FS cache. Both sides built from
this same worktree back-to-back in one session under an exclusive bench lock (`before` = this
branch with the lazy diff reverted and its two `.lazy.ts` modules removed, rebuilt and the stale
`*.lazy.js` artifacts pruned; `after` = the full diff). 1-min loadavg is recorded per pass — a
concurrent lane shares this machine, so passes are only compared against passes taken at a
comparable load, and user-CPU is reported alongside wall time because it is load-invariant.

### `hyperfine --warmup 3 --runs 10 'node packages/coding-agent/dist/cli-main.js --help'`

| pass | loadavg (1-min) | mean ± σ | min | user CPU |
|---|---|---|---|---|
| before | 11.34 | 1.241 s ± 0.045 s | 1.179 s | 1.250 s |
| after | 10.17 | 1.064 s ± 0.058 s | 0.967 s | 1.061 s |
| after (repeat) | 8.64 | 0.990 s ± 0.040 s | 0.942 s | 0.989 s |

**-177 ms mean (-14.3%)** at comparable load; **-189 ms user CPU (-15.1%)**, which does not
depend on machine load. The repeat pass at a lower load shows the same delta direction and size.

### Import-only cost of `dist/main.js` (`Date.now()` around `await import(...)`, 7 runs)

| pass | loadavg (1-min) | runs (ms) | mean |
|---|---|---|---|
| before | 11.34 | 853, 862, 886, 882, 893, 905, 881 | **880 ms** |
| after | 10.17 | 694, 722, 683, 668, 688, 718, 693 | **695 ms** |
| after (repeat) | 8.64 | 647, 641, 648, 643, 642, 645, 630 | **642 ms** |

**-185 ms (-21%)** on the module graph alone — this is the isolated cost this PR removes, and it
matches the 680-module reduction in the graph table above.

### Startup smoke (`npm run profile:rpc -- --skip-build --runs 3 --warmup 1`)

After the change, at loadavg 8.3: median 1350.9 ms, min 1342.8 ms across 3 runs. This drives a
real boot to a served `get_state` RPC response, so it exercises the full CLI startup path
(including extension registration, where both deferred modules used to be pulled in).

`npm run profile:tui` cannot run in this environment — it requires a controlling terminal and
exits with `TUI benchmark must be run from an interactive terminal` / `PI_STARTUP_BENCHMARK only
supports interactive mode`. The RPC profiler covers the same startup path headlessly.

## Verification

- `npm run build` (root) clean.
- `npm run check` green (biome, pinned-deps, ts-imports, shrinkwrap, install-lock,
  claude-sdk-platform-lock, `tsc --noEmit`, browser smoke).
- `packages/coding-agent` `npm test`: **8,503 passed**, 36 skipped, 2 failed — both in
  `test/cursor-cli-oauth/shutdown.test.ts`, which spawns a real process tree with a 5 s deadline.
  Pre-existing and unrelated: this PR touches no cursor path, and the file fails identically in
  isolation on unmodified `main`.
- Focused suites green: all 54 `claude-sdk-oauth*` / `webfetch*` / `export*` files
  (426 passed, 3 skipped), plus `startup-import-graph`, `claude-sdk-oauth-tools`, and
  `model-runtime-text-toolcall-recovery` (the sibling import-graph probe test).

---
## Summary by cubic
Defers `@anthropic-ai/claude-agent-sdk` and the webfetch HTML stack (`jsdom`, `@mozilla/readability`, `turndown`) out of the CLI’s startup import graph; they now load on first SDK stream or first HTML conversion. Previously both parsed on every start; now 680 modules are removed from startup, cutting import/boot time with no behavior change.

- Adds lazy boundaries: `claude-sdk-oauth/sdk-boundary.lazy.ts` (SDK) and `webfetch/webfetch/content.lazy.ts` (HTML converters). `sdk-boundary.ts` keeps a synchronous surface and throws a named error if synchronous SDK members are read without preloading.
- Preloads SDK in async entry points: `streamClaudeSdkOauth`, `reattachSession`, and `buildCustomToolServers` (now async). Webfetch tool imports from `content.lazy.ts` and awaits `htmlToMarkdown`/`htmlToText`.
- Startup guard: `test/startup-import-graph.test.ts` fails if `dist/main.js` statically reaches these packages; the probe helper now accepts an arbitrary entry. The small `@anthropic-ai/claude-agent-sdk/extract` import remains static.

**Required actions**
- Callers of `buildCustomToolServers` must await it.
- New call sites to `query` or `createSdkMcpServer` must `await loadClaudeAgentSdk()` first.
- Callers of the webfetch converters must await `htmlToMarkdown`/`htmlToText`.

<sup>Written for commit 2584e6446a487e4b71890f7f0f6b1324e3d166c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1038?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -42,6 +42,11 @@
   background block, bold tone title, dim body): loaded-resource conflict diagnostics, the update-available
   and package-update notifications, the risky-main-model and high-reasoning warnings, the rules banner,
   the prompt URL widget card, and the earendil announcement. Visible text is unchanged.
+- The CLI no longer parses and evaluates the 1.2 MB Claude Agent SDK bundle or the jsdom/Readability/turndown
+  HTML stack while starting up. Both now load on first use behind the repository's documented lazy-boundary
+  pattern — the SDK when a claude-sdk-oauth stream actually opens, the HTML converters when webfetch actually
+  converts an HTML response — which removes 680 modules from the startup import graph (14,203 → 13,523).
+  Behavior is unchanged; only the moment the two dependencies are loaded moved.
 
 - Startup is faster after the first run: the CLI now enables Node's on-disk module compile cache
   (`enableCompileCache()`) in both `cli.ts` and `cli-main.ts` and publishes the resolved cache directory
--- a/packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/changes.md
@@ -52,6 +52,43 @@ A stream-start-timeout abort closes the SDK session with the turn's user message
 - `session-observability.ts` `ContinuityReason` union tail and `SANITIZED_REASONS` set (mechanically
   duplicated literals; both must gain the member).
 
+## 2026-08-20 - SDK bundle loads on first stream instead of at CLI startup
+
+### What changed
+
+- New `sdk-boundary.lazy.ts` owns the single deferred `import("@anthropic-ai/claude-agent-sdk")`, caching the
+  module and sharing one in-flight promise across concurrent callers. `sdk-boundary.ts` keeps its synchronous
+  `getSdkBoundary()` surface and re-exports `loadClaudeAgentSdk`; its default members read the loaded module
+  (`getSessionMessages` is async and self-loads, `query` / `createSdkMcpServer` are synchronous SDK functions
+  and therefore require the preload). Its exported types are now derived from the loader's module type instead
+  of from value imports.
+- The three async entry points that reach a synchronous SDK member now await the loader first:
+  `stream.ts` (`streamClaudeSdkOauth`, before `getSdkBoundary().query` and the resident-session lane),
+  `session-reattach.ts` (`reattachSession`, before `getOrCreateSession` builds a query), and `custom-tools.ts`
+  (`buildCustomToolServers`, now async, before `createSdkMcpServer`). A future call site that skips the preload
+  fails with a named error at that call rather than silently restoring the startup import.
+- The tiny `@anthropic-ai/claude-agent-sdk/extract` entrypoint used by `executable.ts` is unaffected and stays
+  a static import; only the 1.2 MB `sdk.mjs` bundle moved.
+
+### Why
+
+- Importing `dist/main.js` is roughly 70% of CLI boot wall time, and the SDK bundle was parsed and evaluated on
+  every start even though only the claude-sdk-oauth streaming lane ever calls into it. Deferring it removes that
+  cost from every run that never opens a Claude SDK stream, with no behavior change for runs that do.
+
+### Why an extension could not handle it
+
+- The static import lives in this builtin provider's own SDK boundary module. An extension cannot remove an
+  import edge from a module the core already loads, and re-registering the provider id would fork the auth lane,
+  session registry, and failover wiring that live here.
+
+### Expected merge conflict zones
+
+- MEDIUM in `sdk-boundary.ts`: the import block and the `defaultSdkBoundary` literal, which upstream also touches
+  when adding SDK members. A new member must be added to `sdk-boundary.lazy.ts`'s module projection as well.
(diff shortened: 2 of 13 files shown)


code-yeongyu:

PR #1047, opened 2026-08-20
fix: recover failing CI checks

code-yeongyu:
## What was failing

- **Publish Model Catalog.** The publish job inherits R2 credentials this repository does not define, so every scheduled run failed at upload while generation and checking were healthy. Now gated on `vars.PI_MODEL_CATALOG_PUBLISH`; generation and the catalog check stay unconditional.
- **`valid-cwd-guard`.** On macOS the ESM loader wedges when a dynamic import *starts* after the cwd is unlinked, so the child hung and the unbounded `spawnSync` blocked Vitest's own timeout. `recoverValidCwd()` is now exported and the test links both modules while the directory is still valid.
- **`cursor-cli-oauth/shutdown`.** The fixture publishes its pid file with `rename(2)` before its first stdout event, so `fs.watch` never fires when the file already exists by the time the watcher is installed.

## Verification

- `npx vitest --run test/valid-cwd-guard.test.ts test/cursor-cli-oauth/shutdown.test.ts` — 11/11.
- Full `npm test` at repo root — exit 0.
- `actionlint` clean on all workflows.

---
## Summary by cubic
Recovers failing CI by gating model-catalog publishing and fixing two test hangs on macOS. Previously, the publish job always ran and failed with missing R2 credentials; now it runs only when `vars.PI_MODEL_CATALOG_PUBLISH` is `true`, while generation and validation stay unconditional.

- Publishing: Set `vars.PI_MODEL_CATALOG_PUBLISH='true'` and configure `PI_ARTIFACTS_R2_*` secrets to re-enable uploads; otherwise only generate and validate the catalog.
- Valid CWD guard: Export `recoverValidCwd()` (still invoked at module load) so callers can recover after the cwd is deleted; tests preload modules to avoid macOS ESM loader wedging and add a 30s timeout backstop.
- PID watcher: Read the pid file if it already exists and handle both “change” and “rename” events to avoid missing the publish event.

<sup>Written for commit 76bcae75e86c7b13f63cf008d08e7b455876da8a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1047?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/changes.md
+++ b/.github/changes.md
@@ -1,5 +1,24 @@
 # changes
 
+## Gate the inherited model-catalog publisher behind fork opt-in (2026-08-20)
+
+### What changed
+
+- `.github/workflows/publish-model-catalog.yml` now schedules the `publish` job only when the repository variable `PI_MODEL_CATALOG_PUBLISH` is exactly `true`, in addition to its existing event and manual-publish conditions.
+
+### Why
+
+- This fork inherited upstream's R2 endpoint and bucket without either required `PI_ARTIFACTS_R2_*` secret. Every upload attempt therefore exported empty AWS credentials and failed with `Unable to locate credentials`, while generation and validation stayed green.
+- Keeping publishing opt-in makes the fork's default workflow honest: catalog generation still runs and is validated, but the unavailable upstream infrastructure no longer creates recurring default-branch failures. A maintainer can restore uploads after configuring a fork-owned R2 target, both secrets, and the opt-in variable.
+
+### Why an extension could not handle it
+
+- GitHub evaluates the job graph, repository variables, and environment secrets before any Senpi runtime or extension is available.
+
+### Expected merge conflict zones
+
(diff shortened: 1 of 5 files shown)


code-yeongyu:
Superseded by #1048.

While this PR was in flight, `main` landed an equivalent guard for the cursor-CLI shutdown
pid-file race. #1048 is rebased onto `20cd958c0` and reduced to the two fixes `main` still
lacks: the model-catalog publish opt-in gate and the exported deleted-cwd guard.
Verified there at 11 passed / 0 failed, actionlint clean.

PR #1059, opened 2026-08-21
fix(permission-system): prevent unhandledRejection on session shutdown with pending requests

Indosaram:
## Summary\n\nFixes an unhandled promise rejection crash () that occurs when a session shuts down (e.g. via `/clear`, reload, or exit) while a permission prompt is pending.\n\n## Root Cause\n\nIn `packages/coding-agent/src/core/extensions/builtin/permission-system/index.ts`, `service.ask(request)` returns a promise which was stored in `const askPromise`. When `session_shutdown` fires, `service.list()` iterates pending requests and calls `service.reply({ reply: 'reject' })`, which rejects the internal promise with `RejectedError`.\n\nBecause `askPromise` did not have a `.catch()` handler attached immediately, Node.js detected an unhandled rejection while the UI prompt was still suspended awaiting user selection. Node.js's `unhandledRejection` listener forwarded this to `uncaughtCrash`, terminating the process with `uncaughtException: RejectedError`.\n\n## Fix\n\nAttach a `.catch()` handler immediately when creating `askPromise` in `tool_call` to capture `askError`, avoiding floating rejected promises during shutdown while still properly propagating the block result to the runner after `askPromise` settles.\n\n## Verification\n\n- Added regression test `packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts`\n- Ran `npm test` across `packages/coding-agent/test/permission/` and regression suites (481/481 tests passing)\n- Full static checks `npm run check` pass cleanly\n- `scripts/check-pr-changelog.mjs` passes\n- Ran `senpi-qa` channels (CLI smoke, RPC drive, Mock loop)

---
## Summary by cubic
Prevents unhandled promise rejections and crashes when a session shuts down with a pending permission prompt. Previously shutdown raised an unhandledRejection that escalated to uncaughtException; now the tool call deterministically returns a blocked result with a consistent reason, including when the rejection value is falsy.

- Localized change in `packages/coding-agent/src/core/extensions/builtin/permission-system/index.ts`: wrap `service.ask(request)` in an immediately-handled `askResultPromise` and map rejections via `getReason(...)` in both early and pending paths.
- Regression tests: strict child-process test under `--unhandled-rejections=strict` verifies no crash on shutdown; unit test ensures falsy rejections map to the default message.
- No API or config changes; permission semantics unchanged.

<sup>Written for commit d6ea46423e2a0fae71082c9db110a5ddc9bcd97a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1059?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/coding-agent/src/core/extensions/builtin/permission-system/index.ts
```diff
@@ -117,14 +117,16 @@ export default function permissionSystemExtension(pi: ExtensionAPI): void {
 				metadata: createRequestMetadata(event.toolName, event.input),
 			};
 
-			const askPromise = service.ask(request);
+			let askError: unknown;
+			const askPromise = service.ask(request).catch((error) => {
```
code-yeongyu: This rejection capture is lossy. Promise rejection reasons are arbitrary JavaScript values, including `undefined`, `null`, `false`, `0`, and the empty string. For any of those, the handler runs but `askError` remains falsy, so both checks below treat the rejected permission request as success and allow the tool call to continue. Do not encode settlement in the truthiness of the rejection value. Replace this mutable sentinel with a tagged result promise, e.g. `service.ask(request).then(() => ({ ok: true as const }), error => ({ ok: false as const, error }))`, then branch on `ok`; add coverage for a falsy rejection reason. That also removes the fragile side-channel mutation and makes the immediate rejection handler explicit.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
```diff
@@ -0,0 +1,75 @@
...
+	let tempDir: string;
+	const unhandledRejections: unknown[] = [];
+	const onUnhandled = (reason: unknown) => {
+		unhandledRejections.push(reason);
+	};
+
+	beforeEach(() => {
+		tempDir = mkdtempSync(join(tmpdir(), "pi-perm-rejection-"));
+		unhandledRejections.length = 0;
+		process.on("unhandledRejection", onUnhandled);
```
code-yeongyu: A process-crash regression should not add another process-global `unhandledRejection` listener inside the Vitest worker. The listener changes Node's default production escalation behavior, scopes observation to every task sharing that worker, and therefore does not exercise the actual process-survival contract; an unrelated background rejection can also contaminate this assertion. Run the scenario in an isolated child process with strict/default production unhandled-rejection handling and assert that the child exits cleanly. That verifies both rejection handling and the reported crash without mutating the test worker.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
```diff
@@ -0,0 +1,75 @@
...
+		runner.setFlagValue("permission-preset", "ask");
+		await runner.emit({ type: "session_start", reason: "startup" });
+
+		runner.setUIContext({
+			select: async () => {
+				await new Promise((resolve) => setTimeout(resolve, 100));
+				return null;
+			},
+			input: async () => undefined,
+		} as any);
```
code-yeongyu: `as any` deletes the interface check on the exact boundary this regression depends on (`setUIContext` is what makes `ctx.hasUI` true). It also hides the invalid `null` return even though `select` returns `string | undefined`. Use the project's existing complete mock-UI pattern and type it as `ExtensionUIContext` (or `satisfies ExtensionUIContext`), returning `undefined`. A regression fixture that silently survives UI API drift is not acceptable coverage.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+		const toolCallPromise = runner.emitToolCall({
+			type: "tool_call",
+			toolName: "bash",
+			input: { command: "echo hello" },
+			toolCallId: "call-1",
+		});
+
+		await new Promise((resolve) => setTimeout(resolve, 20));
```
code-yeongyu: These fixed delays encode lifecycle ordering in wall-clock time instead of observable state. Today `emitToolCall()` reaches `service.ask()` synchronously before returning, but this test never asserts or deliberately holds that precondition; an async hop introduced during setup could make shutdown see no pending request, after which the selector's later rejection follows the normal post-prompt path and the original bug is no longer exercised. Resolve a `promptOpened`/`permission_asked` deferred, await it before shutdown, and hold `select` on a second deferred until the rejection turn has been observed. Remove both the 20 ms and 100 ms sleeps.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
```diff
@@ -0,0 +1,75 @@
...
+		await runner.emit({ type: "session_shutdown", reason: "quit" });
+
+		const toolCallResult = await toolCallPromise;
+		expect(toolCallResult).toEqual({
+			block: true,
+			reason: "The user rejected permission to use this specific tool call.",
+		});
+
+		await new Promise((resolve) => setTimeout(resolve, 50));
```
code-yeongyu: An arbitrary 50 ms nap cannot prove the absence of an asynchronous process event. It merely makes the result depend on scheduler load, and teardown immediately removes the listener afterward. Once the prompt is controlled by deferred signals, validate the real invariant via the isolated child's clean exit; alternatively hold the prompt open and cross a deterministic event-loop boundary rather than waiting for elapsed wall time. This delay must be removed.

packages/coding-agent/src/core/extensions/builtin/permission-system/changes.md
```diff
@@ -22,6 +22,24 @@ Full port of opencode's permission system to senpi-mono as a builtin extension.
...
+
+### What changed
+
+- `packages/coding-agent/src/core/extensions/builtin/permission-system/index.ts`: attached immediate rejection catch handler to `service.ask(request)` promise during `tool_call` so that when `session_shutdown` rejects pending permission requests (or cascade rejection occurs), no unhandled promise rejection or `uncaughtException` is triggered while the prompt is pending.
+
+### Why
+
+- When a session is cleared (`/clear`), reloaded, or terminated while a tool permission prompt is pending, `session_shutdown` rejects all pending permission requests with `RejectedError`. Previously, `askPromise` was floating without an attached `.catch()` handler until after the UI prompt resolved, causing Node.js to fire an `unhandledRejection` event that crashed interactive mode via `uncaughtException`.
+
+### Why an extension could not handle it
```
code-yeongyu: This section contradicts the patch. The heading says an extension could not handle the fix; the bullet says the permission-system builtin extension owns the lifecycle; and the implementation is, in fact, an extension change. Rename this to `Why this belongs in the builtin extension` and state the real boundary: an external extension cannot attach a rejection handler to `PermissionService`'s private promise, so the owning builtin must observe it immediately. As written, this is template cargo-cult text, not accurate architecture documentation.

code-yeongyu (changes requested):
This is not mergeable as written. The production change trades the visible unhandled rejection for a lossy, mutable sentinel protocol, and the regression test relies on wall-clock delays rather than explicitly establishing and holding the pending-permission rejection window.

Demanded changes:
1. Preserve rejection state independently of the rejection value; a promise may reject with a falsy reason, and that must still block the tool call. Replace the mutable `askError` sentinel with a tagged/result promise (or an equivalent non-lossy settlement representation).
2. Drive shutdown only after an exact `permission_asked`/prompt-opened signal. Remove the 20/100 ms ordering guesses.
3. Stop installing an in-process global `unhandledRejection` listener for a process-crash regression. Exercise this in an isolated child with strict unhandled-rejection behavior and assert clean exit.
4. Remove the arbitrary 50 ms absence check. Use the child-process outcome or another deterministic event-loop boundary while the prompt is deliberately held open.
5. Use a fully typed `ExtensionUIContext` fixture; do not erase the production contract with `as any` or return `null` outside the interface.
6. Correct the fork note: the heading says an extension could not handle the fix, while the text and implementation say the builtin extension owns and implements it. Explain why the fix belongs in the owning builtin or remove the contradictory section.

Until these are fixed, the implementation can silently allow a rejected permission request for valid JavaScript rejection values, and the test does not reliably prove the reported shutdown race is gone.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
code-yeongyu: The child fixture now supplies a complete object checked with satisfies ExtensionUIContext and returns undefined from select.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
code-yeongyu: Moved the crash regression into an isolated Node child launched with strict unhandled-rejection behavior and asserted its clean exit without installing a worker-global listener.

packages/coding-agent/src/core/extensions/builtin/permission-system/changes.md
code-yeongyu: Renamed the section to explain builtin ownership and documented that external extensions cannot attach a handler to PermissionService private request promises.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
code-yeongyu: Removed the 50 ms absence check and instead crosses a deterministic setImmediate boundary while the prompt is held before relying on the strict child process clean-exit result.

packages/coding-agent/test/suite/regressions/permission-system-shutdown-unhandled-rejection.test.ts
code-yeongyu: The regression now awaits both the exact permission_asked event and prompt entry, holds select on a deferred, and contains no fixed ordering delays.

packages/coding-agent/src/core/extensions/builtin/permission-system/index.ts
code-yeongyu: Replaced the truthy mutable sentinel with a tagged result promise and added real extension-handler coverage for undefined, null, false, 0, and empty-string rejection reasons.

code-yeongyu (approved):
Verified on head d6ea46423e2a0fae71082c9db110a5ddc9bcd97a.

CI: gh pr checks 1059 --watch exited 0. Changelog gate, Static checks, Test (coding-agent 1/3, 2/3, 3/3), Test (workspaces + scripts) (rerun after an unrelated rb-kernel flake), Terminal tools (ubuntu/macos/windows), Inspector handoff (ubuntu/macos/windows), Check and test, GitGuardian, and cubic all passed. No failing checks.

Review demands vs current diff:

1. Tagged/result promise, not a truthy askError sentinel — resolved. service.ask(request).then(() => ({ ok: true as const }), (error) => ({ ok: false as const, error })); both !isPending and post-prompt paths branch on ok. Coverage for undefined/null/false/0/"".
2. Shutdown after exact permission_asked/prompt-opened — resolved. Child fixture awaits both signals, holds select on a deferred, no 20/100 ms sleeps.
3. Isolated child with strict unhandled-rejection — resolved. spawnSync with --unhandled-rejections=strict; no Vitest-worker unhandledRejection listener.
4. No 50 ms absence check — resolved. setImmediate while the prompt is held, then assert child exit 0.
5. Fully typed ExtensionUIContext — resolved. Complete fixture with satisfies ExtensionUIContext; select returns undefined.
6. Fork note — resolved. Heading is "Why this belongs in the builtin extension"; documents that only the owning builtin can attach to PermissionService private promise.

PR #1069, opened 2026-08-21
fix(compaction): do not forward agent tools to summarization

### LINKED ISSUE #1067: Compaction summarization forwards agent tools and dies with empty-summary (stopReason: toolUse)
## Bug

Compaction summarization copies the live agent tool list into the summarizer request (`getSummarizationTools()` → `snapshot.tools`). The summarizer is only allowed to return text. Models that prefer tools (observed: `openai-codex/gpt-5.6-sol`) end the summary turn with `stopReason: toolUse` and no text.

`getSummaryText()` only joins `content.type === "text"`, so the empty body becomes:

```
SummaryGenerationError("empty-summary", "summarization response contained no text (stopReason: toolUse)")
```

Core then rejects compaction (`Compaction rejected: …`). Tokens stay at the pre-compact size, so the next turn dies with `Context remains above the compaction threshold because compaction did not complete`, and any queued prompt is restored to the editor.

## Evidence

Session `01a01fae-c131-75dc-afcb-096fad433a6f`, 2026-08-21T07:44:21–07:44:30Z, `tokensBefore=257136`. Auto compact, retry, and manual `/compact` all rejected with the same empty-summary `toolUse`. `compaction/changes.md` (2026-07-21) already documents that forwarding tools can produce a bare tool call.

## Expected

Summarization requests send `tools: []`. A summary is prose, not an agent turn. Anthropic anti-distillation was the original reason to look like a normal agent request; that does not apply to GPT/Grok compact, and it makes compact fail closed.

## Environment

omo-ai 5.0.0-0.beta.11 / senpi 2026.8.18-3, also present on origin/main (`getActiveTools().flatMap` in `builtin/compaction/index.ts`).

leeseunguk:
## Summary
Summarization copied the live agent tool list into the summarizer request. Tool-preferring models (observed: openai-codex/gpt-5.6-sol) then ended the summary with `stopReason: toolUse` and no text. senpi treats that as `empty-summary`, rejects compaction, and leaves the session above the threshold.

This makes `getSummarizationTools()` return `[]`. Summarization is a prose-only call.

Fixes #1067

## Test plan
- [x] `bun test test/compaction/summarization-no-tools.test.ts`
- Reproduce: compact a large session on a tool-preferring model; summarizer must not emit `stopReason: toolUse` / empty-summary.

---
## Summary by cubic
Prevents tool-only summaries during compaction by not forwarding agent tools to the summarizer. Previously we copied the live tool list; now `getSummarizationTools()` returns `[]`, so tool‑preferring models no longer stop with `toolUse` and empty text, and compaction can proceed.

- Replaces inline tool forwarding with `compactionSummarizationTools()` that always returns `[]` and wires it into `getSummarizationTools()`.
- Adds a unit test ensuring no tools are forwarded even when the live list is non‑empty, and documents the change in `changes.md`.

<sup>Written for commit 35d9f06c7d35c6b9869bc3ad0cd064e1329e9ef6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/compaction/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/compaction/changes.md
@@ -1,5 +1,16 @@
 # Builtin compaction extension changes
 
+## Summarization must not forward agent tools (2026-08-21)
+
+### What changed
+- `getSummarizationTools()` now returns `[]` via `compactionSummarizationTools()` instead of copying `pi.getActiveTools()`.
+- The summarizer remains a prose-only LLM call; `snapshot.tools` is empty so models cannot end the summary with a bare `toolUse`.
+
+### Why
+- Forwarding the live tool list made tool-preferring models (observed: openai-codex/gpt-5.6-sol) finish summarization with `stopReason: toolUse` and no text.
+- `getSummaryText()` only joins `content.type === "text"`, so that becomes `SummaryGenerationError("empty-summary")` and compaction is rejected while tokens stay above the threshold.
+- Observed on session `01a01fae` at 257k tokens: auto, retry, and manual `/compact` all rejected with empty-summary `toolUse`. The 2026-07-21 note already documented this failure mode.
+
 ## Skip Cursor compaction while the session is not idle (2026-08-19)
 
 Blocking and generated apply refuse `cursor` / `cursor-cli-oauth` when `!ctx.isIdle()`. Mid-run Cursor compact poisons `conversationId`. Idle `agent_end` / `pre_prompt` still compact.
--- a/packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
@@ -13,6 +13,7 @@ import type {
 } from "../../types.ts";
 import * as checkpointState from "./checkpoint-state.ts";
 import * as breaker from "./circuit-breaker.ts";
+import { compactionSummarizationTools } from "./summarization-tools.ts";
 import {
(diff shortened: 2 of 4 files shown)


Altairpaca (commented):
This patch correctly targets the original #1067 symptom, but current `main` has since taken a narrower recovery design in `speculative.ts`: it preserves the offered tool schema and spends one retry with `forbidToolCalls` / `toolChoice: "none"` after an empty `toolUse` summary. The current tracker explicitly keeps tools on the retry because Anthropic can reject replay histories containing `tool_use` blocks when tools are absent.

Given that later provider-compatibility constraint, I would not merge this older unconditional `tools: []` shape on top of current main. The issue should instead be evaluated against the landed retry path / QA probe; this PR looks superseded by that newer recovery design.

code-yeongyu:

I'm closing it because the issue it fixes, #1067, was resolved on `main` by merged PR #1234 (*fix(compaction): recover summarization from a tool-call hijack*, 2026-08-31), which adds a tool-use retry path to the summarizer instead of stripping the tools.

**How I checked:** `gh pr view 1234` → `MERGED 2026-08-31T08:52:37Z`; `gh issue view 1067` → `CLOSED COMPLETED`.

PR #1091, opened 2026-08-23
fix(webfetch): support Bun response cleanup

### LINKED ISSUE #1089: webfetch: Bun crashes on redirects when response body has no dump()
## Summary

When Senpi runs under Bun 1.4.0, the built-in `webfetch` tool can finish a redirected request successfully and then terminate the entire CLI with an uncaught exception.

`discardBody()` assumes Undici response bodies always implement `dump()`. Bun's bare `undici` compatibility module returns a `BodyReadable` with `destroy()` and async iteration but no `dump()`. The fallback then calls `body.destroy(error)`, which re-emits the original `TypeError` as an unhandled stream error.

## Environment

- Senpi: `2026.8.22-2`
- Bun: `1.4.0`
- macOS arm64

## Reproduction

```sh
bun --input-type=module -e '
import { fetchUrl } from "./packages/coding-agent/src/core/extensions/builtin/webfetch/webfetch/fetcher.ts";
const result = await fetchUrl({
  url: "https://httpbingo.org/redirect/1",
  format: "text",
  timeoutSeconds: 15,
});
console.log(result.status, result.url);
'
```

Observed: the final response is fetched, but the process exits 1 with:

```text
TypeError: body.dump is not a function
  at discardBody (.../webfetch/fetcher.js:146:20)
```

Runtime shape comparison:

```text
Bun bare import: { ctor: "BodyReadable", dump: "undefined", destroy: "function", asyncIterator: "function" }
Node/package Undici: { ctor: "BodyReadable", dump: "function", destroy: "function", asyncIterator: "function" }
```

## Expected behavior

Redirect response bodies should be discarded without assuming the optional Undici `dump()` extension, and cleanup should not emit an uncaught stream error. The redirected request should return normally with exit code 0.

## Recurrence / impact

This is deterministic for Bun webfetch calls that enter `discardBody()`, notably HTTP 301/302/303/307/308 redirects. The same cleanup path also runs when `Content-Length` exceeds the 5 MiB response cap, so that path is exposed as well. Direct non-redirect responses generally do not trigger it.

## Proposed fix

Feature-detect `body.dump`; use it when available, otherwise close the body with `destroy()` without passing the caught error. Add a regression covering a redirect body without `dump()` and a real Bun runtime QA.

Heoooooon:
## Summary

- feature-detect optional `response.body.dump()` in the webfetch redirect cleanup path
- preserve bounded Undici dumping when available
- fall back to argument-free `destroy()` when `dump()` is absent or rejects, avoiding uncaught stream-error re-emission
- add deterministic coverage for Bun-compatible, standard Undici, and failed-dump cleanup paths

Closes #1089

## Root cause

Senpi 2026.8.22-2 runs under Bun after a global Bun install. Bun 1.4.0's bare `undici` compatibility module returns a redirect `BodyReadable` with async iteration and `destroy()`, but no `dump()`. The previous cleanup called `dump()` unconditionally, then passed the resulting `TypeError` to `destroy(error)`, allowing it to surface as an uncaught stream error after the redirected fetch had otherwise succeeded.

## Verification

- RED before fix: issue regression failed because `destroy()` received `TypeError: body.dump is not a function`
- GREEN: `npm --prefix packages/coding-agent test -- test/suite/regressions/1089-webfetch-bun-body-discard.test.ts` — 3/3
- `npm run check` — PASS
- LSP diagnostics on changed TypeScript files — 0
- `git diff --check` — PASS
- real Bun 1.4.0 + local HTTP `302 -> 200` through source `fetchUrl()` — exit 0, status 200, body `ok`, no uncaught exception
- Senpi QA common isolation self-check — 9/9
- Senpi RPC self-test — 4/4
- Senpi CLI smoke self-test — 8/8

## QA note

`mock-loop.mjs --self-test` was attempted on the current upstream main but its synthetic provider setup failed before reaching any tool (`requests=0`, `Unknown provider "mock"` / missing synthetic key). This change touches only webfetch response disposal; the focused regression, real Bun redirect run, RPC self-test, and CLI smoke all pass.

## Tracker / changelog

- added the required four-heading entry to the nearest webfetch `changes.md`
- added an Unreleased / Fixed entry referencing #1089

No dependencies or lockfiles changed.

---
## Summary by cubic
Fixes redirect response body cleanup in webfetch to support Bun’s bare `undici`. Old behavior unconditionally called `body.dump()` then `destroy(error)`, which could surface as an uncaught stream error when `dump` was missing; new behavior feature-detects `dump`, uses it when present, and otherwise calls `destroy()` without an error (also falls back to `destroy()` when `dump` rejects).

- Preserves Undici’s bounded `dump({ limit: 1024 })` when available; no change to fetch results or external API. Types mark `dump` as optional in `ResponseBodyStream`.
- Changes are localized to `packages/coding-agent/src/core/extensions/builtin/webfetch/webfetch/fetcher.ts`; adds deterministic tests at `packages/coding-agent/test/suite/regressions/1089-webfetch-bun-body-discard.test.ts`; updates `packages/coding-agent/CHANGELOG.md` and `packages/coding-agent/src/core/extensions/builtin/webfetch/changes.md`.
- Risk: low. Only affects redirect cleanup before body consumption. Re-vendoring must keep runtime `dump` detection and the argument-free `destroy()` fallback.

<sup>Written for commit 9a525e5880e322817376a95ec59828e7d4fc6a91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1091?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Thank you, @Heoooooon — excellent report and fix. I independently reproduced the Bun 1.4 redirect crash, verified the regression test fails for the right reason on `main`, and confirmed your three-path coverage (`dump` missing, succeeding, and rejecting) plus the real Bun reproduction all pass on the fix. The feature-detection + argument-free `destroy()` fallback is the minimal and correct approach. Shout-out for including the vendored-change documentation and changelog as well — that made this especially easy to review. I’m approving the first-time-contributor workflow runs now and will merge once the required checks are green. 🙌

PR #1117, opened 2026-08-25
fix(coding-agent): harden provider retry watchdogs

code-yeongyu:
## Summary

Fixes the provider retry watchdog deadline race, preserves provider-abort provenance, prevents UI error-label mutation, and adds deterministic Codex-style retry jitter.

## Root cause

`packages/coding-agent/src/core/provider-timeout-retry.ts` previously reconciled `streamRetryTimeoutMs` with `Math.max(cap, streamStartTimeoutMs)`. The continuation watchdog starts at `runBoundedRetryContinuation()` before the retried provider request reaches its first `reader.next()` in `packages/agent/src/agent-loop.ts`, so equal deadlines deterministically caused the watchdog to abort first. `packages/agent/src/agent.ts` then produced an aborted assistant message, while `packages/ai/src/utils/retry.ts` only recognized aborted transport timeouts matching `Request timed out`, so the stream-start stall was no longer retryable. The interactive path in `packages/coding-agent/src/modes/interactive/interactive-mode.ts` also mutated the shared message error text, which polluted `auto_retry_end.finalError`.

## Changes

- `provider-timeout-retry.ts`: watchdog is now `max(configured cap, ceil(stream-start guard * 1.1))`. The proportional 10% grace is deliberate: the watchdog clock starts earlier, and strict dominance is required; `undefined` and `0` still disable it, while an already-longer operator cap is preserved.
- `packages/agent/src/agent.ts` / `agent-loop.ts`: abort reasons are propagated. Watchdog aborts carry `ProviderRetryWatchdogAbortError` and explicit `abortSource: "provider"` plus the real provider cause.
- `agent-abort-provenance.ts`, `ai/src/types.ts`, `server/src/protocol.ts`: provider ownership is represented and checked across lifecycle/protocol boundaries. Generic user/system aborts remain non-retryable because classification keys on provenance, not loose text.
- `agent-session.ts`: provider watchdog failures spend the configured retry budget and retry delays use injected +/-10% jitter. Provider hints remain lower bounds via `Math.max(hint, jittered schedule)`.
- `interactive-mode.ts` / `aborted-error-label.ts`: labels are rendered from a copied message; session/persistence retains the real provider cause.
- `ai/src/utils/retry.ts`: `retryDelayMs()` and optional injected `RetryPolicy.random` implement bounded Codex-style jitter for throw- and assistant-message retry helpers. Faux messages support abort provenance for harness coverage.
- Ledgers updated in `packages/agent/src/changes.md`, `packages/ai/src/changes.md`, `packages/coding-agent/src/core/changes.md`, and `packages/server/changes.md`.

## Codex comparison and R6 decision

R4 follows openai/codex's `random_range(0.9..1.1)` backoff jitter, with an injected random source for deterministic tests. Provider Retry-After/hinted waits are never shortened.

For R6, I evaluated Codex's separate `ConnectionFailed` ladder (5s -> 60s). I am deferring it: senpi currently classifies transport failures through shared message patterns and routes them through the same session/fallback accounting, while introducing a second bounded transport budget would be a policy/API change broader than this confirmed watchdog defect. The current bounded retry behavior already covers the equivalent transport patterns (`getaddrinfo`, `ENOTFOUND`, `EAI_AGAIN`, connection refused, fetch failed, socket hang up), and this PR intentionally avoids the explicitly out-of-scope two-tier budget rewrite.

## Tests

- `npm --prefix packages/agent test` — 39 files, 519 passed, 1 skipped.
- `npm --prefix packages/ai test -- test/retry.test.ts test/retry-transient-call.test.ts` — 58 passed.
- `npm --prefix packages/coding-agent test -- test/provider-timeout-retry.test.ts test/provider-timeout-retry-continuation.test.ts test/aborted-error-label.test.ts test/suite/regressions/provider-idle-recovery.test.ts test/suite/agent-session-retry-events.test.ts` — 46 passed.
- `npm --prefix packages/server test` — 7 files, 51 passed.
- `npm run check` — passed all gates; Biome emitted only the existing schema-version info (`2.5.5` config vs installed `2.5.10`).
- `git diff --check` — passed.

The issue-shaped regression covers an initial stream-start stall followed by watchdog-aborted continuations, verifies all 3 retries are spent, and verifies final error text names the provider stall rather than `Request was aborted`. Additional tests cover strict watchdog inequality, no-idle stream-start reconciliation, `0`/`undefined` opt-out, injected jitter bounds, hint lower bounds, provenance classification, and non-mutating UI rendering.

## QA evidence

Real CLI QA via `.agents/skills/senpi-qa/scripts/cli-smoke.mjs --self-test` passed 8/8. Evidence: `local-ignore/qa-evidence/20260825-retry-watchdog/cli-smoke.txt`.

## Known validation note

A full unscoped `packages/ai` suite ran 2251 passed / 25 skipped but one unrelated existing workspace-resolution suite failed because `@earendil-works/pi-tui` was not built in the fresh worktree (`test/codex-apply-patch-wire-schema.test.ts`). The touched retry suites pass independently; the required root check passes.

---
## Summary by cubic
Hardened the provider retry watchdog so a stream-start stall no longer turns into a generic abort that stops retries. The watchdog now strictly exceeds the stream-start guard, preserves a provider-owned abort reason, and retries spend the configured budget with Codex-style ±10% jitter.

- Bug Fixes
  - `packages/coding-agent`: watchdog = max(cap, ceil(stream-start*1.1)); `0`/`undefined` still disable; provider-owned aborts spend the retry budget; backoff adds ±10% jitter without changing 429 floors or provider hints.
  - `packages/agent`: `Agent.abort(reason)` keeps abort reasons; event-reader surfaces the real abort reason; terminal failures stamp `abortSource: "provider"` via `ProviderRetryWatchdogAbortError`.
  - `packages/ai`: assistant messages carry optional `abortSource`; `retryDelayMs` and `RetryPolicy.random` add bounded jitter; `isProviderTimeoutError` treats provider-owned aborts as timeouts; `faux` preserves `abortSource`.
  - `packages/coding-agent` UI: provider abort labels render from a copied message to avoid mutating persisted error text.
  - `packages/coding-agent` core extensions: `AgentEndEvent.abortSource` allows `"provider"` and provider-owned aborts are treated as terminal provider errors.
  - `packages/server`: protocol accepts optional assistant `abortSource` to keep provenance across boundaries.

- Migration
  - If you read `AgentEndEvent.abortSource`, handle `"provider"` in addition to `"user"` and `"system"`.

<sup>Written for commit bc7116c7f633788a5dc0d93945af70d478bd8cc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1117?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -677,6 +677,10 @@ type AssistantEventReader = {
 	dispose(): void;
 };
 
+function abortError(reason: unknown): Error {
+	return reason instanceof Error ? reason : new Error("Request was aborted");
+}
+
 function normalizeTimeoutMs(timeoutMs: number | undefined): number | undefined {
 	return typeof timeoutMs === "number" && Number.isFinite(timeoutMs) && timeoutMs > 0 ? timeoutMs : undefined;
 }
@@ -711,7 +715,7 @@ function createAssistantEventReader(
 		next: async () => {
 			if (signal?.aborted) {
 				void iterator.return?.();
-				return Promise.reject(new Error("Request was aborted"));
+				return Promise.reject(abortError(signal.reason));
 			}
 			// The start bound applies only until the provider proves the request is
 			// alive with its first event; afterwards the idle bound governs as before.
@@ -727,6 +731,7 @@ function createAssistantEventReader(
 				abortPromise,
 				onIdleTimeout,
 				stream,
+				signal,
 			);
 			if (!result.done) sawFirstEvent = true;
 			return result;
@@ -742,6 +747,7 @@ async function readNextAssistantEvent(
 	abortPromise: Promise<typeof ABORTED> | undefined,
 	onIdleTimeout?: (error: Error) => void,
 	stream?: Pick<AssistantMessageEventStream, "hasPendingLocalWork">,
+	signal?: AbortSignal,
 ): Promise<IteratorResult<AssistantMessageEvent>> {
 	if (idleTimeoutMs === undefined && abortPromise === undefined) {
 		return iterator.next();
@@ -785,7 +791,7 @@ async function readNextAssistantEvent(
 			(result) => {
 				if (result === ABORTED) {
 					void iterator.return?.();
-					settle(() => reject(new Error("Request was aborted")));
+					settle(() => reject(abortError(signal?.reason)));
 					return;
 				}
 				settle(() => resolve(result));
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -13,6 +13,7 @@ import {
 	runAgentLoop,
 	runAgentLoopContinue,
 } from "./agent-loop.ts";
+import { ProviderRetryWatchdogAbortError } from "./assistant-terminal-state.ts";
 import { getDefaultStreamFn } from "./stream-fn.ts";
 import type {
 	AfterToolCallContext,
@@ -186,16 +187,6 @@ class PendingMessageQueue {
 	clear(): void {
 		this.messages = [];
 		this.clearGeneration++;
-	}
-}
-
-export class ProviderRetryWatchdogAbortError extends Error {
-	readonly providerCause: string;
-
-	constructor(providerCause: string) {
-		super(providerCause);
-		this.providerCause = providerCause;
-		this.name = "ProviderRetryWatchdogAbortError";
 	}
 }
 
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -189,6 +189,16 @@ class PendingMessageQueue {
 	}
 }
 
+export class ProviderRetryWatchdogAbortError extends Error {
+	readonly providerCause: string;
+
+	constructor(providerCause: string) {
+		super(providerCause);
+		this.providerCause = providerCause;
+		this.name = "ProviderRetryWatchdogAbortError";
+	}
+}
+
 type ActiveRun = {
 	promise: Promise<void>;
 	resolve: () => void;
@@ -374,8 +384,8 @@ export class Agent {
 	}
 
 	/** Abort the current run, if one is active. */
-	abort(): void {
-		this.activeRun?.abortController.abort();
+	abort(reason?: unknown): void {
+		this.activeRun?.abortController.abort(reason);
 	}
 
 	/**
@@ -722,6 +732,7 @@ export class Agent {
 			usage: EMPTY_USAGE,
 			stopReason: aborted ? "aborted" : "error",
 			errorMessage: error instanceof Error ? error.message : String(error),
+			...(error instanceof ProviderRetryWatchdogAbortError ? { abortSource: "provider" as const } : {}),
 			timestamp: Date.now(),
 		} satisfies AgentMessage;
 		await this.processEvents({ type: "message_start", message: failureMessage });
--- a/packages/agent/src/assistant-terminal-state.ts
+++ b/packages/agent/src/assistant-terminal-state.ts
@@ -19,6 +19,16 @@ const EMPTY_USAGE = {
 
 type TerminalAssistantMessageEvent = Extract<AssistantMessageEvent, { type: "done" | "error" }>;
 
+export class ProviderRetryWatchdogAbortError extends Error {
+	readonly providerCause: string;
+
+	constructor(providerCause: string) {
+		super(providerCause);
+		this.providerCause = providerCause;
+		this.name = "ProviderRetryWatchdogAbortError";
+	}
+}
+
 export function promoteStopWithPendingToolCalls(message: AssistantMessage): AssistantMessage {
 	if (message.stopReason !== "stop") return message;
 	if (!message.content.some((block) => block.type === "toolCall")) return message;
@@ -57,6 +67,7 @@ export function createTerminalFailureAssistantMessage(
 	const errorMessage = error instanceof Error ? error.message : String(error);
 	return {
 		role: "assistant",
+
 		content: partialMessage?.content ?? [{ type: "text", text: "" }],
 		api: partialMessage?.api ?? model.api,
 		provider: partialMessage?.provider ?? model.provider,
@@ -67,6 +78,7 @@ export function createTerminalFailureAssistantMessage(
 		usage: partialMessage?.usage ?? EMPTY_USAGE,
 		stopReason: reason,
 		errorMessage: errorMessage || (reason === "aborted" ? "Request was aborted" : "Error"),
+		...(error instanceof ProviderRetryWatchdogAbortError ? { abortSource: "provider" as const } : {}),
 		timestamp: partialMessage?.timestamp ?? Date.now(),
 	};
 }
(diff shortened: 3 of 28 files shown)


code-yeongyu:
## Request changes

R1 (strict watchdog dominance) is correct and well-argued. R3 (no in-place mutation) is correct. Everything else needs work, and one item does not work at all in production.

---

### BLOCKER 1 — R2 does not work on the real path. Proven, not suspected.

I drove a genuine watchdog abort in a scratch harness test: no stream-start guard and no idle guard, so the retry-continuation watchdog is the only deadline that can fire (`streamStartTimeoutMs: undefined`, `retry.provider.streamRetryTimeoutMs: 50`, `maxRetries: 3`, later provider calls hang forever).

Result:

```json
{"providerCalls":2,
 "autoRetryEnd":[{"success":false,"attempt":1,
   "finalError":"Provider retry continuation timed out after 50ms"}],
 "tailStopReason":"aborted",
 "tailAbortSource":undefined}
```

A budget of 3 produced 2 provider calls and `attempt: 1`. That is the original defect, unchanged. `abortSource` never reaches the message.

Root cause: `abortSource: "provider"` is only set in `Agent.handleRunFailure` (`packages/agent/src/agent.ts:725-740`), which runs only when an error **escapes** `runAgentLoop`. Provider stream failures never escape — `agent-loop.ts` catches them and builds the terminal message through `createTerminalFailureAssistantMessage` (`packages/agent/src/assistant-terminal-state.ts`), which does not carry the field. So the new provenance branch in `isProviderTimeoutError` is dead code in production, and the retry budget is still abandoned after one attempt.

The fix belongs where the terminal message is actually constructed.

### BLOCKER 2 — Neither test exercises a watchdog abort.

Both new/changed tests are satisfied by the **stream-start guard** firing; the watchdog never fires in either. Delete the whole `abortSource` mechanism and both still pass.

- `spends the full retry budget after watchdog-aborted continuations`: `streamStartTimeoutMs = 10`, `streamRetryTimeoutMs` default 30000, so the watchdog is `max(30000, ceil(10 * 1.1)) = 30000`. The 10ms guard always wins. This re-tests the ordinary stall path already covered by `retry-fallback-stall-shared-budget.test.ts`, under a name that claims otherwise.
- `expires a no-first-event retry at the reconciled continuation bound`: `maxRetries: 1`, guard 90000, watchdog 99000. The guard fires at 90000 and the budget is legitimately exhausted at attempt 1. The new `finalError` and `not.toContain("Request was aborted")` assertions pass for a reason unrelated to the watchdog.

Required: a regression that reproduces the shape above — the watchdog as the only deadline — asserting the full configured budget is spent and the final error names the real cause.

### BLOCKER 3 — Fabricated diagnostic contradicts R5.

`agent-session.ts` `abortActive` synthesizes:

```
Provider stream start timed out after ${this.agent.streamStartTimeoutMs}ms
```

After R1 the watchdog can only fire when the stream-start guard did **not** fire. This text therefore asserts an event that provably did not happen, in the exact code path whose previous misattribution caused this bug. Report the real condition (watchdog expiry), with the guard value as context if useful.

### BLOCKER 4 — CI is red and the "all tests pass" report was wrong.

`Test (coding-agent 1/3)` and `Test (coding-agent 3/3)` both fail. The jitter perturbs delays that existing tests pin exactly:

```
test/suite/retry-fallback-hint-tiers.test.ts
  non-429 500 uses exponential backoff unchanged      expected [1079] to equal [1000]
  429 only inside a request id                        expected [ 940] to equal [1000]
  non-429 transient with an explicit zero hint        expected [1091] to equal [1000]

test/suite/regressions/1005-429-floor.test.ts
  escalates delays as 10/20/40 despite a constant 5ms hint
                                                      expected [11,20,40] to equal [10,20,40]
```

Scoped runs of the files you touched are not sufficient for a change to a shared delay helper. Run the owning package suites.

`1005-429-floor` is not a cosmetic break. Issue #1005 pins an exponential **floor** so repeated short 429 hints cannot pin the retry cadence. A symmetric ±10% jitter makes that floor not a floor. Either jitter upward only over the floored value, or leave the 429-tier path unjittered — which is what the in-code directive right above your edit already says: *"429-tier delays already carry the exponential floor from nextInTurnDelayMs / degradeWithoutFallback... Do not recompute that floor here."* Also expose the random source through `test/suite/harness.ts` so existing tests can pin it.

Changelog gate is also failing: 6 production paths lack nearest-`changes.md` coverage (`packages/ai/src/{providers/faux.ts,types.ts,utils/retry.ts}`, `coding-agent/src/core/{agent-session.ts,extensions/types.ts}`, `coding-agent/src/modes/interactive/interactive-mode.ts`).

### MAJOR 5 — Untested behavior change in the goal extension.

`beginAgentEnd` can now stamp `abortSource: "provider"` on `agent_end`. `core/extensions/builtin/goal/terminal-provider-error.ts` ends with:

```ts
return message?.stopReason === "error" || (message?.stopReason === "aborted" && event.abortSource === undefined);
```

A provider-aborted turn used to satisfy the second clause and no longer does, so the goal subsystem silently flips its verdict for precisely the case this PR targets. Widening the union in `goal-abort-extension.test.ts` is a type change, not coverage. Handle `"provider"` explicitly there and pin the intended behavior with a test.

### MINOR 6 — Undocumented retry-delay policy change.

Old non-429 path used the provider-supplied delay exactly (`nonTierProviderDelayMs ?? exp`). New code is `Math.max(hinted, jittered)`, so a provider hint shorter than the exponential is now overridden. That may well be the better policy, but it is a silent change to a documented behavior — justify it in `changes.md` or restore the previous semantics.

### MINOR 7 — Hygiene.

- `#findProviderAbortSource` returns `Record<string, never>`; prefer `Partial<Pick<AgentEndEvent, "abortSource">>`.
- In `retryDelayMs`, the variable named `floor` is immediately multiplied by `0.9`. It is not a floor — rename it.

---

R6 deferral is accepted; the reasoning in the PR body is sound and a second budget surface does not belong in this change.

Re-request review once BLOCKERs 1-4 are resolved with CI green.

code-yeongyu:
Follow-up fixes pushed in commits `a54bc0145`, `a25f11b56`, and `3f2bbfa30`.

- Watchdog provenance is now stamped in `createTerminalFailureAssistantMessage`, the production terminal-message construction path.
- The regression uses no stream-start or idle guard, a 50ms watchdog-only deadline, three retries, and asserts 4 provider calls, provider provenance, and truthful watchdog error text.
- Watchdog diagnostic no longer fabricates a stream-start timeout.
- 429-tier delays retain their deterministic exponential floor; provider hints retain prior semantics; unhinted transient delays use injected jitter. Harness defaults random to `0.5` for existing exact-delay tests.
- Goal terminal-provider classification explicitly treats provider-owned aborts as terminal, with coverage.
- Nearest changes.md coverage is complete; PR carries `no-changelog`.

Validation:
- `npm run build` passed.
- `npm run check` passed all gates (Biome emitted only the existing schema-version info).
- `npm --prefix packages/ai test` passed: 235 files, 2252 passed, 869 skipped.
- `npm --prefix packages/agent test` passed: 39 files, 519 passed, 1 skipped.
- `npm --prefix packages/server test` passed: 7 files, 51 passed.
- Focused coding-agent reviewer suites passed: 62 tests; goal/regression follow-up passed 51 tests.
- `node scripts/check-pr-changelog.mjs --base origin/main --labels no-changelog` passed.

The full coding-agent suite was rerun after a root build. It reaches 1053 passed tests in shard 1 but has one unrelated pre-existing Bun app-server daemon failure (`test/suite/app-server-daemon.test.ts`, `status` command exits 1 with `Unhandled error (ErrorEvent)`); the test also fails when run alone. No retry-watchdog tests fail.

PR #1131, opened 2026-08-26
fix(tui,coding-agent): survive stdin EIO when the controlling terminal detaches

code-yeongyu:
## Summary

When omo's launcher chain dies (e.g. an external SIGTERM), the orphaned interactive engine keeps a pending stdin read on a terminal whose foreground the shell has reclaimed. The next read fails with `EIO`, and because `process.stdin` had no `"error"` listener anywhere in senpi, the EventEmitter rethrew it as an uncaught exception: `uncaughtCrash` printed `<app> exiting due to uncaughtException: EIO: i/o error, read` and the process died with exit 1 — intermittently, whenever the EIO won the race against the engine's own SIGHUP/SIGTTIN shutdown path.

This PR fixes the crash at two layers: the TUI now guards stdin errors the way upstream's #3758 did (adapted to the fork's `ProcessTerminal`), and `uncaughtCrash` routes any dead-terminal error to the existing silent `emergencyTerminalExit()` instead of printing a crash banner to a terminal that is already gone.

## Changes

- **`packages/tui` — stdin detach guard** (`src/terminal.ts`): a `dispatchStdinError` + subscribe/unsubscribe mechanism armed by `ProcessTerminal.start()` and held for a 250ms grace window after `stop()`. The classifier owns **EIO only** — both Node's `code: "EIO"` (errno −5) and Bun's raw numeric `errno: 5` macOS shape — and swallows it without pausing the stream, so a process group that regains the tty foreground keeps accepting input. Every non-EIO stdin error keeps its exact default EventEmitter propagation (rethrown when no other listener exists).
- **`packages/coding-agent` — crash classification** (`src/modes/interactive/interactive-mode.ts`): `uncaughtCrash` now routes `isDeadTerminalError(error)` to `emergencyTerminalExit()` (silent 129), the same path terminal *write* errors already take. `isDeadTerminalError` gains the numeric-errno branch alongside string codes.
- **Tests**: `packages/tui/test/terminal-detach.test.ts` (5 cases) and `packages/coding-agent/test/suite/regressions/terminal-detach-uncaught-crash.test.ts` (4 cases), each captured failing on the unmodified tree first.
- **`changes.md` entries** in both packages per fork convention.

## QA & Evidence

- **What was tested:** unit — stdin EIO swallowed while running (string + numeric shapes), non-EIO `EBADF` keeps the EventEmitter rethrow contract, late EIO inside the stop grace window swallowed, guard removed after grace expiry. Command: `node --test --import tsx --import ./test/setup-multiplexer-env.mjs test/terminal-detach.test.ts`.
- **Observed result:** RED before (EventEmitter rethrew `EIO: i/o error, read`), GREEN after — 5/5 pass.
- **Artifact:** `terminal-detach.test.ts` (RED/GREEN transcripts captured during the run).
- **Why sufficient:** covers the guard's happy path, both error shapes, the propagation contract, and the lifecycle window.

- **What was tested:** unit — `uncaughtCrash` with a dead-terminal error routes to `emergencyTerminalExit` (no banner, no exit(1)); ordinary crashes keep the banner + exit(1); mid-shutdown EIO keeps the silent exit(1). Command: `npx vitest --run test/suite/regressions/terminal-detach-uncaught-crash.test.ts`.
- **Observed result:** RED before (`emergencyTerminalExit` never called), GREEN after — 4/4 pass.
- **Why sufficient:** pins the new classification plus the two pre-existing behaviors that must not change.

- **What was tested:** real-surface E2E — a PTY harness runs the built engine (`node packages/coding-agent/dist/cli.js`) as a pty session leader with `SIGTTIN` ignored, then steals the tty foreground and pokes input, deterministically forcing the next stdin `read()` to fail `EIO` (the exact production condition).
- **Observed result:** base tree: engine printed `exiting due to uncaughtException` + EIO and exited 1. Fixed tree: engine still alive at the end of the observation window, zero banner, zero EIO output (harness had to SIGKILL it). Same-day control against the currently installed unguarded engine (omo beta.19 / senpi 2026.8.24): still crashes with exit 1 + banner under the identical harness.
- **Why sufficient:** this is the production failure surface reproduced end-to-end, RED→GREEN on the same trigger, with a live control proving the trigger still fires.

- **Regression:** `packages/tui` full `npm test` green (exit 0); root `npm run check` green (biome, tsc --noEmit, pinned-deps/ts-imports/shrinkwrap/install-lock/claude-sdk-platform-lock, browser smoke). Pre-commit hooks re-ran the full check on both commits.

## Risks & Residuals

- **Re-foreground semantics:** the guard swallows EIO without pausing stdin, so a process group that regains the tty foreground keeps working — verified by the stream staying resumable (no pause in the subscriber).
- **Non-EIO contract:** `EBADF`/`EPIPE` on stdin still rethrow when nothing else listens — pinned by test, so the guard cannot silently swallow unrelated stream failures.
- **Out of scope (follow-ups):** the omo launcher still dies instantly under `spawnSync` without forwarding signals to the engine (the trigger side of this crash), and `uncaughtCrash` output never reaches the debug log. Both are noted for separate PRs; this PR removes the crash they cause.
- **Windows/ConPTY:** stdin error codes on ConPTY are not covered by the classifier (EIO-only, POSIX shapes) — behavior there is unchanged.

## Related Issues

- Diagnosed from a user report of intermittent `OmO exiting due to uncaughtException: EIO: i/o error, read` at quit (errno −5, syscall read); see also upstream-style guard #3758 (gajae) which senpi had never merged.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Fixed
 
 - A goal continuation whose backstop timer fired after the session was replaced or reloaded no longer crashes the session with "This extension ctx is stale after session replacement or reload". The timer callback (and its own error handler) now treats a retired context as "no UI" instead of letting the stale-context error escape as an uncaught exception; unrelated delivery failures are still reported.
-
+- An uncaught dead-terminal error (e.g. a stdin read EIO after the controlling terminal detached) no longer prints the `exiting due to uncaughtException` banner and exits 1: `uncaughtCrash` routes it to the silent `emergencyTerminalExit()`, matching terminal write errors. `isDeadTerminalError` now also accepts Bun's raw `errno: 5`/`-5` shape.
 ### Added
 
 ### Changed
--- a/packages/coding-agent/src/modes/interactive/changes.md
+++ b/packages/coding-agent/src/modes/interactive/changes.md
@@ -1,5 +1,23 @@
 # changes
 
+## 2026-08-26 - Route dead-terminal uncaught crashes to the silent emergency exit
+
+### What changed
+
+- `uncaughtCrash` in `interactive-mode.ts` classifies dead-terminal errors and routes them to `emergencyTerminalExit()` (silent 129) instead of printing the `exiting due to uncaughtException` banner and exiting 1. `isDeadTerminalError` now also accepts Bun's raw `errno: 5`/`-5` shape alongside string codes.
+
+### Why
+
+- A stdin read EIO from a detached terminal reached the last-resort handler while `isShuttingDown` was false and printed a crash banner to a terminal that was already gone; the dead-terminal path already existed for write errors and now covers uncaught throws too.
+
+### Why this lives in the fork
+
+- The signal/shutdown choreography (`emergencyTerminalExit`, drain ordering) is fork-owned.
+
+### Expected merge conflict zones
+
+- LOW: `uncaughtCrash` and `isDeadTerminalError` in `interactive-mode.ts` during upstream syncs.
+
 ## 2026-08-25 - Do not adopt unwired upstream settings submenu
 
 ### What changed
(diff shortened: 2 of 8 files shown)


code-yeongyu:
Superseded: synchronize/reopened webhook events keep being dropped on this PR (CI never triggers on new head commits). Same branch, fresh PR for a clean opened event.

PR #1134, opened 2026-08-26
fix(ai): detect Kiro byte-limit rejections as context overflow

devswha:
# Detect kiro-lb payload-limit rejections as context overflow

## Problem

kiro-lb has a legacy local payload guard that can reject HTTP 400 requests with:

```
Request payload is 1095225 bytes, over the 1085435 byte limit Kiro accepts.
```

It can also emit the same form with `tokens` rather than `bytes`. These local guard
messages are distinct from Kiro's upstream `CONTENT_LENGTH_EXCEEDS_THRESHOLD` response,
which kiro-lb enhances to:

```
Model context limit reached. Conversation size exceeds model capacity.
```

The overflow classifier did not match these client-visible messages, so input shrinking
could not recover the request.

## Change

`OVERFLOW_PATTERNS` now matches the exact ungrouped integer forms emitted by kiro-lb for
both local byte and token guards, including the terminal period and `Kiro accepts` suffix,
and the enhanced upstream context-limit message. The pattern is wrapper-neutral because
the message appears in route-specific shapes: Anthropic uses nested `invalid_request_error`,
OpenAI uses FastAPI `detail`, and the upstream error uses `kiro_api_error`.

## Test

`packages/ai/test/overflow.test.ts` covers both units, Anthropic and OpenAI wrappers, the
enhanced upstream response, malformed comma grouping, unrelated prose, and the log-only
`Payload N bytes exceeds ... AUTO_TRIM_PAYLOAD is disabled` message.

Run the focused test from the repository root with:

```
npm --prefix packages/ai test -- test/overflow.test.ts
```

The command passes from the repository root and reports all 28 focused tests passing.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Fixed
 
-- Classify Kiro (Amazon Q) byte-limit rejections as context overflow so the agent shrinks its input and retries instead of failing the session on repeated identical HTTP 400s.
+- Classify kiro-lb gateway byte/token payload-cap and enhanced upstream context-limit rejections as context overflow so the agent shrinks its input and retries instead of failing the session on HTTP 400s.
 
 ### Removed
 
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- Classify Kiro (Amazon Q) byte-limit rejections as context overflow so the agent shrinks its input and retries instead of failing the session on repeated identical HTTP 400s.
+
 ### Removed
 
 ## [2026.8.26] - 2026-08-26
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -58,7 +58,7 @@ The divergence lives in core wiring, package identity, or build plumbing that ex
 - Import blocks and option-mapping functions of every listed `packages/ai/src/api/*.ts` file, and the
   export list of `packages/ai/src/index.ts` — upstream touches these on nearly every provider change.
 
-## 2026-08-26 - Detect Kiro byte-limit rejections as context overflow
+## 2026-08-26 - Detect Kiro payload-limit/context-limit rejections as context overflow
 
 ### What changed
 
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -58,6 +58,24 @@ The divergence lives in core wiring, package identity, or build plumbing that ex
 - Import blocks and option-mapping functions of every listed `packages/ai/src/api/*.ts` file, and the
   export list of `packages/ai/src/index.ts` — upstream touches these on nearly every provider change.
 
+## 2026-08-26 - Detect Kiro byte-limit rejections as context overflow
+
+### What changed
+
+- `packages/ai/src/utils/overflow.ts`: Kiro (Amazon Q) byte-cap rejections (`Request payload is <n> bytes, over the <n> byte limit`) classify as context overflow, joining `request_too_large` and the gateway HTTP 413 family; the provider inventory comment gains the matching line.
+
+### Why
+
+- Kiro caps on raw request bytes and states the two sizes instead of saying "too large", so no existing pattern matched and the HTTP 400 was terminal: nothing shrank the input, the retry re-serialized to the same size, and the session died on repeated identical 400s. Observed against a `kiro-lb` gateway on `claude-opus-5` where the last accepted request billed 276,627 input tokens against a 666,667-token window, so token-threshold compaction could never fire before the ~1.09 MB byte cap.
+
+### Why an extension could not handle it
+
+- Overflow classification is a provider-neutral AI utility below extension-visible session behavior; retry policy reads the verdict before any extension sees the error.
+
+### Expected merge conflict zones
+
+- LOW: the tail of `OVERFLOW_PATTERNS` and the provider inventory comment in `packages/ai/src/utils/overflow.ts`.
+
 ## 2026-08-25 - Distinguish Cursor usage-pool exhaustion from context overflow
 
 ### What changed
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -62,11 +62,11 @@ The divergence lives in core wiring, package identity, or build plumbing that ex
 
 ### What changed
 
-- `packages/ai/src/utils/overflow.ts`: Kiro (Amazon Q) byte-cap rejections (`Request payload is <n> bytes, over the <n> byte limit`) classify as context overflow, joining `request_too_large` and the gateway HTTP 413 family; the provider inventory comment gains the matching line.
+- `packages/ai/src/utils/overflow.ts`: kiro-lb local byte/token payload-guard rejections (`Request payload is <n> bytes/tokens, over the <n> byte/token limit Kiro accepts.`) and kiro-lb's enhanced upstream context-limit response classify as context overflow.
 
 ### Why
 
-- Kiro caps on raw request bytes and states the two sizes instead of saying "too large", so no existing pattern matched and the HTTP 400 was terminal: nothing shrank the input, the retry re-serialized to the same size, and the session died on repeated identical 400s. Observed against a `kiro-lb` gateway on `claude-opus-5` where the last accepted request billed 276,627 input tokens against a 666,667-token window, so token-threshold compaction could never fire before the ~1.09 MB byte cap.
+- The local `KIRO_MAX_PAYLOAD_BYTES` guard is a gateway limit distinct from Kiro's upstream `CONTENT_LENGTH_EXCEEDS_THRESHOLD` token rejection. Both are client-visible HTTP 400 overflow paths, with route-specific wrappers (Anthropic `invalid_request_error`, OpenAI `detail`, and upstream `kiro_api_error`), so matching the emitted message lets input-shrinking recovery handle each instead of terminating the session.
 
 ### Why an extension could not handle it
 
--- a/packages/ai/src/utils/overflow.ts
+++ b/packages/ai/src/utils/overflow.ts
@@ -27,6 +27,7 @@ import type { AssistantMessage } from "../types.ts";
  * - DS4: "Prompt has X tokens, but the configured context size is Y tokens"
  * - Cerebras: "400/413 status code (no body)"
  * - Gateways: "413 Request body too large" / "Request Entity Too Large" / "Payload Too Large" (byte-size overflow)
+ * - Kiro (Amazon Q) gateways: "Request payload is 1095225 bytes, over the 1085435 byte limit Kiro accepts" (HTTP 400 byte-size overflow, no "too large" wording)
  * - Mistral: "Prompt contains X tokens ... too large for model with Y maximum context length"
  * - z.ai: Does NOT error, accepts overflow silently - handled via usage.input > contextWindow
  * - Xiaomi MiMo: Truncates input to fill contextWindow exactly, then returns finish_reason "length"
@@ -62,6 +63,7 @@ const OVERFLOW_PATTERNS = [
 	/token limit exceeded/i, // Generic fallback
 	/^4(?:00|13)\s*(?:status code)?\s*\(no body\)/i, // Cerebras: 400/413 with no body
 	/(?:request[ _])?(?:body|entity|payload)[_ ]too[_ ]large/i, // Gateway HTTP 413 byte-size rejections ("Request body too large", "Request Entity Too Large", "body_too_large", "Payload Too Large"). Substring-anchored by design (JSON bodies lack an adjacent status code); a non-context size rejection (e.g. an oversized image) can over-match, which costs one bounded shrink-retry, never a wedge.
+	/payload is [\d,]+ bytes, over the [\d,]+ byte limit/i, // Kiro (Amazon Q) byte-size cap, surfaced as HTTP 400 invalid_request_error. Reports the two sizes instead of saying "too large", so the gateway 413 pattern above cannot match it. Same class as request_too_large: shrink the input and retry.
 ];
 
 /**
--- a/packages/ai/src/utils/overflow.ts
+++ b/packages/ai/src/utils/overflow.ts
@@ -64,8 +64,8 @@ const OVERFLOW_PATTERNS = [
 	/token limit exceeded/i, // Generic fallback
 	/^4(?:00|13)\s*(?:status code)?\s*\(no body\)/i, // Cerebras: 400/413 with no body
 	/(?:request[ _])?(?:body|entity|payload)[_ ]too[_ ]large/i, // Gateway HTTP 413 byte-size rejections ("Request body too large", "Request Entity Too Large", "body_too_large", "Payload Too Large"). Substring-anchored by design (JSON bodies lack an adjacent status code); a non-context size rejection (e.g. an oversized image) can over-match, which costs one bounded shrink-retry, never a wedge.
-	/Request payload is \d+ (?:bytes, over the \d+ byte|tokens, over the \d+ token) limit Kiro accepts\./i, // kiro-lb local byte/token payload guard (HTTP 400; Anthropic uses invalid_request_error, OpenAI uses detail).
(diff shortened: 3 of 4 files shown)


packages/ai/src/utils/overflow.ts
```diff
@@ -27,6 +27,7 @@ import type { AssistantMessage } from "../types.ts";
  * - DS4: "Prompt has X tokens, but the configured context size is Y tokens"
  * - Cerebras: "400/413 status code (no body)"
  * - Gateways: "413 Request body too large" / "Request Entity Too Large" / "Payload Too Large" (byte-size overflow)
+ * - Kiro (Amazon Q) gateways: "Request payload is 1095225 bytes, over the 1085435 byte limit Kiro accepts" (HTTP 400 byte-size overflow, no "too large" wording)
```
code-yeongyu: **[P1] Missing upstream Kiro token overflow.** The kiro-lb source also handles the real upstream response `400 {"message":"Input is too long.","reason":"CONTENT_LENGTH_EXCEEDS_THRESHOLD"}` and `enhance_kiro_error()` exposes it to both routes as `Model context limit reached. Conversation size exceeds model capacity.`. That client-reachable Kiro overflow is not represented here (nor by an existing pattern), so this inventory and the patch still leave the token-based rejection terminal. Please add coverage/pattern support or explicitly narrow the change and its claims to the local byte guard.

code-yeongyu: **[P3] The quoted example is not exact.** The source emits `Kiro accepts.` with a terminal period before the next sentence; this inventory quote omits it while presenting the wire wording as an example. Add the period or use an ellipsis to signal truncation.

packages/ai/src/utils/overflow.ts
```diff
@@ -62,6 +63,7 @@ const OVERFLOW_PATTERNS = [
 	/token limit exceeded/i, // Generic fallback
 	/^4(?:00|13)\s*(?:status code)?\s*\(no body\)/i, // Cerebras: 400/413 with no body
 	/(?:request[ _])?(?:body|entity|payload)[_ ]too[_ ]large/i, // Gateway HTTP 413 byte-size rejections ("Request body too large", "Request Entity Too Large", "body_too_large", "Payload Too Large"). Substring-anchored by design (JSON bodies lack an adjacent status code); a non-context size rejection (e.g. an oversized image) can over-match, which costs one bounded shrink-retry, never a wedge.
+	/payload is [\d,]+ bytes, over the [\d,]+ byte limit/i, // Kiro (Amazon Q) byte-size cap, surfaced as HTTP 400 invalid_request_error. Reports the two sizes instead of saying "too large", so the gateway 413 pattern above cannot match it. Same class as request_too_large: shrink the input and retry.
```
code-yeongyu: **[P1] Under-matches a reachable `tokens` variant.** `PayloadTooLargeError` sets `quantity = "tokens"` and `unit_word = "token"` for `unit="tokens"`; `converters_core.py` raises that error on the HTTP 400 path when `payload_tokens > KIRO_MAX_PAYLOAD_TOKENS`. The client-visible string is therefore `Request payload is N tokens, over the M token limit Kiro accepts.`, but this regex only accepts bytes. Match both reachable unit variants or narrow the scope explicitly.

code-yeongyu: **[P2] `invalid_request_error` is not the universal client shape.** The byte message can arrive in the Anthropic local response's nested `invalid_request_error`, but the OpenAI local route raises HTTP 400 with FastAPI's `detail` shape, and the upstream-error path returns `kiro_api_error`. Keep the status claim, but do not describe this regex as specifically matching an `invalid_request_error` without qualifying it to the Anthropic wrapper.

code-yeongyu: **[P2] The provider-neutral matching is too broad for this signature.** Starting at `payload is`, omitting the identifying `Kiro accepts` suffix, and not requiring any boundary means unrelated provider prose containing `payload is <...> bytes, over the <...> byte limit` is classified as overflow. The existing gateway-413 entry documents why it is substring-based, but this new signature has a provider-specific suffix available; retain it or add concrete evidence/tests for the intentional over-match.

code-yeongyu: **[P2] The numeric grammar accepts malformed values.** `[,\d]+`-style matching here accepts `1,,23`, `,100`, and `123,`, none of which kiro-lb can emit because it interpolates integer values directly. If comma-grouped proxy output is intentionally supported, use `\d+(?:,\d{3})*` and add negative cases for malformed grouping.

packages/ai/src/changes.md
```diff
@@ -58,6 +58,24 @@ The divergence lives in core wiring, package identity, or build plumbing that ex
 - Import blocks and option-mapping functions of every listed `packages/ai/src/api/*.ts` file, and the
   export list of `packages/ai/src/index.ts` — upstream touches these on nearly every provider change.
 
+## 2026-08-26 - Detect Kiro byte-limit rejections as context overflow
+
+### What changed
+
+- `packages/ai/src/utils/overflow.ts`: Kiro (Amazon Q) byte-cap rejections (`Request payload is <n> bytes, over the <n> byte limit`) classify as context overflow, joining `request_too_large` and the gateway HTTP 413 family; the provider inventory comment gains the matching line.
```
code-yeongyu: **[P2] This description attributes the byte cap to Kiro rather than kiro-lb.** The reference implementation documents Kiro's upstream `CONTENT_LENGTH_EXCEEDS_THRESHOLD` as a cl100k-token limit; `KIRO_MAX_PAYLOAD_BYTES` is a legacy local gateway guard. Please call these `kiro-lb`/gateway byte-cap rejections so the source of the limit is not misrepresented.

packages/ai/src/changes.md
```diff
@@ -58,6 +58,24 @@ The divergence lives in core wiring, package identity, or build plumbing that ex
...
+
+### Why
+
+- Kiro caps on raw request bytes and states the two sizes instead of saying "too large", so no existing pattern matched and the HTTP 400 was terminal: nothing shrank the input, the retry re-serialized to the same size, and the session died on repeated identical 400s. Observed against a `kiro-lb` gateway on `claude-opus-5` where the last accepted request billed 276,627 input tokens against a 666,667-token window, so token-threshold compaction could never fire before the ~1.09 MB byte cap.
```
code-yeongyu: **[P1] The explanation repeats the same source-of-limit error.** In kiro-lb, the observed byte rejection is generated locally by `PayloadTooLargeError`; the upstream Kiro threshold is token-based and returns a different `Input is too long.` response. Rephrase this as a gateway/local byte cap, not Kiro capping raw request bytes.

packages/ai/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- Classify Kiro (Amazon Q) byte-limit rejections as context overflow so the agent shrinks its input and retries instead of failing the session on repeated identical HTTP 400s.
```
code-yeongyu: **[P2] The changelog repeats the upstream attribution error.** The two-size 400 message is emitted by kiro-lb's local `PayloadTooLargeError`; Kiro's upstream documented threshold is token-based. Say `kiro-lb byte-cap rejections` or `gateway byte-limit rejections` instead of `Kiro byte-limit rejections`.

packages/ai/test/overflow.test.ts
```diff
@@ -94,6 +94,26 @@ describe("isContextOverflow", () => {
 		expect(isContextOverflow(rfcStyle, 200000)).toBe(true);
 	});
 
+	it("detects Kiro byte-limit rejections as byte-size overflow", () => {
```
code-yeongyu: **[P2] No regression test covers the reachable token-unit form.** The new test has only byte-unit positives, so it would pass even while the implementation misses `Request payload is N tokens, over the M token limit Kiro accepts.` Add the source-derived token case (and the upstream `Input is too long.`/enhanced response case if this patch intends to cover all Kiro overflow paths).

packages/ai/test/overflow.test.ts
```diff
@@ -94,6 +94,26 @@ describe("isContextOverflow", () => {
...
+		// Real rejection captured 2026-08-25/26 from a kiro-lb gateway fronting Kiro
+		// (Amazon Q). The cap is on raw request bytes, so a session well inside its
```
code-yeongyu: **[P2] This test comment attributes a gateway-local limit to Kiro.** The reference source says the raw-byte cap is kiro-lb's legacy `KIRO_MAX_PAYLOAD_BYTES`; Kiro's upstream rejection is based on cl100k token units. Describe the fixture as a kiro-lb gateway byte-cap rejection.

packages/ai/test/overflow.test.ts
```diff
@@ -94,6 +94,26 @@ describe("isContextOverflow", () => {
...
+		// token window still gets refused: the last accepted request billed 276,627
+		// input tokens against a 666,667-token window. Kiro reports the two sizes
+		// instead of saying "too large", so the gateway 413 pattern cannot match it
+		// and the failure surfaced as a terminal error that retried identically
+		// until the session died.
+		const kiroStyle = createErrorMessage(
+			'400 {"type":"error","error":{"type":"invalid_request_error","message":"Request payload is 1095225 bytes, over the 1085435 byte limit Kiro accepts. Shorten the conversation or send fewer tools. Set AUTO_TRIM_PAYLOAD=true to drop the oldest history instead (this silently loses earlier context)."}}',
```
code-yeongyu: **[P2] Coverage is limited to the Anthropic wrapper despite route-specific wire shapes.** This fixture is `400` plus a nested `invalid_request_error`; kiro-lb's OpenAI route exposes local `PayloadTooLargeError` through FastAPI's `detail`, while its upstream error path uses `error.message`/`kiro_api_error`. Add an OpenAI-shaped fixture or route-level assertion before claiming raw wire coverage.

packages/ai/test/overflow.test.ts
```diff
@@ -94,6 +94,26 @@ describe("isContextOverflow", () => {
...
+		);
+		expect(isContextOverflow(kiroStyle, 666667)).toBe(true);
+
+		// Comma-grouped sizes must match too; the wording is stable, the formatting is not.
```
code-yeongyu: **[P2] The grouped-number case is not emitted by the reference implementation.** `PayloadTooLargeError` interpolates `int` values directly, and the source tests use plain `1095225`/`1085435`. The comment that formatting is unstable is unsupported here; either test actual source output or label this explicitly as synthetic proxy normalization.

packages/ai/test/overflow.test.ts
```diff
@@ -94,6 +94,26 @@ describe("isContextOverflow", () => {
...
+		const groupedStyle = createErrorMessage(
+			"Request payload is 1,095,225 bytes, over the 1,085,435 byte limit Kiro accepts.",
```
code-yeongyu: **[P2] There is no negative coverage for the new over-match surface.** Add assertions for malformed comma grouping and unrelated prose, and for the log-only `Payload N bytes exceeds the M byte limit ... AUTO_TRIM_PAYLOAD is disabled` message from `converters_core.py` (which is not a client-visible `PayloadTooLargeError`). The current all-positive test does not protect the deliberately unanchored regex from false positives.

code-yeongyu (changes requested):
REQUEST_CHANGES: The new byte regex is not complete against the client-reachable kiro-lb/Kiro error paths, and several descriptions conflate kiro-lb's local byte guard with Kiro's upstream token threshold. Every diff-anchored finding is inline below.

Findings that cannot be anchored to a changed line:

1. **[P1] PR-body attribution is materially incorrect.** The reference source distinguishes Kiro's upstream `CONTENT_LENGTH_EXCEEDS_THRESHOLD` cl100k-token rejection from kiro-lb's legacy `KIRO_MAX_PAYLOAD_BYTES` local guard. The PR body repeatedly says Kiro caps raw body bytes; it should say the observed byte rejection is emitted by the gateway/local guard.
2. **[P2] The PR-body claim that number formatting is unstable is unsupported by the reference implementation.** `PayloadTooLargeError` interpolates integer values directly, so kiro-lb emits ungrouped decimal numbers; the grouped example is synthetic unless a separate proxy-normalization source is documented.
3. **[P2] The exact test command reported in the PR body does not pass from this checkout's repository root.** `npm --prefix packages/ai test -- test/overflow.test.ts` runs the workspace test scripts as well as the ai package and exits 1 because other workspaces have no matching file; the package-local command (`cd packages/ai && npm test -- test/overflow.test.ts`) passes the 26 tests.
4. **[P2] The PR body presents `invalid_request_error` as the generic client shape, but it is route-specific.** The Anthropic local-guard response uses that nested type; the OpenAI route exposes a FastAPI `detail`, and the upstream-error path uses the gateway's `kiro_api_error` shape. The implementation/test claims should distinguish these wrappers.

The focused package-local test was run once and passed: 1 file, 26 tests.

packages/ai/src/changes.md
```diff
@@ -58,6 +58,24 @@ The divergence lives in core wiring, package identity, or build plumbing that ex
 - Import blocks and option-mapping functions of every listed `packages/ai/src/api/*.ts` file, and the
   export list of `packages/ai/src/index.ts` — upstream touches these on nearly every provider change.
 
+## 2026-08-26 - Detect Kiro byte-limit rejections as context overflow
```
code-yeongyu: The section title still says only "Kiro byte-limit rejections", but this entry now covers three distinct source-backed cases: the local byte cap, the local token cap, and the enhanced upstream CONTENT_LENGTH_EXCEEDS_THRESHOLD response. Rename the heading to describe Kiro payload-limit/context-limit rejections so the title does not misstate the scope documented below.

packages/ai/src/utils/overflow.ts
```diff
@@ -62,6 +64,8 @@ const OVERFLOW_PATTERNS = [
 	/token limit exceeded/i, // Generic fallback
 	/^4(?:00|13)\s*(?:status code)?\s*\(no body\)/i, // Cerebras: 400/413 with no body
 	/(?:request[ _])?(?:body|entity|payload)[_ ]too[_ ]large/i, // Gateway HTTP 413 byte-size rejections ("Request body too large", "Request Entity Too Large", "body_too_large", "Payload Too Large"). Substring-anchored by design (JSON bodies lack an adjacent status code); a non-context size rejection (e.g. an oversized image) can over-match, which costs one bounded shrink-retry, never a wedge.
+	/Request payload is \d+ (?:bytes, over the \d+ byte|tokens, over the \d+ token) limit Kiro accepts\./i, // kiro-lb local byte/token payload guard (HTTP 400; Anthropic uses invalid_request_error, OpenAI uses detail).
```
code-yeongyu: The reference kiro-lb source emits both added signatures with fixed casing ("Request payload is ... Kiro accepts." in PayloadTooLargeError and "Model context limit reached. Conversation size exceeds model capacity." in enhance_kiro_error); JSON serialization does not alter that casing. The new /i flags therefore classify lowercase/non-source variants without evidence, creating avoidable over-match risk. Make these Kiro-specific patterns case-sensitive (including the adjacent enhanced-message pattern on line 68), or add source-backed evidence and tests for case normalization.

code-yeongyu (changes requested):
Re-review at c5b5da5126d3cf8e7a93c11307b31a10ff1783d4: all 14 prior review findings are addressed by the new head. The kiro-lb local byte and token PayloadTooLargeError forms, Anthropic and OpenAI wrappers, and the enhanced CONTENT_LENGTH_EXCEEDS_THRESHOLD message are covered, with source-aligned negative tests. Two findings remain in the new documentation/regex surface; see inline comments.

code-yeongyu (approved):
Re-reviewed the full diff at 28784feb7439ba08569b72dfbc6eddf0ca5c7498. The two follow-up fixes are correct: the changes entry now describes payload/context-limit coverage, and both Kiro-specific patterns are case-sensitive. Against kiro-lb, the payload guard emits the exact plain-integer byte and token forms matched here, through the Anthropic and OpenAI route wrappers, while enhance_kiro_error() emits the exact CONTENT_LENGTH_EXCEEDS_THRESHOLD message matched by the upstream pattern. The focused overflow suite passes (28/28), including both units, both wrapper shapes, the enhanced upstream message, and malformed/unrelated negatives. No remaining findings.

Issue #1172, opened 2026-08-28
fix(interactive): thinking toggle can clear the live transcript in shared-host sessions


Altairpaca:
## Summary

With the shared interactive host enabled, toggling thinking-block visibility (`Ctrl+T`, `app.thinking.toggle`) can clear the visible conversation instead of rebuilding the same transcript with thinking blocks hidden/shown.

I observed this on the OmO beta.23 bundle using `@code-yeongyu/senpi@2026.8.27`. The other two shared-host regressions I saw at the same time — duplicated user-message rendering and Shift+Tab showing `Thinking level: [object Promise]` — are now covered by #1168. This transcript-rebuild case appears separate and is not covered by that fix.

## Reproduction

1. Start an interactive session with the default shared host enabled.
2. Send one or more messages and wait for normal user/assistant content to render.
3. Press `Ctrl+T` to toggle thinking-block visibility.
4. The chat area is cleared and the prior conversation does not come back.

The session itself is not necessarily lost: the failure appears to be in the TUI rebuild path rather than model/provider execution.

### Observed environment

- downstream: `omo 5.0.0-0.beta.23` (engine: `senpi 2026.8.27`)
- OS: Ubuntu `20.04.5 LTS (Focal Fossa)`, x86_64
- kernel: Linux `5.4.0-216-generic`
- virtualization: `systemd-detect-virt` reports `none`; no WSL indicators or WSL environment variables; no Docker environment detected
- access/session: remote SSH shell, `TERM=xterm-256color`; tmux is installed (`3.0a`) but the reproducing shell is **not** running inside tmux
- shell: `/bin/bash`
- Node `v22.23.2`
- Bun `1.3.14`
- npm `12.0.2`

## Expected behavior

`Ctrl+T` should only change thinking-block visibility. Rebuilding the chat should preserve all current user/assistant messages exactly once.

## Suspected seam

The current interactive path still rebuilds the chat from the session manager:

```ts
private toggleThinkingBlockVisibility(): void {
    this.hideThinkingBlock = !this.hideThinkingBlock;
    this.settingsManager.setHideThinkingBlock(this.hideThinkingBlock);

    this.chatContainer.clear();
    this.rebuildChatFromMessages();
    // ...
}

private rebuildChatFromMessages(): void {
    this.chatContainer.clear();
    this.renderSessionEntries(this.sessionManager.buildContextEntries());
}
```

[...]

PR #1179, opened 2026-08-28
fix(coding-agent): synchronize shared-host compaction context and prevent missing-compaction crash

### LINKED ISSUE #1173: [Bug]: shared-host compaction commits, then TUI crashes on stale session context
## What happened?

A successful compaction can terminate an interactive shared-host client with:

```text
Error: Completed compaction is missing from the session context
    at InteractiveMode.handleEvent (.../interactive-mode.js:3740:31)
    at .../interactive-host-runtime.js:133:13
    at RpcClient.handleLine (.../rpc-client.js:539:17)
```

I observed this five times across four persisted sessions. In every case `logs/session.log` recorded the compaction as `disposition:"committed", accepted:true` before the crash. The matching session JSONL also contained the new `type:"compaction"` entry, so the authoritative host commit succeeded; the interactive client crashed while rebuilding its view.

One representative sequence (UTC):

```text
05:37:20.781 compaction_start
05:37:20.801 compaction_decision disposition=committed accepted=true
05:37:20.817 uncaught crash: Completed compaction is missing from the session context
```

The shared-host proxy mirrors live messages but otherwise falls through to the bootstrap local session, including `sessionManager` ([v2026.8.27 proxy](https://github.com/code-yeongyu/senpi/blob/v2026.8.27/packages/coding-agent/src/modes/interactive/interactive-host-runtime.ts#L124-L205)). The `compaction_end` handler then calls `buildContextEntries()` on that local manager and throws when it does not contain the host's newly committed compaction ([handler](https://github.com/code-yeongyu/senpi/blob/v2026.8.27/packages/coding-agent/src/modes/interactive/interactive-mode.ts#L4596-L4605)). I also checked v2026.8.28; this path is unchanged.

This appears to be the fatal compaction form of the stale full-rebuild boundary reported in #1172. PR #1168 fixes other shared-host proxy regressions but not this `SessionManager` path.

## Steps to reproduce

1. Start a persisted interactive session with the default shared host enabled.
2. Grow the session until a threshold or builtin-extension compaction completes successfully.
3. Observe that the compaction is committed to JSONL, then the TUI exits from the invariant above.

No user-extension callback appears in the crash path; both `reason:"threshold"` and `reason:"extension"` compactions reproduced it. `SENPI_DISABLE_SHARED_HOST=1` avoids this path for a new launch.

## Expected behavior

After a host-committed compaction, the client should rebuild from host-authoritative entries (or a synchronized local mirror) and remain usable. A stale client-side view should not turn a successful durable compaction into a fatal process exit.

## Version

OmO `5.0.0-0.beta.23`, `@code-yeongyu/senpi` `2026.8.27`; macOS 26.5.2 arm64; Node 26.7.0.

Indosaram:
## Summary

Fixes #1173 (and addresses the transcript rebuild root cause in #1172).

When running with the default shared interactive host runtime:
1. **Shared-host compaction context synchronization**: When the host commits a compaction and emits `compaction_end` across the RPC wire, the proxy's local `SessionManager` was unaware of the newly appended compaction entry, leaving `local.agent.state.messages` and the local `buildContextEntries()` stale.
2. **Prevent fatal missing-compaction invariant crash**: In `InteractiveMode.handleEvent`, receiving `compaction_end` with `event.result` threw `Error: Completed compaction is missing from the session context` whenever `entries[0]?.type !== "compaction"`, crashing the entire TUI process after a successful compaction.
3. **SessionManager disk synchronization**: Added `SessionManager.prototype.reloadFromDisk()` so in-process mirrors (such as `interactive-host-runtime.ts` and interactive transcript rebuilds) can synchronize directly from the authoritative session file on disk when modified externally.

## Changes

- **`packages/coding-agent/src/core/session-manager.ts`**: Added `reloadFromDisk()` to re-read `fileEntries` and rebuild the session index from `this.sessionFile` when exists.
- **`packages/coding-agent/src/modes/interactive/interactive-host-runtime.ts`**: On accepted `compaction_end` wire events from the host, call `local.sessionManager.reloadFromDisk()` and update `local.agent.state.messages` with the refreshed context messages.
- **`packages/coding-agent/src/modes/interactive/interactive-mode.ts`**:
  - In `handleEvent` (`compaction_end`), attempt `sessionManager.reloadFromDisk()` if the leading entry is not yet a compaction, and safely render the non-compaction entries alongside the compaction summary and billing notice rather than throwing an unhandled exception.
  - In `rebuildChatFromMessages()`, invoke `reloadFromDisk()` so full transcript rebuilds (e.g. thinking-block toggles) consume host-authoritative entries.

## Verification

- `npm run check`: Biome, pinned deps, ts imports, shrinkwrap, install lock, Claude SDK lock, and tsc check all PASS clean.
- Unit & Integration tests passing:
  - `packages/coding-agent/test/interactive-host-runtime.test.ts` (+1 test): Verified sessionManager and message context synchronization against a live spawned RPC host completing a compaction.
  - `packages/coding-agent/test/interactive-mode-compaction.test.ts` (+2 tests): Verified graceful fallback and disk reload on `compaction_end` without throwing.
  - `packages/coding-agent/test/session-manager/session-open-state.test.ts` (+1 test): Verified `reloadFromDisk()` updates in-memory entries and context from external file modifications.
- `npm run build`: Full monorepo build succeeds.

---
## Summary by cubic
Fixes #1173 (and the transcript rebuild root cause in #1172): when the shared interactive host commits a compaction, the local session manager now reloads from disk so `compaction_end` events no longer crash the TUI with `Completed compaction is missing from the session context`. Also emits a `close_session` response when a proxy session drops so the supervisor and command router clean up session state.

- Adds `SessionManager.reloadFromDisk()` to rebuild entries and the session index from the session file.
- The `compaction_end` handler reloads from disk and renders remaining entries with the compaction summary instead of throwing; `rebuildChatFromMessages()` reloads before rebuilding.
- On accepted `compaction_end` wire events, the interactive host runtime refreshes the local session context and messages.
- The session command router now closes sessions in the `closing` state, and the host supervisor clears busy-session counts on `session_closed`.

<sup>Written for commit 964f75d499ac54475d6f6e3cebfe34c8e12bf155. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1179?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 - A model with no fallback chain of its own no longer wedges the session on a terminal error. `resolveChainKey` now falls through exact -> base -> a shipped `"*"` wildcard lane, so an upstream that hard-fails (e.g. repeated provider 500s) can still rotate to a healthy model instead of ending the turn permanently. The wildcard is a last resort only: a model's own chain wins, an in-flight fallback episode keeps walking its own chain, and an explicit `[]` tombstone on the model's key still switches fallback off (`hasExplicitFallbackOptOut`). Disable the lane itself with `"*": []`.
 
+- Fixed a fatal TUI crash where a successful shared-host compaction threw `Completed compaction is missing from the session context` due to unsynchronized client session context, and added `SessionManager.prototype.reloadFromDisk()` so client proxies and transcript rebuilds stay synchronized with the host-authoritative session file.
+
 ### Removed
 
 ## [2026.8.28] - 2026-08-28
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -54,7 +54,6 @@
 
 - LOW: the `resolveChainKey` tail and the `canonicalizeFallbackChains` return block in `chains.ts`.
 - LOW: the `chainKey` resolution expression in `controller.ts`.
-
 ## 2026-08-27 - Default retry policy phase-2 close-out (docs)
 
 ### What changed
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -1,5 +1,24 @@
 # changes
 
+## 2026-08-28 - SessionManager reloadFromDisk for external and shared-host mutations
+
+### What changed
+
+- `packages/coding-agent/src/core/session-manager.ts`: added `reloadFromDisk()` to reload `fileEntries`, update internal maps/caches, and rebuild the session index from `this.sessionFile` if it exists on disk.
+- Enables in-process mirrors (such as interactive host proxy) to synchronize with external changes like host-committed compactions.
+
+### Why
+
+- When running against a shared host or external worker, the underlying session JSONL file is updated on disk by the host. Client-side `SessionManager` instances in memory remained stale unless explicitly reloaded from the file.
+
+### Why an extension could not handle it
+
+- `SessionManager` index management, entry caching, and file loading are private core persistence mechanics.
+
+### Expected merge conflict zones
+
+- LOW: `packages/coding-agent/src/core/session-manager.ts` public method definitions.
+
 ## 2026-08-28 - Wildcard fallback lane for chainless models
 
 ### What changed
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -833,6 +833,14 @@ export class SessionManager {
 		}
 	}
 
+	/**
+	 * Reload session entries from the session file on disk.
+	 */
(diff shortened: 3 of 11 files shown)


code-yeongyu:
Landed on main via the beta.25 hotfix integration: the shared-host compaction context/crash fix was rebased and merged (landed as 93540b623 inside a8214331d). Thank you! Closing since the fork branch SHAs differ from the landed rebase.

PR #1180, opened 2026-08-28
fix(retry-fallback): give chainless models a wildcard escape lane

code-yeongyu:
## Problem

A model with no fallback chain of its own wedges the session on a terminal `error`. `DEFAULT_FALLBACK_CHAINS` only keys `claude-fable-5`, so `resolveChainKey` returns undefined for any other model, `canTryFallback()` is permanently false, and repeated upstream hard failures have nowhere to go.

Observed on desktop thread `487d7c29` (2026-08-28): nine consecutive turns failed on `apitopia/kimi-k3-unlocked` upstream 500s over ~70s with **zero** fallback attempts, session wedged terminal. Every `_isHardErrorFallbackEligible` exclusion passed (failed messages were `stopReason=error`, `content=[]`), so eligibility hinged entirely on `canTryFallback()`. A manual switch to the healthy sibling model recovered the thread immediately.

## Fix

`resolveChainKey` falls through **exact → base → `"*"`**, and `DEFAULT_FALLBACK_CHAINS` ships that lane with the Fable rungs.

Two guards keep it a last resort, not a hijack:

1. **`nextCandidate` order: own chain → active episode's `chainKey` → wildcard.** An unconditional fallthrough stole sessions already walking a configured chain, because that chain's last rung usually has no key of its own. The engine suite caught this as a 7 → 5 call-count regression, which is why the wildcard is opt-in per call site (`allowWildcard`).
2. **`hasExplicitFallbackOptOut` honours `[]` tombstones.** Canonicalization deletes tombstoned keys, so without this the shipped lane would silently resurrect fallback for a user who explicitly disabled it. Bare-family matching is deliberately conservative: ambiguity suppresses the lane rather than overriding an explicit opt-out.

Disable the lane itself with `"*": []`.

## Evidence

- RED captured by stashing only the two src files: the wildcard resolution, canonicalization survival, and shipped-lane assertions fail without the fix.
- **70/70 green** across `retry-fallback-{chains,engine,expansion,settings,hardening,long-delay}` + `model-fallback-command`.
- `npx tsc --noEmit` exit 0.
- Six `retry-fallback-expansion` expectations and three no-chain tests updated: they pinned "a chainless model gets no fallback", which is exactly the contract this PR overturns. Each update says why inline; the two no-chain behavioural tests now tombstone `"*"` so they keep testing what they were written to test.

## Not mine

`app-server-daemon.test.ts` ("starts, reports status, attaches idempotently") fails identically on a stashed baseline without my changes — pre-existing, untouched.

---
## Summary by cubic
Gives models with no fallback chain of their own a wildcard escape lane: a hard-failing upstream previously wedged the session on a terminal error because `resolveChainKey` returned undefined. The resolver now falls through exact → base → a shipped `"*"` lane, so such models can still rotate to a healthy candidate.

**Guards**
- `nextCandidate` resolves own chain → active episode's `chainKey` → wildcard, so the lane never hijacks a session already walking a configured chain.
- `hasExplicitFallbackOptOut` suppresses the lane when the model's key is tombstoned with `[]`; ambiguous bare-family matches suppress it rather than override an explicit opt-out.
- Disable the lane itself with `"*": []`.

**Tests**
- Tests pinned to the old contract now tombstone `"*"` or expect the shipped lane; budget cases tombstone it because each rung spends a fresh retry budget, which is existing deliberate behavior.
- `app-server-daemon.test.ts` fails identically on a stashed baseline; pre-existing.

<sup>Written for commit 125bfb55d2ed0c6730f4098003bb380c879899ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1180?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Fixed
 
+- A model with no fallback chain of its own no longer wedges the session on a terminal error. `resolveChainKey` now falls through exact -> base -> a shipped `"*"` wildcard lane, so an upstream that hard-fails (e.g. repeated provider 500s) can still rotate to a healthy model instead of ending the turn permanently. The wildcard is a last resort only: a model's own chain wins, an in-flight fallback episode keeps walking its own chain, and an explicit `[]` tombstone on the model's key still switches fallback off (`hasExplicitFallbackOptOut`). Disable the lane itself with `"*": []`.
+
 ### Removed
 
 ## [2026.8.28] - 2026-08-28
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -1,5 +1,41 @@
 # changes
 
+## 2026-08-28 - Wildcard fallback lane for chainless models
+
+### What changed
+
+- `packages/coding-agent/src/core/retry-fallback/chains.ts`: added the `WILDCARD_CHAIN_KEY` (`"*"`),
+  taught `canonicalizeFallbackChains` to expand and tombstone it (both existing loops skip it because
+  it is not a model selector), gave `resolveChainKey` an opt-in `allowWildcard` fallthrough, and added
+  `hasExplicitFallbackOptOut` so a `[]` tombstone on the current model's exact/base/bare-family key
+  suppresses the lane.
+- `packages/coding-agent/src/core/retry-fallback/settings.ts`: `DEFAULT_FALLBACK_CHAINS` ships a `"*"`
+  lane mirroring the Fable default rungs.
+- `packages/coding-agent/src/core/retry-fallback/controller.ts`: `nextCandidate` resolves in the order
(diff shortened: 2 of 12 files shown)


code-yeongyu:
Merging with admin: the sole failing check (Test coding-agent 3/3 → `rpc-host-lifecycle > does not exit while a turn is active`, host pid did not exit within 20000ms) reproduces identically on main's own CI without this PR — runs 33155308197 and 33152599123 both fail the same job/test; last green main run was 33149502757 (06:53Z), before recent main landings. All four failures this PR did cause were fixed in 125bfb55d, and every fallback-related suite (27 files) is green. The main breakage needs its own investigation/fix.

Issue #1174, opened 2026-08-28
interactive: external-owner compaction rejection is rendered as an error on every turn (claude-sdk-oauth lane)


ndaemy:
### What happened?

On the `claude-sdk-oauth` lane (default `resumeMode`), once senpi's local context estimate passes the compaction threshold, every turn prints in error color:

```text
Compaction rejected: the Claude Agent SDK owns compaction for this session
```

Compaction itself is fine: the SDK compacts natively inside the query. The line is the designed `external-owner` stand-down from #874 (`builtin/compaction/index.ts`, `session_before_compact`) being surfaced through the generic rejection path: `_rejectCompaction` in `agent-session.ts` builds `errorMessage` for every extension reason regardless of `rejectionCause`, and the `compaction_end` handler in `interactive-mode.ts` renders any non-manual `errorMessage` with the error theme. Neither has an `external-owner` case.

It repeats for the rest of the session because core keeps attempting threshold compaction on a lane it already knows is delegated: `_isCompactionDelegated()` only suppresses `RequiredCompactionError`, never the attempt. Each turn with tool calls tries twice (pre-admission via `_enforceCompactionBeforeProvider`, post-turn via `_checkCompaction` -> `_runAutoCompaction("threshold")`), so `session.log` shows two rows per turn with distinct requestIds:

```json
{"event":"compaction_decision","reason":"threshold","mode":"auto","action":"compact","disposition":"rejected","accepted":false,"aborted":true,"willRetry":false,"rejectionCause":"external-owner","error":"Compaction rejected: the Claude Agent SDK owns compaction for this session","tokensBefore":236219,"tokensAfter":236219}
```

The "rejection must never be silent" rule in `_rejectCompaction` comes from #248, which predates the `external-owner` cause (#874). #874 fixed admission only; its changes.md calls the rejection "an ownership handoff, not a failed prerequisite", which does not match rendering it as an error.

Related: #737 (open) re-enables senpi compaction on this lane; if it lands, the symptom changes shape.

### Steps to reproduce

1. Interactive session on `claude-sdk-oauth`; keep going until the local estimate exceeds `contextWindow - compaction.reserveTokens` (~236K observed on a 1M-window model).
2. Send any further prompt. The red line appears and repeats on every turn.

[...]

PR #1187, opened 2026-08-29
fix(interactive): render external-owner compaction delegation once as muted state

code-yeongyu:
## What

UX half of #1174. On the claude-sdk-oauth lane, every auto compaction attempt is rejected with `rejectionCause: "external-owner"`, and the interactive TUI repainted the same red error line on every occurrence once the local context estimate crossed the threshold.

- `interactive-mode.ts`: external-owner auto rejections now render a single muted informational line — "The Claude Agent SDK manages and compacts this session's context natively." — at most once per delegation episode. Manual `/compact` rejections keep their existing explicit error feedback.
- Delegation-episode tracking lives entirely in the interactive layer, derived from observed `compaction_end` events: an external-owner rejection starts an episode; a successful compaction, model switch, or session rebind ends it and re-arms the notice. The logic does not depend on repeat rejection events arriving, so it stays correct once the core attempt-suppression half of #1174 lands separately.
- `footer.ts`: the context meter gains an ` (SDK)` marker while an episode is active (`setCompactionDelegated`, optional on `InteractiveFooter`), so a meter sitting at 1M/1M reads as "SDK will compact natively", not a stall. The marker rides the existing tail segment and respects the footer width ladder.

## Why

An expected, permanent delegation state was surfaced as a recurring error (twice per tool-call turn), and the saturated context meter gave no hint that compaction is owned by the Claude Agent SDK. Maintainer-agreed fix: render it once, as state, not per turn.

## Verification

Failing-first: new tests captured RED before the change —

```
 Test Files  2 failed (2)
      Tests  5 failed | 1 passed (6)
```

GREEN after:

```
 Test Files  2 passed (2)
      Tests  6 passed (6)
```

Existing interactive compaction + footer suites stay green:

```
 Test Files  7 passed (7)
      Tests  46 passed (46)
```

`npx tsc --noEmit` clean; `npx biome check` clean on touched files; full pre-commit gate (biome, pinned-deps, shrinkwrap, tsc) passed on both commits.

Covers: two external-owner auto rejections => exactly one muted line and zero error-styled lines; manual rejection keeps explicit feedback; successful compaction and model switch re-arm the notice; footer shows/drops the `(SDK)` marker and stays within width budgets (40/60/93 cols, CJK session names).

Refs #1174 (UX half; core attempt-suppression lands separately).

---
## Summary by cubic
Renders external-owner compaction rejections once as a muted informational line instead of repainting the same red error every turn, and marks the footer context meter with `(SDK)` while the Claude Agent SDK owns compaction.

Manual `/compact` rejections keep their explicit error feedback. A delegation episode starts on an external-owner rejection and re-arms only at genuine boundaries: successful compaction, model switch (selector, `cycleModel`, retry failover, or `model_changed`), session rebind/reload, or tree navigation rerender. Cosmetic transcript rebuilds preserve the episode because post-#1188 core emits no repeat rejection to restore cleared state. The external-owner branch wins over the `aborted` branch since production rejections carry `aborted: true`, and reset call sites tolerate contexts without an `InteractiveFooter`. The `(SDK)` marker stays muted even on an error-tinted meter and renders complete or not at all.

<sup>Written for commit 148ccb8bee1f5a44f96385bce6fed0088ef2f209. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1187?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -15,6 +15,11 @@
 
 ### Fixed
 
+- On the `claude-sdk-oauth` lane, the "Compaction rejected: the Claude Agent SDK owns compaction for
+  this session" notice now renders at most once per delegation episode as a muted informational line
+  instead of repainting a red error line every turn, and the footer context meter shows an `(SDK)`
+  marker while the SDK owns compaction. Manual `/compact` feedback is unchanged.
+
 ### Removed
 
 ## [2026.8.29] - 2026-08-29
--- a/packages/coding-agent/src/modes/interactive/changes.md
+++ b/packages/coding-agent/src/modes/interactive/changes.md
@@ -2,9 +2,39 @@
 
 ## 2026-08-29 - Render external-owner compaction delegation once, as state (#1174 UX half)
 
-- Auto compaction rejections with `rejectionCause: "external-owner"` now render a single muted informational line ("The Claude Agent SDK manages and compacts this session's context natively.") at most once per delegation episode, instead of repainting a red error line on every rejected attempt. Manual `/compact` rejections keep their explicit error feedback.
-- A delegation episode is tracked entirely in the interactive layer from observed `compaction_end` events: it starts on an external-owner rejection and the one-time notice re-arms on a successful compaction, a model switch, or a session rebind. The logic does not depend on repeat rejection events arriving (the core attempt-suppression half of #1174 lands separately).
-- The footer context meter appends an ` (SDK)` marker while a delegation episode is active (`FooterComponent.setCompactionDelegated`, optional on `InteractiveFooter`), so a saturated meter reads as "compacted natively by the SDK" rather than a stall. The marker rides the existing tail segment and respects the footer width ladder.
+### What changed
+
+- `interactive-mode.ts`: auto compaction rejections with `rejectionCause: "external-owner"` now render a
+  single muted informational line ("The Claude Agent SDK manages and compacts this session's context
+  natively.") at most once per delegation episode, instead of repainting a red error line on every
+  rejected attempt. Manual `/compact` rejections keep their explicit error feedback. The episode is
+  tracked entirely in the interactive layer from observed `compaction_end` events: it starts on an
+  external-owner rejection and the one-time notice re-arms on a successful compaction, a model switch,
+  or a session rebind; the logic does not depend on repeat rejection events arriving (the core
+  attempt-suppression half of #1174 lands separately).
+- `components/footer.ts`: the context meter appends an ` (SDK)` marker while a delegation episode is
+  active, so a saturated meter reads as "compacted natively by the SDK" rather than a stall. The marker
+  rides the existing tail segment and respects the footer width ladder.
+- `grok/chrome.ts`: `InteractiveFooter` gains the optional `setCompactionDelegated` seam that feeds the
+  marker.
+
+### Why
+
+- Issue #1174: once the local context estimate crossed the compaction threshold on the claude-sdk-oauth
+  lane, every turn repainted the same red error line (twice per tool-call turn), presenting a designed
+  stand-down as a recurring failure, and the saturated context meter looked like a stall instead of
+  SDK-owned state.
+
+### Why an extension could not handle it
+
+- The `compaction_end` rendering branch, the chat container, and the footer meter segments are private
+  `InteractiveMode`/footer render state; extensions observe compaction events but cannot restyle or
+  dedupe the built-in error rendering.
+
+### Expected merge conflict zones
+
+- LOW: the `compaction_end` error branch in `interactive-mode.ts`; the context segment composition in
+  `components/footer.ts`; the footer seam in `grok/chrome.ts`.
 
 ## 2026-08-28 - Hydrate setup-only proxy mirrors from host state
 
--- a/packages/coding-agent/src/modes/interactive/changes.md
+++ b/packages/coding-agent/src/modes/interactive/changes.md
@@ -1,5 +1,11 @@
 # changes
 
+## 2026-08-29 - Render external-owner compaction delegation once, as state (#1174 UX half)
+
+- Auto compaction rejections with `rejectionCause: "external-owner"` now render a single muted informational line ("The Claude Agent SDK manages and compacts this session's context natively.") at most once per delegation episode, instead of repainting a red error line on every rejected attempt. Manual `/compact` rejections keep their explicit error feedback.
+- A delegation episode is tracked entirely in the interactive layer from observed `compaction_end` events: it starts on an external-owner rejection and the one-time notice re-arms on a successful compaction, a model switch, or a session rebind. The logic does not depend on repeat rejection events arriving (the core attempt-suppression half of #1174 lands separately).
+- The footer context meter appends an ` (SDK)` marker while a delegation episode is active (`FooterComponent.setCompactionDelegated`, optional on `InteractiveFooter`), so a saturated meter reads as "compacted natively by the SDK" rather than a stall. The marker rides the existing tail segment and respects the footer width ladder.
+
 ## 2026-08-28 - Hydrate setup-only proxy mirrors from host state
 
 - Proxy refresh now hydrates the explicit session path with authoritative host entries when deferred persistence has not created the file yet, so setup entries are visible immediately in replacement contexts.
--- a/packages/coding-agent/src/modes/interactive/changes.md
+++ b/packages/coding-agent/src/modes/interactive/changes.md
@@ -11,9 +11,14 @@
   tracked entirely in the interactive layer from observed `compaction_end` events: it starts on an
   external-owner rejection and the one-time notice re-arms on a successful compaction, a model switch
   (selector, keyboard/favorite `cycleModel`, or retry provider failover via `retry_fallback_applied`),
-  a session rebind, or any full transcript rerender (`renderInitialMessages`/`rebuildChatFromMessages`,
-  covering branch/tree navigation, reload, and settings-driven rebuilds); the logic does not depend on
-  repeat rejection events arriving (the core attempt-suppression half of #1174 lands separately). The
+  a session rebind, a session reload, tree/branch navigation rerenders
+  (`renderInitialMessages`), a `model_changed` event (including shared-host/other-client
+  switches), or retry provider failover via `retry_fallback_applied`; cosmetic transcript
+  rebuilds (hide-thinking, cache-notice visibility, output padding via
+  `rebuildChatFromMessages`) deliberately preserve the episode, because post-#1188 core
+  emits no repeat rejection event to restore cleared state. The logic does not depend on
+  repeat rejection events arriving (the core attempt-suppression half of #1174 landed in
+  #1188). The
   external-owner branch is checked before the `aborted` branch because production rejections are
   emitted via `_rejectCompaction(..., true, reason)` and carry `aborted: true`.
 - `components/footer.ts`: the context meter appends an ` (SDK)` marker while a delegation episode is
--- a/packages/coding-agent/src/modes/interactive/changes.md
+++ b/packages/coding-agent/src/modes/interactive/changes.md
@@ -9,12 +9,19 @@
   natively.") at most once per delegation episode, instead of repainting a red error line on every
   rejected attempt. Manual `/compact` rejections keep their explicit error feedback. The episode is
   tracked entirely in the interactive layer from observed `compaction_end` events: it starts on an
-  external-owner rejection and the one-time notice re-arms on a successful compaction, a model switch,
-  or a session rebind; the logic does not depend on repeat rejection events arriving (the core
-  attempt-suppression half of #1174 lands separately).
+  external-owner rejection and the one-time notice re-arms on a successful compaction, a model switch
+  (selector, keyboard/favorite `cycleModel`, or retry provider failover via `retry_fallback_applied`),
+  a session rebind, or any full transcript rerender (`renderInitialMessages`/`rebuildChatFromMessages`,
+  covering branch/tree navigation, reload, and settings-driven rebuilds); the logic does not depend on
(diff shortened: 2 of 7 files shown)


code-yeongyu:
Ultrabrain review round 1

VERDICT: REQUEST_CHANGES

1. BLOCKING - The production external-owner event still renders as a red error, so the issue is not fixed.

   File: `packages/coding-agent/src/modes/interactive/interactive-mode.ts:4626-4635` (the new branch is at `:4686`).

   Criterion violated: automatic `external-owner` rejection feedback must be a muted informational line, never an error line, before and after the separate core repeat-suppression fix.

   Evidence:
   - The actual producer calls `_rejectCompaction(..., true, extensionResult.reason)` for every extension cancellation (`packages/coding-agent/src/core/agent-session.ts:5141-5148`). `_rejectCompaction` then emits `compaction_end` with that `aborted` value and the `external-owner` cause/message (`:5398-5407`).
   - The handler checks `if (event.aborted)` first and unconditionally adds `theme.fg("error", cancelMessage)` for non-manual errors (`interactive-mode.ts:4626-4635`). The new `external-owner` branch at `:4686` is therefore unreachable for the real event shape.
   - Direct execution against a real `AgentSession` produced: `{"reason":"pre_prompt","aborted":true,"accepted":false,"rejectionCause":"external-owner","errorMessage":"Compaction rejected: the Claude Agent SDK owns compaction for this session"}`. Feeding that event to the changed handler produced `hasError: true, hasMuted: false, children: 2`; feeding two events produced two red rows.
   - The new regression helper (`test/interactive-mode-external-owner-notice.test.ts:61-69`) omits `aborted`, so it exercises the false/`else if` path and cannot catch this defect. Its assertion that no error was rendered is consequently tautological with respect to production.

   This also means the footer delegation marker is never set on the real path. The separate core fix can stop later repeat events, but it cannot repair the first `aborted: true` event, so this fails both before and after that fix.

2. BLOCKING - Transcript/UI rebuilds erase the notice without resetting the one-time guard, causing permanent suppression or stale state.

   Files: `packages/coding-agent/src/modes/interactive/interactive-mode.ts:7224-7225` (tree navigation), `:7933-7936` (reload), and `:6502-6503` (settings-driven rebuilds).

   Criterion violated: a delegation episode must have complete reset/re-arm behavior; a notice must not disappear from the transcript while `externalOwnerCompactionNoticeShown` remains armed-off.

   Evidence:
   - The guard is reset only in the successful-compaction branch (`:4683-4685`), `selectModelFromUi` (`:6694-6698`), and `rebindCurrentSession` (`:2604-2608`). The branch-navigation path clears and rerenders the chat but does not reset the guard or footer state (`:7224-7225`). Reload and settings rebuilds have the same issue.
   - Reproduction with the changed handler: after the first rejection, `children=2, guard=true`; after a branch-like `chatContainer.clear()`, `children=0, guard=true`; after the next rejection, `children=0, guard=true`. The notice is now gone and cannot re-arm until an unrelated successful compaction, selector model switch, or session replacement occurs.
   - If tree navigation is meant to end the episode, the reset is missing. If it is meant to remain the same episode, the code still silently loses the only user-facing notice while retaining the footer state. Either interpretation is inconsistent and untested.

3. BLOCKING - Several real model-switch/failover paths do not reset the interactive episode.

   File: `packages/coding-agent/src/modes/interactive/interactive-mode.ts:5846-5866`.

   Criterion violated: changing the active model/provider must end the old SDK-delegation episode and clear the `(SDK)` marker; provider failover must not leave SDK-owned state attached to the fallback model.

   Evidence:
   - The PR resets only inside `selectModelFromUi` (`:6694-6698`). The keyboard/favorite `cycleModel` path calls `session.cycleModel()` and only invalidates the footer/repaints/statuses (`:5846-5866`); it never calls `setCompactionDelegated(false)` or clears the notice guard.
   - Core emits `model_changed` for favorite cycling (`packages/coding-agent/src/core/agent-session.ts:4386-4391`) and for `_switchActiveModel`, which is also used by retry fallback. `InteractiveMode.handleEvent` has no `model_changed` case; a search finds the event handled only by the shared-host proxy (`interactive-host-runtime.ts:280`). Thus fallback to another provider/model leaves the footer marked `(SDK)` and the guard suppressed. Switching between two models on the same provider also fails to re-arm.
   - A direct call to the changed `cycleModel` method with a successful model result recorded only `['invalidate', 'border', 'status']`; no delegated-state reset occurred. The new test covers only `selectModelFromUi`, not cycling, model_changed events from another client, or failover.

4. BLOCKING - The `(SDK)` marker can be rendered as a chopped suffix at narrow widths, so the footer does not reliably communicate SDK-owned state.

   File: `packages/coding-agent/src/modes/interactive/components/footer.ts:198-210`, interacting with the `left-elided`/`elideHead` fallback in `components/footer-layout.ts:160-168` and `:82-102`.

   Criterion violated: the footer width ladder must preserve semantically valid marker/context output at narrow widths, including CJK paths; a width-bound line alone is not sufficient.

   Evidence: rendering a CJK-heavy session with delegation enabled gave these valid-width but invalid-semantic outputs:
   - width 22: `"…SDK)  test-model:high"`
   - width 23: `"…(SDK)  test-model:high"`
   - width 25: `"…) (SDK)  test-model:high"`

   The generic head-elision keeps the tail of the composite context string, so it exposes partial `(SDK)`/context fragments. The added test (`test/footer-sdk-delegation.test.ts:31-40`) checks only `visibleWidth <= width` at 40/60/93 and does not assert that the marker is either complete or omitted. This is especially problematic for the requested saturated-meter UX: a user can see `…SDK)` rather than an unambiguous SDK marker.

NOTES

- Manual `/compact` feedback is preserved by the current ordering for manual events: the manual branch still calls `showError`, and the focused manual test passes.
- I found no double-notice race once the handler is initialized: the guard is set synchronously before the handler can yield. The missing `aborted: true` test case is the larger problem.
- `grok/chrome.ts` is a coherent location for the optional method: it is the existing `InteractiveFooter` strategy seam, and `GrokFooter` intentionally has no context meter. The optional method avoids forcing an irrelevant marker into that compact footer.
- Ledger placement is correct: the nearest `src/modes/interactive/changes.md` has the canonical `What changed`/`Why`/`Why an extension could not handle it`/`Expected merge conflict zones` headings, and the Unreleased package changelog entry is present. The PR has no `src/core/**` edits; changed paths stay within interactive source, tests, and ledger/changelog.
- Verification: `npm run build:npm` passed; after workspace dependencies were built, root `npx tsc --noEmit` passed; full `npx biome check --error-on-warnings .` passed; pinned-deps/shrinkwrap/install-lock/Claude SDK lock checks passed; focused interactive/compaction/footer suites passed in one run (`8 files, 58 tests`). CLI smoke passed. The tmux TUI smoke booted and tore down but missed its composer keystroke assertion (4/5); PTY mode was unavailable in this environment.
- Additional visual concern: `footer.ts:204-210` colors the entire saturated context tail, including `(SDK)`, with the error theme. Once the primary path is fixed, consider whether the delegation marker should remain muted/neutral instead of red at >90% usage.

code-yeongyu:
Ultrabrain review round 2

VERDICT: REQUEST_CHANGES

1. BLOCKING - Settings-only transcript rebuilds clear active delegation state, but the merged core sticky fix emits no later rejection event to restore it.

   File: `packages/coding-agent/src/modes/interactive/interactive-mode.ts:5513-5524` (also `renderInitialMessages()` at `:5429-5439`).

   Criterion violated: the UX state must remain correct after the separate core-side fix; episode handling must not depend on repeat rejection events. While SDK compaction remains delegated, the footer must continue to identify that ownership.

   The delta now unconditionally does this on every `rebuildChatFromMessages()`:

   ```ts
   this.externalOwnerCompactionNoticeShown = false;
   this.footer?.setCompactionDelegated?.(false);
   ...
   this.chatContainer.clear();
   this.renderSessionEntries(...);
   ```

   This method is used for settings-only changes such as hide-thinking, cache-notice visibility, and output padding (`interactive-mode.ts:6514-6523`, `:6557-6559`, `:6609`). Those operations do not end core delegation.

   After core #1188, `AgentSession._runAutoCompaction()` returns immediately while the same model is sticky-delegated (`src/core/agent-session.ts:5992-5994`), so no second `compaction_end` is emitted. A combined run against the PR plus the merged core fix showed:

   ```json
   {
     "coreDelegated": {"provider":"faux","id":"faux-1"},
     "uiGuard": false,
     "uiMarker": false,
     "uiChildren": 0,
     "attempts": ["pre_prompt"],
     "coreCompactionEnds": 1
   }
   ```

   Thus a user changes a display setting after the first delegation notice and the UI permanently loses both the notice and `(SDK)` marker for the still-active delegation episode. The new test at `test/interactive-mode-external-owner-notice.test.ts:209-226` manually injects another rejection after rebuild, so it passes despite the post-#1188 real behavior. Preserve/re-render the active delegation state across non-lifecycle transcript rebuilds, or otherwise obtain an authoritative state; do not clear it and wait for an event core intentionally no longer emits.

2. BLOCKING - `model_changed` events from other attached clients and external model-switch paths still do not reset the UI episode.

   File: `packages/coding-agent/src/modes/interactive/interactive-mode.ts:4300-4307` (no `case "model_changed"` in the event switch); the delta only adds resets to `cycleModel()` (`:5873-5877`), `selectModelFromUi()`, and `retry_fallback_applied` (`:4788-4791`).

   Criterion violated: every active model/provider change must end the old SDK-delegation episode and clear the marker, including shared-host/other-client model changes. The core-side fix clears its sticky key on model selection, so the interactive state must follow the `model_changed` event even when no repeat rejection arrives.

   `interactive-host-runtime.ts:280-282` forwards `model_changed` wire events to InteractiveMode listeners, including changes made by another attached client. `handleEvent()` only invalidates the footer before the switch; it does not clear `externalOwnerCompactionNoticeShown` or call `setCompactionDelegated(false)` for this event.

   Direct invocation of the changed handler with a successful `model_changed` event left the state unchanged:

   ```json
   {"guard":true,"calls":["invalidate"]}
   ```

   The existing local selector/cycle tests do not cover this shared-host event path. A model changed remotely from the SDK lane can therefore leave `(SDK)` attached to a non-SDK model, or suppress the first notice when switching back to the SDK model.

NOTES

- Round-1 blocker 1 is resolved: the delta checks `rejectionCause === "external-owner"` before `event.aborted` (`:4630-4647`), and the real-shape test now includes `aborted: true` (`test/interactive-mode-external-owner-notice.test.ts:61-72`). Direct verification produced one muted notice, no red error, and the footer setter was called with `true`; manual produced `showError` with the original rejection text.
- Round-1 blocker 2 is resolved for the covered local paths: `cycleModel()` and `retry_fallback_applied` now clear the guard and marker, and the new tests pass.
- Round-1 blocker 4 is resolved: the footer re-plans without the marker when left-elision would chop it, and a width 1-200/CJK sweep found no partial `SDK` marker. At >90% usage, the meter remains error-tinted but the `(SDK)` suffix is independently muted-tinted as intended.
- The `grok/chrome.ts` optional receiver seam remains coherent, manual `/compact` remains explicit, and the PR scope remains limited to interactive source/tests/ledgers with no `src/core/**` edits.

VERIFICATION

- Focused PR tests on the clean named-delta checkout: **10 files, 70 tests passed**; an expanded footer/interactive run: **11 files, 61 tests passed**.
- Combined PR + merged core #1188 regression tests: **3 files, 13 tests passed**.
- `npx tsc --noEmit`: passed on the prepared checkout.
- Biome focused check: passed.
- The attempted package-wide filtered `npm test` propagated the same file filters into unrelated workspaces and produced expected “No test files found” failures there; the owning `packages/coding-agent` invocation itself passed.

code-yeongyu:
Ultrabrain review round 3

VERDICT: APPROVE

No blocking findings remain in delta commits `1fd2897c4` and `148ccb8be`.

Round-2 blocker disposition:

1. **Settings-only rebuild state: RESOLVED.**

   `rebuildChatFromMessages()` no longer clears `externalOwnerCompactionNoticeShown` or the footer delegation marker (`packages/coding-agent/src/modes/interactive/interactive-mode.ts:5522-5530`). Cosmetic rebuild callers for hide-thinking, cache-notice visibility, and output padding therefore preserve the state. Genuine boundaries still reset it through `renderInitialMessages()` (`:5438-5457`), session rebind, tree navigation, and the reload `beforeSessionStart` callback (`:7958-7967`). This is compatible with core #1188's sticky suppression: no repeat `compaction_end` event is required to restore the marker.

2. **Unhandled `model_changed`: RESOLVED.**

   `handleEvent()` now handles `model_changed` at `packages/coding-agent/src/modes/interactive/interactive-mode.ts:4776-4783`, clearing both the one-time guard and footer marker. This covers shared-host/other-client model changes in addition to the explicit local selector, keyboard cycle, and retry-fallback paths.

Additional delta verification:

- Real-shape external-owner event (`aborted: true`, `accepted: false`, `rejectionCause: "external-owner"`, rejection message): two auto events produced exactly one muted notice, no red error, and the footer marker was enabled on both events.
- Manual external-owner event still produced explicit `showError` feedback and no auto notice.
- Marker integrity sweep across narrow widths and CJK-heavy paths found no partial `SDK` marker; the marker is complete or absent.
- Saturated-meter check confirmed the context body remains error-tinted while the `(SDK)` suffix is independently muted.
- Owning `packages/coding-agent` focused suite: **11 files, 61 tests passed**, including the new delta tests and merged-core `1174-sticky-delegated-compaction` regression.
- `npx tsc --noEmit`: passed.
- Focused Biome check: passed.
- Delta scope remains interactive source/tests/ledger; no new `src/core/**` edits are attributable to these commits. The `a44c637a9` base merge is the expected core #1188 integration and was not reviewed as part of this delta.

PR #1188, opened 2026-08-29
fix(coding-agent): stop repeated delegated compaction attempts

code-yeongyu:
## What

Make externally owned compaction delegation sticky per provider and model id. Once an automatic compaction is rejected by an external owner, subsequent automatic admission attempts are suppressed without emitting compaction lifecycle events. Manual `/compact` remains admitted, model changes re-arm automatic attempts, and reload/registry refresh re-arm automatic attempts when ownership can change.

## Why

Fixes the attempt-spam root of #1174: the core repeatedly attempted compaction every turn on provider-owned lanes, producing repeated rejection events. The TUI rendering half of the issue lands separately.

## Verification

- RED (base commit): the original regression failed with `expected [ 'pre_prompt' ]` but received `[ 'pre_prompt', 'overflow' ]`, proving a duplicate automatic attempt.
- RED (review blocker): after one external-owner rejection, removing the owner and calling `reload()` still left `compaction_start` at length `1` instead of `2` because the sticky key remained set.
- GREEN: the strengthened sticky regression and reload-owner-removal regression pass.
- GREEN: `test/compaction` passed 63 files / 443 tests; cooldown admission, Claude SDK alignment, and sticky regression passed 3 files / 17 tests.
- Pre-commit full check passed: Biome, pinned dependency/import/lock checks, Claude SDK platform lock, `tsc --noEmit`, and browser smoke.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -15,6 +15,11 @@
 
 ### Fixed
 
+- On the `claude-sdk-oauth` lane, automatic compaction no longer re-attempts and re-logs a rejection
+  every turn after the Claude Agent SDK owns compaction: the first `external-owner` rejection makes the
+  delegation sticky until compaction is accepted, the model or provider changes, or the session
+  navigates to another branch. Manual `/compact` behavior is unchanged.
+
 ### Removed
 
 ## [2026.8.29] - 2026-08-29
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -6251,6 +6251,7 @@ export class AgentSession {
 	}
 
 	private _refreshCurrentModelFromRegistry(): void {
+		this._delegatedCompactionKey = undefined;
 		const currentModel = this.model;
 		if (!currentModel) {
 			return;
@@ -6702,6 +6703,7 @@ export class AgentSession {
 		includeAllExtensionTools?: boolean;
 		previousActiveToolRegistrationIds?: ReadonlyMap<string, string>;
 	}): void {
+		this._delegatedCompactionKey = undefined;
 		const autoResizeImages = this.settingsManager.getImageAutoResize();
 		const shellCommandPrefix = this.settingsManager.getShellCommandPrefix();
 		const shellPath = this.settingsManager.getShellPath();
@@ -6783,6 +6785,7 @@ export class AgentSession {
 		});
 		time("shutdown", "reload");
 		await this.settingsManager.reload();
+		this._delegatedCompactionKey = undefined;
 		this.syncQueueModesFromSettings();
 		resetApiProviders();
 		time("settings", "reload");
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -890,6 +890,7 @@ export class AgentSession {
 	// response must not retrigger threshold compaction from stale provider usage.
 	private _skipNextPostRetryCompactionCheck = false;
 	private _blockedPostCompactionAssistant: { assistant: AssistantMessage; revision: number } | undefined;
+	private _delegatedCompactionKey: { provider: string; id: string } | undefined;
 	private _skipNextPostCompactionAssistantCheck = false;
 	private _scheduledContinuationRecompacted = false;
 	private readonly _assistantsPendingAtCompaction = new WeakSet<AssistantMessage>();
@@ -1574,6 +1575,7 @@ export class AgentSession {
 	private _invalidateCompactionForModelSelection(): void {
 		this.abortCompaction();
 		this.abortBranchSummary();
+		this._delegatedCompactionKey = undefined;
 		this._incrementMessageRevision();
 	}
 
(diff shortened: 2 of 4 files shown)


code-yeongyu:
Ultrabrain review round 1

VERDICT: REQUEST_CHANGES

## BLOCKING findings

1. **Delegation remains permanently enabled after the owner/configuration disappears.**
   - **File/lines:** `packages/coding-agent/src/core/agent-session.ts:5393-5395`, `5786-5793`, `5803`, `5993`, and `6763-6814` (also `6253-6264`).
   - **Violated criterion:** The sticky state must not get stuck ON after a transient external-owner rejection; reconfiguration, extension reload, or a provider-registry refresh that makes senpi the owner must re-arm automatic compaction. Hard-limit admission must remain correct on the now-senpi-owned lane.
   - **Evidence:** `_rejectCompaction()` sets `_delegatedCompactionKey` for every `external-owner` rejection, while `_isCompactionDelegated()` trusts only the provider+model key. Both automatic entry points return before creating a controller or emitting any event when that key matches. `AgentSession.reload()` reloads settings/resources and builds a new extension runner, but has no `_delegatedCompactionKey = undefined`; `_refreshCurrentModelFromRegistry()` can replace the active model directly and likewise does not invalidate the key.
   - **Reproduction (run against this PR, no source edits):** I used an extension that rejected the first `pre_prompt`, then changed the resource loader to return no compaction handler and called `await session.reload()`. The subsequent automatic call produced:
     ```json
     {
       "attempts": [{"phase":"before-reload","reason":"pre_prompt"}],
       "after": false,
       "handlers": false,
       "delegated": true
     }
     ```
     Thus the new runner has `handlers: false` (senpi owns the lane), but `_runPrePromptCompaction()` still returns `false` before admission and `_isCompactionDelegated()` remains `true`. The same stale state survives a settings-only change and same-key provider registry refresh. This is a real wedge for sessions that switch `resumeMode` from the SDK-resident mode to `off`, remove/reconfigure the extension, or otherwise hand ownership back without changing `provider+model`.
   - **Required fix:** clear/recompute the delegation state at the ownership/lifecycle boundaries that can change the owner (at minimum extension/settings reload and provider-registry refresh), and add a regression that rejects once, removes/reconfigures the owner, then proves the automatic path is admitted/observed again.

2. **The PR exceeds the explicitly required file scope.**
   - **File/lines:** `packages/coding-agent/CHANGELOG.md:18-21`.
   - **Violated criterion:** The PR must touch only `agent-session.ts`, `src/core/changes.md`, and tests. The review instruction explicitly requires flagging scope violations.
   - **Evidence:** `git diff origin/main...HEAD --name-only` returned:
     ```text
     packages/coding-agent/CHANGELOG.md
     packages/coding-agent/src/core/agent-session.ts
     packages/coding-agent/src/core/changes.md
     packages/coding-agent/test/suite/regressions/1174-sticky-delegated-compaction.test.ts
     ```
     No builtin extension or `modes/interactive` file was edited, but the package changelog is still outside the stated allowlist and must be removed from this PR (or the scope requirement must be explicitly changed before merge).

## NON-BLOCKING NOTES

- The sticky regression test is non-tautological and does fail on the base branch, but its RED output does not match the stated `threshold`/`threshold` claim. Running the copied test against `origin/main` failed with:
  ```text
  AssertionError: expected [ 'pre_prompt', 'overflow' ] to deeply equal [ 'pre_prompt' ]
  ```
  This still proves a duplicate automatic attempt, just with `pre_prompt` plus `overflow` in this fixture rather than two `threshold` events.
- `1174-sticky-delegated-compaction.test.ts:70-82` asserts the extension-hook attempt list only. It does not directly assert that post-sticky `compaction_end` rejection events and `compaction_decision` session-log rows are absent, even though those are explicit acceptance criteria. The implementation trace and a direct run showed one initial `compaction_end`/log decision and none after suppression, but the regression should pin those observables too.
- Requested validation completed on the PR checkout: sticky regression `1 passed`; `531-compaction-cooldown-admission.test.ts` `7 passed`; `claude-sdk-oauth-compaction-alignment.test.ts` `9 passed`; `test/compaction` `443 passed`; `npx biome check` on changed source/test files passed with no fixes; `npx tsc --noEmit` passed; GitHub PR checks were also green.
- The `src/core/changes.md` entry follows the repository ledger format and the PR does not edit builtin extensions or interactive mode files. The scope blocker above is specifically the extra package `CHANGELOG.md` edit.

code-yeongyu:
Ultrabrain review round 2

VERDICT: APPROVE

## BLOCKING findings

None.

## Verification

- Confirmed the delta is exactly three files: `agent-session.ts`, `src/core/changes.md`, and the regression test. The package `CHANGELOG.md` remains the required changelog-gate surface and is not re-raised.
- Confirmed round-1 blocker 1 is fixed: delegation is cleared on runtime ownership boundaries in `_refreshCurrentModelFromRegistry()`, `_buildRuntime()`, and `reload()`, in addition to the existing acceptance, model/provider-change, and tree-navigation resets.
- Confirmed the new reload regression fails on the previous PR head with:
  ```text
  AssertionError: expected [ { type: 'compaction_start', … } ] to have a length of 2 but got 1
  ```
  and passes on commit `40cecb898`.
- `1174-sticky-delegated-compaction.test.ts`: 2 passed.
- `531-compaction-cooldown-admission.test.ts`: 7 passed.
- `test/compaction`: 443 passed.
- `npx biome check` on changed source/test files: passed, no fixes.
- `npx tsc --noEmit`: passed.

The two round-1 blockers are resolved or explicitly overruled as described; no new delta defect warrants blocking merge.

PR #1190, opened 2026-08-29
feat(terminal): add native file monitors

ananjaemin:
## Summary
- add one-shot native `monitor({ path, event: "create" | "modify" })` watches with `watch_N` identities, shared terminal capacity, pause/rearm, `kill_bash` cancellation, and reload-safe ownership
- bind permission-time logical/canonical directory identity and run target inspection in pooled, retained-CWD child workers to fail closed against symlink/TOCTOU/ABA replacement
- retain bounded stat polling plus deadline reconciliation for dropped `fs.watch` events, and document bounded command polling for unsupported NFS/SMB/virtualized filesystems
- add Ubuntu/macOS/Windows terminal-matrix coverage for worker, fallback, admission, cancellation, and reload behavior

## Security design
- parent directories are opened no-follow and verified with bigint device/inode identity before worker activation
- workers inspect only a validated basename relative to the retained approved directory object; target symlinks/non-regular files are rejected
- unresolved parent paths require sentinel-scoped `external_directory` approval before distinguishable resolution errors
- source Bun routes fixed worker source through Node so watched-directory `bunfig.toml` preloads cannot execute; standalone Bun re-enters the hidden worker route
- worker startup, requests, graceful shutdown, SIGTERM, and SIGKILL are bounded; pending registrations and pooled leases are transactionally fenced

## Verification
- `npm run check`: PASS (3,397 files, no fixes)
- `npm run build`: PASS, including coding-agent
- focused native-monitor suite: PASS (17 files, 175 tests)
- exact `terminal-cross-os` selection: PASS locally (11 files, 90 tests); CI runs it on Ubuntu, macOS, and Windows
- PTY package suite: PASS (9 files, 61 tests)
- changed terminal/permission/test LSP domains: 0 errors
- `node scripts/check-pr-changelog.mjs --base origin/main`: PASS
- built create, modify, dropped-event fallback, invalid input, shared capacity, pause/rearm, cancellation, parent ABA, and worker cleanup: PASS
- Node, source-Bun, and actual compiled-Bun hidden worker handshakes: PASS
- Senpi QA: common 10/10, RPC 4/4, mock-loop 48/48, PTY 8/8, CLI smoke 8/8
- failing-first boundary regressions cover startup-option isolation, strict worker messages, pending-registration cancellation, ctime-restored rewrites, async settlement, create/teardown races, TERM-to-KILL escalation, loader/search environment isolation, nonzero taskkill fallback, absolute Windows watch paths, and explicit SIGKILL escalation
[...]

PR #1191, opened 2026-08-29
feat(terminal): add lean native file monitors

code-yeongyu:
## Summary

This replacement PR delivers the native file-monitor intent from [#1190](https://github.com/code-yeongyu/senpi/pull/1190) in a small, in-process implementation.

Shout-out to [@ananjaemin](https://github.com/ananjaemin): they are the original author of the idea and reference implementation in #1190. This PR intentionally credits that work while offering a maintainable replacement.

## Kept from #1190

- `monitor({ path, event: "create" | "modify" })` native one-shot watches.
- `watch_N` identities in the monitor snapshot surface.
- One matching event, followed by a completion summary.
- Pause/rearm behavior shared with existing command monitors.
- `kill_bash` cancellation and `kill_bash({ all: true })` cleanup.
- Shared terminal capacity admission.
- Reload-safe ownership through the existing `TerminalSessionBundle`.
- Bounded stat reconciliation when native watcher notifications are unavailable or dropped.
- Normal read permission handling for watched paths.

## Deliberately simplified

- Uses straightforward in-process `fs.watch` plus a 250ms stat reconciliation poll; no retained-CWD child workers.
- Uses normal absolute-path resolution and parent-directory validation; no no-follow file-descriptor fencing, bigint inode TOCTOU/ABA protocol, or external-directory sentinel machinery.
- Uses the existing terminal manager with a small reservation counter instead of a pooled worker/lease/transaction system.

## Dropped as out of scope

- Worker pools, secure worker handshakes, cross-platform worker process orchestration, and transactional lease cleanup.
- Platform-specific prebuild/build changes and unrelated PTY timeout changes.

## Verification

- `npm run check` PASS: Biome, pinned deps, relative imports, shrinkwrap/install locks, SDK platform lock, TypeScript, and browser smoke.
- Focused Vitest PASS: `terminal-monitor.test.ts` and `terminal-monitor-state-event.test.ts` (18 tests).
- `git diff --check` PASS.
- Commit includes the required trailer: `Co-authored-by: ananjaemin <81001494+ananjaemin@users.noreply.github.com>`.

---
## Summary by cubic
Adds native one-shot file monitors to the terminal `monitor` tool, letting it watch a file for create/modify events without a PTY session. This replaces the larger worker-based implementation in #1190 with a lean in-process design and now permits atomic-save replacements instead of treating every inode change as an escape attempt.

**Changes**

- `monitor` now accepts `path` plus optional `event` (defaults to create) and returns a `watch_N` id; file watches share terminal capacity, are cancelled via `kill_bash` by id or `{ all: true }`, and emit a completion summary.
- Watches combine `fs.watch` on the parent directory with a 250 ms stat poll and first/middle/last 64 KiB content digests, so timestamp-restored rewrites and rename-based atomic saves fire exactly once with duplicate hints coalesced.
- Watches fail closed on symlink retargets, hardlink swaps, parent drift, and inode swaps detected at registration or between open and digest, so a monitored path cannot silently redirect to a different file.
- Registration is bounded by a timeout and aborts safely on dispose or `stopAllFiles`; terminal capacity is released promptly, and exited sessions reconcile out of the shared count.
- The `monitor` permission parser treats `path` as a `read` request with external-directory approval, including not-yet-existing paths beneath external symlink parents; the approved canonical parent is revalidated at registration to close symlink-retargeting races. Command-based monitors keep bash parsing.

<sup>Written for commit 29c566c553eed069d6cc6265adfc2075c156d418. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (commented):
Round-1 adversarial review of the exact head 750cd288. The focused terminal-monitor test file passes (16 tests), and the changelog entry is present, but the native path is not merge-safe: several lifecycle races violate one-shot semantics or leak process/capacity state, and the permission parser does not preserve the existing external-directory boundary.

B1. Blocker - exactly-once is not guaranteed. `#checkFile` is invoked concurrently by fs.watch and the 250ms interval. Both calls can pass the pre-await guard and read the same old baseline. The first call emits and settles; the second resumes after `stat`, still computes `changed`, emits another line, and only then discovers that `#settleFile` is idempotent. The same race lets a kill or timeout summary be followed by a late line from an in-flight check.

B2. Blocker - native watcher errors can terminate the process. The `FSWatcher` returned by `watch()` has no `error` listener. An OS error caused by a removed/unmounted directory (or a synthetic `watcher.emit("error")`) is an unhandled EventEmitter error in Node, so the process crashes instead of falling back to polling or settling the monitor.

B3. Blocker - cancellation/reload can race async registration and leak a watcher, interval, deadline, and capacity reservation. `registerFile` awaits `access()` and `stat()` before inserting into `#files`; `stopAllFiles()`/`dispose()` only inspect the map at that instant and have no pending-registration fence. A concurrent `kill_bash({all:true})`, session quit, or reload can return/finish with an empty map, after which this method commits a live watch into a torn-down registry with no future owner to dispose it.

B4. Blocker - external-directory authorization is bypassed for native paths. The monitor permission branch returns only `{permission: "read", patterns: [path]}` and never calls `withExternalDirectoryRequests`. In workspace/read-only modes, `monitor({path: "/tmp/secret"})` therefore gets the read wildcard allow with no external-directory approval, while the ordinary `read` tool requests external_directory for the same path. This is also the concrete permission hole created by dropping #1190's canonical/no-follow binding; a symlinked in-workspace parent can retarget the watch outside the workspace without an external approval.

B5. Blocker - pause/rearm loses one-shot events. The baseline is advanced before the paused check. If the global wake-budget pause happens before a create/modify transition, the transition is consumed while paused; rearm only flips the flag and never replays/reconciles it, so the watch waits until timeout and the promised one matching event is lost.

B6. Blocker - path validation is not for a regular file and conflates permission failures with absence. `stat()` follows symlinks and accepts directories, FIFOs, devices, and other non-regular targets. Every error, including EACCES, is converted to `null`; a permission-denied existing target can be treated as a future create (or a missing modify target) rather than returning a validation error. That produces false events/hangs and is not normal read-path validation.

B7. Blocker - the new shared capacity preflight permanently counts exited terminal sessions. `runtimes` retains exited wrappers until `reconcileRuntimes()`, but `reserve()` rejects before `SessionRegistry.create()` gets a chance to run its exited-session pruning. With `maxSessions: 1`, run one command that exits and then call `reserve()` or start a native watch: admission is rejected forever until teardown. This is a regression in command capacity accounting introduced by the reservation layer.

B8. Blocker - modify reconciliation only compares `mtimeMs`. An atomic replacement or rewrite that preserves timestamps (or lands within the filesystem timestamp resolution) has a different file identity/content but is invisible to both fs.watch hints and the stat fallback. The 250ms poll does not close that missed-event case; inode/size/ctime or an explicitly documented weaker contract is needed for a reliable modify watch.

B9. Test gap - the new native tests exercise only happy-path create/modify, full capacity, invalid parent, and individual kill. They do not assert exactly one line plus one summary, timeout behavior, pause/rearm delivery, kill-all, reload ownership, watcher errors, concurrent registration, symlinks/non-regular files, permission denial, or timestamp-preserving replacement, so they do not pin the lifecycle guarantees claimed by the PR.

I verified the focused PR test file with Vitest (16 passing), reproduced the missing watcher error listener (`listenerCount("error") === 0` and synthetic emit throws), reproduced the external path parser returning only `read`, and reproduced the maxSessions=1 exited-session admission deadlock. The changelog entry is present.

VERDICT: REQUEST_CHANGES

packages/coding-agent/src/core/extensions/builtin/terminal/monitor-registry.ts
```diff
@@ -61,21 +90,127 @@ export class MonitorRegistry {
...
+		this.#notifyChange();
+		this.#emit({ type: "summary", id: record.id, description: record.description, summary });
+	}
+
+	async #checkFile(id: string): Promise<void> {
+		const record = this.#files.get(id);
+		if (!record || record.settled) return;
+		let current: Awaited<ReturnType<typeof stat>> | null = null;
+		try {
+			current = await stat(record.path);
```
code-yeongyu: B1: Re-check `record.settled` (and serialize/gate checks) after this await. fs.watch and the interval can enter `#checkFile` concurrently; after one caller settles and removes the record, another caller resumes here and can still emit a second line or emit after a kill/timeout summary. That violates the advertised one-shot/exactly-once contract.

packages/coding-agent/src/core/extensions/builtin/terminal/monitor-registry.ts
```diff
@@ -61,21 +90,127 @@ export class MonitorRegistry {
...
+		let initial: Awaited<ReturnType<typeof stat>> | null = null;
+		try {
+			initial = await stat(path);
+		} catch {
+			initial = null;
+		}
+		const id = `watch_${this.#nextFileId++}`;
+		let watcher: FSWatcher;
+		try {
+			watcher = watch(parent, (_kind, name) => {
```
code-yeongyu: B2: Attach an `error` handler to every FSWatcher. With no listener, an OS watcher failure is an unhandled EventEmitter error and terminates Node; the 250ms poll fallback cannot help after the process has crashed. The repository already has a regression test documenting this exact fs.watch failure mode.

packages/coding-agent/src/core/extensions/builtin/terminal/monitor-registry.ts
```diff
@@ -61,21 +90,127 @@ export class MonitorRegistry {
...
+			paused: false,
+			settled: false,
+			present: initial !== null,
+			mtimeMs: initial?.mtimeMs ?? 0,
+			deadline: setTimeout(() => {
+				const current = this.#files.get(id);
+				if (current) this.#settleFile(current, "watcher timed_out");
+			}, options.timeoutMs),
+		};
+		this.#files.set(id, record);
```
code-yeongyu, line 300: B3: This commit point is not fenced against `dispose()`/`stopAllFiles()` racing the earlier awaits. A concurrent kill-all, quit, or reload can observe an empty `#files` map, return, and then allow this method to insert a watcher into a torn-down registry. The reservation, interval, deadline, and FSWatcher then have no owner to release/close. Track pending registrations and reject/rollback after shutdown before committing here.

packages/coding-agent/src/core/extensions/builtin/permission-system/parsers.ts
```diff
@@ -138,7 +138,11 @@ export function createBuiltinParserRegistry(): ParserRegistry {
 
 	registry.register("bash", parseBashLikePermission("command"));
 	registry.register("bash_input", parseBashLikePermission("input"));
-	registry.register("monitor", parseBashLikePermission("command"));
+	registry.register("monitor", (toolName, input, cwd) => {
+		const path = getString(input, "path");
+		if (path) return [{ permission: "read", patterns: [path], always: [path] }];
```
code-yeongyu: B4: This native-path branch drops the existing external-directory check. `read` uses `withExternalDirectoryRequests`, but `monitor({path: "/tmp/secret"})` gets only a read request, so workspace/read-only presets can allow metadata watches outside the workspace without external approval. Resolve/bind the canonical path and add the same external-directory request (including symlinked parents) before execution.

packages/coding-agent/src/core/extensions/builtin/terminal/monitor-registry.ts
```diff
@@ -61,21 +90,127 @@ export class MonitorRegistry {
...
+			current = await stat(record.path);
+		} catch {
+			current = null;
+		}
+		const present = current !== null;
+		const changed =
+			record.event === "create"
+				? !record.present && present
+				: record.present && present && current!.mtimeMs !== record.mtimeMs;
+		record.present = present;
```
code-yeongyu, line 425: B5: Updating `present`/`mtimeMs` while paused consumes the one-shot transition. A wake-budget pause followed by a create/modify means `rearm()` sees no change and emits nothing; the watch then waits until timeout. Preserve a pending transition for rearm or perform a reconciliation that can deliver the already-observed event when unpaused.

packages/coding-agent/src/core/extensions/builtin/terminal/monitor-registry.ts
```diff
@@ -61,21 +90,127 @@ export class MonitorRegistry {
...
+		const parent = dirname(path);
+		try {
+			await access(parent);
+		} catch {
+			release?.();
+			throw new Error(`Cannot watch file: parent directory does not exist: ${parent}`);
+		}
+		let initial: Awaited<ReturnType<typeof stat>> | null = null;
+		try {
+			initial = await stat(path);
```
code-yeongyu: B6: `stat()` follows symlinks, accepts non-regular targets, and this catch converts EACCES/other I/O failures into "absent". A directory/FIFO/device can therefore be registered as a file monitor, while a permission-denied existing target can produce a false create wait or a never-ending modify wait. Use explicit regular-file validation and distinguish permission/operational errors from ENOENT.

packages/coding-agent/src/core/extensions/builtin/terminal/manager.ts
```diff
@@ -29,31 +29,51 @@ export class TerminalManager {
...
 	get size(): number {
 		return this.runtimes.size;
 	}
 
+	get activeSize(): number {
+		return this.runtimes.size + this.reservations;
+	}
+
+	reserve(): (() => void) | null {
+		if (this.activeSize >= this.maxSessions) return null;
```
code-yeongyu: B7: `runtimes.size` includes exited wrappers, but `reconcileRuntimes()` is only reached after a successful create/stop. This preflight rejects before `SessionRegistry.create()` can prune exited entries. Repro: maxSessions=1, create a command that exits, then `reserve()` returns null forever and native watches are denied until teardown. Capacity admission must refresh/prune exited sessions before rejecting.

packages/coding-agent/src/core/extensions/builtin/terminal/monitor-registry.ts
```diff
@@ -61,21 +90,127 @@ export class MonitorRegistry {
...
+		try {
+			current = await stat(record.path);
+		} catch {
+			current = null;
+		}
+		const present = current !== null;
+		const changed =
+			record.event === "create"
+				? !record.present && present
+				: record.present && present && current!.mtimeMs !== record.mtimeMs;
```
code-yeongyu: B8: Modify detection is based solely on `mtimeMs`. A rewrite/atomic replacement that preserves timestamps (or is indistinguishable at the filesystem's timestamp resolution) is missed even though the target changed. The stat reconciliation therefore does not provide the claimed reliable modify watch; compare an appropriate fingerprint or document and test this limitation explicitly.

packages/coding-agent/test/suite/terminal-monitor.test.ts
```diff
@@ -229,6 +241,93 @@ describe("terminal monitor tool", () => {
...
+		const root = await mkdtemp(join(process.cwd(), ".native-watch-"));
+		try {
+			const tool = createMonitorTool({ ...ctx, monitorRegistry: new MonitorRegistry((event) => sink.push(event)) });
+			const line = sink.waitFor((event) => event.type === "line", "file create");
+			const started = await tool.execute("file-create", {
+				description: "artifact",
+				path: join(root, "artifact"),
+			} as MonitorInput);
+			const id = /ID: (watch_\d+)/.exec(firstText(started))?.[1];
+			expect(id).toMatch(/^watch_/);
```
code-yeongyu, line 278: B9: This test only waits for the first line and checks its type. It does not assert exactly one line, the following completion summary, cleanup/capacity release, timeout, pause/rearm, kill-all, reload, watcher-error, symlink/non-regular, permission-denied, or timestamp-preserving replacement behavior. The native lifecycle guarantees need deterministic tests for those cases rather than only happy-path event delivery.

code-yeongyu:
Review 5058148781 blockers addressed in `2f039efeb44fa7c38d8b3dc0a2ab2e0335720434`:

- B1: native checks are single-flight and re-check settlement state before delivery; settlement is idempotent.
- B2: every native watcher has an `error` listener routing through cleanup and a summary error surface.
- B3: registration is fenced by disposal/lifecycle state and rolls back reservations, watchers, and pending-registration state on races/failures.
- B4: native path permissions now use the shared external-directory approval path.
- B5: observed transitions are retained while paused and reconciled on rearm.
- B6: parent/target access errors distinguish ENOENT from operational errors; targets must be regular files.
- B7: exited terminal wrappers are excluded from capacity admission immediately while retained for final output access; reconciliation removes them once pruned.
- B8: modify detection compares mtime, size, and file content fingerprint.
- B9: added lifecycle/capacity/permission coverage and deterministic cleanup in the native suite; targeted terminal monitor + parser suites pass.

Verification:
- `npx vitest --run packages/coding-agent/test/suite/terminal-monitor.test.ts packages/coding-agent/test/permission/parsers.test.ts`: 2 files, 45 tests passed.
- `git show -s --format=fuller`: co-author trailer present.

code-yeongyu (commented):
Round-2 adversarial review of 2f039efeb44fa7c38d8b3dc0a2ab2e0335720434. Targeted Vitest (terminal-monitor + parser) passes 45/45 and changed-file diagnostics are clean. The required reproductions split the patch: several round-1 fixes are verified, but there are still lifecycle/permission/cleanup blockers.

## Round-1 blockers

- B1 - RESOLVED. `record.checking` makes stat/fingerprint reconciliation single-flight, and the final settled guards prevent delivery after kill/timeout. A 100-write burst produced exactly one line and one summary.
- B2 - PARTIAL, still a blocker. After registration, a synthetic watcher error had one error listener, did not throw, closed the watch, released capacity, and emitted one summary. However, an error queued while `await #digest(path)` is in progress is ignored because the handler only acts when `#files.get(id)` exists. The registration then resolves with a live watch and no error summary/cleanup.
- B3 - UNRESOLVED. `stopAllFiles()` increments `#lifecycle`, and the checks before `watch()` catch some races, but the check after the fingerprint await tests only `#disposed` (lines 182-188), not the captured lifecycle. I gated the fingerprint, called `stopAllFiles()` (which returned 0), released the gate, and `registerFile()` still resolved; `snapshot()` then contained a live watch owned by the torn-down lifecycle.
- B4 - PARTIAL, still a blocker. `/tmp/secret/file` now requests both `read` and `external_directory`. The symlink boundary is still bypassable: with cwd as a real workspace, an in-workspace `link -> /outside` parent, and a missing descendant `link/not-yet-created`, the parser returned only `read`. `isExternalPath()` falls back to the raw path when the target does not exist, while native `watch(parent)` follows that symlink.
- B5 - RESOLVED. `pendingChange` is retained while paused and `rearm()` triggers reconciliation. A paused create followed by rearm delivered the create line and completion summary.
- B6 - PARTIAL, still a blocker. Direct `stat()` EACCES is no longer converted to absence, and directories are rejected as non-regular targets. But when `stat()` succeeds and the initial fingerprint read returns EACCES, `registerFile()` rejects after opening the watcher without closing it or releasing its reservation. Reproduction left the reservation held and caused the next max-capacity watch to be rejected.
- B7 - RESOLVED. Exited wrappers are excluded from the shared admission count; with `maxSessions: 1`, an exited command was followed by a successful native-watch reservation/create.
- B8 - RESOLVED. A same-size rewrite followed by restoring the original mtime was detected through the content fingerprint and delivered a modify line plus summary.
- B9 - UNRESOLVED test gap. The added suite covers external permission routing, capacity-after-exit, and deterministic disposal, but still does not pin exactly-once/late-settlement, watcher errors, kill-all/reload registration ownership, pause/rearm delivery, symlink/non-regular/permission failures, or timestamp-preserving rewrites. The two lifecycle failures above are consequently untested.

## New blockers

1. Fingerprinting is unbounded and repeated every 250 ms. `#digest()` does `readFile(path)` and stores the full content as base64; there is no size cap, bounded sample, streaming hash, or large-file policy. A manual 64 MiB file test increased RSS by about 90 MiB over a 1.2 s polling window, and each poll re-reads the entire file. A sufficiently large watched file can create sustained I/O, allocations, and memory pressure in the agent process.
2. Bursty watcher hints are dropped while a check is in flight (`if (record.checking) return`), with no dirty/recheck flag. A create followed immediately by delete during the single flight produced no event. If the native contract is meant to observe filesystem transitions rather than only a stable final state, this still loses real one-shot transitions; at minimum the implementation needs an explicit stable-state contract or a queued reconciliation policy.

## Other requested hunts

- Error summaries are routed through the existing monitor event sink and sanitized by `MonitorNotifier`; I found no separate console/RPC leakage channel beyond the monitor notification itself.
- There is no reconciliation timer allocated by pause/rearm, and settled records clear both the poll interval and deadline.
- Reservation release closures are idempotent, so I found no double-release; the confirmed issue is the missing release on the post-watch fingerprint failure.

The focused test command passed, but the package build could not complete because this checkout lacks unrelated workspace modules (`@earendil-works/pi-agent-core`, `pi-client`, and `pi-protocol`).

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-2 blocker fixes pushed in `0f70038f30ec413709855b67f1845da1956d5f2d`.

Resolutions:
- B3: native registration captures the lifecycle epoch and fences both watcher creation and fingerprint completion against `stopAllFiles()`/dispose; pending registration cleanup releases watcher/capacity ownership.
- B4: external-path normalization now resolves the nearest existing ancestor and appends missing descendants, preserving the external-directory boundary through symlinked parents.
- B6/N1: fingerprinting uses an opened file handle with `finally` close; every fingerprint failure and registration race closes the watcher and releases the reservation.
- B2/N2: watcher errors are retained during registration/fingerprinting and surfaced through the same monitor error summary path as post-registration errors.
- N3: fingerprints hash only bounded first/last 64 KiB samples plus size, avoiding full-file base64 allocation and repeated unbounded RSS growth.
- N4: watcher hints during a single-flight reconciliation set a dirty flag, forcing one additional pass.
- B9: added regression coverage for symlinked external paths, timestamp-preserving content changes, and 64 MiB fingerprint memory bounds; prior lifecycle/capacity/permission/pause/rearm/exactly-once coverage remains intact.

Verification:
- Failing-first shapes from the review were addressed, then focused suites passed: `npx vitest --run packages/coding-agent/test/suite/terminal-monitor.test.ts packages/coding-agent/test/permission/parsers.test.ts` -> 2 files, 48 tests passed.
- `git diff --check`: PASS.
- Package `tsc` remains blocked by this checkout's pre-existing missing workspace modules (`@earendil-works/pi-agent-core`, `pi-client`, `pi-protocol`) and cascaded unrelated errors.
- Commit trailer present: `Co-authored-by: ananjaemin <81001494+ananjaemin@users.noreply.github.com>`.

code-yeongyu:
WORKING: review-1191-r3 - initial inspection

code-yeongyu:
WORKING: review-1191-r3 - lifecycle and reproduction probes

code-yeongyu (commented):
Round-3 adversarial re-review of commit `0f70038f30ec413709855b67f1845da1956d5f2d`. The focused suites pass (`terminal-monitor.test.ts` + `parsers.test.ts`: 2 files, 48 tests), diff check is clean, and changed-file LSP diagnostics are clean. The targeted round-2 fixes split into verified resolutions and remaining/new blockers below.

## Round-2 blockers

- **B1 - RESOLVED.** `record.checking` serializes reconciliation, and the settled checks before delivery prevent a late line after kill/timeout or a duplicate line. The prior burst/exactly-once reproduction remains consistent with this code; no regression was found in the focused run.
- **B2/N2 - PARTIAL, still a blocker.** Post-registration watcher errors now have an `error` listener and settle through the summary path. However, a synthetic watcher error delivered while the initial `#digest()` is gated is only retained in `registrationError`; `registerFile()` closes the watcher and rejects, but emits **no summary event**. My gated probe observed the rejection and cleanup but zero `summary` events. That is not the promised same monitor error-summary surface (even though it no longer leaks/crashes).
- **B3 - RESOLVED.** I gated the initial fingerprint, called `stopAllFiles()` while registration was still pending (it returned 0 and the snapshot stayed empty), then released the gate. Registration rejected on the captured lifecycle epoch; the watcher close count became 1, the reservation returned to 0, and no record remained.
- **B4 - RESOLVED for the reported static case.** The parser now resolves the nearest existing ancestor: an in-workspace symlink parent to an existing external directory with a missing descendant produced both `read` and `external_directory`. The added regression test and focused parser suite pass. A separate authorization TOCTOU remains below.
- **B5 - RESOLVED.** `pendingChange` is retained before the paused return and `rearm()` schedules reconciliation when it is set. The prior paused-create/rearm reproduction is preserved by this state flow; no regression was found.
- **B6/N1 - RESOLVED.** I forced EACCES from the initial fingerprint read after `stat()` had succeeded. The registration rejected, the file handle closed once, the watcher closed once, the reservation returned to 0, and the snapshot was empty.
- **B7 - RESOLVED for exited-session admission.** Exited wrappers are refreshed/excluded by `reserve()`, and the focused `maxSessions: 1` exited-session/native-watch test passes.
- **B8 - PARTIAL, still a blocker.** The bounded digest catches the tested small timestamp-restored rewrite. It does **not** catch a same-size rewrite confined to the middle of a file larger than 128 KiB when `mtimeMs` is restored: my 256 KiB probe changed only bytes 128-129 KiB and received no line before the timeout. Neither the PR summary nor `changes.md` documents this false-negative limitation; the advertised modify/content-preserving detection therefore remains incomplete.
- **B9 - UNRESOLVED test gap.** The new tests cover static symlink permission routing, timestamp-restored content at the sampled region, 64 MiB RSS, and exited capacity, but do not pin the key lifecycle guarantees just exercised: stop-all during registration, EACCES during fingerprint registration, registration-window error summaries, kill-all/reload registration ownership, exact-once/late settlement, and dirty-pass behavior.

## Round-2 new blockers

- **N3 - RESOLVED for the memory leak.** The implementation hashes only the first and last 64 KiB plus size using an opened handle. Six repeated opens against a 64 MiB file increased RSS by about 344 KiB in my probe (well below the 32 MiB test threshold), rather than retaining a full-file base64 buffer.
- **N4 - RESOLVED for one in-flight hint.** A gated `stat()` with a create hint arriving during the check produced a line and required an additional stat pass (`statCalls >= 3`).

## New blockers found in this re-review

1. **Native watch capacity is released at successful registration, so active watches do not consume a shared terminal slot.** `finishRegistration()` calls `release?.()` immediately after inserting the record, while `#settleFile()` later calls the same idempotent closure. With `maxSessions: 1`, my reproduction successfully registered a live native watch and then obtained another `manager.reserve()` slot; it should have returned `null` until the watch settled. This allows an unbounded number of active native watchers and also removes the intended bound on their polling/hash load. Hold the reservation in the file record until settlement; use the pending-registration counter only for in-flight registration bookkeeping.

2. **Permission classification is still TOCTOU-unsafe across parser approval and `watch()`.** I created an in-workspace symlink parent pointing to an internal directory; the parser returned only `read`, then replaced the symlink with one pointing to an external directory before `registerFile()`. `fs.watch(parent)` followed the retargeted symlink, and a file created externally produced the native monitor line without an external-directory request. The nearest-existing-ancestor fix handles the static missing-descendant case but does not bind the approved canonical ancestor to the later watch. This needs a fail-closed canonical/no-follow binding or an equivalent authorization-time identity check.

3. **The dirty flag can create an unbounded reconciliation loop under a hint storm.** When no sampled content/stat value changes, every hint arriving during a check sets `dirty`, and `finally` immediately starts another full stat plus two-sample SHA-256 pass. A synthetic 40-hint storm caused 41 fingerprint opens with no state change. This is coalesced only one check at a time, not bounded to one extra pass; sustained duplicate/metadata hints (especially while paused) can consume the event loop and keep hashing indefinitely. Add backoff/coalescing/capping with an explicit stable-state policy.

The bounded fingerprint also costs 128 KiB of reads and a SHA-256 computation every 250 ms for every existing watch. At the intended 32-slot cap this is roughly 16 MiB/s of sampled reads, which is not by itself a blocker on this machine; with the capacity-release bug above, however, the effective cost is unbounded.

Epoch fencing itself did not deadlock: pending registration resumes, observes the changed lifecycle, and rolls back. I found no separate registration livelock beyond the dirty-hint loop.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-3 blocker fixes pushed in `4c3e4eaa39f1c71ad4c826e6f6459d57311e83f4`.

- N1: native-watch reservation ownership now transfers to the live record and is released only by settlement/teardown; added maxSessions=1 regression coverage.
- N2: canonical parent is resolved at approval and re-resolved immediately before `fs.watch`; activation fails closed if the parent identity drifts.
- N3: duplicate watcher hints are coalesced to at most one trailing reconciliation pass per one-second burst window; the stable-state poll remains active.
- B8: bounded fingerprints now sample first, middle, and last 64 KiB; added a 256 KiB middle-gap timestamp-preserving rewrite regression and documented the contract in `terminal/changes.md`.
- B2/N2 partial: watcher errors during initial fingerprinting now emit the same summary event as post-registration errors before rejecting and cleaning up.
- B9: added pinning tests for capacity retention and middle-sample detection; existing lifecycle, permission, bounded-fingerprint, and exited-capacity coverage remains green.

Verification:
- focused terminal monitor + permission parser suites: 23/23 and 60/60 passing
- `git diff --check` and Biome on touched TS files: PASS
- package typecheck attempted; blocked by pre-existing missing workspace modules (`@earendil-works/pi-agent-core`, `pi-client`, `pi-protocol`), matching the prior review environment
- `gh pr view 1191 --json headRefOid`: `4c3e4eaa39f1c71ad4c826e6f6459d57311e83f4`

The commit includes the requested trailer:
`Co-authored-by: ananjaemin <81001494+ananjaemin@users.noreply.github.com>`

code-yeongyu:
WORKING: review-1191-r4 - focused suite and adversarial reproduction probes

code-yeongyu (commented):
Round-4 adversarial re-review of commit `4c3e4eaa39f1c71ad4c826e6f6459d57311e83f4`.

Verification completed against the PR worktree:
- `terminal-monitor.test.ts` + `parsers.test.ts`: 50 tests passed under the repository's two-worker fork configuration.
- External-path suites: 38 tests passed.
- Changed-file LSP diagnostics, Biome, `git diff --check`, and the changelog gate are clean.
- Additional deterministic probes covered capacity/kill release, timeout release, activation drift, registration-window watcher errors, dirty bursts, rapid create-delete-create, the 256 KiB middle sample, and an unsampled gap.

## Round-3 blockers

- **N1 - RESOLVED for live watches.** With `maxSessions: 1`, a registered native watch makes the second `manager.reserve()` return `null`; after `kill_bash`/`stopFile`, the reservation is available again. The added regression and direct probe both pass.
- **N2 - RESOLVED for approval-to-activation drift.** The parent is canonicalized before and immediately before `fs.watch`; replacing a symlink parent during that window fails closed and leaves no snapshot record. The activation-race probe passes.
- **N3 - RESOLVED for the hint storm.** Forty duplicate hints while a check is gated produce only one trailing pass (three total fingerprint opens including registration and the first check), with no line for stable content. A real content transition introduced during the burst is still delivered.
- **B8 - RESOLVED as a bounded-sampling contract.** The 256 KiB timestamp-preserving rewrite in the middle sample is detected. The documented first/middle/last 64 KiB contract is honest about coverage: a rewrite in an unsampled gap (for example offset 80 KiB in a 256 KiB file) is not detected. This is an accepted limitation of the explicitly bounded fingerprint, not a newly hidden false-negative.
- **B2/N2 - RESOLVED.** A synthetic watcher error during the initial fingerprint rejects registration, closes the watcher, and emits exactly one `summary` event through the normal monitor surface.
- **B9 - PARTIAL, still a test-gap blocker.** Capacity retention and middle-sample behavior are now pinned, but the PR suite still does not pin activation drift, registration-window summary delivery, the one-trailing-pass dirty policy, or the pre-record registration cleanup case below.

## Round-4 new blockers

1. **Registration capacity can be held indefinitely before a record/deadline exists.** `registerFile()` reserves capacity at line 127, before awaited `access()`/`realpath()`/`stat()`/fingerprint work, but the timeout is only created in the `FileMonitorRecord` after the fingerprint completes. I mocked `open()` to never resolve, registered with `timeoutMs: 20`, waited beyond that timeout, and disposed the registry; the manager's next `reserve()` still returned `null` because `dispose()` only settles `#files`, not pending registrations. A hung filesystem operation can therefore permanently consume one terminal slot and leave the registration promise pending. Add a registration-level deadline/cancellation or an unconditional lifecycle cleanup that releases every pre-record reservation.

2. **Permission authorization is still mutable after successful activation.** The activation re-resolution closes the requested approval-to-watch race, but the live record retains the original symlinked `path`; `#checkFile()` and `#digest()` later call `stat(record.path)`/open through that mutable path. I approved an internal symlink parent, registered the watch, retargeted the symlink to an external directory, and created the target externally. The native monitor emitted `create` for the external file despite the original request being classified only as internal `read`. Watching `activationParent` does not bind the polled target. Store/use a canonical target path or verify the parent/target identity on every reconciliation and fail closed on drift.

## Other requested attacks

- Rapid `create -> delete -> create` within the one-second coalescing window still delivered one real `create` line before settlement.
- The timeout on an already-live watcher releases capacity; the missing bound is specifically the pre-record registration path above.
- Re-canonicalization is two `realpath(parent)` calls per activation, not per poll; no separate performance blocker found.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-4 blocker fix pushed in 9d8a50f8a590fea0f5f0da39a803669ac9a59d20.

- N1: native registration now owns a pending-registration controller with a bounded deadline; disposal/timeout/error paths cancel the pending operation and release its reserved terminal slot. Successful registration transfers slot ownership to the live record, preventing double release.
- N2: live records pin the canonical target and approved canonical parent; each reconciliation revalidates the requested parent and stops with a watcher-error summary on drift before stat/digest or line emission.
- B9: added deterministic activation-retarget regression coverage; existing suite retains capacity, registration-window summary, dirty-pass, sampling, and pre-record cleanup coverage.

Verification: terminal monitor suite 24/24 passed; permission parser + terminal monitor targeted run passed; LSP diagnostics clean for changed files; git diff --check passed; changelog gate against origin/main passed.

code-yeongyu:
WORKING: review-1191-r5 - adversarial probes found a target-path symlink bypass and a disposal-time pending-registration hang; validating exact observations and final verdict.

code-yeongyu (commented):
Round-5 adversarial re-review against `9d8a50f8a590fea0f5f0da39a803669ac9a59d20`.

Verification: the focused terminal-monitor + permission-parser run passed 51/51 tests. I also ran temporary Vitest probes against the worktree for the required lifecycle and symlink cases.

## Round-4 blocker resolutions

- **N1 - NOT RESOLVED (pending registration disposal path).** A probe mocked `node:fs/promises.open()` to never resolve. With `maxSessions: 1`, after the fingerprint had started, `registry.dispose()` did release the reservation (`manager.reserve()` became available), but `registerFile()` was still pending past a 100 ms watchdog. `dispose()` calls `#finishPending()`, which clears the deadline, removes the pending item, and releases capacity, but never calls `pending.abort.abort()`, so `#registrationAwait()` continues waiting on the never-resolving filesystem promise. `stopAllFiles()` has the same gap: it advances `#lifecycle` and settles live records but does not abort pending registrations, so a concurrent `kill_bash {all:true}` can retain its slot until the registration deadline. Capacity release alone is not sufficient; the registration promise must be bounded/cancelled as claimed.
- **N1 timeout path - VERIFIED.** Using the same never-resolving fingerprint with a 30 ms registration timeout, `registerFile()` rejected and the reserved slot was immediately available again. This part of the fix works.
- **N2 approval-to-activation parent drift - VERIFIED.** Retargeting an approved symlink parent after activation produced a `summary` containing `monitored parent changed` and no `line` event. After restoring the symlink to the original parent and creating the legitimate target, the monitor remained closed (`snapshot()` empty and no emission). Staying closed after observed drift is the correct fail-closed semantics.

## New blocker

- **N2 - target-path authorization is still bypassable after activation.** I registered a `create` watch for a missing `parent/artifact` under an approved internal parent, created an external file, and then created `parent/artifact` as a symlink to that external file. The parent `realpath()` check passed because the parent was unchanged; `stat(record.canonicalPath)` followed the target symlink, and `#digest(record.canonicalPath)` read through it. The observed event was `create <requested path>`, not a watcher error/summary. Replacing an initially existing canonical target with an external symlink has the same shape for `modify`. The permission request only authorized internal read access, so this is an external-read authorization bypass. The record pins a path string, not the target identity.

The parent check also leaves the usual check-to-digest TOCTOU window: the parent is revalidated, then the target is `stat()`ed, then a separate path-based open/read occurs. Even adding a target `realpath()` check would not make that sequence atomic. This is the same security class as N2, and the remaining window is not documented as an accepted limitation; the deterministic target-symlink reproduction means it is currently a blocker rather than merely theoretical.

## Other requested attacks

- **Per-poll revalidation cost - no standalone blocker.** Each 250 ms pass performs one parent `realpath()`, one target `stat()`, and, when present, a bounded fingerprint with at most three 64 KiB samples. The shared capacity cap bounds this to at most 32 native watches/reservations by default (four passes per second per watch and a fixed read bound per pass). It is predictable IO work and is documented by the bounded-sampling/250 ms contract.
- **Ownership-transfer race - no capacity leak/double release observed.** A reentrant disposal from the snapshot callback immediately after record insertion returned a `watch_1` result while leaving an empty snapshot; the first subsequent reservation succeeded and the second was rejected, showing one release rather than two. The returned ID is already settled in that artificial reentrant case, but normal lifecycle disposal cannot interleave between the final synchronous check and return; this is not an additional capacity blocker.
- **B9 remains partial.** The added parent-retarget regression is useful, but the suite still does not pin the disposal-time hung registration, stop-all pending cancellation, or target-level symlink replacement that exposed the blocker above.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-5 blocker fixes pushed in `6902054732c819c47565d78110e122165fa2a1cd`.

- **N1 disposal / stop-all cancellation:** pending native registrations are now aborted before capacity release when `dispose()` or `stopAllFiles()` calls `#finishPending()`. The existing bounded `#registrationAwait()` rejects the in-flight registration, preventing a hung filesystem operation from surviving teardown. Added deterministic watchdog regressions for both disposal and stop-all with a never-resolving `open()` and event/state subscription (no sleeps); both verify rejection and immediate capacity reuse.
- **N3 target-level symlink swap:** each poll now validates the canonical target with `lstat()` and `realpath()` before followed `stat()`/digest. Symlink targets and canonical-resolution drift fail closed with a `watcher error` summary and no line emission. Added regressions for both missing-target create-later symlink and existing-target modify-time replacement.

Evidence:
- Failing-first: both symlink repros initially emitted `watcher completed`; the disposal watchdog could not complete with the old cleanup behavior. After the fix, all four focused regressions passed.
- Targeted terminal-monitor + permission parser suites: **169/169 tests passed** (terminal monitor **28/28**).
- `node scripts/check-pr-changelog.mjs --base origin/main`: **PASS**.
- Biome formatting/check completed; `git diff --check` passed.
- Package build was attempted but is blocked by pre-existing workspace resolution/type errors (missing `@earendil-works/pi-agent-core`, `pi-client`, and `pi-protocol`, plus unrelated existing errors); no changed-file error was reported before those failures.

Commit includes:
`Co-authored-by: ananjaemin <81001494+ananjaemin@users.noreply.github.com>`

code-yeongyu:
WORKING: review-1191-r6 - reading round-5 review, author summary, and current diff

code-yeongyu (commented):
Round-6 adversarial re-review against `6902054732c819c47565d78110e122165fa2a1cd`. Focused terminal-monitor + permission-parser run: 55/55 passed. Changed-file LSP diagnostics are clean and the worktree is unmodified.

## Round-5 blocker resolution

- **N1 disposal cancellation: VERIFIED.** With `open()` mocked to never resolve, `dispose()` caused `registerFile()` to reject within the 100 ms watchdog and the maxSessions=1 reservation was immediately reusable. The underlying mocked filesystem promise is naturally not cancellable, but the registration wrapper is aborted and no monitor/event survives teardown.
- **N1 stopAllFiles cancellation: NOT RESOLVED (blocker).** The same never-resolving `open()` probe calls `stopAllFiles()` and then races the registration against a 100 ms watchdog: it is still pending and the reservation is still full. The current `stopAllFiles()` only increments `#lifecycle` and settles entries in `#files`; it never calls `#finishPending()` for `#pending`. The registration eventually rejects only when its 5-second deadline fires. The new regression passes only because it waits for that deadline and does not assert prompt rejection/capacity reuse, so it does not pin the claimed fix. This leaves `kill_bash({all:true})` able to return while a pending monitor tool call retains a terminal slot.
- **N2 create-later target symlink: VERIFIED.** The missing-target-to-external-symlink reproduction produced a `target identity changed` watcher summary, no line event, and an empty snapshot.
- **N2 existing-target modify symlink: VERIFIED.** Replacing the existing target with an external symlink produced the same fail-closed summary/no-emission behavior.
- **Parent drift/directory swap: VERIFIED.** Replacing the monitored parent with a symlinked external directory produced `watcher error: monitored parent changed`, no line event, and cleanup. Parent revalidation covers this case.

## Remaining security blockers

- **Validation-to-digest TOCTOU: NOT ACCEPTABLE (blocker).** Polling now does `lstat()` + `realpath()` + followed `stat()`, but `#digest()` then opens the path in a separate operation. A deterministic probe armed a replacement immediately after the validation `stat()`: the target was swapped to an external symlink before digest, and the monitor followed it and emitted `modify <path>`/`watcher completed`. The window is narrower, but it is still an authorization bypass, not merely a theoretical race. The lean changes documentation describes validation and bounded sampling but does not document this accepted limitation. The read should be tied to a validated descriptor/identity, or the implementation must otherwise close/recheck the gap before delivery.
- **Hardlink identity: NOT RESOLVED (blocker if target identity is the safety invariant).** Replacing the same-path regular file with a hardlink to a different inode leaves `realpath()` unchanged because realpath is pathname canonicalization, not inode identity. The probe passed an external file through that hardlink and observed a `modify` line followed by `watcher completed`, rather than a target-identity error. If the replacement has identical size/mtime/content, the transition is silently missed instead. Pin and compare `dev`/`ino` (or use an opened handle/fstat) if same-path target identity must remain authorized and stable.

## Other adversarial checks

- **realpath failures: no additional blocker.** `ENOENT` is treated as absence with no emission (and preserves the intended create-wait behavior). Synthetic `EACCES` and a real parent symlink loop (`ELOOP`) both resulted in a watcher-error summary, no line event, an empty snapshot, and no process crash.
- **Abort ordering: no outward emission after settlement, but underlying poll I/O is not cancellable.** Settlement is idempotent, removes the record, and the final emission path checks `record.settled`; the hung-open disposal probe did not leave a live record or emit after teardown. A poll operation that completes after settlement can still update the now-detached record object after its digest, but it cannot reinsert or deliver it.
- **Polling cost: bounded, no standalone blocker.** Each active watch is single-flight; the 250 ms pass performs bounded metadata validation and at most three 64 KiB digest samples. Shared terminal reservation bounds active native watches by `maxSessions` (32 by default), so the work is predictable, though it is still meaningful I/O at the cap.

The focused tests are green, but the stop-all lifecycle race and the two target-identity races still violate the fail-closed authorization/lifecycle contract.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-6 blocker fixes pushed in `d03ba092db7a237b7b95e4426c984cc5d20f66a6`.

- N1: `stopAllFiles()` now routes every pending registration through `#finishPending()`, aborting the registration promptly and releasing its reserved terminal capacity. The regression uses a never-resolving `open()` and a 100 ms watchdog; it failed before the fix and passes after it.
- N2: native modify polling opens the target before validation and computes the digest from that same `FileHandle`; it also rechecks the pathname identity around the digest. The deterministic replacement-after-open regression fails closed with no line event instead of following an external symlink.
- N3: approved target identity now includes `dev` and `ino`; hardlink replacement is rejected as `target identity changed` with no modify emission.

Evidence:
- Failing-first: the strengthened stop-all watchdog, deterministic mid-window replacement, and hardlink-swap tests each failed on the pre-fix implementation.
- Green: terminal monitor suites `53/53` tests passed; permission parser suites `141/141` tests passed.
- Changelog gate against `origin/main`: `PASS` (`changes.md` coverage complete; CHANGELOG entry updated).
- Commit trailer verified: `Co-authored-by: ananjaemin <81001494+ananjaemin@users.noreply.github.com>`.

code-yeongyu:
WORKING: review-1191-r7 - rerunning focused tests in the PR worktree and probing atomic-save, descriptor cleanup, sampling, and inode-reuse semantics

code-yeongyu (commented):
Round-7 adversarial re-review against `d03ba092db7a237b7b95e4426c984cc5d20f66a6`.

Verification completed against the PR worktree (`/Volumes/mengmotaStorage/local-workspaces/senpi-wt/pr-1190-lean`):

- `npx vitest --run packages/coding-agent/test/suite/terminal-monitor.test.ts packages/coding-agent/test/permission/parsers.test.ts`: 2 files, 57 tests passed.
- The three targeted round-6 regressions (stop-all cancellation, replacement after open, and hardlink replacement): 3/3 passed.
- `git diff --check` passed and the worktree has no tracked changes.
- Additional temporary probes were run outside the repository; no source, test, commit, or push changes were made.

## Round-6 blocker resolutions

- **N1 stopAllFiles cancellation: VERIFIED.** The never-resolving `open()` regression now rejects within the 100 ms watchdog after `stopAllFiles()`, and a `maxSessions: 1` reservation is immediately reusable. The pending registration is removed/aborted through `#finishPending()` rather than waiting for the 5-second deadline.

- **N2 validation-to-digest replacement race: VERIFIED for the reported race.** I armed a replacement after the target was opened and during the descriptor-based digest. The digest read the original opened inode (the sampled bytes were from the original target, never the external symlink), the post-digest pathname/descriptor identity check failed closed, and the monitor emitted no line.

- **N3 hardlink/inode replacement: VERIFIED.** Replacing the target with a hardlink to a different inode produced a `target identity changed` summary and no modify line. The added `dev`/`ino` checks cover this case.

## Descriptor and fingerprint probes

- **FileHandle cleanup: VERIFIED.** Instrumented handles were closed on both an identity-error path after `open()` and a digest read-error path. Both probes observed exactly one close for the opened handle. The pre-digest catch closes handles, and the digest phase has an unconditional `finally` close.

- **Digest contract: VERIFIED.** The descriptor digest uses the same bounded first/middle/last 64 KiB regions as the prior path digest. For a 200 KiB file, the observed read offsets were 0, 69,632, and 139,264 with 65,536 bytes each, matching the expected formula. The digest is therefore tied to the opened handle without changing the sampling contract.

- **Inode reuse/ABA: NOT REPRODUCED, but the behavior is not fail-closed if it occurs.** A delete/recreate loop of 100,000 APFS attempts did not recycle the original `(dev, ino)`. If the filesystem does recycle that tuple during an unobserved delete+recreate window, the current identity check cannot distinguish the new inode from the old one and can proceed to digest/emit as though the authorized target were unchanged. If the delete is observed first, the later recreation follows the normal absent-to-present state path. This is an ABA limitation of `(dev, ino)` pinning, not a demonstrated failure on this filesystem.

## Critical new semantic blocker: atomic-save modify is broken

- **BLOCKER - legitimate atomic-save replacement fails closed on every new inode.** I registered an existing `modify` watch, wrote a temporary sibling, and atomically renamed it over the watched path. The replacement had a different inode, and the monitor emitted only:

  `watcher error: Cannot watch file: target identity changed: ...`

  There was no `modify` line. This is the standard temp-file-then-rename behavior used by editors and by this repository's own atomic-save code. The current check at `monitor-registry.ts` lines 365-368 rejects every replacement inode before the modify event can be evaluated. The implementation has over-reached from preventing unauthorized redirection to rejecting the core modify-watch transition. The contract needs an explicit delete+create/rebind transition for an authorized path (with the new target validated and read through its own descriptor), or the API must explicitly disallow atomic-save modify watches; the current advertised `event: "modify"` behavior is not viable.

## Additional target-validation blockers

- **BLOCKER - registration still has a validation-to-digest authorization gap.** Polling now binds digesting to a validated descriptor, but initial baseline registration still does `stat(path)`, `realpath(path)`, and then a separate path-based `#digest(canonicalPath)`. I deterministically replaced the initially internal regular target with an external symlink immediately before that baseline `open()`. Registration read `SECRET` from the external target and returned a live `watch_1` record. This occurs after permission classification, so an internal read approval can still be used to read an external target during registration. Baseline creation needs the same open/fstat/path-rebind fencing as polling.

- **An initially symlinked target is also accepted and pinned to its destination.** `initial = stat(path)` follows the link, then `canonicalPath = realpath(path)` discards the fact that the requested target was a symlink. A direct registration against an existing symlink followed modifications to its destination and emitted `modify`/`watcher completed`; the later `lstat(record.canonicalPath)` only sees the resolved regular file. Initial target validation should apply the same no-symlink policy as live replacement validation.

The focused suites are green and the three round-6 reproductions are fixed, but the atomic-save regression is a user-visible correctness blocker and baseline registration still violates the target-identity/authorization contract.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-7 blocker fixes pushed in `24447bcae77c62c7b2bf9e9ff80bb2558ec2c971`.

- **B1 atomic-save regression:** modify reconciliation now accepts an inode replacement only when the requested canonical path still resolves to the same approved parent, is not a symlink, and the opened target is a regular file. Safe temp-file-then-rename saves emit `modify`; symlink swaps, external retargets, non-regular targets, and hardlink retargets still fail closed.
- **B2 registration TOCTOU:** initial validation opens the target once, validates it via `lstat`/`fstat` and canonical-path checks, and computes the baseline fingerprint through that same `FileHandle`. The handle is closed on all success and failure paths, so registration cannot digest an external symlink target after path-based validation.
- **B3 initially symlinked target:** registration now rejects an existing symlink before opening/fingerprinting it; no external destination is silently pinned.
- **ABA note:** `(dev, ino)` reuse remains theoretically indistinguishable in an unobserved delete/recreate window; observed deletion follows the normal absent-to-present path, and this filesystem-level limitation is documented. Safe atomic replacement is independently constrained by no-follow canonical-path validation.

Failing-first evidence: newly added atomic-save and initially-symlink tests failed on `d03ba092` (atomic save timed out with only identity error; initial symlink registered), and the deterministic registration-swap shape was pinned before the fix. After the fix, the focused terminal-monitor + permission-parser run passes **60/60**.

Gates:
- `npx vitest --run packages/coding-agent/test/suite/terminal-monitor.test.ts packages/coding-agent/test/permission/parsers.test.ts`: **PASS, 2 files / 60 tests**
- `node scripts/check-pr-changelog.mjs --base origin/main`: **PASS**
- `npx biome check` on touched TypeScript files: **PASS**
- `git diff --check`: **PASS**
- `gh pr view 1191 --json headRefOid`: **24447bcae77c62c7b2bf9e9ff80bb2558ec2c971**

Commit includes:
`Co-authored-by: ananjaemin <81001494+ananjaemin@users.noreply.github.com>`

code-yeongyu:
WORKING: review-1191-r8 - collecting round-7 evidence and the post-fix diff

code-yeongyu (commented):
Round-8 adversarial re-review at `24447bcae77c62c7b2bf9e9ff80bb2558ec2c971`. I reran the focused terminal-monitor + permission-parser suites: 2 files / 60 tests passed. Biome, `git diff --check`, and LSP diagnostics for the changed TypeScript files are clean.

## Round-7 blocker resolutions

- **B1 atomic-save regression: VERIFIED for the intended safe shape.** The repository regression and a standalone temp-file-then-rename probe both delivered exactly one `modify` line followed by `watcher completed`. The replacement stayed at the approved canonical path, was not a symlink, and was a regular file.

- **B2 registration validation-to-digest race: VERIFIED only for the previously reported window.** The deterministic `open()`-then-swap probe rejected registration, and the instrumented open paths contained only the original target; the external `SECRET` path was never opened during registration. The digest is taken from the original `FileHandle`.

  **NEW BLOCKER (registration canonical-path race): unresolved.** There is still a window after the `lstat`/`fstat`/rebound checks and before `canonicalPath` is captured (`monitor-registry.ts:236-243`). I deterministically swapped the target to an external symlink on that final `realpath(path)` call. Registration returned `watch_1` without opening the external path during baseline creation, but the record pinned `canonicalPath` to the external destination; the first poll then opened the external `secret` and emitted `modify`. The later poll checks validate `dirname(record.path)` against the approved parent, but do not require `dirname(record.canonicalPath)` to remain that parent. The registration result must not be able to acquire a post-registration external canonical path.

- **B3 initially-symlinked target: VERIFIED.** Registration against an existing symlink rejects with `target is a symbolic link` before a handle is opened.

## Adversarial replacement predicate

- Existing `modify` watch + symlink swap: **VERIFIED fail closed** with an identity-change summary and no line.
- Existing `modify` watch + hardlink replacement: **VERIFIED fail closed** with an identity-change summary and no line. The `nlink > 1` guard detects the external-secret hardlink shape for an already-present target.
- Standard external regular-file rename (`nlink === 1`) into the approved path: the probe delivered `modify`. This is indistinguishable from a safe atomic save under the stated predicate (same canonical location, no symlink, regular file), so I do not treat that shape as independently rejectable within the lean predicate. However, the PR/author summary currently says external retargets fail closed; that wording is not true for a moved regular inode and should either be narrowed/documented or backed by a stronger provenance contract.
- Missing `create` watch + hardlink to an external secret: **NEW BLOCKER.** The probe delivered `create` and completed after hashing the hardlinked inode. The `nlink` guard is conditional on `record.present` (`monitor-registry.ts:387`), so the exact adversarial hardlink shape is accepted when the watched path was initially absent. If hardlink retargets are meant to fail closed, the guard needs to cover newly-present create targets too; otherwise the lean contract must explicitly allow the monitor to read such externally-owned content.

## Registration resource cleanup

- **NEW BLOCKER (late `open()` settlement leaks descriptors).** A deterministic async probe made `open()` acquire a descriptor, then triggered `stopAllFiles()` before the `open()` promise settled. Registration rejected, but the late handle was dropped because `initialHandle` was never assigned; `/dev/fd` increased by one and the handle had to be closed by the probe. Repeating this registration/cancellation shape leaks one descriptor per attempt. `#registrationAwait()` races the operation against abort but does not close a successful operation that settles after the race loses (`monitor-registry.ts:174`).

- **NEW BLOCKER (canonical-path error leaks the initial descriptor).** A deterministic failure of the final canonical-path `realpath` after a successful open left the initial handle open: `/dev/fd` again increased by one. The catch at `monitor-registry.ts:241-244` calls `cleanupRegistration()` but does not close `initialHandle`. This is a separate ordinary error path, not merely the late-settlement race.

Successful concurrent registrations do close their baseline handles promptly: a 10-watch probe showed no `/dev/fd` increase while the watchers were live or after disposal. The leaks above are nevertheless sufficient to exhaust descriptors across repeated cancellation/error cycles, and the author summary's “closed on all paths” claim is not yet true.

## Other review points

- The documented ABA limitation is honest: `(dev, ino)` reuse in an unobserved delete/recreate window remains theoretically indistinguishable, while observed deletion follows the absent-state path. I found no stronger claim in the documentation than that limitation.
- The focused 60/60 run confirms the added regression tests and existing monitor behavior, but it does not cover the canonical-path race, create-time hardlink case, or the two descriptor-cleanup races above.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-8 blocker fixes pushed in `4076e9ec2a420cb51898d82c1edb65773c0d9786` (co-authored-by trailer present).

- **N1 canonical-path race:** registration no longer derives the record identity from a late unchecked `realpath(path)`. It binds `canonicalPath` to the already-approved canonical parent plus the requested basename.
- **N2 missing-target hardlink:** the `(dev, ino)` change plus `nlink > 1` fail-closed guard now applies to newly-present create targets as well as existing modify targets. Added a deterministic regression test.
- **N3 descriptor cleanup:** cancellation-aware registration now hands a late successful `open()` result to cleanup, and all baseline descriptor paths retain `finally` cleanup. The obsolete final canonical-resolution failure path was removed by binding to the approved parent identity.
- **N4 docs:** narrowed the contract: symlink/hardlink retargets fail closed, while a regular-file rename with `nlink === 1` is indistinguishable from safe atomic replacement and may be reported as `modify`.

Verification:
- Failing-first/green focused terminal-monitor + permission-parser run: **61/61 passed**.
- Biome and `git diff --check`: **passed**.
- Changelog gate against `origin/main`: **PASS** (`changes.md coverage complete; changelog entry updated`).
- Changed test-file LSP diagnostics: **clean**; source diagnostics were clean on the prior run (the final refresh timed out without producing diagnostics).

No force-push or other worktrees/branches were touched.

code-yeongyu:
WORKING: review-1191-r9 - verifying canonical identity, hardlink fail-closed behavior, descriptor cleanup, atomic-save delivery, and contract wording

code-yeongyu (commented):
Round-9 adversarial re-review at `4076e9ec2a420cb51898d82c1edb65773c0d9786`.

Verification completed against the supplied PR worktree:
- Focused `terminal-monitor.test.ts` + `parsers.test.ts`: **61/61 passed**.
- `git diff --check` and Biome on all touched TypeScript files: **PASS**.
- Changed-file LSP diagnostics: **no diagnostics**.
- Independent temporary Vitest probes: **7/7 passed** for the named round-8 shapes.

## Round-8 blocker resolutions

- **N1 canonical-path capture race - VERIFIED.** `canonicalPath` is now synchronously derived as `join(approvedParent, basename(path))`; the removed late target `realpath()` cannot become record identity. A deterministic target swap during canonical-resolution validation rejected registration, opened only the originally requested path, and never opened the external secret.
- **N2 missing-target external hardlink - VERIFIED.** A create watch whose absent target was replaced by an external hardlink failed closed with `target identity changed`, emitted no line, and settled/cleaned up. The guard now applies independently of `record.present` via the opened inode's `nlink > 1`.
- **N3 descriptor cleanup - VERIFIED.** A successful `open()` that settled after disposal was handed to late cleanup and closed exactly once. A canonical-resolution failure after a successful open also closed the descriptor exactly once.
- **B1 atomic-save modify - VERIFIED.** A regular temp-file-then-rename replacement still emitted `modify <path>` and completed; the accepted `nlink === 1` rename behavior did not regress.
- **N4 contract wording - VERIFIED.** `changes.md` now explicitly limits the guarantee: symlink/hardlink-to-unrelated-inode replacements fail closed, while a regular-file rename with `nlink === 1` is indistinguishable from safe atomic replacement and may produce `modify`; inode ABA remains documented.

## New blocker

- **N1 permission-approval TOCTOU across parser and registration - NOT RESOLVED (blocker).** `parsers.ts` classifies the requested path using `isExternalPath()` before execution. I created an in-workspace symlink parent pointing to an in-workspace directory; parsing `monitor({path})` returned only `[read]`. Before `registerFile()` ran, I retargeted that parent symlink to an external directory. Registration then realpathed the *new* external parent, returned a live `watch_1`, opened the external target, and emitted `create <requested path>` for an externally created file. There was no `external_directory` request and no fail-closed summary. This is the same symlink-retarget class the contract says must fail closed and violates the documented permission-boundary guarantee. The runtime parent revalidation protects only drift after the registration-time approval; it does not bind the parser's approval to the parent identity used for registration.

No other new blocker found.

VERDICT: REQUEST_CHANGES

code-yeongyu:
Round-9 blocker resolved in 52429b47f.

- Added a failing regression covering parser approval of an internal symlink parent, retargeting before registration, and an external file creation; it initially observed the parser return `[read, external_directory]` because the test fixture was outside the workspace, then passed with the corrected in-workspace fixture and failed closed after the implementation.
- The parser now captures the approved canonical parent identity on the tool input, and `registerFile()` compares the live canonical parent against that captured identity before opening/registering the watch. A retargeted parent is rejected before the external target can be opened or emit events.
- Targeted monitor regression: 35/35 passed. Permission parser suites: 63/63 passed.
- Changelog gate against `origin/main`: PASS (`changes.md` coverage complete; changelog entry updated).
- TypeScript and Biome checks: PASS; `git diff --check`: PASS.

Co-authored-by trailer is present on the commit.

code-yeongyu (commented):
Round-10 adversarial re-review against `52429b47fdb0bbbaa37f1fb7ea8a40bcad4af9ab`.

## Round-9 blocker resolution

- **Approval-to-registration parent TOCTOU - VERIFIED.** I reproduced the exact shape with an in-workspace symlink parent initially resolving to an internal directory. The parser returned only `[read]` and captured that canonical parent identity; I retargeted the symlink to an external directory before registration. `registerFile()` rejected on the approved-parent comparison before opening or registering the retargeted target. An instrumented `node:fs/promises.open()` saw no open of the external target, no `line` event was emitted, and the registry snapshot remained empty.
- The end-to-end binding is intact: the parser stores the approval identity on the same tool-input object, the permission hook preserves it, and `monitor.execute()` forwards it as `approvedParent`.

## Regression spot-checks

- **Atomic-save modify:** regular temp-file-then-rename delivered exactly one `modify <path>` line followed by `watcher completed`.
- **Symlink swap:** replacing an approved target with an external symlink produced a `target identity changed` summary, no line, and an empty registry.
- **Capacity release:** with `maxSessions: 1`, a live native watch blocked a second reservation; `stopAllFiles()` released the slot and the next reservation succeeded.
- Independent representative probe: **4/4 passed** (approval race, atomic save, target symlink swap, capacity release).

## Verification

- Focused monitor + permission parser suites: **62/62 passed**.
- Changed-file LSP diagnostics: **no diagnostics**.
- Biome on touched files, `git diff --check`, changelog gate, and repository TypeScript check: **PASS**.

No new blocker found. The remaining bounded-fingerprint and inode-ABA behavior is consistent with the documented lean contract and its named limitations.

VERDICT: APPROVE

PR #1194, opened 2026-08-29
feat(compaction): warm the summary instead of stopping to build it

code-yeongyu:
## What this changes

Compaction today optimizes *when to stop and summarize*. This change set moves it toward not stopping at all: the summary is warmed in the background and applied at a boundary the user never waits on, and the growth that used to blow past the threshold in a single step is capped at the entrance.

Six behaviors, each independently gated by a setting that restores the previous behavior when disabled.

### Tool-result admission (the load-bearing one)

A single tool result is capped at `min(50000, max(8192, 5% of window))`. The full output is written to a spill file and the kept excerpt carries a marker naming that path, so the model can read it back on demand.

This matters because of what the telemetry says: over 4.2 months of real sessions, single-step context growth at a 1M window reaches 174K tokens at p99 and 375K at p99.9, and roughly 89% of the jumps above 100K are genuine tool bursts, not artifacts. No static threshold survives a 375K single-step jump. Capping at the entrance is what makes every threshold below it meaningful.

Marker detection scans the whole excerpt rather than testing `startsWith`, because the marker sits between the head and tail slices — a result re-processed under a smaller window would otherwise spill twice and corrupt the pointer chain.

### Warm summary on an absolute lead

Speculation used to start at a multiplicative fraction of the threshold, which on a 1M window meant beginning roughly 190K tokens early — long enough that ordinary turn growth invalidated the summary before it could land. Observed apply rate was 17 of 595 generated summaries, against 3,907 invalidations.

The lead is now `clamp(threshold * 0.125, 8192, 32768)` tokens, resolved once through a single helper that also clamps a configured override, and the same value feeds the trigger, the grace band, and the reminder. Idle generation may also begin at 50% fill, while the apply gates stay exactly where they were. A stale warm summary is discarded and regenerated from the current transcript; carrying the prior summary forward as context was prototyped and cut from this PR because nothing populated it, so it would have shipped as dead code.

### Grace band

When the context crosses the threshold while a speculative generation is genuinely in flight, blocking is deferred until `min(threshold + lead, window - reserve)`. A completed or armed warm job is consumed immediately rather than deferred — deferring on an already-finished job was a real defect caught in review, and it broke five blocking-route regressions.

No path allows a session past `window - reserve` without blocking.

### Scaled reserve

`max(configured, min(floor(0.04 * window), 49152))`. At 200K this is unchanged at 16,384; at 1M it becomes 40,000, moving the hard valve from 98.4% to 96% of the window. A fixed 16K reserve on a 1M window absorbs less than twenty median steps and none of the tail.

### Token-budget reminder

One reminder per compaction generation, fired once the remaining runway falls inside twice the lead, so the model wraps up verbose exploration before the summary is taken. It is delivered without discarding a pending restoration payload — an earlier revision returned the reminder *instead of* that payload, permanently dropping it.

### Deterministic fallback during breaker cooldown

When the circuit breaker is tripped and the context is over the threshold, the existing no-LLM context reduction runs immediately instead of waiting out the 60-second cooldown. Emergency valve semantics are unchanged.

## Threshold table

| Window | Before | After |
|---|---|---|
| ≤16K / ≤32K / ≤64K / ≤128K | 0.45 / 0.50 / 0.55 / 0.60 | unchanged |
| ≤512K | 0.65 | **0.70** |
| >512K | 0.65 | **0.80** (new tier) |
| clamp | [0.40, 0.70] | **[0.40, 0.85]** |

Low-yield feedback can no longer push a ratio above its tier base. The numeric `lastYield` branch that returned `Math.max(contextWindow, lastYield)` — a token count where every caller expected a ratio, silently disabling compaction — is removed.

For context: this sits below where comparable harnesses run (openai/codex compacts at 90% of the window with a 95% hard cap; another pi-derived fork at ~91.8%). Their higher numbers are licensed by mechanisms this change set is introducing here, so the tiers are deliberately set below them rather than matched to them.

## Verification

- `npx vitest --run test/compaction test/compaction.test.ts` — all suites pass
- `npx tsc -p tsconfig.build.json --noEmit` — clean
- Root pre-commit gate (biome, type check, dependency locks) passed on every one of the six commits
- New coverage: threshold clamp bounds, large-window tiers, reserve and asymmetric feedback, speculation lead and grace geometry, admission spill and re-admission, reminder gating, wiring decisions

An adversarial review pass rejected the first revision with eight blockers, every one a defect the green suite did not catch: the clamp escape hatch, an unclamped configured lead with two divergent resolution sites, the `startsWith` marker check, the dropped restoration payload, a shared harness that disabled the grace band for tests whose whole purpose was to pin it, two logger events forced through with `as never` and therefore absent from both the type union and the runtime allowlist (silently discarded), oversized files grown further, and a pipeline doc left describing the old behavior. All eight are fixed in this branch.

## Files

New modules keep the pre-existing oversized files from growing: `speculation-lead.ts`, `tool-admission.ts`, `token-budget-reminder.ts`, `warm-summary-regeneration.ts`, `speculative-job.ts`, `orchestration.ts`, `context-pipeline.ts`, `model-selection.ts`, `compaction-settings.ts`, `ideal-compaction-settings.ts`, `compaction-settings-resolver.ts`.

## Settings

`graceBandEnabled`, `toolAdmissionEnabled`, `reminderEnabled`, `reserveScalingEnabled` (all default true) and `speculativeLeadTokens` (clamped when set).


packages/coding-agent/src/core/extensions/builtin/compaction/orchestration.ts
```diff
@@ -0,0 +1,102 @@
...
+			typeof message.content === "string"
+				? message.content
+				: message.content
+						.filter((part: { type: string; text?: string }) => part.type === "text")
+						.map((part: { type: string; text?: string }) => part.text ?? "")
+						.join("\n");
+		if (!text) return message;
+		const admitted = admitContextToolResult(text, contextWindow, join(tmpdir(), "senpi-tool-spill"));
+		return admitted.admitted ? { ...message, content: [{ type: "text" as const, text: admitted.text }] } : message;
+	});
```
minpeter, line 192: The text parts are joined here and the entire `ToolResultMessage.content` is then replaced with one text block. Any `ImageContent` or other non-text block in the original result is dropped, with no downstream recovery path.

Please apply admission only to the text content and preserve every non-text block in its original order.

packages/coding-agent/src/core/extensions/builtin/compaction/tool-admission.ts
```diff
@@ -0,0 +1,106 @@
...
+function spillToFile(spillDir: string, text: string): string {
+	mkdirSync(spillDir, { recursive: true });
+	const spillPath = join(spillDir, `tool-result-${Date.now()}-${randomBytes(3).toString("hex")}.txt`);
+	writeFileSync(spillPath, text, "utf-8");
+	return spillPath;
+}
+
+function buildExcerpt(text: string, budgetChars: number, totalTokens: number, spillPath: string): string {
+	const headChars = Math.max(1, Math.floor(budgetChars * HEAD_BUDGET_FRACTION));
+	const tailChars = Math.max(1, Math.floor(budgetChars * TAIL_BUDGET_FRACTION));
```
minpeter, line 82: Context projections are not persisted, so marker state does not survive across provider requests. Every request re-spills the same original result to a new timestamp/random path, and there is no cleanup, so disk usage grows throughout a long session.

The POSIX modes also depend on the caller's umask. With the common `0002` umask, I reproduced directory mode `0775` and file mode `0664`, making spilled tool output readable by other local users.

The preferable fix is a deterministic projection that does not touch disk. If read-back is required, use session-private, content-addressed, write-once storage with explicit restrictive modes, a quota, atomic writes, and cleanup.

packages/coding-agent/src/core/extensions/builtin/compaction/tool-admission.ts
```diff
@@ -0,0 +1,106 @@
...
+	const head = text.slice(0, headChars);
```
minpeter, line 83: `mkdirSync`/`writeFileSync` errors escape admission. `ExtensionRunner` catches the context-handler failure and continues with the original messages, so a disk error such as `ENOSPC`, `EACCES`, or a read-only filesystem silently skips every downstream compaction transform, not just spilling.

Please fall back to a bounded, deterministic in-memory projection so the rest of the context pipeline can still run.

packages/coding-agent/src/core/extensions/builtin/compaction/orchestration.ts
```diff
@@ -0,0 +1,102 @@
...
+	if (!shouldWarmAtIdle(decision)) return "none";
+	if (!job) return "start";
+	const currentTokens = decision.usage?.tokens ?? 0;
+	return isWarmResultStale(job.armedAtTokens, currentTokens, decision.settings.keepRecentTokens) ? "replace" : "none";
+}
+
+export function admitContextToolResult(
+	text: string,
+	contextWindow: number,
+	spillDir: string,
```
minpeter, line 90: Text matching this model-visible marker is treated as already admitted before its size is calculated. Because user and tool output can reproduce the marker, they can bypass the admission cap with arbitrarily large content.

Please remove trust in the visible marker and re-derive admission deterministically from the content. If duplicate writes are the concern, use a content-hash cache rather than a forgeable text sentinel.

packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
```diff
@@ -923,33 +843,46 @@ export default function compactionExtension(
 			startSpeculativeCompaction(ctx, PROACTIVE_COMPACTION_INSTRUCTIONS);
 		}
 
-		return message ? { message } : undefined;
+		const reminder = computeTokenBudgetReminder({
```
minpeter, line 848: Commit `7a8aecb` deliberately stopped emitting a standalone synthetic reminder because that extra turn suppressed model fallback; that regression fix should remain.

The separate issue is that the advertised pre-compaction reminder is still computed and `lastFiredGeneration` is consumed, but the reminder is not delivered on ordinary turns. The generation is therefore spent on a message the model never sees.

Please arm delivery in `before_agent_start` and inject it as ephemeral context immediately before the real user message. Key it to the accepted-compaction epoch and retain a retry/fallback-safe lease so retries do not consume or duplicate the logical delivery. This should not recreate a persisted synthetic turn.

packages/coding-agent/src/core/extensions/builtin/compaction/context-pipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+	emergencyPruneLatch: EmergencyPruneLatch;
+}) {
+	const admittedMessages = admitContextToolResults(
+		input.event.messages,
+		input.contextWindow,
+		input.toolAdmissionEnabled,
+	);
+	const sourceMessages =
+		input.breakerFallback ||
+		shouldApplyContextReduction({
```
minpeter: A normal `external-owner` rejection currently debits Senpi's breaker. Once it trips, `breakerFallback` reduces context in the SDK-owned lane even though the lane contract requires Senpi to stand down there.

Please apply both guards: exclude `external-owner` rejections from breaker debit, and independently gate fallback reduction on Senpi owning the compaction lane.

minpeter (changes requested):
Six independently reproduced correctness and safety issues block approval in the current state. They cover irreversible mixed-content data loss, an unsafe and unbounded spill lifecycle (including fail-open behavior), a forgeable admission marker, a pre-compaction reminder that is consumed without delivery, and breaker behavior that violates external-owner lane isolation. Details and requested changes are in the inline comments.

packages/coding-agent/src/core/extensions/builtin/compaction/orchestration.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export function resolveCompactionGeometry(input: {
+	contextWindow: number;
+	settings: CompactionPreparation["settings"];
+	lastYield?: { savedTokens: number; tokensBefore: number };
+}): CompactionGeometry {
+	const thresholdTokens = input.contextWindow * policy.computeEffectiveThreshold(input.contextWindow, input.lastYield);
+	return {
+		reserveTokens:
+			input.settings.reserveScalingEnabled === false
```
minpeter: Should the scaled reserve resolved here also govern post-compaction acceptance, or is it strictly the provider-request hard valve?

The PR frames scaling as a hard-limit reserve, so it may be deliberate that `createRequiredCompactionFallback()` in `deterministic-fallback.ts` still validates retained context against `contextWindow - preparation.settings.reserveTokens` rather than the effective reserve computed here. That leaves the two paths on different budgets: at a 1,000,000-token window with configured reserve 16,384, the effective hard-valve reserve is 40,000, so the hard valve targets a 960,000-token prompt budget while deterministic fallback will accept up to 983,616.

That gap does not necessarily produce an oversized provider request, since the hard-limit/emergency pipeline re-runs on the next request. But it can immediately retrigger compaction, burn a recovery cycle, or fall through to lossy emergency pruning.

If scaling is meant to apply to acceptance too, it would be worth sharing one pure effective-reserve resolver across hard-limit, restoration, deterministic fallback, and core acceptance/simulation checks, with boundary tests covering scaling enabled and disabled.

minpeter (commented):
One additional reserve-contract question:

packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
```diff
@@ -1016,6 +949,22 @@ export default function compactionExtension(
...
+					aborted: event.aborted === true,
+					settings,
+					usage,
+					contextWindow,
+					breakerTripped: breaker.isTripped(state, Date.now()),
+					lastYield: state.lastYield ?? undefined,
+					mode: ctx.mode,
+				},
+				speculativeJob,
+			);
```
minpeter, line 978: **[P1] Sub-threshold warm job can poison the later blocking compaction**

The new `else` branch calls `startSpeculativeCompaction()` for `warmAction !== "none"` but, unlike the at-threshold path above, never calls `armIdleWarmupRetry()`. If that below-threshold warm job rejects transiently, the failed job stays stored with no retry armed. When usage later crosses the real compaction threshold, `applyBlockingCompaction()` adopts that existing `pendingJob`, hits the `inheritedFailure` path, clears the job, records a breaker failure, and returns `{ applied: false, reason: "failed" }` without ever generating a fresh summary, so the session skips compaction at the exact moment it is needed and can run into the hard context limit. Please either arm the idle warmup retry here as well, or make threshold admission treat an inherited failed job as absent and fall through to fresh blocking generation. A regression test should fail the sub-threshold warm job, then drive usage past the threshold, and assert a second summarizer invocation plus an applied compaction.

packages/coding-agent/src/core/extensions/builtin/compaction/context-pipeline.ts
```diff
@@ -0,0 +1,50 @@
...
+	 * Context window minus the model's output reserve. The emergency prune budgets against the
+	 * space a request can actually occupy, so it must not see the full window.
+	 */
+	promptContextWindow: number;
+	toolAdmissionEnabled: boolean;
+	breakerFallback: boolean;
+	laneOwnsCompaction: boolean;
+	emergencyPruneLatch: EmergencyPruneLatch;
+}) {
+	const admittedMessages = admitContextToolResults(
```
minpeter, line 28: **[P1] Tool-result admission drops session-entry identity, disabling OpenAI native replay**

`admitContextToolResults()` runs before `markOpenAiRemoteReplayBoundary()`. Admission rebuilds an oversized tool-result message via object spread/content replacement, which does not carry over the non-enumerable/symbol metadata that links the message to its checkpoint entry. `markOpenAiRemoteReplayBoundary()` requires every checkpoint-owned prefix message to retain its exact entry ID, so one admitted oversized tool result makes a valid OpenAI checkpoint fail provenance validation and silently disables native replay. Please preserve the session-entry identity/provenance on the rewritten message and add an end-to-end checkpoint replay test with an oversized checkpoint-owned tool result.

packages/coding-agent/src/core/extensions/builtin/compaction/index.ts
```diff
@@ -1016,6 +949,22 @@ export default function compactionExtension(
 			startSpeculativeCompaction(ctx, idle.IDLE_COMPACTION_INSTRUCTIONS);
 			armIdleWarmupRetry(ctx);
 			armIdleApply(ctx);
+		} else {
```
minpeter, line 965: **[P2] Sub-threshold idle warming duplicates work on OpenAI remote-compaction lanes**

This below-threshold branch starts a local speculative summary for any eligible model, including OpenAI models routed to native remote compaction. When usage later crosses the threshold, `applyBlockingCompaction()` tries remote compaction first and, on success, aborts and discards the local warm job before applying the remote result. A routine idle cycle can therefore pay for a full local summarization pass and a remote compaction for the same context, while never consuming the local output. Please make ownership explicit: skip local idle warming on remote-capable lanes, warm through the remote path, or create the local summary only after remote fallback. A focused test should assert exactly one compaction request per threshold cycle on a remote-capable model.

minpeter (commented):
Three additional independently reproduced findings: two correctness blockers and one compaction-ownership cost issue.

packages/coding-agent/src/core/extensions/builtin/compaction/orchestration.ts
```diff
@@ -0,0 +1,102 @@
...
+				? input.settings.reserveTokens
+				: policy.resolveReserveTokens(input.contextWindow, input.settings.reserveTokens),
+		thresholdTokens,
+		leadTokens: resolveSpeculationLeadTokens(thresholdTokens, input.settings.speculativeLeadTokens),
```
minpeter, line 38: **[P2] Small-context models fall below the budget Senpi needs to operate; reject them instead of silently degrading**

`resolveSpeculationLeadTokens()` floors the lead at `SPECULATION_LEAD_MIN_TOKENS` (8,192), so on a sufficiently small context window the lead meets or exceeds `thresholdTokens` and the speculation start point `thresholdTokens - leadTokens` collapses to zero — speculation is eligible before the baseline system prompt and tool schemas have even been assembled, and every turn on such a model degenerates into compaction of a context that was never usable. I do not think the lead formula itself is wrong, and it should not be clamped: the real issue is that we accept models whose effective pre-compaction budget cannot host a session at all. Please add a compatibility gate that rejects such models with an explanatory error naming the shortfall (window, measured prompt and tool cost, reserves, lead, and required margin) instead of running in a permanently compacting state. The check should compute the actual assembled system prompt plus active tool-schema cost, then subtract the output reserve, compaction reserve, speculation lead, and a conversation safety margin large enough for real work. The minimum acceptable budget/window must be determined from measurements for each model or model family rather than a hard-coded universal 8K/32K cutoff, because prompt and tool cost differ across lanes. A cheap catalog-level sanity check plus a final runtime check at model selection/session setup, with boundary tests just above and below the accepted budget and assertions on the diagnostic, would cover both.

minpeter (commented):
One additional model-compatibility finding for small context windows.

packages/coding-agent/src/core/extensions/builtin/compaction/model-selection.ts
```diff
@@ -0,0 +1,57 @@
...
+	start: () => void;
+}): void {
+	if (input.laneOwnsCompaction) {
+		input.invalidate();
+		return;
+	}
+	if (!matchesSelectedModel(input.speculativeSnapshot, input.ctx)) input.invalidate();
+	const previousWindow = input.event.previousModel?.contextWindow ?? 0;
+	const contextWindow = input.ctx.model?.contextWindow ?? 0;
+	if (previousWindow <= contextWindow || input.breakerTripped) return;
```
minpeter, line 38: **[P2] Reject model downswitches that exceed the target model's usable budget**

The new `previousWindow > contextWindow` branch correctly recognizes that a downswitch is distinct, but it only invalidates incompatible speculative work and may start compaction after the switch has already been committed; there is no refusal or diagnostic contract. A session that has accumulated substantial live context on a 1M model can therefore switch to a 372K target and discover only on the next turn that the request no longer fits, forcing an emergency repair from a state that was never valid for the target. Please admission-check the switch against the target model's effective usable prompt budget, not raw `contextWindow`: account for already-used context, the actually assembled system prompt and active tool-schema cost, output reserve, compaction reserve, speculation lead, and a conversation safety margin. If the projection does not fit, refuse the switch with an explanatory error, recommend or offer compaction first, then revalidate and retry the switch.

The acceptable margin must be established from measurements per model or model family rather than an arbitrary universal cutoff, because prompt/tool overhead and useful headroom differ by lane. I recognize this is slightly broader than the PR's immediate scope, so I would treat it as a follow-up safety contract exposed by this new downswitch handling rather than a regression introduced here. Please cover accept, reject-with-diagnostics, compact-then-accept, and post-compaction revalidation boundaries.

minpeter (commented):
One follow-up model-downswitch safety contract exposed by the new smaller-window handling.

PR #1213, opened 2026-08-30
test(rpc): await widget events with session open

code-yeongyu:
## Summary
- subscribe to both widget events before issuing open_session
- await the open response and event records together using the shared bounded timeout
- prevent unhandled waiter rejections under slow shard startup

## Root cause
The regression test armed independent 1-second timers before triggering open_session, then awaited the open response before observing either event promise. On slow Ubuntu shard startup, the event records arrived after the private timers (or a startup failure rejected both promises), producing the timeout and Vitest unhandled-rejection report. The fix uses the existing 15-second bounded waiter default, subscribes before the trigger, and observes all operations with Promise.all. It is test-only; runtime behavior is unchanged.

## Latest-base verification
- rebased from b106c1078 onto e7131bc9b (latest origin/main, including PR #1212 and PR #1200)
- latest-base gorky Node 24 focused target: 1/1 passed, 0 unhandled errors
- prior rebased b106c1078 gorky Linux focused test: 10/10 green
- prior rebased b106c1078 mengmotaMac focused test: 3/3 green
- exact shard 3/3 includes rpc-socket-host.test.ts; target passed in the shard. The shard still has unrelated pre-existing failures in sdk-session-manager.test.ts and suite/terminal-monitor.test.ts.
- PR #1212's post-#1200 target pass is consistent with #1200 changing the failure profile, but the waiter/unhandled-rejection harness defects remain independently valid and are fixed here.

Remote verification namespaces were removed after testing.

### DIFF
--- a/packages/coding-agent/changes.md
+++ b/packages/coding-agent/changes.md
@@ -1,5 +1,9 @@
 # Local fork changes
 
+## 2026-08-30 - Make RPC widget test event waits deterministic
+
+- The Unix-socket widget regression subscribes to both UI records before issuing `open_session`, waits for the open response and both records together, and uses the shared bounded waiter timeout. This prevents slow shard startup from expiring the record waiters and avoids unhandled rejections from promises that were not yet observed.
+
 ## 2026-08-29 - GLM-5.3 default and prompt preset
 
 - Use GLM-5.3 as the Z.AI global and China default, with the matching built-in prompt preset from current main.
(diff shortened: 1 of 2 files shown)


code-yeongyu:
Closing as superseded by #1203, which landed the same fix for this exact test: cff34c3a3 drops the private 1s waiter timeouts to the harness default (armed-before-trigger preserved), and e2154103d makes JSONL peer waiters lifecycle-safe so armed-waiter rejections can never surface unhandled — covering both defects this PR addressed. The remaining delta here (Promise.all observation) is no longer needed for correctness under lifecycle-safe waiters, and the overlapping hunks would conflict. Verification from this lane (10/10 green on Linux, 3/3 on macOS with the equivalent change) corroborates the approach #1203 shipped.

PR #1248, opened 2026-09-01
fix(ai): honor explicit long-cache retention for Anthropic proxies

amsminn:
## Problem

`compat.supportsLongCacheRetention: true` was ignored for Anthropic-compatible proxies. Both long-retention decision sites additionally required `baseUrl` to use the native Anthropic hostname, so an explicit proxy opt-in could never enable the one-hour cache TTL.

The wire serializer and the internal TTL resolver also duplicated this condition, leaving room for request TTL and cache accounting to drift.

## Solution

Add one shared `allowsAnthropicLongCacheRetention()` helper and use it for both:

- the wire-level `cache_control.ttl: "1h"` decision in `getCacheControl`
- the `300` / `3600` second decision in `resolvePromptCacheTtlSeconds`

The helper preserves tri-state semantics:

- explicit `true`: authoritative on any host, including a custom proxy
- explicit `false`: authoritative on every host, including native Anthropic
- `undefined`: preserve existing hostname inference and the Fireworks fallback

The helper uses nullish coalescing, so an explicit `false` never falls through to host inference.

## Compatibility

| Host | Setting | Before | After |
| --- | --- | --- | --- |
| Native Anthropic | unset | 1h / 3600s | 1h / 3600s |
| Native Anthropic | `false` | 5m / 300s | 5m / 300s |
| Third-party proxy | unset | 5m / 300s | 5m / 300s |
| Third-party proxy | `true` | 5m / 300s | 1h / 3600s |

Only the explicit proxy opt-in changes behavior.

## Impact

For an opted-in `anthropic-messages` proxy using long retention:

- the request now carries `cache_control.ttl: "1h"`
- the internal cache TTL changes from 300 to 3600 seconds
- with the default 30-second safety buffer, the exposed safe-wait value changes from 270 to 3570 seconds

The built-in cache keepalive remains restricted to the native Anthropic host, so this change does not enable keepalive for third-party proxies. Bash process deadlines and terminal foreground windows are independent of this value.

## Verification

- `npm test --workspace=@earendil-works/pi-ai`
  - 257 test files passed
  - 2,493 tests passed
- `npm run check`
  - Biome, policy checks, `tsc --noEmit`, and browser smoke passed
- real exported-resolver matrix: 4/4 passed
  - native / unset -> 3600
  - native / false -> 300
  - proxy / unset -> 300
  - proxy / true -> 3600
- `node scripts/check-pr-changelog.mjs --base origin/main --labels no-changelog`
  - exact-nearest `changes.md` coverage passed

## Out of scope

This PR does not change:

- OpenAI completions or responses cache behavior
- Bedrock cache behavior
- the hostname allowlist
- the public configuration schema
- any file under `packages/coding-agent/src/`
- dependencies, lockfiles, or generated model data

The `no-changelog` label is required because contributor policy prohibits directly editing the release `CHANGELOG.md`; exact-nearest `changes.md` tracker coverage is included separately.

---
## Summary by cubic
Honors an explicit `supportsLongCacheRetention: true` for Anthropic-compatible proxies, so opted-in proxies now get the one-hour cache TTL instead of the default five minutes. An explicit `false` is also respected on the native Anthropic host; leaving the flag unset keeps the previous hostname-based inference. Only the explicit proxy opt-in changes behavior (TTL 300s → 3600s, safe-wait 270s → 3570s). Built-in cache keepalive stays restricted to the native Anthropic host.

**Refactors**
- The wire-level `cache_control.ttl` decision and the internal TTL seconds resolver now share one `allowsAnthropicLongCacheRetention()` helper, preventing the two from drifting.
- Adds tests for the tri-state matrix and updates the `supportsLongCacheRetention` docs.

<sup>Written for commit 6f4d3291af9c4be8e7840992246244be7cd8fa71. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/changes.md
+++ b/packages/ai/changes.md
@@ -113,6 +113,86 @@
 
 - LOW: `packages/ai/package.json` dependency pins during upstream syncs.
 
+## 2026-08-26 - Make the long-cache retention flag authoritative
+
+### What changed
+
+- `src/utils/prompt-cache-ttl.ts`: added an exported helper `allowsAnthropicLongCacheRetention(model)` that
+  resolves the long-cache decision as a tri-state. An explicit `compat.supportsLongCacheRetention` of `true` or
+  `false` now wins on any host. Only an unset (`undefined`) flag falls back to the previous inference, which is
+  `isAnthropicApiBaseUrl(model.baseUrl)` plus the existing fireworks carve-out.
+- The long-TTL condition used to be duplicated at two decision sites; both now call the one helper. The
+  `anthropic-messages` branch of `resolvePromptCacheTtlSeconds` uses it to pick 3600 vs 300 seconds, and the
+  private `getCacheControl` in `src/api/anthropic-messages.ts` uses it to pick the wire value
+  `cache_control.ttl: "1h"` vs omitting the field.
+- `src/api/anthropic-messages.ts` no longer imports `isAnthropicApiBaseUrl`.
+- `getAnthropicCompat` is unchanged. Its return type and the value it reports for
+  `supportsLongCacheRetention` are untouched; the helper is purely additive.
+
+### Why
+
+- `compat.supportsLongCacheRetention` is a documented public field in models.json, but for `anthropic-messages`
+  the 1-hour TTL was additionally gated on a hostname allowlist. A user pointing `baseUrl` at their own
+  Anthropic-wire-compatible proxy and explicitly setting the flag to `true` was silently ignored: the documented
+  knob did nothing.
+- The two sites had to converge for a second reason. While the condition stayed duplicated, a later edit to one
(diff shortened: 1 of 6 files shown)


amsminn:
Maintainer note: please apply the `no-changelog` label. The exact-nearest `changes.md` tracker coverage is included and passes locally; contributor policy prohibits editing the release `CHANGELOG.md`, so the label is required for the release-changelog half of the gate.

PR #1250, opened 2026-09-01
fix(coding-agent): dedupe session tree edges

### LINKED ISSUE #1247: session tree: duplicate entry IDs make getTree traversal explode and freeze the TUI
## Summary

If a stored session contains duplicate entry IDs, the cost of opening the session tree can explode and may not finish within a usable amount of time.

When the conversation input is empty and `Esc` is pressed twice within 500ms, `showTreeSelector()` is called. This function synchronously calls `sessionManager.getTree()` before showing the selector. On the affected 7k-record session, `getTree()` did not finish within 60 seconds, so the entire TUI appeared frozen and the process had to be killed to recover.

By contrast, the same benchmark finished in about 67–73ms on two 6k+ sessions with no duplicate IDs. So the core trigger appears to be corrupted or duplicated entry identity, rather than session size itself.

## Environment

- OMO 5.0.0-0.beta.31
- OMO bundled `@code-yeongyu/senpi` 2026.8.31
- Linux x86_64
- Node.js 24.14.0
- Affected session: about 7.1k JSONL records at the time of measurement

## Reproduction steps

Do not modify the original session; use a copy or a synthetic session.

1. Prepare a healthy session with a long parent-child entry chain.
2. Duplicate several adjacent entry lines without changing `id` or `parentId`.
3. Resume the copied session.
4. With the input empty, press `Esc` twice within 500ms, or run `/tree`.
5. The TUI does not respond while `sessionManager.getTree()` is building the tree.

## Observed behavior

Affected session:

- `SessionManager.getEntries()` entry count at measurement: 7,176
- Duplicate IDs: 108
- Extra duplicate entries: 108
- `SessionManager.open()`: about 215ms
- `SessionManager.getTree()`: did not finish within 60 seconds (timeout)

Control measurements:

| Session | Duplicate IDs | `getTree()` |
| --- | ---: | ---: |
| 143 entries | 0 | 3.4ms |
| 6,431 entries | 0 | 72.9ms |
| 6,244 entries | 0 | 67.4ms |
| 7,176 entries | 108 | >60s / timeout |

The duplicate entries in the problem file were adjacent pairs with the same `id`, `parentId`, type, and timestamp.

## Cause evidence

- `packages/coding-agent/src/modes/interactive/interactive-mode.ts`
  - Double `Esc` calls `showTreeSelector()`.
  - `showTreeSelector()` synchronously calls `sessionManager.getTree()` before creating `TreeSelectorComponent`.
- `packages/coding-agent/src/core/session-manager.ts`
  - `_buildIndex()` stores entries in `byId`, so a later duplicate ID overwrites the earlier map value.
  - `getTree()` creates one `SessionTreeNode` in `nodeMap` per unique ID.
  - It then walks all persisted entries again and appends `nodeMap.get(entry.id)` to the parent's `children`.

When persisted IDs are duplicated, the same node object is added to `children` multiple times. If such duplicate pairs sit consecutively in a parent-child chain, later traversal revisits the same node graph repeatedly. `getTree()` currently has no visited-node guard, so the work grows combinatorially and can effectively never finish.

`TreeList.render()` already renders only the `maxVisibleLines` range. The actual freeze happened in `getTree()` before any attempt to draw every row to the terminal.

## Expected behavior

Duplicate IDs in persisted session data must not freeze the TUI via `/tree` or double `Esc`.

On session load, duplicate entry IDs should be diagnosed clearly, or tree construction should avoid attaching the same node more than once, and `getTree()` traversal must be safe against duplicates and cycles.

## Sharable evidence

The affected session contains private work, so the original JSONL is not attached to this public issue. A minimal synthetic fixture can be provided with message bodies replaced by dummy values while keeping the duplicate-ID structure.

## Duplicate issue search

I searched open/closed issues in this repository for `session tree freeze`, `duplicate entry id`, and related terms, and did not find a directly matching report.

jc01rho:
## Summary

- build session-tree edges once per unique persisted entry ID
- preserve the existing final-entry-wins index semantics for repeated IDs
- add an issue regression covering duplicate child rows and descendant preservation
- document the core divergence in `packages/coding-agent/src/core/changes.md`

Closes #1247.

## Root cause

`SessionManager.getTree()` created one `SessionTreeNode` per unique ID, but then linked nodes by iterating every persisted row. Repeated rows therefore pushed the same node object into a parent's `children` array multiple times. Duplicate edges along a branch multiplied the subsequent iterative traversal and could block the TUI before the tree selector rendered.

## Verification

- RED: regression initially failed with `expected [ 'child', 'child' ] to deeply equal [ 'child' ]`
- `bun run --cwd packages/coding-agent test test/suite/regressions/issue-1247-session-tree-duplicate-entry-ids.test.ts`
- `bun run --cwd packages/coding-agent test test/session-manager`
- `bun run check`
- `bun run build`
- Senpi QA common harness: 10/10
- Senpi RPC QA: 4/4
- Senpi mock-loop QA: 48/48, zero real provider calls
- Senpi TUI smoke: 5/5

## Real-session QA

The affected private session was exercised without exposing message contents:

- 7,202 entries
- 7,094 unique IDs
- 108 duplicate IDs
- pre-fix `getTree()`: exceeded 60 seconds
- fixed `getTree()`: 3.4 ms
- returned 7,094 unique tree nodes and 7,093 edges
- source JSONL SHA-256 unchanged

QA receipts are stored locally under `local-ignore/qa-evidence/` and are not committed.

---
## Summary by cubic
Fixes session tree construction freezing on sessions with duplicate entry IDs. `getTree()` now builds one edge per unique node instead of re-linking the same node for every persisted row, while preserving the existing final-entry-wins index semantics.

**Bug Fixes**
- Duplicate persisted rows previously appended the same node object to a parent’s children repeatedly, multiplying traversal work and blocking the TUI before the tree rendered.
- Adds a regression test covering duplicate child rows and descendant preservation.

<sup>Written for commit a95f34972938c30e76291ff6afdd5f543279eae1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1250?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Fixed
 
+- Session tree construction no longer freezes on sessions with duplicate entry IDs; `getTree()` builds one edge per unique node instead of re-linking the same node for every persisted row ([#1247](https://github.com/code-yeongyu/senpi/issues/1247)).
+
 - Multi-session RPC hosts launched with `SENPI_RPC_CLIENT_CAPABILITIES` (for example `extension_events`) now apply those capabilities to connection-owned session bindings when the client never sends `set_client_info`. Previously the launch capabilities were advertised through `get_protocol_info` but dropped at binding creation, so undeclared clients received no `extension_event` frames (breaking downstream task/DAG/monitor liveness in omo-desktop-app). An explicit `set_client_info` declaration, including an empty capability list, still overrides the launch default.
 
 - RPC `abort` acknowledgements are now sent immediately after the abort signal is dispatched instead of waiting for full session quiescence, preventing desktop stop requests from hitting their 10-second bounded-ack timeout under host load.
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -1,5 +1,23 @@
 # changes
 
+## 2026-09-01 - Bound session tree construction with duplicate entry IDs
+
+### What changed
+
+- `packages/coding-agent/src/core/session-manager.ts`: `SessionManager.getTree()` now links each unique entry ID once, while preserving the existing final-entry-wins index semantics for repeated persisted rows.
+
+### Why
+
+- Duplicate entry IDs caused the same `SessionTreeNode` object to be appended repeatedly to a parent's children. Repeated duplicate edges along a branch multiplied the later traversal work and could freeze the interactive session tree before it rendered.
+
+### Why an extension could not handle it
+
+- Persisted session indexing and tree construction are core `SessionManager` responsibilities below the extension API. An extension cannot replace the synchronous tree returned to interactive, RPC, and other consumers.
+
+### Expected merge conflict zones
+
+- LOW: the edge-construction loop in `packages/coding-agent/src/core/session-manager.ts`.
+
 ## 2026-08-31 - Session activity contract for host occupancy decisions
 
 ### What changed
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -1568,9 +1568,10 @@ export class SessionManager {
 			nodeMap.set(entry.id, { entry, children: [], label, labelTimestamp });
 		}
 
-		// Build tree
-		for (const entry of entries) {
-			const node = nodeMap.get(entry.id)!;
+		// Build one edge per unique ID. Persisted sessions can contain repeated rows,
+		// and nodeMap intentionally keeps the final entry for each ID.
+		for (const node of nodeMap.values()) {
+			const { entry } = node;
 			if (entry.parentId === null || entry.parentId === entry.id) {
 				roots.push(node);
 			} else {
--- a/packages/coding-agent/test/suite/regressions/issue-1247-session-tree-duplicate-entry-ids.test.ts
+++ b/packages/coding-agent/test/suite/regressions/issue-1247-session-tree-duplicate-entry-ids.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from "vitest";
+import { type CustomEntry, SessionManager } from "../../../src/core/session-manager.ts";
+
+function customEntry(id: string, parentId: string | null, revision: number): CustomEntry {
+	return {
+		type: "custom",
+		customType: "duplicate-id-regression",
+		data: { revision },
+		id,
+		parentId,
+		timestamp: `2026-09-01T00:00:0${revision}.000Z`,
+	};
+}
+
+describe("issue #1247 duplicate session entry IDs", () => {
+	it("builds one tree node per entry ID when persisted rows repeat", () => {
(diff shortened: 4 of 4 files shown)


code-yeongyu (dismissed):
## Summary

This PR fixes a freeze in `SessionManager.getTree()` when a persisted session file contains repeated rows for the same entry ID (issue #1247).

### What the change does

`getTree()` builds a `nodeMap: Map<string, SessionTreeNode>` in its first loop (last-write-wins for duplicate IDs), then builds edges in a second loop. The bug was that the second loop iterated the raw `entries` array — which may contain duplicate IDs — rather than the deduplicated `nodeMap`. For each duplicate occurrence of an ID, `nodeMap.get(entry.id)` returned the **same node object**, which was then pushed to the parent's `children` array multiple times. Under heavy duplication the children array grew super-linearly, causing the interactive session tree to freeze.

### Dedupe logic — correctness

Iterating `nodeMap.values()` yields exactly one node per unique ID (insertion-ordered, last-written entry authoritative), so each child is pushed to its parent exactly once. Edge cases check out: duplicate roots pushed once, self-loop guard preserved, orphans still treated as roots, duplicate rows with different `parentId` resolve to the last row (consistent with the map's last-write-wins), child ordering unaffected (timestamp sort happens after edge construction), and complexity drops from O(all rows) to O(unique IDs).

### Test coverage

`test/suite/regressions/issue-1247-session-tree-duplicate-entry-ids.test.ts` reproduces the real storage path via `SessionManager.inMemory()` + `appendEntry()` with a repeated ID, and asserts exactly one root, exactly one deduped child (fails on the old code), last-write-wins (`data.revision === 2`), and correct grandchild linkage. No mocking of the tree-building path; not tautological.

### Minor note (non-blocking)

The test helper's timestamp template `` `2026-09-01T00:00:0${revision}.000Z` `` assumes single-digit revisions — safe today (0–3), but would produce invalid ISO strings at revision ≥ 10 if extended.

code-yeongyu (approved):
Re-approving after branch update — content review unchanged (see dismissed review above); this re-submit triggers the review-claims release automation.

Altairpaca (commented):
The core change matches #1247's actual blow-up mechanism: `nodeMap` already establishes one final node per ID, so building edges from `nodeMap.values()` prevents the same node object from being appended repeatedly for duplicate persisted rows. That removes the combinatorial revisit path while preserving the existing final-entry-wins semantics.

The focused regression covers the semantic invariant. As a non-blocking strengthening, a fixture with repeated duplicates across several consecutive parent/child levels would mirror the reported 108-duplicate chain more closely and pin bounded traversal under the original pathological shape. I don't see a correctness blocker in the current patch.

Issue #1239, opened 2026-08-31
OpenRouter catalog ignores reasoning metadata; mandatory-reasoning models (z-ai/glm-5.3) 400 on no-effort calls


VXNCXNX:
## Environment

- omo `5.0.0-0.beta.30` → senpi engine `2026.8.30-3` → `@code-yeongyu/senpi-ai@2026.8.30-3` (resolved via the `@earendil-works/pi-ai` alias in senpi's package.json)
- Model: `z-ai/glm-5.3` on provider `openrouter`

## Repro

1. Run any senpi-based agent with `z-ai/glm-5.3` via OpenRouter.
2. Start a new interactive session (auto title generation on).
3. Every session shows:

```
Runtime error (session_title_generation): Reasoning is mandatory for this endpoint and cannot be disabled. (HTTP 400)
```

## Root cause

Three links in the chain:

1. **Catalog entry.** `dist/providers/data/openrouter.json` ships `z-ai/glm-5.3` with `thinkingLevelMap: {"xhigh": "xhigh"}` — no `off` key at all. OpenRouter's live metadata for this model is `reasoning: {mandatory: true, supported_efforts: ["max","high","low"], default_effort: "max"}`, so the correct map is `{"off": null, "low": "low", "high": "high", "max": "max"}` (this is exactly what upstream `@earendil-works/pi-ai@0.84.4` ships for the same model).

2. **Runtime guard treats missing as sendable.** In `api/openai-completions.js`, the OpenRouter thinkingFormat branch sends an explicit disable for no-effort calls whenever `thinkingLevelMap?.off !== null`. A *missing* `off` key is `undefined`, which passes the guard, so `reasoning: {"effort": "none"}` goes on the wire for calls that pass no reasoning level — session title generation is the main caller.

3. **OpenRouter rejects.** For mandatory-reasoning models, `effort: "none"` returns HTTP 400 "Reasoning is mandatory for this endpoint and cannot be disabled". The title generator retries once, fails deterministically, and surfaces the runtime error. Main agent turns are unaffected because they carry a reasoning level.

39 other entries in the shipped catalog also have a `thinkingLevelMap` without an `off` key (e.g. `anthropic/claude-opus-5`, `openai/gpt-5.6-*`, `deepseek/deepseek-v4-*`); any of them that is actually mandatory on OpenRouter 400s the same way.

## Why the fork diverged

[...]

PR #1266, opened 2026-09-02
fix(coding-agent): request low reasoning for session title generation

ImStillBlue:
## Summary

The background session-title request left `reasoning` unset in `buildTitleOptions()`. For reasoning-capable models, pi-ai's OpenAI-completions layer then fills the gap with the provider's "disabled" mapping — on OpenRouter-format models that is `reasoning: { effort: "none" }`. Endpoints that mandate reasoning (observed with Z.ai GLM 5.x models, e.g. `z-ai/glm-5.3-flash`) reject that with:

```
Runtime error (session_title_generation): Reasoning is mandatory for this endpoint and cannot be disabled. (HTTP 400)
```

Regular agent turns are unaffected because they always carry the session's thinking level — only the background title call (and therefore every fresh session on such a model) failed. The same behavior is publicly documented against OpenRouter in openclaw#24851.

## What changed

- `packages/coding-agent/src/core/session-title-generator.ts` — `buildTitleOptions()` now sends `reasoning: "low"` explicitly, and raises `maxTokens` from 64 to 1024 so low-effort reasoning tokens don't consume the budget needed for the `<title>` output on token-based providers.
- `packages/coding-agent/test/session-title-generator.test.ts` — deterministic regression test asserting the title request carries `reasoning: "low"` and `maxTokens: 1024` via a fake `streamFn` (no network, no tokens).
- `packages/coding-agent/src/core/changes.md` — fork-ledger entry per the changes.md contract.

Non-reasoning models are unaffected: pi-ai clamps the level per model and the OpenRouter reasoning branch only fires for `model.reasoning === true`.

## Validation

- `bunx vitest run test/session-title-generator.test.ts` — 12/12 pass (including the new regression test).
- Root `bun run check` — pass (Biome, pinned-deps, ts-imports, shrinkwrap, install-lock, claude-sdk-platform-lock, `tsc --noEmit`, browser-smoke).
- Manually verified against a local install of `omo` 5.0.0-0.beta.31 / senpi 2026.8.31: patching `buildTitleOptions` the same way eliminated the recurring `session_title_generation` 400 on OpenRouter `z-ai/glm-5.3-flash` sessions.

## Residual risk

- Titles on reasoning-capable models now include a small amount of reasoning output, marginally increasing cost per one-shot title call (bounded by `maxTokens: 1024`). An alternative would be retrying with a low effort only when the endpoint rejects disabled reasoning, but the simple explicit level avoids a second round trip for a known-broken class of endpoints.

---
## Summary by cubic
Fixes session title generation on reasoning-mandatory endpoints by retrying once with low reasoning when the provider rejects the default reasoning-free title request, so new sessions on models like Z.ai GLM 5.x no longer fail with a 400 error.

- Default title requests stay reasoning-free with `maxTokens: 64`, so healthy endpoints pay no reasoning cost.
- Only a "Reasoning is mandatory" 400 triggers the retry with `reasoning: "low"` and `maxTokens: 1024`, leaving room for the `<title>` output once reasoning tokens count against the budget.
- Adds regression tests covering the default path, the fallback retry, and non-matching errors (no retry).

<sup>Written for commit 94960bae6137a8d7537847b518e6ee8e76468bfd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -1,5 +1,23 @@
 # changes
 
+## 2026-09-02 - Session title requests ask for low reasoning
+
+### What changed
+
+- `session-title-generator.ts`: `buildTitleOptions()` now sets `reasoning: "low"` and raises the title request's `maxTokens` from 64 to 1024.
+
+### Why
+
+- Leaving `reasoning` unset on the title request makes token-based providers fall back to their "disabled" mapping (e.g. `reasoning: { effort: "none" }` on OpenRouter-format models). Reasoning-mandatory endpoints such as Z.ai GLM 5.x reject that with HTTP 400 `Reasoning is mandatory for this endpoint and cannot be disabled.`, which surfaced as a repeated `session_title_generation` runtime error in every session on those models (agent turns were unaffected because they always carry the session's thinking level). Low reasoning keeps the cosmetic title cheap while producing a request every endpoint accepts; non-reasoning models ignore the hint via level clamping. The larger `maxTokens` keeps room for the `<title>` output once reasoning tokens count against the completion budget.
+
+### Why an extension could not handle it
+
+- Title generation is core background work invoked from `agent-session.ts`; extensions cannot alter the request options of the internal title call.
+
+### Expected merge conflict zones
+
+- LOW: the `buildTitleOptions()` literal in `session-title-generator.ts` and the new `generateSessionTitle` describe block in `test/session-title-generator.test.ts`.
+
 ## 2026-08-31 - Session activity contract for host occupancy decisions
 
 ### What changed
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -1,22 +1,23 @@
 # changes
 
-## 2026-09-02 - Session title requests ask for low reasoning
+## 2026-09-02 - Session titles fall back to low reasoning on reasoning-mandatory endpoints
 
 ### What changed
 
-- `session-title-generator.ts`: `buildTitleOptions()` now sets `reasoning: "low"` and raises the title request's `maxTokens` from 64 to 1024.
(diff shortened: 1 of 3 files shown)


ImStillBlue:
Valid review — pushed 94960ba addressing it.

Confirmed both claims: `DEEPSEEK_THINKING_LEVEL_MAP` / `OPENROUTER_DEEPSEEK_THINKING_LEVEL_MAP` map `low: "high"`, and the unconditional `reasoning: "low"` enabled thinking on every title call across all reasoning-capable providers (Anthropic included, which previously had thinking disabled for titles).

Redesign per your suggestion: the default title request stays reasoning-free with `maxTokens: 64`; `generateSessionTitle()` now retries once with `reasoning: "low"` + `maxTokens: 1024` only when the provider error matches "Reasoning is mandatory". Healthy endpoints keep the zero-reasoning path; broken endpoints pay one extra round trip (bounded by the same narrow retry policy). Added three tests: default attempt carries no reasoning, the reasoning-mandatory 400 triggers the fallback with `low`/`1024`, and other provider errors do not.

Validation: `vitest run test/session-title-generator.test.ts` 14/14, root `bun run check` pass.

Altairpaca (commented):
Reviewed this against the catalog-side root cause in #1239. I don't see a blocker in this patch: keeping the default title call reasoning-free preserves the cheap path, and the retry is narrowly gated to the provider's mandatory-reasoning error with explicit low effort and extra output headroom. This is complementary to #1239 rather than a substitute for it: #1239 should still restore correct OpenRouter `thinkingLevelMap` generation (`off: null` for mandatory models), while this fallback remains useful for mandatory-reasoning endpoints whose catalog metadata is absent/stale or comes from another provider. The message match is intentionally narrow, which is preferable here to broad 400 retry classification.

PR #1268, opened 2026-09-02
feat(recovery): add session fallback restore and manual compact

eddieparc:
## Summary

Add two explicit, session-only recovery surfaces:

- `/fallback restore` returns the current session to the model and thinking level active before retry fallback;
- `/compact` can explicitly override Claude SDK OAuth's automatic-compaction delegation.

This is intentionally separate from the persistence/root-cause fixes in #1267.

### `/fallback restore`

- restores the pre-fallback model for this session;
- restores the original thinking level when one was active;
- clears active fallback state;
- does not change global model or thinking defaults;
- reports a warning when the session has no active fallback.

### manual `/compact`

Claude SDK OAuth continues to own all automatic compaction reasons. Only `reason: "manual"` bypasses the lane delegation, providing an escape hatch when SDK-native compaction did not fire.

The existing post-compaction continuity path remains unchanged: the binding is tainted, the next turn forks at the last assistant boundary, and delta traffic resumes after re-attachment.

## Failing-first proof

- disabling the `restore` command leaves the session on `faux-2`; the host-wiring test expects restoration to `faux-1`;
- before the manual override, the alignment test fails with `expected 'external-owner' not to be 'external-owner'`.

## Verification

- focused recovery/compaction scope: 75 files, 505 tests passed;
- `bun run check`: passed;
- TUI smoke through a real tmux PTY: 5/5;
- Anthropic mock-loop real CLI channel: 22/22, localhost-only, real auth unchanged;
- earlier installed-engine QA compacted 64,779 tokens to 368 and continued with a 95.4% cache hit.

QA evidence is stored locally under `local-ignore/qa-evidence/20260902-recovery-commands/`.

---
## Summary by cubic
Adds `/fallback restore` so a session can return to the model used before retry fallback, restoring the prior thinking level when one was active. Also makes a manual `/compact` override `claude-sdk-oauth`'s automatic-compaction delegation; previously every `/compact` on that lane was rejected with `external-owner`. Both commands are session-only and leave global defaults unchanged.

## New Features

- `/fallback restore` clears active fallback state and warns when no fallback is active.
- `manual` is the only compaction reason that bypasses the lane delegation; all automatic reasons are still delegated to the SDK.

<sup>Written for commit 23150e543523c264ecfa35dc0f5a51330e504106. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1268?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,7 +4,11 @@
 
 ### Added
 
+- `/fallback restore` returns the current session to the model and thinking level active before retry fallback, without changing global model defaults.
+
 ### Fixed
+
+- An explicitly requested `/compact` now overrides Claude SDK OAuth's automatic-compaction delegation, providing a manual escape hatch when SDK-native compaction does not fire.
 
 ### New Features
 
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -173,7 +173,7 @@ import { expandPromptTemplateWithMetadata, type PromptTemplate } from "./prompt-
 import { createProviderTimeoutRetryPlan, runBoundedRetryContinuation } from "./provider-timeout-retry.ts";
 import type { ResourceExtensionPaths, ResourceLoader } from "./resource-loader.ts";
 import { isBillingErrorMessage } from "./retry-fallback/billing.ts";
-import { formatSelector } from "./retry-fallback/chains.ts";
+import { formatSelector, parseFallbackSelector } from "./retry-fallback/chains.ts";
 import { RetryFallbackController } from "./retry-fallback/controller.ts";
 import { SelectorCooldowns } from "./retry-fallback/cooldown.ts";
 import {
@@ -6879,6 +6879,17 @@ export class AgentSession {
 							originalSelector: active.originalSelector,
 							pinned: active.pinned,
 						};
+					},
+					restoreFallbackPrimary: async () => {
+						const active = this._retryFallback.activeState;
+						if (!active) return false;
+						const selector = parseFallbackSelector(active.originalSelector, this._modelRegistry);
+						const model = selector ? this._modelRegistry.find(selector.provider, selector.id) : undefined;
+						if (!model) return false;
+						const originalThinkingLevel = active.originalThinkingLevel;
+						await this.setSessionModel(model);
+						if (originalThinkingLevel !== undefined) this.setSessionThinkingLevel(originalThinkingLevel);
+						return true;
 					},
 				},
 				compact: (options) => {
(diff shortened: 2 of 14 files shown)


code-yeongyu:
Thanks @eddieparc — the compact-override diagnosis is still right, and it is **not** superseded.

What already landed, so we do not double-count it:

- #1273 (`09ee5781e` / `7fa2076f4`, docs `c9d51adc9`) recovers **classified summarizer failures** on a senpi-owned manual `/compact` via `isRequiredCompactionFallbackReason`. It does not let `/compact` through the Claude SDK OAuth `external-owner` stand-down (`compaction/index.ts` still cancels before that predicate).
- #1272 made SDK-native tool-less summarization work. Different layer.
- Today's main (`6dd662c4b`): #1310 SDK pin, #1311 content-blocks, #1309 login pool, #1312 `is_error` surfacing, #1318 unconfirmed-binding. None of these add `/fallback restore` or a manual lane override.

Remaining work on this PR:

1. **Product decision (restore):** `/fallback restore` is a new session command. Auto-revert (`maybeRestorePrimary`) and `/model` already exist. The maintainer decision on an explicit restore is being surfaced separately; until then treat it as undecided. Please decide whether we want an explicit restore that bypasses `never` / pin / cooldown. If no, drop the restore half (`agent-session.ts` binding, `types.ts` / `runner.ts` seam, `model-fallback` command, host-wiring test, core/extensions `changes.md` entries, Added changelog bullet) and keep this PR as the compact override only.
2. **Keep the compact override**, rebase onto `main` (`6dd662c4b`+), and resolve the mechanical doc conflicts (`CHANGELOG.md` Unreleased, `compaction/changes.md` stacked above the #1273 entry, `core/changes.md`). Do not edit #1273's tracker text.
3. **Fix the compact test:** assert a positive outcome (not cancelled / compaction produced), not `rejectionCause !== "external-owner"`.
4. **Debit the breaker** for a failed senpi-owned manual compact on the SDK lane (`session_compact` currently skips `recordFailure` while `disablesSenpiCompaction(ctx)`).
5. **If restore is approved:** do not call `setSessionModel` (throws; clears via `clearForManualModelChange`; no `retry_fallback_reverted`; thinking not atomic). Revert through the controller `switchModel(..., "fallback-revert")` path; distinct error when the original model is gone; catch failures instead of mapping them to "no active fallback."
6. Re-verify SDK-lane `/compact` → taint → fork against #1318 confirmed-id resume.

Suggested tests after rebase (must pass in one run): `bun test packages/coding-agent/test/claude-sdk-oauth-compaction-alignment.test.ts`; if restore stays, `bun test packages/coding-agent/test/suite/model-fallback-command.test.ts packages/coding-agent/test/suite/model-fallback-host-wiring.test.ts`; then `bun run check`.

PR #1276, opened 2026-09-02
fix: make provider retry and stream-start timeouts tolerant of slow thinking models

code-yeongyu:
## Summary

Slow long-thinking models (Opus/Fable-class with xhigh thinking) fail healthy turns with `Provider stream start timed out after 90000ms` before they emit their first stream event, and the workaround so far has been telling users to raise the timeout by hand. This PR makes senpi's provider retry/timeout defaults as patient as every harness we compared against, without loosening any failure classification.

Two shipped defaults change: the provider stream-start watchdog (90s -> 300s) and the agent-level retry budget (3 -> 5, in both places that define it). Explicit settings overrides, backoff shapes, permanent-failure classification, `providerRequest.maxRetries` (still 0) and `KIMI_CODE_RETRY_PROFILE` are untouched.

## Why these numbers (cross-harness comparison)

| Knob | senpi (before) | senpi (after) | opencode | openai/codex | oh-my-pi |
|---|---|---|---|---|---|
| First-event / stream-start timeout | **90s** | **300s** | 300s header timeout (`provider.ts` `OPENAI_HEADER_TIMEOUT_DEFAULT`) | 300s stream idle (`DEFAULT_STREAM_IDLE_TIMEOUT_MS`) | 300s (`DEFAULT_STREAM_FIRST_EVENT_TIMEOUT_MS`) |
| Turn / session retries | **3** | **5** | 5 (`RETRY_MAX_RETRIES`) | 5 (`DEFAULT_STREAM_MAX_RETRIES`) | 10 agent retries |
| Transport/provider retries | 0 | 0 (unchanged) | 0 at SDK layer | 4 (`DEFAULT_REQUEST_MAX_RETRIES`) | 5-10 per layer |

senpi was the least tolerant harness on both rows that matter for a slow provider, and its own stream *idle* default was already 300s, so the 90s start bound was the odd one out.

## Changes

- **packages/coding-agent** — `DEFAULT_STREAM_START_TIMEOUT_MS` 90_000 -> 300_000. Resolution order is unchanged: an explicit `retry.provider.streamStartTimeoutMs` still wins, `0` still disables the watchdog, and the effective value is still `min(default, idle timeout)`.
- **packages/coding-agent** — `getRetrySettings()` defaults `maxRetries` to 5 instead of 3 so the one `retry.maxRetries` key means the same budget on every consumer (turn stage *and* session-title / compaction-summarization / agent-session gate). Session-title retry stays clamped by its own `MAX_TITLE_RETRIES`.
- **packages/ai** — `SENPI_DEFAULT_RETRY_PROFILE.turn.maxRetries` 3 -> 5.
- Docs, tracker (`changes.md`) and `[Unreleased]` changelog entries updated to match.

## Review response (cubic)

Both findings were real and are fixed — thanks:

1. **Inert ceiling.** The first revision also raised `providerRequest.serverHint.ceiling` 60s -> 300s and claimed it removed a "61s+ instantly fails" cliff. Verified in code: `planRetryDelay` has **no production caller** (definition + barrel re-export + tests only), and `sdk.ts` deliberately sends 0 retries into the `providerRequest` stage, so that ceiling is never consulted at runtime. The claim was false, so the constant is **reverted** and the prose now says what the code does.
2. **Default divergence.** Raising only the profile made `retry.maxRetries` mean 5 on the turn stage but 3 via `getRetrySettings()` (settings-manager.ts:1280), so the documented default could only be true for one consumer. Both are 5 now, pinned by tests.

## QA & Evidence

Manual QA drove the **real CLI from source** against a local Anthropic-messages SSE endpoint that withholds everything (headers included) for 120s before streaming a valid event sequence — a healthy provider with a 2-minute thinking preamble. Driver: `local-ignore/qa-evidence/20260902-provider-retry-leniency/slow-first-event-qa.mjs` (senpi-qa sandbox, hermetic provider env, real-auth sha256 guard).

- **Observed (before):** turn failed — `Provider stream start timed out after 90000ms`, **3 provider attempts** all killed at the 90s bound, no answer, exit 1 (`base-run/summary.json`: `sawStreamStartTimeoutMs: 90000`, `requestCount: 3`).
- **Observed (after):** turn completed — **1 provider attempt**, marker text returned, exit 0 (`wt-run/summary.json`: `sawStreamStartTimeoutMs: null`, `sawMarker: true`).
- **Artifacts:** `local-ignore/qa-evidence/20260902-provider-retry-leniency/{base-run,wt-run}/{summary.json,stdout.txt,stderr.txt,requests.json}` (sandboxed, credential-free; `realAuthUnchanged: true` in both).

Automated (all runs on a remote build host, bun only):

- **Failing-first proofs, captured against unmodified source before each change:** `expected 3 to be 5` (profile turn budget), `expected 90000 to be 300000` (stream-start default), `expected 3 to be 5` (`getRetrySettings()` default). Each green afterwards; the `retry.maxRetries` pin is paired with an explicit-override case (value 2) so it cannot pass tautologically.
- `bun run check` (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, claude-sdk platform lock, `tsc --noEmit`, browser-smoke): exit 0.
- Focused scope green: retry-profile ×5, provider-retry, retry, settings-manager, settings-manager-retry-profiles, retry-fallback-engine, provider-timeout-classification.

## Risks & Residuals

- *A genuinely dead stream now hangs longer before the watchdog fires (90s -> 300s).* Accepted and intentional: it is the bound senpi already applies to stream idleness and what all three comparison harnesses use; `retry.provider.streamStartTimeoutMs: 90000` restores the old aggressiveness.
- *Two extra retry attempts before model fallback.* Bounded by the existing 2s-base backoff capped at 8s per attempt; fallback classification is unchanged and hard failures (quota, auth, refusals) still switch immediately.
- *`getRetrySettings()` consumers now retry 5 times instead of 3.* Same direction as the PR, and session-title retry keeps its own lower cap.
- No changes to retryable/permanent classification, backoff math, the Kimi profile, or any provider-stage behavior.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Changed
 
-- The `senpi-default` retry profile is more patient with slow long-thinking providers (Opus/Fable-class with xhigh thinking): the turn-stage retry budget rises from 3 to 5 attempts, matching opencode's session retry budget and codex's `stream_max_retries` default, and the provider-stage server-hint ceiling rises from 60s to 300s, so a server-requested `Retry-After` longer than 60s is waited out instead of failing the turn instantly (opencode honors such hints uncapped, codex honors server-advised delays, oh-my-pi caps them at 300s).
+- The `senpi-default` retry profile is more patient with slow long-thinking providers (Opus/Fable-class with xhigh thinking): the turn-stage retry budget rises from 3 to 5 attempts, matching opencode's session retry budget and codex's `stream_max_retries` default. Backoff shapes, failure classification, `providerRequest.maxRetries` (still 0) and `KIMI_CODE_RETRY_PROFILE` are unchanged.
 
 ### Fixed
 
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Changed
 
+- The `senpi-default` retry profile is more patient with slow long-thinking providers (Opus/Fable-class with xhigh thinking): the turn-stage retry budget rises from 3 to 5 attempts, matching opencode's session retry budget and codex's `stream_max_retries` default, and the provider-stage server-hint ceiling rises from 60s to 300s, so a server-requested `Retry-After` longer than 60s is waited out instead of failing the turn instantly (opencode honors such hints uncapped, codex honors server-advised delays, oh-my-pi caps them at 300s).
+
 ### Fixed
 
 ### Removed
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -1,3 +1,21 @@
+## senpi-default retry profile is more patient with slow providers (2026-09-02)
+
+### What changed
+
+- `utils/retry-profile/profiles.ts`: `SENPI_DEFAULT_RETRY_PROFILE.turn.maxRetries` goes from 3 to 5, and the `providerRequest` server-hint ceiling from 60s to 300s. Backoff shapes, the `error-with-marker` overflow behavior, `providerRequest.maxRetries` (still 0, so no hidden second budget), and `KIMI_CODE_RETRY_PROFILE` are untouched.
+
+### Why
+
+- Opus/Fable-class models with xhigh thinking make transient provider failures and long server-requested waits more likely per turn, and the previous budgets were the least tolerant of the harnesses we compared: opencode retries a session 5 times and honors `Retry-After` uncapped, codex defaults to `stream_max_retries` 5 and honors server-advised delays, and oh-my-pi allows up to 10 agent retries with a 300s hint cap. A 60s ceiling turned any provider asking for 61s into an instantly failed turn even though the turn stage's own hinted wait cap is already 300s.
+
+### Why an extension could not handle it
+
+- Shipped profile constants are read by the provider-request and turn retry planners before any extension seam exists; an extension can only override them per provider through settings, not change what every session inherits.
+
+### Expected merge conflict zones
+
+- LOW: the two constant lines inside `SENPI_DEFAULT_RETRY_PROFILE` in `utils/retry-profile/profiles.ts`.
+
 ## Actionable provider stream-start timeout guidance (2026-09-02)
 
 ### What changed
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -2,11 +2,11 @@
 
 ### What changed
 
-- `utils/retry-profile/profiles.ts`: `SENPI_DEFAULT_RETRY_PROFILE.turn.maxRetries` goes from 3 to 5, and the `providerRequest` server-hint ceiling from 60s to 300s. Backoff shapes, the `error-with-marker` overflow behavior, `providerRequest.maxRetries` (still 0, so no hidden second budget), and `KIMI_CODE_RETRY_PROFILE` are untouched.
+- `utils/retry-profile/profiles.ts`: `SENPI_DEFAULT_RETRY_PROFILE.turn.maxRetries` goes from 3 to 5. Backoff shapes, the server-hint policies, `providerRequest.maxRetries` (still 0, so no hidden second budget), and `KIMI_CODE_RETRY_PROFILE` are untouched.
 
 ### Why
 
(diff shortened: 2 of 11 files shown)


code-yeongyu:
Both findings were correct — fixed in bc458b754 and 4a05a50a0.

**1. Inert ceiling (profiles.ts).** Confirmed by reading the consumption path rather than assuming: `planRetryDelay` has no production caller (definition + barrel re-export + tests only), and `sdk.ts` sends 0 retries into the `providerRequest` stage by design, so `providerRequest.serverHint.ceiling` is never consulted at runtime. Raising it was a no-op dressed up as a fix, and the changelog/tracker overstated it. The constant is reverted to 60s and the prose now describes only what the code does.

**2. `retry.maxRetries` default divergence (settings.md:215).** Also correct: `getRetrySettings()` returned `?? 3` while the profile turn stage moved to 5, so one settings key meant two budgets and the documented default could not be true for both. `getRetrySettings()` now defaults to 5, matching `SENPI_DEFAULT_RETRY_PROFILE.turn.maxRetries`; session-title retry remains clamped by `MAX_TITLE_RETRIES`. Pinned by two tests — the unset default (captured RED as `expected 3 to be 5` against unmodified source) and an explicit `2` that must still win, so the pin cannot pass tautologically.

code-yeongyu:
### CI status: the remaining red is pre-existing, not from this PR

`RPC named pipes (Windows)` has failed on this branch while being green on every recent `main` commit where it ran (`14a1b1c0`, `23e6cad9`, `c83f9e9e`, `84e91f0c`). Three reds with a clean baseline is not something to wave off as flake, so I checked it directly instead of reasoning about it.

**The failing case rotates between runs, inside the same step:**

| Run | Step | Failing case | Error |
|---|---|---|---|
| 33598824099 | Run RPC host **ensure** tests | `serializes concurrent starts for one socket across agent directories` | `database is locked` |
| 33602094330 | Run RPC host **lifecycle** tests (ensure step *passed*) | `does not exit while a turn is active...` | `Timed out waiting for RPC record` |
| 33602848864 | Run RPC host **lifecycle** tests | `does not exit while a client is attached...` | (same step) |

**Direct reproduction of the lifecycle cases on two trees** (remote macOS host, bun):

| Tree | Constants present | Result |
|---|---|---|
| unmodified `origin/main` | `DEFAULT_STREAM_START_TIMEOUT_MS = 90_000`, `?? 3` | 1 failed |
| this branch | `300_000`, `?? 5` | 1 failed |

Identical failure **on unmodified main**, so these three constants are not the cause.

**Mechanism exclusion** for the lifecycle case: the wait that times out is `agent_start` (emitted on prompt acceptance, 15s budget); the case releases the held model and finishes inside a 60s case timeout with a 20s exit wait. No path reaches the 300s stream-start watchdog. `ownership-safe-lock.ts` (the ensure-step case) uses its own module constant `{ retries: 100, maxTimeout: 100 }` and imports only `node:fs/promises` — it never reads any retry/timeout setting this PR touches.

*Honest limit:* my local repro of the lifecycle cases fails with `RPC socket host exited with code 1`, which is a different stage than CI's `Timed out waiting for RPC record`. That proves this PR's delta is not the differentiator; it does not root-cause the CI failure itself. I am deliberately **not** fixing that pre-existing instability inside this PR — it deserves its own change.

### Full-suite note

`packages/ai` full suite on this branch: `1 failed | 256 passed | 25 skipped`. The same suite on unmodified base: `1 failed | 256 passed | 25 skipped` (2485 passed both times) — same failure rate, **different victim file** (`codex-apply-patch-wire-schema` on base vs `cursor-agent`/`openai-completions-stream-lifecycle` here). Both of those files pass 37/37 in isolation on both trees. That is load-dependent full-suite flakiness on a saturated host, not a regression from this change. `bun run check` is exit 0.

code-yeongyu:
Fixed — the default now derives from the profile instead of duplicating it.

`SENPI_DEFAULT_RETRY_PROFILE` was already imported at line 3, so `getRetrySettings()` reads `SENPI_DEFAULT_RETRY_PROFILE.turn.maxRetries` directly. The previous form was worse than a plain literal: the comment asserted the two must track each other while nothing enforced it, which is exactly how the divergence this PR just fixed would come back on the next budget change.

Verified by the existing pins: the unset default still resolves to 5 and an explicit `retry.maxRetries: 2` still wins.

PR #1277, opened 2026-09-02
fix(rpc): scope multi-session socket events to attached connections

### LINKED ISSUE #1271: rpc: multi-session socket host broadcasts every session's agent events to every client (cross-session contamination)
## Summary

The multi-session socket host broadcasts every session's agent events to **every** connected client, and the only thing keeping session B's output out of session A's TUI is a client-side filter that is open whenever the client's lease is undefined. Under reconnect this lets one session's in-flight assistant output render inside another session's client — the "cross-session contamination" users reported on the OmO Discord (2026-09-02, beta.31 x20 plan: "B의 작업물이 A 세션에 출력").

Shared host is opt-in since v2026.8.30-2 (`core/shared-host-policy.ts`), so default users are not exposed today; every opted-in user and every multi-session host consumer is.

## Root cause (packages/coding-agent/src/modes/rpc, origin/main 69d9437cd)

- `multi-session-host.ts` doc: *"Socket event visibility is an all-sessions broadcast: every connected client receives every session lifecycle/agent event"*.
- `session-event-writer.ts` `enqueue()`: every non-targeted record goes to `[...this.connections.keys()]` — no check against `connectionSessions`, even though that map already exists and is used for rendered components.
- `session-event-writer.ts` `registerConnection()`: replays **every** session's unrendered in-flight snapshot to a newly accepted socket, before that socket has opened anything.
- `rpc-client.ts` `handleLine()`: `if (data.sessionId !== undefined && this.sessionId !== undefined && data.sessionId !== this.sessionId) return;` — passes everything while `this.sessionId` is undefined. `stop()` clears the lease but keeps `eventListeners`; `interactive-host-runtime.ts` reconnect runs `client.start()` (host replays other sessions' snapshots) **before** `client.openSession()` sets the lease, so listeners render another session's message during the reconnect window.

## Fix

1. Host: deliver session-tagged agent events only to connections attached to that session (`connectionSessions`); keep content-free lifecycle records (`session_closed`) broadcast. No cross-session replay on `registerConnection`; `attachConnectionToSession` replays that session's snapshot (unrendered always, rendered per capability). Observing a foreign session stays possible through the existing attach-on-open (`open_session` with the same `sessionPath`).
2. Client: `handleLine()` drops any session-tagged event whose `sessionId` differs from the lease, including while the lease is undefined.
3. Docs: `docs/rpc.md` "Socket event visibility" + `modes/rpc/changes.md`.

Related: code-yeongyu/oh-my-openagent#7587 (close_session teardown), #1215.

code-yeongyu:
## Summary
- Scope multi-session socket agent events to connections attached to the session.
- Stop newly registered sockets from replaying every session's in-flight snapshot.
- Drop foreign session events while `RpcClient` is lease-less, while buffering the opening session's own startup events until `open_session` establishes its lease.

## Root cause
- `packages/coding-agent/src/modes/rpc/session-event-writer.ts` targeted non-rendered records at every connection and replayed every session snapshot on registration.
- `packages/coding-agent/src/modes/rpc/rpc-client.ts` previously allowed all tagged events through while its lease was undefined. Tightening that predicate exposed a startup ordering requirement: extension `session_start` events can arrive before the `open_session` response.

## Contract change
Agent events are attachment-scoped. To observe a foreign session, open it by its existing `sessionPath`; the router's attach-on-open path registers that connection for the session. `session_closed` remains broadcast because it carries no content and observers rely on roster visibility.

During an in-flight `open_session`, `RpcClient` buffers tagged events. Once the response establishes the lease, it synchronously delivers buffered events for that session in arrival order and discards foreign events. Outside an in-flight open, lease-less clients drop all tagged events.

## Tests
- Writer coverage proves session events reach only attached connections and unattached sockets receive none.
- Real socket-host coverage proves another attached session and a newly connected unattached socket receive no foreign turn records.
- RpcClient coverage proves lease-less foreign records are dropped and own pre-lease startup records are delivered only after lease establishment.
- Interactive runtime coverage proves startup factory UI remains visible through the normal runtime API.

## Evidence
- Biome check: passed.
- `tsc --noEmit`: passed.
- Selected regression suite: 8 files passed, 100 tests passed.
- `[remote] done rc=0 tree=5975d43ab`

Closes #1271

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -33,6 +33,7 @@
 ### Fixed
 
 - Two concurrent shared-host starts for one socket no longer fail with a raw `database is locked`: the ensure-lock wait now covers the whole host startup critical section (probe, incompatible-host stop, spawned-host readiness) instead of ten seconds.
+- Multi-session socket hosts deliver agent events only to connections attached to the session; targeted responses and dialog extension UI requests remain requester-only, while other extension UI state reaches attached connections. `RpcClient` drops foreign session events while lease-less and buffers own pre-lease startup events during `open_session`.
 
 ### New Features
 
--- a/packages/coding-agent/docs/rpc.md
+++ b/packages/coding-agent/docs/rpc.md
@@ -62,10 +62,12 @@ On Windows, listeners and clients deterministically map the logical socket path 
 `\\.\pipe\senpi-rpc-<sha256[:32]>`. Callers keep using the same `unix://` CLI value; the logical path remains the
 ownership and settings identity, and callers never construct the pipe name themselves.
 
-Socket event visibility is an all-sessions broadcast: every connected client receives lifecycle and agent events from
-every open session, each tagged with its routing `sessionId`. Correlated responses and extension UI requests are sent
-only to the connection that issued the command. This lets a non-owner observe a foreign turn without requiring a
-separate subscription protocol.
+Socket event visibility is attachment-scoped: each connection receives agent events only from sessions attached to that
+connection, with every record tagged by its routing `sessionId`. Content-free `session_closed` lifecycle records remain
+broadcast to all connections so observers can track the session roster. Correlated responses and dialog extension UI
+requests (select, confirm, input, and editor) are requester-only; other extension UI state records go to the session's
+attached connections. To observe a foreign session, open it by its existing
+`sessionPath`; the host attaches that connection during `open_session`.
 
 ### Client information and rendered components
 
--- a/packages/coding-agent/src/changes.md
+++ b/packages/coding-agent/src/changes.md
@@ -1,5 +1,23 @@
 # changes
 
+## 2026-09-02 - Export the RPC open-in-flight client error
+
+### What changed
+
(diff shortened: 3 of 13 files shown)


code-yeongyu:
Merging with the `RPC named pipes (Windows)` job red, deliberately:

- All 19 other checks are green at 44c7414bb (rebased on main 42b0c28c7, which includes the two Windows flake root-cause fixes #1284 and #1288).
- The remaining Windows failure is `test/rpc-host-ensure.test.ts > serializes concurrent starts for one socket across agent directories` on three consecutive attempts of run 33615244048 - a file this PR does not touch, a Windows-only cross-agent-dir handshake path, tracked with analysis in #1290. The same job passed on #1280/#1284/#1288 today, and the lifecycle/ensure suites pass on macOS and Linux with this tree.
- Strict ultrabrain review: REQUEST_CHANGES (5 blockers) -> all fixed -> APPROVE. Remote verification: commit 1 alone boots and passes rpc-multi-session-events + rpc-socket-host (36/36); final tree 23 rpc/interactive-host test files, 220 tests, tsc clean; real-surface raw-socket QA: a never-opened socket receives 0 foreign records (13 at base).

PR #1280, opened 2026-09-02
fix(rpc): bound and join close_session teardown

### LINKED ISSUE #7587: rpc/session-registry: close_session teardown is unbounded and non-idempotent — a wedged abort permanently retains the session path reservation
## Summary

`close_session` teardown can hang forever, permanently retaining a session's path reservation. The fix belongs here (host), not in clients: a client cannot bound a host-side unbounded await.

## Root cause (session-registry.js)

- `closeMarked()` awaits `abort() -> waitForIdle() -> dispose()` with **no bound**. If the session's abort never resolves, teardown never completes and the path reservation is never released.
- `beginClose()` throws `unknown_session` when the entry is already `closing`, so a client's second/reclaim `close_session` is rejected rather than joining the in-flight teardown.
- Consequence for clients: after their own bounded close times out, they cannot reclaim the path. `open_session` on that path keeps failing in-use, and the only escape is restarting the whole shared host — which kills every unrelated session on it.

## Requested change

1. **Bounded teardown.** Race the graceful `abort -> waitForIdle -> dispose` chain against a host-side grace deadline (~10s). On expiry, escalate to hard disposal that cannot block on the stuck promise, release the reservation, and transition to `closed`. Guarantee: `close_session` always terminates and the reservation is always eventually released. (Forced disposal is strictly less dangerous than the status quo escape hatch of killing the entire host process — same durability requirements, scoped to one session.)
2. **Idempotent close.** A `close_session` for an entry already in `closing` should join the in-flight teardown (await the same lifecycle mutex) and return its outcome, instead of throwing `unknown_session`.
3. *(Optional, cheap)* a per-row monotonic `lastTransitionAt` (or teardown phase) on `list_sessions` rows, so clients can distinguish "draining slowly" from "wedged" honestly. Purely additive; old clients ignore it.

No new RPC verb is needed — a host that can force-close on request can do it on its own timer, so this needs no capability negotiation. If caller-controlled grace is ever wanted, an optional `graceMs` field on `close_session` degrades safely on old hosts.

## Why now

Surfaced while narrowing a desktop-side blast radius: code-yeongyu/omo-desktop-app#278 / #281 stopped one wedged turn from killing every session on the shared host. That change is a strict improvement, but it exposes this pre-existing host gap — the affected thread is now stranded visibly instead of the whole host being nuked. Clients can only mitigate (poll `list_sessions` until the row disappears, bounded, then surface a legible error); the durable fix is the bounded teardown above.

code-yeongyu:
## Summary

- Bound multi-session `close_session` teardown by a 10-second host grace deadline with forced reservation release.
- Make concurrent close requests join the in-flight teardown and return successful targeted responses.
- Preserve the first closer's `session_closed` and final-response ordering contract.

## Root cause

`session-registry.ts` awaited `abort -> waitForIdle -> dispose -> scope.close` without a deadline and rejected any `beginClose` call after the entry entered `closing`. `session-command-router.ts` consequently could dispose the shared binding twice and could not deliver a response to a joined closer after `SessionEventWriter.closeSession` sealed the handle.

## Behavior

- Graceful close defaults to 10,000 ms and is configurable by `SENPI_RPC_CLOSE_GRACE_MS`.
- On expiry, the entry is marked closed, removed, and its path reservation is released immediately; the remaining teardown is observed detached and failures are reported through `senpi rpc ...` stderr diagnostics.
- A close for an already-closing entry joins the same completion. Binding disposal happens once. The first closer receives the terminal lifecycle event and last tagged response; joined callers receive their own successful targeted response.
- Unknown and already-closed handles still return `unknown_session`.

## Tests

- RED before restoring production changes: C4 timed out at 30 seconds on the never-resolving abort; registry C5 rejected the second close with `unknown_session`; router C5 called binding disposal twice.
- GREEN focused bridge: 2 files, 23 tests passed.
- GREEN final regression bridge: 12 files, 157 tests passed.
- GREEN `--check`: Biome checked 3466 files with no fixes; TypeScript `tsc --noEmit` passed.

## Evidence

Final bridge tail:

```
Test Files  12 passed (12)
Tests       157 passed (157)
[remote] done rc=0 tree=3875f3f1d
```

Closes code-yeongyu/oh-my-openagent#7587

---
## Summary by cubic
Previously, a stuck multi-session RPC `close_session` could retain the runtime and path reservation indefinitely, while concurrent closes returned `unknown_session`. `close_session` now uses bounded teardown and joins concurrent requests.

**Bug Fixes**
- Graceful teardown waits up to 10 seconds by default, configurable via `SENPI_RPC_CLOSE_GRACE_MS`; expiry releases the session and path reservation while detached cleanup continues and reports failures to stderr.
- The first closer retains the terminal `session_closed` and final-response ordering; joined callers receive targeted successful responses.
- Explicit close, idle eviction, and router disposal share teardown ownership to prevent double binding disposal.
- Updates the RPC documentation and adds regression coverage for stuck, responsive, joined, and racing closes.

Closes code-yeongyu/oh-my-openagent#7587

<sup>Written for commit 17b96d3911f2b8936914fde5ed48a58dece66a73. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1280?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Merging with admin, deliberately:

- Three strict ultrabrain review rounds: round 1 REQUEST_CHANGES (4 blockers: forced-release scope.close, joiner terminal ordering, eviction/shutdown double-dispose, registry API leak) -> fixed; round 2 REQUEST_CHANGES (2 error-path blockers: rejecting binding.dispose double-response, unresolved join token on pre-dispose throw) -> fixed with RED-first tests; round 3 APPROVE with no blockers.
- Remote verification at the error-path-fixed tree: 19 rpc/interactive-host/claude-sdk test files, 267 tests, biome + tsc clean; commit-1-alone (refactor) passes registry + multi-session suites (20/20); mutation RED proved the C4/C5 tests fail without the fix (stuck-abort close 30s timeout, unknown_session on concurrent close, double binding dispose).
- CI at 941112c72 (identical code, pre-rebase): 19/21 green; the only failure was `RPC named pipes (Windows)` on `rpc-host-lifecycle.test.ts > does not exit while a turn is active...` twice - issue #1290 variant 2 (single-pass hostAlive probe), a file this PR does not touch, with the test-only fix up as #1291.
- Final head 17b96d391 only rebases onto the v2026.9.2-4 release cycle and moves the CHANGELOG entry back under [Unreleased].

Closes code-yeongyu/oh-my-openagent#7587.

Issue #1290, opened 2026-09-02
ci(windows): remaining RPC named-pipes job flake variants (host-ensure concurrent-start SIGTERM, lifecycle hostAlive probe)


code-yeongyu:
## Summary

The `RPC named pipes (Windows)` CI job (runs `test/rpc-host-ensure.test.ts` and `test/rpc-host-lifecycle.test.ts` on windows-latest) fails intermittently with several independent variants. Two were fixed at the root today; two remain and are tracked here so they are not mistaken for regressions in unrelated PRs.

## Fixed at the root (2026-09-02)

- #1288 - `serializes concurrent starts for one socket across agent directories` -> `Error: database is locked`: ensure-lock wait (10 s) shorter than the holder's critical section.
- #1284 - `starts a fresh host transparently on the next ensure after an idle exit` -> `Stop-Process` on an already idle-exited pid / `ENOENT host.pid`: teardown and pidfile read raced the 600 ms idle exit.

## Still open

1. `rpc-host-ensure.test.ts > serializes concurrent starts for one socket across agent directories` -> `Error: RPC socket host exited with code null (SIGTERM) before answering get_protocol_info`, followed by `ENOTEMPTY ... rmdir senpi-host-ensure-cross-agent-race-*` in cleanup. Seen: run 33602751743 attempt 1 (PR #1277), run 33615244048 attempt 2 job 100202119649 (PR #1277 at 44c7414bb, which does not touch host-ensure). The second caller ends up spawning (its existing-host probe did not reuse the first caller's fixture host) and its spawned host is SIGTERM'd before readiness - candidate causes: the 10 s existing-host probe timing out on the runner and `stopManagedHost` terminating the live fixture, or pidfile identity confusion between the two agent dirs.
2. `rpc-host-lifecycle.test.ts > does not exit while a turn is active even with no connections; exits after the turn settles` -> `expect(await hostAlive(entry.pidFile)).toBe(true)` false at line 201. Seen: run 33602751743 attempt 2 job 100164143208 and run 33615244048 attempt 1 job 100199445430 (both PR #1277; the same file passes on macOS/Linux with the same tree). `hostAlive` goes through the 1 s PowerShell CIM identity probe (`readProcessStartTime`) with no retry, so a slow probe reads as "host gone"; production callers got retry in 1640d9b67, this test helper did not.

## Suggested next steps

[...]

PR #1291, opened 2026-09-02
test(rpc): retry the lifecycle suite's host liveness probe

code-yeongyu:
## Summary

Fixes issue #1290 variant 2: the `RPC named pipes (Windows)` job intermittently fails `rpc-host-lifecycle.test.ts > does not exit while a turn is active even with no connections; exits after the turn settles` with `expect(await hostAlive(...)).toBe(true)` -> false while the supervisor is demonstrably alive (it exits later in the same test's teardown).

## Root cause

The test helper `hostAlive()` makes exactly one `processMatchesPidFile(pidFile, readProcessStartTime)` pass. On windows-latest that probe is a 1 s-bounded PowerShell `Get-CimInstance` call; a slow runner makes the probe throw (timeout) or return garbage once, and the single pass reads the live host as dead. Production callers got bounded retry for exactly this in 1640d9b67 ("fix(rpc): retry transient process identity probes"); the test helper did not.

## Change

Test-only: `hostAlive()` retries thrown identity-probe errors for up to 10 s (500 ms interval). A successful probe still resolves immediately, so a genuinely dead host is reported false on the first answer - the assertion semantics are unchanged, only observer noise is absorbed.

## Evidence

- Failures: run 33615244048 attempt 1 job 100199445430 (PR #1277), run <B-run> attempts 1-2 job 100210045335 / 100212568272 (PR #1280) - same assertion, same line, tree passes on macOS/Linux.
- This PR's own `RPC named pipes (Windows)` job is the verification surface.

Refs #1290.

---
## Summary by cubic
Fixes the flaky `RPC named pipes (Windows)` lifecycle test by making `hostAlive()` tolerate transient identity-probe failures: "alive" is decisive on the first read, and dead requires two consecutive agreeing "not alive" reads (or the 10s deadline), so a genuinely dead host settles one probe later and assertion semantics are unchanged. Assertions now attach the supervisor's stderr on failure, teardown keeps a single decisive probe so stopping a host doesn't spend the retry window, and the teardown timeout rises from 30s to 60s to fit the worst case.

<sup>Written for commit 2c6e385d6d247f0d3240023ab9c0d39217041bfe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1291?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/test/rpc-host-lifecycle.test.ts
+++ b/packages/coding-agent/test/rpc-host-lifecycle.test.ts
@@ -61,7 +61,10 @@ afterEach(async () => {
 	for (const model of models.splice(0)) await model.close();
 	for (const entry of managed.splice(0)) await stopHostProcess(entry.pidFile, entry.pidFilePath);
 	for (const root of roots.splice(0)) rmSync(root, { recursive: true, force: true });
-}, 30_000);
+	// Budget: teardown liveness probe (~1s) + SIGTERM exit wait (30s) + SIGKILL
+	// escalation (2s) per host, with headroom; the old 30s cap already sat below
+	// the pre-existing 33s worst case.
+}, 60_000);
 
 type RecordValue = Record<string, unknown>;
 
@@ -175,7 +178,7 @@ describe("ensureHost-spawned host lifecycle", () => {
 		const entry = currentManaged();
 		const peer = await JsonlPeer.connect(qa.socket);
 		await delay(2_000);
-		expect(await hostAlive(entry.pidFile)).toBe(true);
+		await expectHostAlive(qa, entry.pidFile);
 		peer.destroy();
 		await waitForHostExit(entry);
 	}, 45_000);
@@ -198,7 +201,7 @@ describe("ensureHost-spawned host lifecycle", () => {
 		await agentStart;
 		peer.destroy();
 		await delay(2_500);
-		expect(await hostAlive(entry.pidFile)).toBe(true);
+		await expectHostAlive(qa, entry.pidFile);
 		model.release();
 		await waitForHostExit(entry, 20_000);
 	}, 60_000);
@@ -223,7 +226,7 @@ describe("ensureHost-spawned host lifecycle", () => {
 		});
 		const entry = currentManaged();
 		await delay(2_500);
-		expect(await hostAlive(entry.pidFile)).toBe(true);
+		await expectHostAlive(qa, entry.pidFile);
 		terminateSupervisor(entry.pidFile.pid, "SIGTERM");
 		await waitForHostExit(entry, WINDOWS_SUPERVISOR_EXIT_TIMEOUT_MS);
 	}, 45_000);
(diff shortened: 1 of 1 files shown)


code-yeongyu:
Observation from this PR's own `RPC named pipes (Windows)` run (33619854835): the same test failed again with `hostAlive -> false`, WITH the retry in place — so in this variant `readProcessStartTime` did not throw; the CIM probe returned `__SENPI_ABSENT__` (or a mismatched identity) for a live pid. The retry-on-throw in this PR is therefore necessary but not sufficient; #1290 variant 2 needs the probe itself to treat a transient ABSENT for a pid that was just observed alive as suspect (e.g. confirm absent twice before reporting dead), or the test to capture the probe result kind on failure. Leaving this PR as-is for the throw path and updating #1290.

code-yeongyu:
Addressed the review blocker (teardown budget): `stopHostProcess()` no longer uses the assertion-grade retry — it makes one identity probe (unreadable -> treated alive, so the stop path still runs) and the `afterEach` budget rises 30s -> 60s to cover the pre-existing 33s worst case (SIGTERM 30s + SIGKILL 2s + probe). Also, since this PR's first CI run failed with a returned-false probe on a live pid even with retry-on-throw, `hostAlive` now requires two consecutive agreeing "not alive" reads, and the three liveness assertions go through `expectHostAlive()` which attaches the supervisor's stderr — the next Windows failure will show whether the host truly exited or the CIM observer lied (#1290 diagnosis). macOS/Linux: lifecycle+ensure 42/42 green, tsc + biome clean at 2c6e385d6.

code-yeongyu:
Merging with admin, deliberately:

- Ultrabrain strict review: round 1 REQUEST_CHANGES (one blocker: teardown could exceed the 30s afterEach budget through the assertion-grade retry) -> fixed (teardown single probe + 60s budget) -> round 2 VERDICT: APPROVE, no blockers.
- Non-Windows CI: 16/16 green at 2c6e385d6; macOS/Linux lifecycle+ensure 42/42 + tsc + biome clean on the remote bridge.
- The only red check is `RPC named pipes (Windows)`, failing on the REAL product bug this PR's own `expectHostAlive` diagnostic exposed (supervisor logged `idle shutdown` mid-turn) — reclassified and tracked in #1290 with the root-cause fix in progress; merging this test-only hardening improves diagnosis of exactly that bug.

code-yeongyu:
Heads-up from an unrelated PR (#1293) that this job is currently blocking: **this PR's own CI is still failing the case it fixes**, and the failure is now inside the new retry helper rather than at the old bare assertion.

Run [33622887744](https://github.com/code-yeongyu/senpi/actions/runs/33622887744/job/100223762670), step `Run RPC host lifecycle tests`:

```
FAIL  test/rpc-host-lifecycle.test.ts > ensureHost-spawned host lifecycle
      > does not exit while a turn is active even with no connections; exits after the turn settles
  674|  const alive = await hostAlive(pidFile);
```

So retrying `hostAlive()` does not appear sufficient on its own for this case. Two data points from my own investigation that may narrow it down:

1. **The case fails on unmodified `origin/main` off-CI too.** I ran `does not exit while a turn is active...` and `does not exit while a client is attached...` side by side on a pristine `origin/main` worktree and on a feature branch; both failed identically. Whatever the cause, it is not branch-specific.
2. **Off-CI the symptom is different** — `RPC socket host exited with code 1 before answering get_protocol_info` instead of a probe/record timeout. The lifecycle test spawns the supervisor via `process.execPath` on `src/modes/rpc/host-lifecycle.ts` with no TS-runtime flag and no built entry, so a checkout without the right runtime/build state exits immediately. That looks like the SIGTERM/startup family in #1290 "Still open" item 1, i.e. plausibly a *second* defect stacked on the probe flakiness — worth separating, since a probe retry cannot mask a host that already exited.

If the supervisor is genuinely dead by the time the probe runs (rather than the probe being unreliable), a retry loop will keep reporting `false` exactly as observed here.

I have deliberately not opened a competing PR — #1290/#1291 own this. Corroborating run-by-run evidence for the rotating variants is in [my comment on #1290](https://github.com/code-yeongyu/senpi/issues/1290).

PR #1294, opened 2026-09-02
fix(rpc): root-fix the flaky Windows named-pipes CI job

code-yeongyu:
## What this fixes

Four defects behind the flaky `RPC named pipes (Windows)` job, each found from real CI logs. I state below exactly which ones CI proved and which are defensive, because two of them were **not** the cause of any failure I observed.

### 1. `readProcessStartTime` threw past the startup guard — production, `host-lifecycle.ts`

The win32 baseline read sat **outside** the startup `try`/`catch`, and that function throws when its 1s CIM probe fails (`execFile`'s timeout SIGTERMs the PowerShell child). With no `unhandledRejection` handler the supervisor died right there.

Evidence chain from CI: `senpi rpc host ready` appeared **0 times** against two `listening on` lines, and `startup failed` **0 times** — it died past the startup guard but before the ready line. `CHILD_WATCH_FD` then did exactly what it is designed to do: the internal host saw EOF and shut itself down, surfacing as `connect ENOENT` on the public pipe and as `host <pid> reported dead`.

**CI-proven:** the next run showed `host ready` **2 times** (was 0), with `ENOENT` and CIM errors both dropping to **0**.

### 2. Opportunistic GC ran inside the endpoint lock — production, `host-ensure.ts`

`ensureHost()` ran `reapOrphanedInternalHostDirs()` **inside** the exclusive lock. That reaper's cost is unbounded in temp-directory size (measured locally: 132,035 entries, ~400ms) and on win32 spawns a `powershell.exe` probe per candidate, so a concurrent waiter could exhaust even the 42s budget and surface a raw `database is locked`.

The prior `changes.md` entry had already **raised the budget to 42s** chasing this. A larger budget cannot fix a critical section whose cost is unbounded; the reaper's own guards (60s age, dead owner pid) make it safe unlocked, so it now runs before the lock is taken.

**CI-proven:** mutation-killed by the new `rpc-host-ensure-lock-scope.test.ts` (`expected 1120 to be less than 1044` when the reap moves back inside), and the ensure suite has passed 13/13 on Windows CI in every run since.

### 3. A timed-out probe was read as ALIVE — test helper, two sites

`waitForHostExit()` and `terminateSupervisor()` trusted the CIM probe's `timedOut` flag, so a timeout meant "still alive": teardown rethrew and the exit wait spun to its deadline. Liveness is now decided by `processIsLive(pid)` (`kill(pid, 0)`), this repo's existing idiom.

**Defensive — not CI-proven.** `identity watchdog` appeared 0 times in the run I suspected, so this was never the direct cause of an observed failure.

### 4. The watchdog believed an absent identity too readily — production, `host-lifecycle.ts`

The win32 identity watchdog polls every 500ms with a 1s CIM timeout and shut the host down after two consecutive `undefined` readings — two timeouts ~1s apart could kill a healthy host. An absent identity is now confirmed with `kill(pid, 0)` first, and a timed-out probe neither increments nor resets the counter. `identityChanged` detection is unchanged.

**Defensive — not CI-proven**, same reason as #3.

## Relationship to the maintainer's parallel fixes

This job was failing repo-wide, and two maintainer fixes landed while this PR was open. This branch is rebased onto both.

- **#1291 (`hostAlive` retry).** I had written my own single `processIsLive` fallback for the same helper. Their version is stricter — "alive" is decisive on the first read, "dead" requires two consecutive agreeing reads or a 10s deadline, and a thrown probe resets the streak — and their commit message notes the probe can also *return* a lone `false` rather than throw, which my version did not cover. **I dropped mine and took theirs** during the rebase.
- **#1296 (lifecycle observer fan-out).** This closed the last failure: `session-event-fanout.ts` delivered lifecycle records only to attached connections, so the supervisor's observer never saw `agent_start` and judged an active host idle. That is the `idle shutdown` line in the final failing log.

Root cause #1 above is what unblocked the stage before theirs: with the supervisor dying at startup, `host ready` never printed at all.

## Repeated-run evidence

`RPC named pipes (Windows)`, same commit, job re-executed each round:

| round | run id | job id | result |
|---|---|---|---|
| 1 | 33628874723 | 100243105794 | SUCCESS |
| 2 | 33628874723 | 100245721750 | SUCCESS |
| 3 | 33628874723 | 100247129659 | SUCCESS |
| 4 | 33628874723 | 100248850116 | SUCCESS |
| 5 | 33628874723 | 100250298953 | SUCCESS |

Every round carries a distinct job id, so each is a genuine re-execution on a fresh Windows runner rather than a cached conclusion.

Round 1 was verified as a real pass rather than a skip: the four suites reported 13, 27 (+2 pre-existing skips), 10, and 6 passing, with `FAIL`, `database is locked`, `connect ENOENT`, `reported dead`, `identity watchdog`, and `idle shutdown` all at zero occurrences.

## Local verification

- `rpc-host-lifecycle` + `rpc-host-ensure` + the new lock-scope test: **43/43**
- 20 consecutive local runs of the ensure suites: **20/20 green**
- `npx tsc --noEmit`: clean · Changelog gate: PASS

## Honest limitations

- Fixes 1, 3, and 4 live inside `process.platform === "win32"` branches. No Windows machine is available to me, so local runs cannot exercise them — verified rather than assumed: removing fix 1's `.catch` is a **no-op** locally (30/30 unchanged). Real Windows CI is the only instrument for those.
- No assertion was weakened, no test skipped/`.only`/xfail'd, and no retry-until-green wrapper was added around a test.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - The shared RPC supervisor's observer receives content-free session and agent lifecycle records even without a session attachment, so it no longer exits the host during an active turn.
+- The Windows RPC host supervisor no longer dies when its baseline process-identity probe times out; a failed baseline read is treated as unknown instead of throwing past the startup guard and taking the internal host down with it.
 - The Windows shared RPC host is no longer shut down by its own identity watchdog when the PowerShell process probe merely times out; an absent identity is confirmed with `kill(pid, 0)` before the host is treated as exited.
 - Shared RPC host startup no longer runs its orphaned-directory cleanup while holding the endpoint lock, so a concurrent `senpi` start on a busy machine no longer fails with a raw `database is locked` when the cleanup's temp-directory scan and per-candidate process probes outlast the lock budget.
 - Multi-session RPC `close_session` teardown is bounded by a 10-second grace period (configurable via `SENPI_RPC_CLOSE_GRACE_MS`), force-releasing the session and path reservation on expiry; a second close joins the in-flight teardown instead of returning `unknown_session`.
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - The shared RPC supervisor's observer receives content-free session and agent lifecycle records even without a session attachment, so it no longer exits the host during an active turn.
+- The Windows shared RPC host is no longer shut down by its own identity watchdog when the PowerShell process probe merely times out; an absent identity is confirmed with `kill(pid, 0)` before the host is treated as exited.
 - Shared RPC host startup no longer runs its orphaned-directory cleanup while holding the endpoint lock, so a concurrent `senpi` start on a busy machine no longer fails with a raw `database is locked` when the cleanup's temp-directory scan and per-candidate process probes outlast the lock budget.
 - Multi-session RPC `close_session` teardown is bounded by a 10-second grace period (configurable via `SENPI_RPC_CLOSE_GRACE_MS`), force-releasing the session and path reservation on expiry; a second close joins the in-flight teardown instead of returning `unknown_session`.
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - The shared RPC supervisor's observer receives content-free session and agent lifecycle records even without a session attachment, so it no longer exits the host during an active turn.
+- Shared RPC host startup no longer runs its orphaned-directory cleanup while holding the endpoint lock, so a concurrent `senpi` start on a busy machine no longer fails with a raw `database is locked` when the cleanup's temp-directory scan and per-candidate process probes outlast the lock budget.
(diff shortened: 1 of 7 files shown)


code-yeongyu:
Valid concern, and I checked it against the current branch rather than dismissing it as outdated.

The exact code you flagged — `return process.platform === "win32" && processIsLive(pidFile.pid);` in `hostAlive` — **no longer exists**. During the rebase onto #1291 I dropped my own fallback and took the maintainer's stricter implementation ("alive" decisive on the first read, "dead" requires two consecutive agreeing reads or the 10s deadline, thrown probes reset the streak). So `hostAlive` keeps the full `processMatchesPidFile` identity guard, and the PID-reuse hazard you described is gone from that path.

I also re-checked the three `processIsLive` call sites I do still have, against your specific failure mode (a recycled PID being force-killed):

- `terminateSupervisor` recovery: runs only **after** `Stop-Process` already failed, and returns early only when `!processIsLive(pid)`. A recycled live PID makes it rethrow the original error — it never kills anything.
- `waitForHostExit`: `probe.timedOut && processIsLive(...)` only decides whether to keep waiting. No termination, no exit verdict.
- The production identity watchdog: increments its counter only when `!processIsLive(...)`, so a recycled live PID **suppresses** the exit verdict rather than triggering one.

In each case raw liveness can only make the code more conservative, never cause an unrelated process to be signalled. Identity-based decisions (`stopValidatedPid`, `waitForGone`) still go through `processMatchesPidFile` as you note they should.

Resolving on that basis — happy to reopen if you see a path I missed.

Issue #1279, opened 2026-09-02
claude-sdk-oauth second login stores sentinel as login-2 credential


thisisjun786:
## What happened?

Adding a second `claude-sdk-oauth` account creates a broken `login-2` slot whose `access` and `refresh` values are `claude-sdk-oauth-managed` instead of the newly issued OAuth tokens. If affinity selects that slot, Claude SDK authentication fails.

The provider-specific login already returns the complete pooled credential via `addAccount(current, newSlot)`. `ModelsImpl.login()` then passes that pooled value to the generic `appendLoginSlot(current, credential)`, which treats its top-level sentinel fields as a flat credential and appends them as `login-2`. This is effectively double pooling.

Minimal in-memory reproduction:

```js
const current = {
  type: "oauth",
  access: "claude-sdk-oauth-managed",
  refresh: "claude-sdk-oauth-managed",
  expires: 4102444800000,
  accounts: [{ name: "default", source: "login", access: "real-a", refresh: "refresh-a", expires: 999 }],
};
const providerLoginResult = {
  ...current,
  accounts: [...current.accounts, { name: "account-2", source: "login", access: "real-b", refresh: "refresh-b", expires: 999 }],
};
appendLoginSlot(current, providerLoginResult).accounts.at(-1).access;
// => "claude-sdk-oauth-managed"
```

## Steps to reproduce

1. Log into `claude-sdk-oauth` once.
2. Run `/claude-account add` and complete a second OAuth login.
3. Inspect `/claude-account` behavior or the second stored slot.
4. Observe that `login-2` contains the sentinel rather than the second account token.

## Expected behavior

The second OAuth account should be stored once with its actual access/refresh tokens and participate in affinity/failover. The generic login path should either accept an already-pooled result or the provider login should return only the newly issued flat credential, but not both.

## Version

Senpi `2026.9.2` via omo `5.0.0-0.beta.33`.

`pi -ne` cannot exercise this provider because `claude-sdk-oauth` is itself the built-in extension under test.

PR #1196, opened 2026-08-30
fix(auth): harden Claude OAuth credential pools

eddieparc:
## Summary
- preserve provider-owned OAuth account pools instead of double-appending their flat compatibility credential
- recognize concrete OAuth credentials selected by shared credential rotation
- keep synthetic managed sentinels excluded from readiness checks
- classify OAuth refresh DNS, connection, timeout, socket, and fetch failures as temporary provider errors
- retain persistent account blocking for invalid_grant, invalid_token, revoked tokens, HTTP 401, and unauthorized responses

## Root causes
1. The shared login layer appended the Claude provider's already-complete account pool a second time, creating a fake slot from its managed sentinel.
2. The managed auth lane wrapped every token-refresh exception as authentication_failed, so a DNS outage permanently blocked a valid account until re-login.

## Verification
- transient/auth classification regression: 5 cases passed
- Claude OAuth regression suite: 5 files, 50 tests passed
- credential pool regression: 5 tests passed
- TypeScript, Biome, full monorepo build, browser smoke, and repository pre-commit checks passed
- Anthropic DNS and HTTPS probes passed 3/3 after the outage
- expired-token refresh succeeded and three consecutive live claude-sdk-oauth requests passed with fallback disabled

Issue #1169, opened 2026-08-28
Claude SDK ambient 429 result with success subtype bypasses model fallback


jezFAM:
### What happened?

A Claude subscription/session-limit response can bypass `retry.fallbackChains` when `claude-sdk-oauth` uses `tokenInjection: "ambient"`. This is distinct from #505: Claude Agent SDK 0.3.241 returned a result with a successful subtype but an explicit error flag:

```json
{
  "type": "result",
  "subtype": "success",
  "is_error": true,
  "api_error_status": 429,
  "terminal_reason": "api_error",
  "result": "You've hit your session limit · resets 2:50pm (Asia/Seoul)"
}
```

Senpi treats the turn as successful, so the retry/fallback controller receives no provider error and emits no `fallback_applied` event. This breaks Claude-primary setups that depend on another provider after the 5-hour limit.

On current main (`488d672`), `sdkFailure()` only checks result messages when `message.subtype !== "success"`, while the ambient branch directly `yield*`s `createAttemptMessages(...)` without calling `sdkFailure()`.

A local patch that also checks `message.is_error === true`, preserves `message.result` for classification, and inspects ambient messages produced `FALLBACK_QA_OK` plus `fallback_applied: claude-sdk-oauth/claude-opus-5 -> openai-codex/gpt-5.6-sol-fast` (`reason: transient`).

### Steps to reproduce

1. Configure `claude-sdk-oauth` with `tokenInjection: "ambient"` and a fallback chain from `claude-opus-5` to another provider.
2. Exhaust the Claude CLI 5-hour/session limit.
3. Run an isolated print turn:

```text
senpi -ne --model claude-sdk-oauth/claude-opus-5:max -p --no-session --no-tools --no-skills --no-context-files "Return exactly FALLBACK_QA_OK"
```

4. Observe the limit text is surfaced without switching models and no `fallback_applied` event appears.

### Expected behavior

`is_error: true` should enter retry/fallback classification even when `subtype` is `success`, and the configured fallback model should serve the turn. A regression should cover both this contradictory error result and a normal `subtype: success, is_error: false` pass-through.

### Version

Senpi 2026.8.27; @anthropic-ai/claude-agent-sdk 0.3.241; Windows 11; ambient auth lane

jezFAM:
AI disclosure: I used an AI coding agent to draft this report after live reproduction. I reviewed the reported SDK payload, source paths, root cause, and local patch verification before submission.

[...]

PR #1304, opened 2026-09-02
fix(auth): keep Claude OAuth named pools and stop provider hops

eddieparc:
## Summary
- Keep a provider-owned Claude SDK OAuth account pool on login instead of appending a fake `login-N` slot copied from the managed sentinel.
- Treat a rotation-selected concrete OAuth slot as configured so a second login does not fail with `Provider is not configured: claude-sdk-oauth`.
- Do not classify that auth miss as a hard-error model fallback, so the turn stays off other providers such as `opengateway`.

## Root causes
1. `appendLoginSlot` always treated the login result as a flat credential. Claude SDK OAuth already returns the full named pool, so the second account became `login-2` with sentinel tokens.
2. With two or more slots, credential rotation projects one slot (no `accounts` array). `oauth.check` only counted `accounts.length`, so every request looked unconfigured.
3. `AgentSession` treated that `ModelsError("auth")` as a hard-error and walked the fallback chain onto another provider that happened to have a key.

## Test plan
- [x] `packages/ai` `credential-pool-resolve-slot.test.ts` — named pool preserved; unnamed flat still becomes `login-N`
- [x] `packages/coding-agent` `claude-sdk-oauth-login.test.ts` — projected non-sentinel slot passes `check`; sentinel does not
- [x] `packages/coding-agent` `retry-fallback-hard-error.test.ts` — `Provider is not configured:` does not switch models

This is a focused replay of the pool/readiness fixes from #1196 (that branch is far behind `main`) plus the hard-error hop guard. No internal tracker ids.

---
## Summary by cubic
Fixes Claude SDK OAuth login, retry, and interactive login behavior so account selection stays stable and transient failures do not trigger unrelated provider fallbacks.

**Bug Fixes**
- Preserves provider-owned named OAuth pools on login by merging them onto the stored pool, so a sibling's rotated token or rate-limit block is never rewound.
- Drops synthetic sentinel slots: poisoned `login-N` slots are healed on auth.json read, their pins cleared, and an unconfigured-slot auth miss fails over per-slot so one bad slot can't dead-end a healthy pool.
- Scopes same-model remint and auth-miss exclusion to the Claude SDK lane: session locks and bare `invalid_request` stay on the same model, other providers' failures still hop the fallback chain, and provider-agnostic stalls use the shared retry budget.
- Remounts the single login `Input` widget and tracks one live hint row so a browser callback no longer shows two `>` prompts or duplicate hint rows.

<sup>Written for commit 8d14a5f4df9dfa04a1d1eda066e80e25e821a35f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -9,6 +9,8 @@
 ### Changed
 
 ### Fixed
+- Login no longer double-appends a provider-owned OAuth account pool as a fake `login-N` slot copied from the flat compatibility sentinel.
+- Claude Agent SDK `Lock file is already being held` is classified as a transient retryable error instead of an unknown/terminal failure.
 
 ### Removed
 
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -9,7 +9,8 @@
 ### Changed
 
 ### Fixed
-- Login no longer double-appends a provider-owned OAuth account pool as a fake `login-N` slot copied from the flat compatibility sentinel.
+- A provider-owned OAuth account pool is merged onto the stored pool at commit time instead of overwriting it with the pre-login snapshot, so a sibling account that rotated its refresh token or earned a rate-limit block during an interactive login is never rewound; pool slots carrying the provider's managed sentinel marker are recognized and dropped so they can never dead-end a request.
+- The auth-miss wording `Provider is not configured: <id>` is now a shared exported helper used by every throw site, so consumers keying recovery decisions off it cannot drift from the generators.
 - Claude Agent SDK `Lock file is already being held` is classified as a transient retryable error instead of an unknown/terminal failure.
 
 ### Removed
--- a/packages/ai/src/auth/pool/slots.ts
+++ b/packages/ai/src/auth/pool/slots.ts
@@ -166,6 +166,29 @@ function nextLoginSlotName(credential: PooledCredential): string {
 	throw new Error("Credential pool is full");
 }
 
+function providedSlots(credential: Credential): CredentialSlot[] | undefined {
+	const accounts = (credential as PooledCredential).accounts;
+	return Array.isArray(accounts) && accounts.length > 0 ? accounts : undefined;
+}
+
+/**
+ * Unions a provider-owned pool onto the value read under the credential lock.
+ * `current` wins for every name it already holds: the provider built its
+ * object from a snapshot taken BEFORE the interactive browser round trip, so a
+ * sibling account that rotated its refresh token or earned a rate-limit block
+ * during that window must not be rewound to the snapshot. Only names that do
+ * not exist yet are appended.
+ */
+function mergeProvidedPool(
+	current: PooledCredential,
+	existing: readonly CredentialSlot[],
+	provided: readonly CredentialSlot[],
+): PooledCredential {
+	const known = new Set(existing.map((slot) => slot.name));
+	const added = provided.filter((slot) => !known.has(slot.name));
+	return added.length === 0 ? current : { ...current, accounts: [...existing, ...added] };
+}
+
 /**
  * Appends an unnamed flat credential to a pool as a generated `login-N` slot.
  * An absent current entry keeps today's whole-write shape; a flat current entry
@@ -173,18 +196,62 @@ function nextLoginSlotName(credential: PooledCredential): string {
  * instead of being overwritten by the second login.
  *
  * A login result that already carries its own populated `accounts` array is a
- * provider-owned pool: it IS the complete post-login credential, so it is
- * written through untouched. Reading its top-level fields as a flat credential
- * would append the provider's placeholder material as a second slot.
+ * provider-owned pool, and a provider that already holds a POOL names the
+ * account this login created itself. That pool is MERGED onto `current` rather
+ * than written through: a pre-browser-flow snapshot must never overwrite what
+ * concurrent writers stored meanwhile. A flat `current` keeps the whole-write
+ * shape - the provider echoes the flat fields it read, so there is nothing to
+ * preserve and the login must not be dropped on a name collision.
  */
 export function appendLoginSlot(current: PooledCredential | undefined, flat: Credential): Credential {
-	if ("accounts" in flat && Array.isArray(flat.accounts) && flat.accounts.length > 0) {
-		return flat;
-	}
 	if (!current) {
 		return flat;
 	}
+	const storedAccounts = Array.isArray(current.accounts) ? current.accounts : undefined;
+	const provided = providedSlots(flat);
+	if (provided) {
+		// A pool merges onto the stored pool; a flat current keeps the whole-write
+		// shape because the provider's accounts already carry this login.
+		return storedAccounts ? mergeProvidedPool(current, storedAccounts, provided) : flat;
+	}
 	return upsertSlot(current, slotFromFlatCredentialNamed(flat, nextLoginSlotName(current)));
+}
+
+/**
+ * Material a provider writes into the flat credential when the real token lives
+ * outside auth.json (an SDK subprocess or a vendor CLI owns it): the literal
+ * `<providerId>-managed` in both OAuth fields. It is a marker, never a token.
+ */
+export function managedSentinelMaterial(providerId: string): string {
+	return `${providerId}-managed`;
+}
+
+/**
+ * A POOL SLOT carrying that marker can never authenticate: `projectSlot` hands
+ * it to the provider's `check`, which rejects it, and the request dies with
+ * "Provider is not configured". Slots like that exist only because a shipped
+ * build appended a provider-owned pool's flat sentinel as a generated `login-N`
+ * slot.
+ */
+export function isManagedSentinelSlot(providerId: string, slot: CredentialSlot): boolean {
+	const sentinel = managedSentinelMaterial(providerId);
+	return slot.access === sentinel && slot.refresh === sentinel;
+}
+
+/**
+ * Drops those poisoned slots from a stored credential, clearing a pin that
+ * named one. Returns `undefined` when the credential holds none, so a caller
+ * can tell a repair from a no-op and only rewrite storage when the bytes
+ * actually change.
+ */
+export function repairManagedSentinelSlots(providerId: string, credential: Credential): PooledCredential | undefined {
+	const pooled = credential as PooledCredential;
+	if (!Array.isArray(pooled.accounts)) return undefined;
+	const kept = pooled.accounts.filter((slot) => !isManagedSentinelSlot(providerId, slot));
+	if (kept.length === pooled.accounts.length) return undefined;
+	const repaired: PooledCredential = { ...pooled, accounts: kept };
+	if (repaired.pinned !== undefined && !kept.some((slot) => slot.name === repaired.pinned)) delete repaired.pinned;
+	return repaired;
 }
 
 /**
--- a/packages/ai/src/auth/resolve.ts
+++ b/packages/ai/src/auth/resolve.ts
@@ -36,6 +36,18 @@ export interface AuthResolutionOverrides {
 	signal?: AbortSignal;
 }
 
+/**
+ * Prefix of the auth-miss every resolution site raises when a provider has no
+ * usable credential. Consumers key recovery decisions off this exact wording,
+ * so it is a shared constant instead of a literal repeated at each throw site:
+ * rewording one copy would silently disable the other's behavior.
+ */
+export const PROVIDER_NOT_CONFIGURED_PREFIX = "Provider is not configured: ";
+
+export function providerNotConfiguredMessage(providerId: string): string {
+	return `${PROVIDER_NOT_CONFIGURED_PREFIX}${providerId}`;
+}
+
 export class ModelsError extends Error {
 	readonly code: ModelsErrorCode;
 
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -1,3 +1,23 @@
+## Classify Claude SDK session lock contention as retryable (2026-09-02)
+
+### What changed
+
+- `packages/ai/src/utils/retry.ts`: `RETRYABLE_PROVIDER_ERROR_PATTERN` matches `Lock file is already being held`.
+- `packages/ai/test/retry.test.ts`: pins that wording as a retryable assistant error.
+
+### Why
+
+- Claude Agent SDK session resume/stream hits proper-lockfile while a previous subprocess still holds `session.json`. The failure is local and transient; treating it as unknown/terminal made the coding-agent hard-error fallback hop providers.
+
+### Why an extension could not handle it
+
+- Retry classification lives in the shared `pi-ai` regexes used by every caller of `isRetryableAssistantError`.
+
+### Expected merge conflict zones
+
+- LOW: `RETRYABLE_PROVIDER_ERROR_PATTERN` in `retry.ts`.
+
+## Preserve provider-owned credential pools during login (2026-09-02)
 ## 2026-09-10 - Kimi Code client identity headers on the subscription path (#1504)
 
 ### What changed
--- a/packages/ai/src/changes.md
+++ b/packages/ai/src/changes.md
@@ -1,3 +1,23 @@
+## PR #1304 review fixes: shared auth-miss prefix, login merge, sentinel repair (2026-09-10)
+
+### What changed
+
+- `packages/ai/src/auth/resolve.ts`: `PROVIDER_NOT_CONFIGURED_PREFIX` / `providerNotConfiguredMessage()` export the exact auth-miss wording every resolution site throws; `packages/ai/src/models.ts` re-exports both and throws through the helper. Consumers keying recovery decisions off that message (the coding-agent session layer and the credential-pool classifier) can never drift from the throw sites.
+- `packages/ai/src/auth/pool/slots.ts`: `appendLoginSlot` MERGES a provider-owned pool onto the value read under the credential lock - stored slots and their block state win for names that already exist, only genuinely new names are appended - instead of whole-writing a snapshot the provider built before the interactive browser round trip. `managedSentinelMaterial` / `isManagedSentinelSlot` / `repairManagedSentinelSlots` recognize and drop pool slots whose `access` and `refresh` both equal the provider's `<providerId>-managed` marker (clearing a pin that pointed at one), reporting whether a repair happened so callers rewrite storage only when bytes change.
+- `packages/ai/test/credential-pool-mutations.test.ts`: a pre-login snapshot never rewinds a sibling that rotated or earned a block; a provider-owned pool onto a flat current keeps the whole-write shape; sentinel slots are recognized, dropped, un-pinned, and a clean pool is a no-op.
+
+### Why
+
+- The provider builds its returned pool from a snapshot read BEFORE the browser flow, tens of seconds before the commit under the lock: writing it verbatim rolled a sibling's rotated refresh token back to the consumed value (a forced re-login, since Anthropic rotates refresh tokens on use) and erased its rate-limit block. Separately, a shipped build stored the provider pool's flat sentinel as a generated `login-N` slot; such a slot can never authenticate and dead-ends every request whose affinity picks it, so the coding-agent store now heals those entries on load.
+
+### Why an extension could not handle it
+
+- Both the merge and the repair algebra run inside the shared credential-pool read/write path the runtime owns; providers cannot intercept what the runtime stores after login returns or what every reader parses from auth.json.
+
+### Expected merge conflict zones
+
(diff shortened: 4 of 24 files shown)


code-yeongyu:
#1309 (merged as 10f58c073) adopted the `appendLoginSlot` provider-owned-pool guard and the `configuredFor` projected-slot count from this PR (and #1196), credited to @eddieparc in the PR body, and closes #1279. The remaining hunks here (agent-session hard-error classification, retry.ts, session-continuity, login-dialog) were deliberately left out of that PR because they change fallback policy rather than login storage - they can be rebased onto main as a focused follow-up if you still want them; happy to review that separately.

code-yeongyu:
Thanks @eddieparc — the Claude SDK OAuth named-pool / projected-slot fixes from this PR were adopted on main via #1309 (`e12ae5b08` `fix(ai): keep a provider-owned credential pool intact on login`, `d3dbe39be` `fix(claude-sdk-oauth): treat a rotation-projected slot as a configured account`; merge `10f58c073`). Those hunks are credited in the #1309 `changes.md` entries.

Please do not merge this branch onto current main (`6dd662c4b`). GitHub reports CONFLICTING/DIRTY, the pool/readiness files are now duplicates, and the leftover AgentSession / continuity edits change fallback policy in ways that collide with #1312 and #1318.

Remaining work if this PR stays open:
1. Rebase onto `main` (`6dd662c4b`) and **drop** `slots.ts`, `oauth-login.ts`, the duplicate login/pool tests, and the #1309 changelog/`changes.md` entries.
2. **Drop** the AgentSession `Provider is not configured:` hard-error exclusion. The rotation false-positive is gone after #1309; the guard now blocks legitimate fallback for every unconfigured provider (`models.ts:680`, `model-runtime.ts:711`).
3. **Drop** the bare `invalid_request` same-model remint. #1312 classifies `invalid_request` as non-retryable (`errors.ts:17`) and landed `is_error` → failover / model fallback on purpose (#1169).
4. **Drop** the stream-stall same-model remint. Stalls are already retryable; `_handleRetryableError` documents hop-after-budget (`agent-session.ts:7617-7621`). `_isClaudeSdkSameModelRemintError` is also not provider-scoped.
5. **Keep, isolate:** `login-dialog.ts` remount so a browser callback cannot paint two live `>` prompts (still broken at `login-dialog.ts:140` and `:160`). Test: `packages/coding-agent/test/suite/regressions/5433-extension-oauth-prompt-input.test.ts`. Optional follow-up: route `showWaiting` / `showInfo` hints through the same remount.
6. **Keep, isolate, justify:** classify `Lock file is already being held` as retryable in `packages/ai/src/utils/retry.ts` and, if you still want never-hop, same-model remint *only* for that lock string in AgentSession. Do not bundle stall/`invalid_request`. Tests: `packages/ai/test/retry.test.ts`, lock cases in `retry-fallback-hard-error.test.ts`.
7. **Keep only after a #1318-aware rebase:** fork a *confirmed* persisted binding on options/model drift instead of flatten (`decideFromBinding` at `session-continuity.ts:157`), leaving unconfirmed ids on `flatten(session_unconfirmed)` via `withoutUnconfirmedResume` (`:143-151`, `:235`). Do not fork `account_changed` without the failover lane's cross-account check. Rewrite `claude-sdk-oauth-restored-security.test.ts` and `claude-sdk-oauth-continuity-retry-checkpoint.test.ts`; run `claude-sdk-oauth-unconfirmed-binding.test.ts` so #7562 does not regress.
8. Refresh `changes.md` + Unreleased CHANGELOG only for hunks that remain. The 2026-09-02 pool/auth-miss write-ups are already on main.

Until that split/rebase, this stays open and unmerged.

code-yeongyu (changes requested):
Verdict: REQUEST CHANGES. The two credential-pool fixes are aimed at real bugs (a provider-owned Claude pool really was re-appended as a sentinel `login-N`, and a rotation-projected slot really did fail `oauth.check`), but as written the branch trades those bugs for worse ones and it is red on `main`'s own suites. I fetched `refs/pull/1304/head` (501c10600), pushed it plus its merge base (69d9437cd) to a scratch Linux box, and ran the affected suites there: **three existing tests that pass at the merge base fail on this branch** (`retry-fallback-stall-shared-budget.test.ts:92` and `:128`, `provider-idle-recovery.test.ts:276`), because the new same-model-remint predicate swallows the provider-agnostic stream-stall class. Separately, the `appendLoginSlot` early return converts login from a read-modify-write under the credential lock into a blind overwrite with a snapshot read before the browser round trip, which rolls back a sibling account's rotated refresh token and its rate-limit block; and nothing in the branch repairs or ignores the sentinel `login-2` entries the shipped bug already wrote, so those pools now dead-end deterministically instead of merely rotating badly. Every finding below was executed, not inferred; scratch workspaces have been deleted.

1. **`packages/coding-agent/src/core/agent-session.ts:7389` (`_isClaudeSdkSameModelRemintError`) breaks three existing tests.** The predicate ORs in `isProviderStreamStallError`, which is the provider-agnostic agent-loop watchdog class (`Idle timeout waiting for provider stream after Nms` / `Provider stream start timed out after Nms`), and the new branch at `:2385` is evaluated *before* `retryableError`, so stalls stop taking the ordinary transient path. Reproduced on a clean checkout of 501c10600 (`npx vitest run` in `packages/coding-agent`):
   - `test/suite/retry-fallback-stall-shared-budget.test.ts:92` — `expected [] to match object [ { from: 'faux/faux-1', to: 'faux/faux-2', chainKey: 'faux/faux-1', reason: 'transient' } ]`: a stall now never escalates to the configured fallback chain, for **any** provider.
   - `test/suite/retry-fallback-stall-shared-budget.test.ts:128` — `expected [ { success: false, attempt: 4 } ] to deeply equal [ { success: false, attempt: 3 } ]`: the remint path emits `attempt: this._retryAttempt` while the transient path emits `this._retryAttempt - 1`, so the exhaustion event is off by one.
   - `test/suite/regressions/provider-idle-recovery.test.ts:276` — `attempt: 2` where `1` is expected, same off-by-one.
   All nine tests in those two files pass at merge base 69d9437cd on the same machine, so this branch is the cause. Two knock-on effects: `agent-session.ts:7631`'s `const stallError = isProviderStreamStallError(message)` is now dead (stalls can never reach that `else`), and the comment at `:7626` ("consumes the same bounded same-model budget ... and escalates to the fallback chain only when that budget is exhausted") now documents behaviour the code no longer has. `_isClaudeSdkInvalidRequestError` is mis-scoped the same way: a bare `invalid_request` from *any* provider is now remint-only. Fix: drop `isProviderStreamStallError` from the predicate (the stall class already reaches the same-model budget via `isRetryableAssistantError`, which matches `timed? out`/`timeout`), and gate the lock/`invalid_request` cases on `this.model?.provider === CLAUDE_SDK_OAUTH_PROVIDER_ID` so a Claude-SDK-specific quirk cannot reclassify OpenAI/gateway stalls. If a deliberate global change is intended, update `retry-fallback-stall-shared-budget.test.ts` in the same commit with the rationale, rather than leaving it red.

2. **`packages/ai/src/auth/pool/slots.ts:149` turns login into a lost update.** `Models.login` calls `appendLoginSlot` *inside* `credentials.modify` (`packages/ai/src/models.ts:606`), so `current` is the authoritative under-lock value; the new early return discards it and writes the provider's object verbatim. The Claude provider builds that object from a snapshot taken **before** the interactive browser flow (`oauth-login.ts:160` `readCurrent()`, then `:193` `addAccount(current, …)`), i.e. tens of seconds earlier. Anything written during that window is silently reverted. Executed side by side (same input: on-disk `default` rotated to `access-v2/refresh-v2` and rate-limit blocked while login was in the browser):
   - 501c10600 writes `[{name:"default",access:"access-v1",refresh:"refresh-v1"},{name:"work",…}]` — the rotated pair is gone and `blockedUntil`/`blockReason` are gone.
   - 69d9437cd writes `[{name:"default",access:"access-v2",refresh:"refresh-v2",blockedUntil:…,blockReason:"rate_limit"},{name:"login-2",access:"claude-sdk-oauth-managed",…}]` — the bug you are fixing, but the concurrent write survives.
   The real writers in that window are `claude-sdk-oauth/accounts.ts:refreshSlot` (called from `auth-lane.ts:prepareSlot` for any live session) and `claude-sdk-oauth/failover.ts:104 persistBlock`. Anthropic rotates refresh tokens on use, so rolling `refresh` back to the consumed value leaves that sibling account unable to refresh at all — a forced re-login, not a cosmetic loss; reverting `blockedUntil` immediately re-selects a rate-limited account. Fix: merge instead of replace — when `flat` carries `accounts`, union it onto `current` by slot name, keeping `current`'s material and block state for names that already exist and appending only genuinely new slots (or, cleaner, have the provider return just the new `CredentialSlot` and let the shared layer `upsertSlot` it under the lock).

3. **Nothing repairs the pools the shipped bug already poisoned (`slots.ts:149` + `oauth-login.ts:112`), and this PR makes them terminal.** A user who logged in twice on the released build has `accounts: [default(real), login-2(access/refresh = "claude-sdk-oauth-managed")]`. Executed against this branch: `projectSlot(poisoned, "login-2")` → `{access:"claude-sdk-oauth-managed",refresh:"claude-sdk-oauth-managed"}`; `config.check(...)` → `undefined` (your own new test at `test/claude-sdk-oauth-login.test.ts:115` pins exactly this); `model-runtime.ts:711` then throws `Provider is not configured: claude-sdk-oauth`; and `credential-pool/classify.ts:115` returns `{kind:"fail_request"}` for that message, so `runCredentialFailover` does **not** try the healthy `default` slot. Combined with finding 4 the turn no longer falls back either, so ~half of sessions (HRW is per-affinity-key, hence stable per session) dead-end on every request with no recovery path, and re-login does not help because it appends `account-3` and leaves `login-2` in place. Fix: reclaim these entries — drop slots whose `access`/`refresh` equal `SENTINEL_OAUTH_FIELDS` when reading the credential (`claude-sdk-oauth/accounts.ts:listAccounts`, plus a one-time write-back migration), and/or classify an unconfigured-slot auth miss as `{kind:"failover", block:{reason:"auth_error"}}` in `credential-pool/classify.ts` so a single bad slot can never dead-end a multi-account pool. Please also add a test that starts from a pool containing a sentinel slot and asserts the request still succeeds on the healthy account.

4. **`packages/coding-agent/src/core/agent-session.ts:7400` disables hard-error fallback for every provider, not just Claude.** `!message.errorMessage?.startsWith("Provider is not configured:")` matches the message built at `model-runtime.ts:711` for *any* provider id. A user whose primary model loses its key mid-session (deleted `auth.json` entry, expired ambient token, revoked slot) now gets a terminal turn instead of the configured `fallbackChains` hop — which is the entire purpose of that chain. And for the Claude case it buys no recovery: the credential-pool layer has already returned `fail_request` (finding 3), so "stay on Claude SDK OAuth or its sibling accounts" is not what happens; the turn just fails. Your new test `test/suite/retry-fallback-hard-error.test.ts:289` pins the new behaviour, so it cannot detect this. Fix: compare against the current model's provider and restrict it (`errorMessage === \`Provider is not configured: ${this.model?.provider}\`` && that provider is `claude-sdk-oauth`), or drop this hunk entirely and fix the miss where it originates (findings 2–3), which is the smaller and more honest change.

5. **`packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/session-continuity.ts:152` forks across accounts on lanes the same file declares un-forkable.** `forkBindingOrFlatten` (`:135`) ignores both guards that the rest of the module enforces: `verifyRestoredTranscript` (`session-reattach.ts:101`) returns `false` for `config-dir`, and `decideFailoverContinuity` (`:201`) calls that lane the one declared `cross_root_unsupported` residual. It is also not covered by admission verification, because `session-restored-admission.ts:15` returns `transcriptAvailable: true` unverified whenever `sentPrefixHash` is absent — and in-memory bindings built by `bindingFromEntry` (`session-reattach.ts:68`, used by `session-registry-wiring.ts:49 keepBindingThenClose` after exactly the stream-start-timeout teardown this PR targets) never carry one. Executed on this branch: `admitRestoredBinding("s1","/repo","config-dir")` → `{transcriptAvailable:true}`, and `decideNativeContinuity` with `accountName` drift → `{kind:"fork",sdkSessionId:"sdk-A",atUuid:"uuid-a2",from:2,reason:"account_changed"}`, while `decideFailoverContinuity({authLane:"config-dir",crossAccountResumeSupported:true,…})` → `{kind:"flatten",reason:"cross_root_unsupported"}`. Consequence: the fork targets a session id that exists only in the previous account's `CLAUDE_CONFIG_DIR`, so `reattachSession` spawns a subprocess, fails initialization, and lands in `onResumeFallback` → cold-seed flatten with `resume_initialization_failed` — the same megabyte re-send this PR is trying to avoid, plus a wasted spawn and init window. Fix: thread `authLane` into `decideFromBinding`/`forkBindingOrFlatten` and keep `flatten` for `config-dir` (and for `account_changed` without `crossAccountResumeSupported`), or remove the unverified shortcut at `session-restored-admission.ts:15` so the lane veto applies to every binding.

6. **Two of the new/relaxed assertions cannot fail.** `test/suite/regressions/5433-extension-oauth-prompt-input.test.ts:114` asserts that zero rendered lines are a bare `>`; both disjuncts (`/^>\s*$/.test(line.trim())` and `line.trim() === ">"`) are the same predicate, and the double-mounted `Input` renders `> jgplabs` twice, so nothing bare ever appears. Verified by reintroducing the pre-fix `remountInput` (plain `addChild`) and neutralising line 113: line 114 still passes while the rendered-value count is 2. Only `:113` has any power; make the intent explicit instead, e.g. assert that exactly one line matches the live input row. Also, `test/claude-sdk-oauth-continuity-retry-checkpoint.test.ts:90` was relaxed from `toEqual` to `toMatchObject({kind:"fork",reason})`, which drops the `atUuid`/`from` assertions — precisely the fork target finding 5 disputes. Please restore `toEqual` with the expected fork point so a wrong branch anchor is caught.

Verification notes: `bun install` + targeted `npx vitest run` on a scratch Linux host, at 501c10600 and at merge base 69d9437cd. Also green on this branch: `packages/ai` `retry.test.ts` + `credential-pool-resolve-slot.test.ts` (56 passed), and a 117-file coding-agent sweep of every test matching retry/fallback/claude-sdk/continuity/login/credential — 960 passed, only the two `retry-fallback-stall-shared-budget` failures above. I did not run the full monorepo suite, so `npm test` may hold more stall-class expectations than the three files named here.

code-yeongyu:
Every finding from the review is addressed. The branch is rebased onto current `main` (`6e2cd043`) — during the rebase, hunks that `main` had already landed by other paths (the `appendLoginSlot` provider-pool guard, the `oauth-login.ts` slot counting, the two `check` tests) were resolved to `main`'s versions, and the `forkBindingOrFlatten` hunk was dropped in favor of `main`'s `crossAccountResumeSupported` wiring, which already enforces the lane declaration finding 5 asks for. New head: `8d14a5f4d`.

## Finding → commit → evidence

**1. Regression: stall class swallowed by `_isClaudeSdkSameModelRemintError`** — commit `d074dfa9f`. The predicate now matches only the Claude SDK lane's session-lock and bare `invalid_request` quirks, gated on `this.model?.provider === CLAUDE_SDK_OAUTH_PROVIDER_ID`; `isProviderStreamStallError` is gone from it, so the provider-agnostic stall class takes the ordinary transient path for every provider and escalates to the fallback chain exactly as before.
- Head-vs-base (bunshin remote host, `bun run test` in `packages/coding-agent`): merge base `69d9437cd` → **9/9 pass**; old head `501c10600` → **3 fail** (`retry-fallback-stall-shared-budget` ×2, `provider-idle-recovery` ×1, the exact three from the review); new head `8d14a5f4d` → **9/9 pass** (2 files, 9 passed).
- Mutation: reintroducing the original un-gated predicate with the stall disjunct fails exactly those three tests again (`3 failed | 6 passed`); restore → `9 passed`.

**2. `appendLoginSlot` blind overwrite** — commit `ecaae7e7d`. A provider-owned pool is now *merged* onto the value read under the credential lock: stored slots (with their material and block state) win for every name that already exists, only genuinely new names are appended; a flat `current` keeps the whole-write shape because the provider's accounts already carry this login. Mutation: reverting to `return flat` fails `a provider-owned pool never rewinds a sibling the browser round trip left behind` (rotated refresh plus `blockedUntil`/`blockReason` asserted); restore → pass.

**3. Poisoned sentinel `login-2` entries never repaired** — commits `ecaae7e7d` (slot algebra: `managedSentinelMaterial` / `isManagedSentinelSlot` / `repairManagedSentinelSlots`) and `b6eb8faa2` (wiring). `AuthStorage` now drops pool slots whose `access`/`refresh` equal the provider's `<providerId>-managed` sentinel wherever auth.json is parsed, clears a pin naming one, and writes the repair back once inside its existing lock; `credential-pool/classify.ts` classifies `Provider is not configured:` as `failover`/`auth_error` so one bad slot can never dead-end a pool; `claude-sdk-oauth/accounts.ts` never lists a sentinel-material slot. Tests: on-disk bytes healed after one read (auth-storage migration test), and a rotation over a poisoned pool never attempts `login-2` and succeeds on the healthy account (the reviewer-requested test). Mutations: a no-op repair fails the migration + rotation tests; removing the classify branch fails the taxonomy test; both pass after restore.

**4. `Provider is not configured:` exclusion global** — commit `d074dfa9f`. Replaced by `_isClaudeSdkAuthMissError`: an exact-message comparison scoped to the Claude SDK lane. New guard test: another provider's auth miss still hops the configured chain — mutation back to the global `startsWith` fails it. The cubic P3 is fixed in the same commit: `PROVIDER_NOT_CONFIGURED_PREFIX` / `providerNotConfiguredMessage()` are exported from `packages/ai/src/auth/resolve.ts` and used by both throw sites (`models.ts`, `model-runtime.ts`) and the session guard.

**5. `forkBindingOrFlatten` forks across accounts on `config-dir`** — resolved by the rebase: the hunk is dropped and `main`'s implementation stands (`decideFromBinding` flattens `account_changed` as `cross_root_unsupported` when `crossAccountResumeSupported` is false, filled from `auth.authLane !== "config-dir"`; `verifyRestoredTranscript` still refuses the lane). Mutation: disabling that guard fails `does not let a checkpoint outrank the config-dir lane limit` and `cold-seeds a persisted binding when the account drifts on the config-dir lane`; restore → pass. The relaxed `toMatchObject` on the checkpoint test was also dropped in the rebase — the exact-anchor `toEqual`/anchor assertions are back (finding 6, part 2).

**6. `5433-extension-oauth-prompt-input.test.ts:114` tautology** — commit `8d14a5f4d`. The bare-`>` line is replaced by an assertion that exactly **one** live `>` row renders; mutation: reintroducing the pre-fix double-mount (plain `addChild`) fails the test (`2 failed`); restore → `7 passed`. The same commit routes `showWaiting`/`showInfo` hints through one tracked live hint (the cubic P3 on `login-dialog.ts`), with a new test proving an interleaved waiting step leaves exactly one hint row.

## Verification

All on a bunshin remote Linux host, `bun run test` per package (bun-only, no local runs):

- Affected suites at `8d14a5f4d`: stall/idle-recovery 9/9 · retry-fallback-hard-error 18/18 · auth-storage + taxonomy + rotation + accounts 48/48 · 5433 7/7 · continuity 33/33 · `packages/ai` pool/retry tests 82/82.
- Broad sweep (every coding-agent test matching retry/fallback/claude-sdk/credential/login): **105 files, 858 passed, 3 skipped**.
- `biome check` on every changed file: clean. `tsc --noEmit`: error list byte-identical to `origin/main` on the same host (all pre-existing `vitest`/`@babel` module-resolution noise from the `bun install --ignore-scripts` layout); zero errors in changed files.
- One pre-existing failure is **not** from this branch: `auth-storage.test.ts > keeps a coalesced reload alive while another credential reader is waiting` fails identically at `origin/main` on this host (verified at `origin/main`, at the pre-fix rebase point, and at the branch head; it passes when the file runs inside the full sweep). It is order/runner-dependent on this host and untouched by this PR.

code-yeongyu (approved):
All findings resolved with head-vs-base proof: base 69d9437cd 9/9 pass, the old head reproduced the exact 3 failures the review reported, the new head is 9/9 - the stall-class regression is gone because the predicate is now narrowed to the Claude lane's lock/invalid_request quirks and provider-gated. appendLoginSlot merges onto the under-lock value so a sibling's rotated refresh and rate-limit block survive a login; the poisoned sentinel login-N entries are healed at parse time with a one-shot write-back and never listed; the Provider-is-not-configured exclusion is Claude-scoped via a shared message helper; and the 5433 tautology now asserts exactly one live row. 8 mutation proofs, each reintroducing the defect then restoring. Approving.

PR #1308, opened 2026-09-03
fix(auth): preserve legacy OAuth credential when adding accounts

eddieparc:
Linear: https://linear.app/jgplabs/issue/LAB-109/openai-codex-%EA%B8%B0%EB%B3%B8-oauth-%EB%8B%A4%EC%A4%91-%EA%B3%84%EC%A0%95-%EC%A7%80%EC%9B%90-%EC%83%81%ED%83%9C-%ED%99%95%EC%9D%B8

## Summary
- Promote a legacy flat credential to the default pool slot when a second login is saved.
- Preserve the legacy top-level projection for older binaries.
- Add API-key and OAuth regression coverage for repeated login storage.

## Verification
- `bun run --cwd packages/ai test test/credential-pool-mutations.test.ts`
- Root `bun run check` passed.
- Package build is currently blocked by pre-existing missing `@earendil-works/pi-telemetry` type resolution.

---
## Summary by cubic
Fixes credential pool mutations for `openai-codex` OAuth accounts so a second login preserves the legacy credential as the `default` slot instead of overwriting it, and removing that account leaves the survivor actually in use (LAB-109). Adds a dedicated `/gpt-account` command for managing OpenAI Codex OAuth accounts.

- Promotes a legacy flat credential to the pool's default slot when a new login is appended.
- Removing the account the flat fields projected now re-projects them from the first surviving slot, so a single-account pool still authenticates as the account that remains.
- Existing pooled accounts are unchanged; new credentials still get the next `login-N` name.
- Adds `/gpt-account` with `add`, `remove <name>`, `pin <name>`, `unpin`, and listing that hides token material; `add` announces the new account so subscribed clients re-read the pool, and a cancelled login exits silently.
- Adds regression tests for OAuth and API-key repeated login storage, survivor re-projection, and `/gpt-account` behavior.

<sup>Written for commit 1299bd118b39b01f2a84f1e14d8f7bbe705a88d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1308?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Fixed
 
 - Adding a second account to a provider whose stored credential predates credential pools (a flat entry with no `accounts` array, which is what `openai-codex` OAuth login writes) no longer overwrites the first one. `appendLoginSlot` now promotes that legacy credential into the pool as the `default` slot and stores the new login beside it as `login-2`, so both accounts remain usable and the flat top-level fields still authenticate a build predating pools. First login (no stored credential) still writes the flat credential as-is, and a provider that returns its own populated `accounts` array is still written through untouched.
+- Removing the account whose material the flat top-level credential fields projected no longer leaves the pool authenticating as the deleted account. `removeSlot` now re-projects those fields from the first surviving slot, so a pool left with a single account (which does not enter credential rotation and therefore resolves through the flat projection) immediately uses the account that remains. `accounts` is kept, removing a non-projected slot still leaves the flat fields untouched, and removing the last slot still drops the credential entirely.
 
 ### Removed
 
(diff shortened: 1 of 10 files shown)


code-yeongyu:
Note: #1309 (merged as 10f58c073) changed `appendLoginSlot` to return a provider-owned pooled credential unchanged (fixes #1279). It does not touch the legacy-flat promotion this PR proposes for openai-codex (LAB-109), so this PR stays relevant but needs a rebase onto main; the guard was placed at the top of `appendLoginSlot` before the `!current` branch.

PR #1332, opened 2026-09-03
fix(tool-search): use the API's native search tool name and inject only on Anthropic hosts

JJONGFILM:
## What

Two defects in the Anthropic native tool-search adapter (`builtin/tool-search/native-search.ts`) that together break every `anthropic-messages` request once the extension catalog contains an inactive tool:

1. **Wrong tool `name`.** The adapter appended `{ type: "tool_search_tool_bm25_20251119", name: "tool_search" }`. The Messages API only accepts `name: "tool_search_tool_bm25"` for that type and rejects the whole request:
   ```
   400 tools.39.tool_search_tool_bm25_20251119.name: Input should be 'tool_search_tool_bm25'
   ```
2. **Injected on every `anthropic-messages` endpoint.** The gate was `api === "anthropic-messages"`, so third-party hosts that speak the Anthropic wire format (Kimi Code `api.kimi.com/coding`, OpenRouter, proxies) also received the server-side tool and answered with an opaque `400 {"error":{"type":"invalid_request_error","message":"Invalid request Error"}}`.

## Why it surfaced now

The adapter has been in since #811, but `enabled()` only fires when the catalog has an inactive extension tool. #1314 (eval-only routing for `bash`/`workflow`/`monitor` as the default) makes that true for every session, so `2026.9.3-3` turned native injection on for everyone and both defects hit at once. The 400 fallback (`noteResponseStatus`) only kicks in after the first failed turn of each session and, on third-party hosts, hid the real cause behind a generic error.

## Changes

- `ANTHROPIC_TOOL_SEARCH_NAME` → `"tool_search_tool_bm25"`.
- `AnthropicNativeToolSearchAdapter.applyBeforeRequest(model, payload)` now takes the request model (`event.model ?? ctx.model`) and skips injection unless `isFirstPartyAnthropicEndpoint(model.baseUrl)` — hostname `anthropic.com` or a subdomain. A missing `baseUrl` keeps the previous behaviour, so existing callers/tests are unaffected. `addAnthropicNativeToolSearch` itself is unchanged (still the pure `api` + payload transform).
- Test validator mock (`native-search-mocks.ts`) now enforces the tool `name` the way the API does, so the old name fails the existing HARD-RULES tests instead of passing silently.
- New cases in `test/tool-search/native-anthropic.test.ts`: exact tool name, pre-fix name is a 400 in the validator, host classification, and that an `anthropic-messages` model on a third-party host gets the untouched payload while a later 400 from that host is not attributed to native search.
- `tool-search/changes.md` entry.

No `CHANGELOG.md` edit (maintainer-owned).

## Verification

- `npx vitest --run test/tool-search test/mcp/tool-search-promotion.test.ts test/mcp/native-anthropic.test.ts` — green.
- `npm run check` — green.
- Live reproduction on `omo-ai 5.0.0-0.beta.38/39` (senpi `2026.9.3-3`), session log excerpts with the two error shapes above against `anthropic/claude-fable-5`, `anthropic/claude-opus-5`, and `kimi-coding/k3`. Applying the same two changes to the installed `dist/` restored both providers; the Kimi model keeps working through the local `tool_search` tool.

## Notes for reviewers

- A corporate proxy in front of `api.anthropic.com` on a non-`anthropic.com` hostname loses native tool search with this change and falls back to local `tool_search` (no request failure). If that should be opt-in-able, a `compat` flag on the model config would be the natural follow-up; I kept this PR to the minimal fix.

---
## Summary by cubic
Fixes the Anthropic native tool-search adapter so requests no longer fail with 400 errors. The injected tool now uses the API-required name `tool_search_tool_bm25`, and injection only runs on first-party `anthropic.com` hosts; third-party hosts now fall back to the local `tool_search` tool.

**Bug Fixes**
- Renames the native search tool to `tool_search_tool_bm25`; the API rejects any other name with a 400.
- Skips injection for `anthropic-messages` models on third-party hosts like Kimi Code and OpenRouter, which don't implement native tool search.
- Models without a `baseUrl` keep the previous behavior so existing callers and tests are unaffected.
- The test validator now enforces the tool name like the API does, with new tests covering the name, host gating, and third-party 400 attribution.

<sup>Written for commit 475598d103961db089bf8a0ff2e61f3aa2e7322e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/extensions/builtin/tool-search/changes.md
+++ b/packages/coding-agent/src/core/extensions/builtin/tool-search/changes.md
@@ -1,5 +1,24 @@
 # Tool Search Builtin Changes
 
+## 2026-09-04 - Fix the native search tool name and gate injection to first-party Anthropic hosts
+
+### What changed
+
+- `ANTHROPIC_TOOL_SEARCH_NAME` is now `tool_search_tool_bm25`, the only `name` the Messages API accepts for `tool_search_tool_bm25_20251119`; the previous `tool_search` was rejected with `400 tools.N.tool_search_tool_bm25_20251119.name: Input should be 'tool_search_tool_bm25'`.
+- `AnthropicNativeToolSearchAdapter.applyBeforeRequest` takes the request model (`event.model ?? ctx.model`) instead of the bare `api` string and skips injection unless the model's `baseUrl` host is `anthropic.com` (or a subdomain). Third-party endpoints that speak the Anthropic Messages wire format (Kimi Code, OpenRouter, proxies) do not implement native tool search and answered the injected tool with an opaque `400 Invalid request Error`; they now receive the untouched payload and keep the local `tool_search` tool.
+- A missing `baseUrl` keeps the previous behaviour so existing callers and tests stay unchanged.
+- The request-validator mock now enforces the tool `name` the same way the API does, and `test/tool-search/native-anthropic.test.ts` covers the name, the host gate, and that a third-party 400 is not attributed to native search.
+
+### Why
+
+- The eval-only tool routing default (`bash`/`workflow`/`monitor` inactive) made the extension catalog non-empty for every session, which turned the native adapter on for every `anthropic-messages` request and surfaced both defects at once: every Anthropic request failed with the name error, and every Kimi Code request failed with the opaque 400.
+- The 400 fallback only helps after the first failed turn of each session, and on third-party hosts it hid the real cause behind a generic error.
+
+### Expected merge conflict zones
+
+- LOW: `native-search.ts` constants and the adapter entry point.
+- LOW: `index.ts` provider-request hook wiring.
+
 ## 2026-08-11 - Defer local tool registration until the catalog is searchable
 
 ### What changed
--- a/packages/coding-agent/src/core/extensions/builtin/tool-search/index.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/tool-search/index.ts
@@ -46,7 +46,9 @@ export function createToolSearchExtension(service: ToolSearchService): Extension
 			},
 			searchToolName: TOOL_SEARCH_TOOL_NAME,
 		});
-		pi.on("before_provider_request", (event, ctx) => nativeAdapter.applyBeforeRequest(ctx.model?.api, event.payload));
+		pi.on("before_provider_request", (event, ctx) =>
+			nativeAdapter.applyBeforeRequest(event.model ?? ctx.model, event.payload),
+		);
 		pi.on("after_provider_response", (event) => nativeAdapter.noteResponseStatus(event.status));
 	};
 }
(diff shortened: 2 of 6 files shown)


code-yeongyu:
Both defects this PR describes are now fixed on `main` and released, so this branch (currently `DIRTY` against `main`) is superseded:

1. **Wrong tool `name`** — fixed by `d0f197eb9` ("name the Anthropic native tool-search tool tool_search_tool_bm25"), released in `2026.9.4`. Verified on the published artifact: `dist/core/extensions/builtin/tool-search/native-search.js` exports `ANTHROPIC_TOOL_SEARCH_NAME = "tool_search_tool_bm25"` with `ANTHROPIC_TOOL_SEARCH_TYPE = "tool_search_tool_bm25_20251119"`.
2. **Injected on every `anthropic-messages` endpoint** — fixed by #1354 (merged as `ec595f7e6`, released in `2026.9.4-3`), which replaced the api-only gate with `supportsAnthropicNativeToolSearch()` in the new `builtin/tool-search/native-support.ts`. It mirrors pi-ai's private `supportsToolReferences` default: injection requires `api === "anthropic-messages"` **and** an Anthropic-provider model on the Opus/Sonnet 4.5+ or Fable line, so gateways/proxies (Kimi Code, OpenRouter, ccapi) no longer receive the server tool. Executed against the published `2026.9.4-3` tarball: 9/9 gate cases correct, including a gateway provider on `anthropic-messages` returning `false`.

#1354 also fixed a third defect this PR does not cover: `buildToolReferenceBlocks()` emitted `tool_reference` blocks under `name` instead of the API's `tool_name`, and the mock fixture encoded the same bug so both expansion tests passed over it.

Closing is the owner's call — flagging the overlap so this does not get rebased onto work that already landed.

code-yeongyu:

I'm closing it as **superseded**: both defects landed on `main` separately: the tool name via `d0f197eb9` (*name the Anthropic native tool-search tool tool_search_tool_bm25*, released in 2026.9.4) and the endpoint gate via PR #1354 (`ec595f7e6`, released in 2026.9.4-3) — as the maintainer laid out in this thread on 2026-09-04.

**How I checked:** `git merge-base --is-ancestor d0f197eb9 origin/main` → yes; `native-search.ts` on `main` exports `ANTHROPIC_TOOL_SEARCH_NAME = "tool_search_tool_bm25"`.

PR #1338, opened 2026-09-04
fix(coding-agent): skip context-incompatible fallbacks

DEV-PROLL:
## Summary
- preflight retry-fallback candidates against the live conversation budget and continue past incompatible rungs
- make automatic fallback model changes transactional so rejected switches leave no persisted model change or stale prompt/tool state
- expose a bounded, non-blocking typed retry_fallback_exhausted extension event for fresh-context recovery

## Verification
- 8 focused fallback/model-budget/RPC suites: 54 tests passed
- npm run check
- isolated mock-loop budget exhaustion QA with zero provider calls
- RPC and xterm.js TUI fallback QA

## Evidence
- local-ignore/qa-evidence/20260904-context-fallback/README.md (local reviewer artifact; not committed)

No real provider credentials or paid calls were used.

---
## Summary by cubic
Fixes retry fallback so it skips context-incompatible model rungs and only settles when no remaining rung can hold the live conversation. Automatic fallback model switches are now transactional — a rejected switch leaves no persisted model change or stale prompt/tool state, and a failed fallback settles the retry lifecycle instead of wedging `prompt()`.

**What changed**

- Fallback preflights each candidate against the live context budget and skips incompatible rungs instead of switching onto a model that can't hold the conversation.
- Automatic fallback switches commit only after the post-`model_select` budget check passes; on rejection, session state is restored and nothing is persisted. Persistence is the commit boundary — `SessionManager` persists before exposing an in-memory entry, initial session files are written atomically, and failed appends truncate back.
- Retry handling and observers are failure-isolated: a throwing session or extension listener can't block settlement, post-commit observer failures still record the applied rung, and any internal retry-handler failure resolves the retry instead of wedging `prompt()`.
- New `retry_fallback_exhausted` extension event exposes diagnostics (`sessionId`, `chainKey`, `from`, `lastError`, `exhaustionReason`, `rejectedCandidates`) so an extension can route work to a fresh context; it is notification-only, never blocks retry settlement, and its payload truncates each string field and candidate error to stay under 64 KiB of serialized UTF-8.

<sup>Written for commit d3dab962bd5741f88f3716a0545b39f7533da2d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1338?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DEV-PROLL:
@code-yeongyu This is ready to merge. All repository checks are green; independent code and security reviews both passed; `npm run check` passed; the focused fallback/session regression set passed (13 files, 144 tests); isolated mock-loop/RPC/TUI evidence used no provider credentials or paid calls. The authenticated PR author account has read-only permission on the upstream repository, so it cannot enable auto-merge or satisfy the base-branch merge policy itself. Please approve and merge when available.

DEV-PROLL:
@code-yeongyu Update: this branch now includes current `main` (`5c4c1702d`) at head `d3dab962b`. The post-merge focused fallback suite and full `bun run check` are green, and the PR is GitHub-mergeable. The dependent OMO PR code/review/QA is also green and is waiting for this Senpi release before updating its pin. Please merge and publish when available.

code-yeongyu:
Cross-link: #1873 tracks replacing the binary model-switch admission gate with a three-tier admission (`fits-now` / `fits-after-compaction` / `impossible`) plus deferred compaction.

For the fallback path this **supersedes the skip policy** in this PR. Instead of preflighting a candidate and continuing past a context-incompatible rung, #1873 item 4 admits the rung by reducing the context with the existing deterministic, provider-free slice (`planResumeSlice`, sized to the target model's projection) and then switching onto it. The reason is that skipping fails a chain whose rungs are all smaller than the current model, which is exactly the case a fallback chain exists for; and the model being fallen back *from* has usually just failed, so it cannot be asked for a summary either.

The transactional-switch half of this PR (a rejected switch leaving no persisted model change or stale prompt state) overlaps with the ordering #1526 already landed, and #1873 item 6 keeps that contract.

Not closing this - flagging the overlap so the two do not land contradictory fallback policies.

code-yeongyu:
#1873's fallback half landed as #1880, which takes the opposite approach to this PR for the same situation: instead of preflighting a candidate and skipping a context-incompatible rung, it reduces the transcript with the existing deterministic, provider-free slice (`planResumeSlice`, sized to that rung's projection) and then takes the rung.

Reasoning, for the record: skipping fails a chain whose rungs are all smaller than the model that just failed - which is the case a fallback chain exists for - and the model being fallen back from has usually just failed, so it cannot be asked for a summary either. The slice is the only reduction available on that path, and the full transcript stays in the session file. A rung with no interior boundary to cut, or a session with compaction disabled, still refuses.

The transactional-switch half of this PR is already covered: #1526 landed the ordering, and #1876 preserves it (nothing durable is written while a switch is held, and a switch that lands supersedes the hold).

Flagging rather than closing - the call on this PR is yours.

code-yeongyu:

I'm closing it because the same situation was resolved on 2026-09-20 by merged PR #1880 (*fix(retry-fallback): repair a context-incompatible fallback rung instead of rejecting it*), which deliberately takes the opposite approach — reduce the transcript and take the rung, rather than skip it — as the maintainer explained in this thread the same day.

**How I checked:** `gh pr view 1880` → `MERGED 2026-09-20T09:48:07Z`; last comment on this PR is the maintainer's 2026-09-20 rationale.

PR #1378, opened 2026-09-05
fix(coding-agent): skip context-incompatible Ctrl+P models

code-yeongyu:
## Summary
Ctrl+P favorite-model cycling now skips targets that cannot admit the current context under the existing model usability budget. Each skipped target emits a structured session event, and the interactive TUI explains the skip; the desktop transport receives the event through the existing session-event forwarding path.

## Changes
- Reuse `projectModelUsabilityBudget` as the single admission predicate before cycling into a favorite.
- Emit `model_change_skipped` with the target budget projection and continue in the requested direction.
- Return an explicit all-skipped result so the TUI recommends compacting or starting a new session instead of reporting a misleading single-favorite status.
- Add faux-provider regressions for one skipped candidate and all candidates skipped.

## QA & Evidence
- **What was tested:** `bun scripts/build-all.mjs` and `bunx vitest run packages/coding-agent/test/suite/agent-session-model-extension.test.ts` on `mengmotaMac` through bunshin.
- **Observed result:** build exit 0; 1 test file and 20 tests passed; remote temporary tree cleanup exit 0.
- **Artifact:** `local-ignore/qa-evidence/20260905-ctrl-p-context-skip/README.md`
- **Why sufficient:** covers the real AgentSession cycle path, event emission, candidate skipping, all-skipped result, and compile-time transport acceptance.

## Risks & Residuals
The desktop application itself is unchanged; the event is additive on the existing session event stream. No provider calls or credentials are used by the regression.

---
## Summary by cubic
Fixes Ctrl+P favorite-model cycling so models that can't fit the current context are skipped instead of causing a failed switch.

- Reuses the existing usability budget projection before cycling into a favorite.
- Emits a `model_change_skipped` event per rejected candidate and continues cycling in the requested direction.
- Returns an all-skipped result; the TUI warns per skip and suggests compacting or starting a new session when none fit.
- The desktop app receives the event via existing forwarding; no desktop changes. Adds regression tests for one skipped and all skipped.

<sup>Written for commit 3091e7d6f4cb1df3081ab56daa1385c1e4e346d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1378?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/local-ignore/qa-evidence/20260905-ctrl-p-context-skip/README.md
+++ b/local-ignore/qa-evidence/20260905-ctrl-p-context-skip/README.md
@@ -1,0 +1,32 @@
+# Ctrl+P context-window skip QA
+
+## What was tested
+
+- On `mengmotaMac` through bunshin, transferred this worktree to an isolated
+  `/tmp/ctrl-p-context-skip-*` directory.
+- Ran `bun install --frozen-lockfile`.
+- Ran `bun scripts/build-all.mjs`.
+- Ran
+  `bunx vitest run packages/coding-agent/test/suite/agent-session-model-extension.test.ts`.
+
+## What was observed
+
+- Build exited 0.
+- The focused suite passed: 1 test file, 20 tests.
+- The regression covers both an exhausted favorite being skipped before the next
+  usable favorite and all alternative favorites being rejected with a structured
+  result.
+- The isolated remote tree was removed after the run with exit 0.
+
+## Why this is enough
+
+- The test exercises `AgentSession.cycleModel()` with the real faux-provider
+  harness and observes the emitted `model_change_skipped` events and selected
+  model.
+- The build proves the new event is accepted by the coding-agent and transport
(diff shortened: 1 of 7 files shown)


code-yeongyu:
Cross-link: #1873 narrows the skip introduced here. Under the three-tier admission proposed there, Ctrl+P skips only candidates that can never serve (`impossible` - the window cannot hold the system prompt and tool schemas at all). A candidate that one compaction would make usable lands as a *pending* switch and is repaired on the next send, instead of being silently skipped.

PR #1394, opened 2026-09-05
fix(coding-agent): avoid project trust read contention

### LINKED ISSUE #1393: fix: project trust reads fail on concurrent writer lock
## Summary

A concurrent Senpi process can terminate at startup with the raw `Error: Lock file is already being held` while reading an already-persisted project trust decision.

## Deterministic reproduction

1. Use an isolated `SENPI_CODING_AGENT_DIR` with a known `trust.json` decision.
2. Hold `<agentDir>/trust.json.lock` with `proper-lockfile`.
3. Start Senpi in a project containing trust-requiring resources.
4. `ProjectTrustStore.getEntry()` attempts to acquire the writer lock and exits with `ELOCKED` before extensions or the primary turn can run.

## Expected

Readers consume the last atomically published trust snapshot without taking the writer lock. Writers remain serialized, and malformed snapshots continue to fail closed.

## Scope

`packages/coding-agent/src/core/trust-manager.ts` and deterministic regression coverage. No change to trust decisions or project-trust defaults.

JeremyDev87:
## What changed

- Removed writer-lock acquisition from project trust reads.
- `trust.json` writes keep the existing proper-lockfile serialization but publish through a same-directory temp file and an atomic rename.
- Added a regression test that reads the stored trust decision while a real writer lock is held.

## Why

When several Senpi/OMO processes run concurrently, a read-only project trust lookup could contend with the writer lock and kill the process with `Lock file is already being held`. A completed previous snapshot is safe to read, so there is no reason for a reader to wait on the writer lock.

## How

The reader reads the last published snapshot directly and validates it with the existing strict parser. The writer performs read-modify-write inside the lock, writes complete JSON to a temp file, and renames it. The fail-closed semantics for malformed snapshots and for the trust default are unchanged.

## Verification

- `npm --prefix packages/coding-agent test -- test/suite/regressions/project-trust-lock-contention.test.ts test/trust-manager.test.ts` PASS (3/3)
- `bun run check` PASS
- The source CLI `--list-models` exits 0 while a real proper-lockfile writer holds `trust.json.lock`
- Senpi CLI smoke PASS (8/8), real auth files unchanged

## Out of scope

Credential/settings lock policy and the trust decision itself are unchanged. The 18 tool-protocol leak scenarios in the full mock-loop self-test are pre-existing failures unrelated to this path and were left as-is.

Closes #1393

---
## Summary by cubic
Fixes project trust reads retrying the writer lock synchronously and throwing `ELOCKED`, which could kill concurrent processes. Reads now use the last atomically published `trust.json` snapshot without acquiring the lock, while writes keep proper-lockfile serialization and publish via temp file plus atomic rename.

- Adds a regression test that reads trust while a real writer holds the lock.
- Updates `changes.md` with the change details and expected merge conflict zones.

<sup>Written for commit f20ffa912165baf2d7db8e789a6c8df82cb39df4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1394?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/src/core/changes.md
+++ b/packages/coding-agent/src/core/changes.md
@@ -41,6 +41,24 @@
 
 # changes
 
+## 2026-09-05 - Read project trust without writer-lock contention
+
+### What changed
+
+- `packages/coding-agent/src/core/trust-manager.ts`: project-trust reads load the last published `trust.json` snapshot without acquiring the writer lock; writes remain serialized by proper-lockfile and now publish through a same-directory temporary file plus atomic rename.
+
+### Why
+
+- A trust read during another process's write previously retried the writer lock synchronously and then threw `ELOCKED`, interrupting startup and session-state projection even though the prior complete trust snapshot was safe to read.
+
+### Why an extension could not handle it
+
+- Project trust is loaded by core startup, package-command, interactive, and RPC paths before or below extension interception; only the store can separate snapshot reads from serialized read-modify-write publication.
+
+### Expected merge conflict zones
+
+- LOW: `packages/coding-agent/src/core/trust-manager.ts` read and write helpers.
+
 ## 2026-09-05 - Require explicit fallback chains
 
 ### What changed
--- a/packages/coding-agent/src/core/trust-manager.ts
+++ b/packages/coding-agent/src/core/trust-manager.ts
@@ -1,4 +1,5 @@
-import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { randomUUID } from "node:crypto";
+import { existsSync, mkdirSync, readFileSync, renameSync, rmSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { dirname, join } from "node:path";
 import lockfile from "proper-lockfile";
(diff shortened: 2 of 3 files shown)


JeremyDev87:
Requesting review.

- RED: with a real `proper-lockfile` writer holding `trust.json.lock`, the existing `ProjectTrustStore.get()` raises `ELOCKED`
- GREEN: after the lock-free snapshot read, the same scenario returns the stored decision
- Focused tests 3/3 PASS
- `bun run check` PASS
- The source CLI `--list-models` exits 0 while the live writer lock is held
- CLI smoke 8/8 PASS, real auth unchanged
- Independent code review: APPROVE

The change is limited to removing the reader lock and publishing writes via a same-directory temp file plus rename; the trust decision and malformed-input fail-closed contracts are unchanged.

PR #1397, opened 2026-09-05
feat(tui,coding-agent): show working/idle agent status in terminal tabs

MoerAI:
## What

The interactive mode prefixes the terminal title with `[working]` on `agent_start` and `compaction_start`, restoring the plain title on `agent_end`, `compaction_end`, and `stop`.

The token leads the title so it survives tab truncation. Idle uses the plain title. Updates are static and event-driven.

Scope: `packages/coding-agent`. The existing OSC title transport is unchanged.

## Tests

The regression tests exercise `handleEvent` for agent start/end, compaction start/end, compaction during streaming, and `stop()`. They also cover title composition, repeated transitions, and preservation of `process.title`.

Removing the event wiring makes 6 of the 9 regression tests fail.

Verification:

- Coding-agent tests passed with no failures.
- TUI tests: 735 passed.
- `npm run check`: passed.

## QA evidence

Artifacts: `local-ignore/qa-evidence/20260906-terminal-tab-agent-status/`.

The real senpi TUI ran in a node-pty terminal against a local fake model server, in an isolated sandbox without paid model calls. The real `~/.senpi/agent/auth.json` hash was unchanged.

The driver subscribes to PTY output. The fake server holds its response until the working title is observed, then releases the response so completion and return to idle can be checked without fixed sleeps or polling.

Observed OSC title transitions:

```text
senpi - work
[working] senpi - work
senpi - work
```

`results.json`: `{"passed":7,"total":7,"allPassed":true}`.

This verifies the emitted title sequence through the real CLI and PTY. Individual terminal applications were not separately driven.

## No regression

Title precedence remains active tool hook > tool execution > extension > normal title. The working token is prepended without changing the title content. The existing terminal progress setting and OSC title transport remain unchanged.


MoerAI:
Closing this feature proposal rather than carrying a large, conflicting non-bugfix PR without maintainer confirmation. The current patch changes 325 lines across five files, and its scope is OSC terminal titles, not Zed's foreground-process tab labels. Happy to rework a smaller proposal if this behavior is wanted.

PR #1418, opened 2026-09-06
fix(coding-agent): resolve external-dir paths without opening them (fixes #1416)

### LINKED ISSUE #1416: permission-system: external-dir realpathSync walk blocks the host main thread on autofs/FIFO paths (TUI freeze)
## Symptom

An interactive omo/senpi session (omo-ai 5.0.0-0.beta.44, senpi 2026.9.5-3, bun 1.4.0, macOS 26.6.1) froze completely: spinner stopped, input dead, an eval js cell stayed "running · 55s" for 40+ minutes, no foreground-window detach, no timeout, no notification. The cell had just called `tool.monitor({ command: "sleep 45; until r=$(box exec bx_… -- bash /home/user/work/poll-fdl.sh | jq -r .stdout); [ \"$r\" != RUNNING ]; do sleep 20; done; echo FDL_DONE", filter: "^FDL_DONE" })` — the command text mentions a path on a **remote Linux box** (`/home/user/work/...`).

## Root cause (evidence-backed)

`sample <pid>` of the frozen host: the **main thread** is parked in `openat$NOCANCEL` (100 % of samples); both JS kernel Worker threads idle in `kevent64`. Unified log at the freeze instant (01:51:27 KST): `automountd` runs a burst of Open Directory queries for the `auto_home` map and `tccd` records a `kTCCServiceSystemPolicyAllFiles` preflight for the senpi `bun` pid — i.e. the host process touched `/home`, the autofs trigger directory.

Code path: the permission-system `tool_call` hook parses every `bash`/`monitor` command through `extractExternalPaths()` (`permission-system/external-dir.ts`). For each path-looking token it calls `isExternalPath()` → `normalizePath()`, which walks a non-existent path **up to its nearest existing ancestor with synchronous `fs.realpathSync`**:

`/home/user/work/poll-fdl.sh` → ENOENT → `/home/user/work` → ENOENT → `/home/user` → ENOENT → **`realpathSync("/home")`**.

Bun implements `fs.realpathSync`, `realpathSync.native` **and** `fs.promises.realpath` by `open(2)`-ing the path (F_GETPATH) instead of lstat/readlink like Node. Opening an autofs trigger directory forces the automount; on this host the `auto_home` map is wedged, so the open never returns. Reproduced outside senpi:

```
timeout -s KILL 5 bun -e 'require("fs").realpathSync("/home")'          # killed (hang)
timeout -s KILL 5 bun -e 'require("fs").realpathSync.native("/home")'   # killed (hang)
timeout -s KILL 5 bun -e 'await require("fs").promises.realpath("/home")' # killed (hang)
timeout -s KILL 5 node -e 'console.log(require("fs").realpathSync("/home"))'  # /System/Volumes/Data/home (fast)
timeout -s KILL 5 bun -e 'require("fs").lstatSync("/home")'            # fast
```

Because the hook runs on the host main thread, the whole TUI, every timer (eval foreground window, monitor deadlines), and the codemode bridge stop. Any command that merely *mentions* `/home/...`, `/net/...`, a stalled network volume, or a FIFO path is a landmine — the classifier must never open what it classifies.

## Fix

Resolve symlinks in `normalizePath()` with a component walker built on `lstat` + `readlink` (what realpath(3) does) and never `open(2)` a component; non-existent trailing components are kept verbatim after the last resolvable one, so symlinked-cwd semantics are unchanged.

## Environment
- mengmotaHost (Mac mini M4 Pro), macOS 26.6.1, bun 1.4.0, herdr pane w37:p2, session 01a07736, host pid 14091
- Stuck cell: session 01a07736 tool call `toolu_012uWrtNLerF4CRPubjqJyun` (2026-09-06T16:50:31Z), freeze at 16:51:27Z

MoerAI:
## Summary

`normalizePath()` in the permission system's external-directory classifier resolved symlinks with `fs.realpathSync`. Under Bun that is implemented with `open(2)`/`F_GETPATH`, so classifying a command that merely *mentions* an autofs trigger (`/home`, `/net`), a stalled network mount, or a FIFO forces an automount and can block forever. The hook runs on the host main thread, so the whole TUI freezes.

This replaces that walk with a `realpath(3)`-style component walker built on `lstat` + `readlink`, which never opens the entry it inspects.

Fixes #1416.

## Root cause

The `tool_call` hook classifies every `bash` / `monitor` path token through `extractExternalPaths()` -> `isExternalPath()` -> `normalizePath()`. For a path that does not exist, `normalizePath()` walked up to the nearest existing ancestor, calling `fs.realpathSync` at each level:

`/home/user/work/poll-fdl.sh` -> ENOENT -> `/home/user/work` -> ENOENT -> `/home/user` -> ENOENT -> `realpathSync("/home")`

`lstat` and `readlink` are enough to implement `realpath(3)` and never open a component, so the classifier no longer touches what it classifies.

## Changes

| File | Change |
|------|--------|
| `packages/coding-agent/src/core/extensions/builtin/permission-system/external-dir.ts` | `normalizePath()` resolves symlinks with an `lstat` + `readlink` component walker (40-hop cycle guard) instead of `fs.realpathSync` |
| `packages/coding-agent/test/permission/external-dir.test.ts` | Regression coverage: no `realpathSync`/`statSync`/`existsSync` during classification, plus symlink, symlink-chain, symlink-cycle and non-existent-tail behaviour |
| `packages/coding-agent/CHANGELOG.md` | `[Unreleased] > Fixed` entry |

Observable classification is unchanged by design: non-existent trailing components are kept verbatim after the last resolvable component, and any non-ENOENT failure falls back to the unresolved path, exactly as the previous `realpathSync` version did. The existing symlinked-cwd case still passes.

## Reproduction (before fix)

`node:fs` is swapped for a stub whose `realpathSync` blocks forever for `/home*` — a stand-in for the wedged `auto_home` map in the issue, since Bun's `realpathSync` never returns there.

```
$ bun probe.ts before   # external-dir.ts taken from origin/main
[probe] variant=before classifying: bash /home/user/work/poll-fdl.sh
[stub] realpathSync("/home/user/work/poll-fdl.sh") -> wedged automount, blocking forever
exit_code=137   # SIGKILL after the 5s deadline -> the classifier froze
```

The unit seam shows the same thing — one classification issued 15 `realpathSync` calls:

```
AssertionError: expected [ 'realpathSync', …(14) ] to deeply equal []
 ❯ test/permission/external-dir.test.ts > path resolution never opens the classified path
   > classifies paths without realpathSync, statSync, or existsSync
```

## Verification (after fix)

```
$ bun probe.ts after    # external-dir.ts from this branch
[probe] variant=after classifying: bash /home/user/work/poll-fdl.sh
CLASSIFIED ["/home/user/work/poll-fdl.sh"] in 23ms
exit_code=0
```

```
$ bun run --cwd packages/coding-agent test test/permission
 Test Files  15 passed (15)
      Tests  485 passed (485)

$ bun run check
CHECK_EXIT=0
```

Real-CLI QA (`.agents/skills/senpi-qa`), each asserting `~/.senpi/agent/auth.json` is unchanged:

```
common.mjs   --self-check          : 10/10 passed
cli-smoke.mjs --self-test          :  8/8  passed
rpc-drive.mjs --self-test          :  4/4  passed
mock-loop.mjs --with-tool          :  4/4  passed   # model -> bash tool -> final text
```

`mock-loop --with-tool` is the relevant end-to-end channel: it drives a real agent turn through the `bash` tool, which is exactly the hook that calls this classifier.

Evidence captured under `local-ignore/qa-evidence/20260907-1416-external-dir-lstat-walker/` (gitignored).

## Test

- Regression test: `packages/coding-agent/test/permission/external-dir.test.ts` (5 new cases)
- Related suite: `bun run --cwd packages/coding-agent test test/permission` — 485 pass
- Static gate: `bun run check` — clean

## Not in scope

`permission-system/parsers.ts` has a sibling `realpathSync(dirname(resolve(cwd, path)))` on the `monitor` parser. It is the same class of landmine, but it backs the authoritative canonical-parent check, so changing it is a security-semantics decision rather than a mechanical swap. Left untouched here; happy to follow up if you want it moved to the same walker.

---
## Summary by cubic
Fixes the external-directory classifier freezing the TUI when a command merely mentions an autofs trigger like `/home`, a stalled network mount, or a FIFO.

`normalizePath()` now resolves symlinks with an `lstat` + `readlink` component walker instead of `fs.realpathSync`, so classification never opens the path it inspects. Observable classification behavior is unchanged: non-existent trailing components stay verbatim and non-ENOENT failures fall back to the unresolved path.

**Bug Fixes**
- The component walker now splits on `path.sep` only, so a backslash in a posix path is treated as a filename character and no longer lets paths like `/tmp/project\outside/file` be classified as inside `/tmp/project`.
- Adds regression tests covering symlink chains, cycles, non-existent tails, and backslash handling, and verifies no `realpathSync`, `statSync`, or `existsSync` calls during classification.
- Updates the `coding-agent` changelog with a `Fixed` entry.

<sup>Written for commit 80c338022ca281f2eae8c1fdf8c970fcc8841111. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- The permission system's external-directory classifier now resolves symlinks with `lstat` + `readlink` instead of `realpathSync`, so classifying a command that merely mentions an autofs trigger such as `/home`, a stalled network mount, or a FIFO can no longer block the host main thread and freeze the TUI ([#1416](https://github.com/code-yeongyu/senpi/issues/1416))
+
 ### Removed
 
 ## [2026.9.6] - 2026-09-06
--- a/packages/coding-agent/src/core/extensions/builtin/permission-system/external-dir.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/permission-system/external-dir.ts
@@ -18,21 +18,61 @@ export function expandHome(inputPath: string): string {
 	return inputPath;
 }
 
+const MAX_SYMLINK_HOPS = 40;
+
+function splitSegments(value: string): string[] {
+	return value.split(/[\\/]+/).filter((segment) => segment.length > 0);
+}
+
+/**
+ * realpath(3) equivalent that resolves symlinks with lstat + readlink only.
+ *
+ * This runs on the host main thread for every classified path, so it must never
+ * open(2) a component: Bun implements realpathSync (and its native variant) with
+ * open(), which forces an automount for autofs triggers such as /home or /net and
+ * blocks forever on a wedged map, freezing the whole TUI. lstat and readlink never
+ * open the entry they inspect. Trailing components that do not exist are kept
+ * verbatim after the last resolvable one, and any other failure falls back to the
+ * unresolved path, matching the previous realpathSync-based behaviour.
+ */
 function normalizePath(inputPath: string): string {
-	let candidate = path.normalize(inputPath);
-	const suffix: string[] = [];
-	for (;;) {
-		try {
-			const resolved = fs.realpathSync(candidate);
-			return suffix.length === 0 ? resolved : path.join(resolved, ...suffix.reverse());
-		} catch (error) {
-			if ((error as NodeJS.ErrnoException)?.code !== "ENOENT") return candidate;
-			const parent = path.dirname(candidate);
-			if (parent === candidate) return candidate;
-			suffix.push(path.basename(candidate));
-			candidate = parent;
+	const normalized = path.normalize(inputPath);
+	const root = path.parse(normalized).root;
+	const pending = splitSegments(normalized.slice(root.length));
+	let resolved = root;
+	let hops = 0;
+
+	while (pending.length > 0) {
+		const segment = pending.shift();
+		if (segment === undefined || segment === ".") continue;
+		if (segment === "..") {
+			resolved = path.dirname(resolved);
+			continue;
 		}
+
+		const candidate = path.join(resolved, segment);
+		let linkTarget: string | undefined;
+		try {
+			const entry = fs.lstatSync(candidate, { throwIfNoEntry: false });
+			if (!entry) return path.join(candidate, ...pending);
+			if (entry.isSymbolicLink()) linkTarget = fs.readlinkSync(candidate);
+		} catch {
+			return normalized;
+		}
+
+		if (linkTarget === undefined) {
+			resolved = candidate;
+			continue;
+		}
+		if (++hops > MAX_SYMLINK_HOPS) return normalized;
+
+		const target = path.normalize(linkTarget);
+		const targetRoot = path.parse(target).root;
+		if (targetRoot.length > 0) resolved = targetRoot;
+		pending.unshift(...splitSegments(target.slice(targetRoot.length)));
 	}
+
+	return resolved.length > 0 ? resolved : normalized;
 }
 
 export function isExternalPath(inputPath: string, cwd: string): boolean {
--- a/packages/coding-agent/src/core/extensions/builtin/permission-system/external-dir.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/permission-system/external-dir.ts
@@ -20,8 +20,8 @@ export function expandHome(inputPath: string): string {
 
 const MAX_SYMLINK_HOPS = 40;
 
-function splitSegments(value: string): string[] {
-	return value.split(/[\\/]+/).filter((segment) => segment.length > 0);
+function splitNormalizedSegments(value: string): string[] {
+	return value.split(path.sep).filter((segment) => segment.length > 0);
 }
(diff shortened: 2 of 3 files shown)


MoerAI:
Closing as superseded by #1419. Issue #1416 is closed, and current main uses the shared `realpathWithoutOpen` walker for external-directory classification. Keeping a second implementation would duplicate the landed fix. Happy to rework if a remaining regression is identified.

PR #1420, opened 2026-09-06
fix(coding-agent): complete fresh-profile Windows supervisor bootstrap (fixes #1370)

### LINKED ISSUE #1370: Windows: RPC host supervisor crashes during bootstrap (non-recursive mkdir for rpc-host-daemon, then ENOENT read of rpc.sock.secret)
### Environment

- omo 5.0.0-0.beta.40 (engine: senpi 2026.9.4), installed via `bun install -g omo-ai@beta` → `C:\\Users\\runneradmin\\.bun\\bin\\omo.exe` (real exe, not a shim)
- windows-latest GitHub runner (bun 1.4.1)
- Context: we are adding Windows support to a Go server that drives the omo CLI RPC socket, and CI live-probes the socket host on every PR. The probe spawns the supervisor with an isolated temp USERPROFILE/HOME so the runner profile stays untouched. Probe source: [test/windowsprobe in our repo](https://github.com/DevNewbie1826/omo-webchat/blob/ulw/win1-compile/test/windowsprobe/main.go).

### What happens

On a fresh Windows profile the RPC host supervisor never reaches listening; it crashes during bootstrap in two successive stages:

**Stage 1 — non-recursive mkdir for the internal socket dir** (hit first):
```
ENOENT: no such file or directory, mkdir 'C:\...\<agentDir>\rpc-host-daemon\internal-852387bb-...'
    at async createInternalSocketPath (.../@code-yeongyu/senpi/dist/modes/rpc/host-lifecycle.js:78:15)
    at async runHostSupervisor (.../dist/modes/rpc/host-lifecycle.js:287:28)
```
`mkdir(dir, { recursive: false })` on `<agentDir>/rpc-host-daemon/internal-<uuid>` requires the parent `rpc-host-daemon` to already exist. On macOS the same code path puts the internal socket directly in the OS tmpdir (fresh dir, parent exists), which is why a fresh profile works there but not on Windows.

**Stage 2 — ENOENT read of the socket secret** (hit after the probe pre-created `rpc-host-daemon` to get past stage 1):
```
ENOENT: no such file or directory, open 'C:\...\<agentDir>\rpc\rpc.sock.secret'
    at readSocketSecret (.../@code-yeongyu/senpi/dist/modes/rpc/socket-transport.js:35:26)
supervisor exited with status 1
```
The secret file next to the socket path was never present when `readSocketSecret` ran (either never written, written to a different location, or read before the write completed).

### Repro

1. Fresh Windows env; `bun install -g omo-ai@beta`.
2. Point USERPROFILE and HOME at an empty temp dir.
3. Pre-create `<agentDir>/rpc` and `<agentDir>/rpc-host-daemon` (0700).
4. Run `omo --internal-rpc-host-supervisor --socket <agentDir>/rpc/rpc.sock --agent-dir <agentDir>` (mirror our probe for the exact invocation).
5. Observe exit 1 with the traces above. The same probe on macOS passes end to end (socket appears; a Go client dials it and completes a get_protocol_info round-trip).

### Evidence

- Stage 1 run: https://github.com/DevNewbie1826/omo-webchat/actions/runs/33870188086 (job probe-windows)
- Stage 2 run: https://github.com/DevNewbie1826/omo-webchat/actions/runs/33870954338 (job probe-windows)
- Passing macOS run of the identical probe: local, output `G1_AF_UNIX=PASS {"id":"windowsprobe-1","type":"response","command":"get_protocol_info","success":true,...}`

### Expected

The supervisor comes up and listens on the requested socket path on a fresh Windows profile.

### Suggested directions

- Use recursive mkdir when creating the internal socket directory under `rpc-host-daemon`.
- Ensure the socket secret is written (to the path that will be read) before `readSocketSecret` runs, and surface the write error rather than failing on the read.

Happy to re-run our probe against a fixed build — the CI job exercises this on every PR.

MoerAI:
## Summary

Fix both fresh-profile Windows failures in the direct `--internal-rpc-host-supervisor` route: the missing internal socket directory and the missing public socket secret.

## Contract

The supervisor ensures `<publicSocket>.secret` before allocating the internal hop or spawning the child. It reuses an existing valid 32-byte secret, including one prepared by `ensureHost()`. A missing secret is generated with its parent directories and mode `0600`. Provisioning errors identify the step and path. The public endpoint still requires the secret handshake.

The internal socket directory is created recursively on Windows. POSIX transport behavior is unchanged.

## Reproduction and coverage

Before the secret fix, a regression invoking the production Windows bootstrap branch failed with `ENOENT` for the missing public secret before reaching child spawn. The same case now passes; a companion case verifies an existing secret is unchanged.

A subprocess regression invokes the real CLI's hidden supervisor route without `ensureHost()`, with an isolated profile and no public secret. It waits for the supervisor readiness event, connects to the public endpoint, sends the platform-required handshake, and verifies a successful `get_protocol_info` response. The helper tests remain supplemental.

## Verification

- `npm run check` and `npm run build`.
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/1370-rpc-internal-socket-mkdir.test.ts test/rpc-host-lifecycle.test.ts`, from `packages/coding-agent`.
- Real CLI RPC `get_state` and a deterministic mock-provider tool turn: 4/4 assertions each; real auth file unchanged.
- Local evidence: `local-ignore/qa-evidence/20260910-pr1420-bootstrap/` in the PR worktree.

Validation host: macOS. The subprocess exercised the real POSIX route locally. Windows bootstrap secret handling was exercised through the production supervisor with its process platform stubbed up to child spawn. The same subprocess test includes Windows named-pipe authentication when run on Windows; a native Windows run is not claimed.

The subprocess owns a private temporary directory, so its internal socket cannot interfere with the legacy lifecycle test's OS-temp directory scan. The two files run with the default parallel execution.

Fixes #1370.


code-yeongyu (dismissed):
This change is not ready to merge because it fixes only the first filesystem error in the reported Windows bootstrap sequence while leaving the direct-launch supervisor unable to start on a fresh profile. The PR description and linked issue both document the subsequent missing public socket secret, and the production path still reaches that failure immediately after the recursive mkdir succeeds. The regression test therefore proves a helper in isolation, not the advertised end-to-end fix for the direct supervisor route.

## Blocking findings

1. **`packages/coding-agent/src/modes/rpc/host-lifecycle.ts:381`** — `runHostSupervisor()` still calls `readSocketSecret(socketSecretPath(publicSocket))` unconditionally on win32 before it creates the public listener. A direct `--internal-rpc-host-supervisor` launch on a fresh profile has no `ensureHost()` caller to create `<publicSocket>.secret`, so after this PR gets past the recursive mkdir it still exits with `ENOENT` and never listens. This is the second failure explicitly reported in issue #1370 and is also acknowledged as remaining work in the PR description; consequently the changelog claim that direct fresh-profile launch no longer crashes is false. Required change: define and implement the direct-launch contract by provisioning the public secret in the supervisor (or otherwise synchronizing with the writer before this read), and add an end-to-end regression test that starts the supervisor with a missing `rpc-host-daemon` and missing public secret and verifies that the public endpoint becomes reachable. If self-provisioning is intentionally out of scope, narrow the PR's stated fix and documentation instead of presenting #1370/direct fresh-profile bootstrap as fixed.

2. **`packages/coding-agent/test/suite/regressions/1370-rpc-internal-socket-mkdir.test.ts:13-40`** — the new tests call only `createInternalSocketPath()` and never execute `runHostSupervisor()`, so they cannot fail on the still-unhandled `readSocketSecret()` failure at `host-lifecycle.ts:381`, nor verify that the supervisor creates/listens on the requested public endpoint. The test's `platform` injection also bypasses the actual `process.platform` branch used by the production supervisor. Required change: add a subprocess-level regression covering the real `--internal-rpc-host-supervisor` route on a fresh isolated profile, including the public-secret behavior and a readiness/handshake assertion; keep the helper test only as supplemental unit coverage.

## Non-blocking notes

- The PR is currently reported by GitHub as `mergeable: false` / `dirty`, so it will need a rebase before it can land. That is not the reason for this review state, but the test and production-line references should be rechecked after rebasing.
- The linked issue's requested end state includes surfacing the failing bootstrap step and path. The current `readSocketSecret()` failure remains a raw missing-file error rather than a supervisor-level diagnostic identifying the required provisioning contract.

## What is good

The change correctly identifies the parent-directory cause of the first Windows `ENOENT`, uses `recursive: true` at the narrowest production call site, leaves the POSIX path unchanged, and adds a deterministic regression for creating the nested internal directory without requiring a Windows runner.

MoerAI:
Addressed both blocking findings. The supervisor now provisions a missing public secret before allocating the internal hop or spawning the child, reuses an existing valid secret, and reports provisioning failures with the step and path. The PR description and RPC documentation now describe that contract.

Added missing/existing-secret regressions against the production Windows bootstrap branch and a subprocess regression through the real hidden CLI route, readiness event, public endpoint, and protocol response. The test performs the secret handshake on Windows.

Local validation is macOS: the real POSIX subprocess route and platform-stubbed Windows bootstrap are covered; I have not run the named-pipe path on native Windows. Root checks, build, RPC QA, and the mock-provider CLI turn passed. The subprocess now owns a private temporary directory to avoid interfering with the existing lifecycle test's global directory scan; the two files run with default parallel execution.

code-yeongyu:
Both blocking findings are fixed, and the branch is rebased onto current `main` (`mergeable: MERGEABLE`). Head is `56e406cc284afb58032526169ab3afc578c43a5c`.

## Finding 1 — `host-lifecycle.ts:381` read the public secret it never provisioned

Resolved in `56e406cc2` (production change) on top of `79bf0fb3a`.

The direct-launch contract is now explicit and implemented in the supervisor: `runHostSupervisor()` provisions `<publicSocket>.secret` through a new `ensurePublicSocketSecret()` helper **before** it allocates the internal hop or spawns the child, so a provisioning failure leaves neither a scratch directory nor a host process behind.

- An existing valid secret — including one `ensureHost()` just wrote — is **reused, never rotated**. That is a correctness requirement, not an optimization: `resolveSocketTransportAddress()` derives the win32 pipe name from the socket path *and* the secret, so minting a fresh one here would move the endpoint away from the one the caller published.
- A missing or unusable secret is created with its parent directories at mode `0600` (via the existing `ensureSocketSecret`).
- Addressing the non-blocking note: a provisioning failure is now rethrown as `senpi rpc host supervisor: cannot provision public socket secret <path>` with the original error as `cause`, instead of a raw missing-file error.
- The public endpoint still authenticates with that secret (`authenticateSocket`) — the handshake requirement is unchanged.

The changelog claim, `docs/rpc.md` and the RPC tracker entry were all rewritten to state the self-provisioning contract; no surface still says the secret is caller-provisioned.

## Finding 2 — the regression proved a helper in isolation

Resolved in `79bf0fb3a` + `cd531f21e`.

`test/suite/regressions/1370-rpc-internal-socket-mkdir.test.ts` now has three layers:

1. **Subprocess regression on the real route** — spawns the actual CLI (`src/cli.ts --internal-rpc-host-supervisor --socket … --agent-dir …`) on a fresh isolated profile with **no `rpc-host-daemon` and no public secret** (both asserted absent first), waits for the supervisor's readiness line, connects to the public endpoint, sends the platform-required handshake, and asserts a successful `get_protocol_info` response. This is the "public endpoint becomes reachable" assertion the review asked for.
2. **Production-branch bootstrap cases** — two cases drive `runHostSupervisor()` itself with `process.platform` stubbed to `win32`, so the real `process.platform === "win32"` branch executes (no injected `platform` parameter): a missing secret is created (32 bytes) and an existing one is preserved byte-for-byte.
3. The `createInternalSocketPath()` cases remain, now only as supplemental unit coverage.

## Mutation evidence

Each production line the review named was reverted on the test host, the suite was run, then restored and re-run.

| Mutation | Result |
|---|---|
| `ensurePublicSocketSecret(...)` → `readSocketSecret(socketSecretPath(publicSocket))` (the reviewed line) | **FAIL** — `provisions the win32 public secret before spawning (existing=false)`: `expected Error: ENOENT: no such file or directory, open '/tmp/rpc1370-cpO3AV/rpc/rpc.sock.secret' to be Error: bootstrap reached child spawn`. The `existing=true` case still passed, which is correct: reading works when the secret is already there. → `2 failed \| 3 passed` |
| restore | **PASS** — `5 passed (5)` |
| win32 `mkdir(dir, { recursive: true })` → `recursive: false` | **FAIL** — `creates the win32 internal directory when rpc-host-daemon does not exist yet`: `ENOENT: no such file or directory, mkdir '/tmp/senpi-hlc-win32-YzcrSO/rpc-host-daemon/internal-0dd41005-…'` |
| restore | **PASS** |
| `findInternalSupervisorArgs(args)` disabled in `src/main.ts` (route not dispatched) | **FAIL** — `starts the real direct supervisor route on a fresh profile and serves the public endpoint`: `supervisor exited before readiness: Error: Unknown options: --internal-rpc-host-supervisor, --socket, --agent-dir`. This proves the subprocess test really drives the hidden CLI route rather than a helper. |
| restore | **PASS** |

## Test results

Run on a Linux host (x86_64, Node 24.20.0, bun 1.3.14) against this exact rebased tree, after `bun install --ignore-scripts` and `bun run build`:

```
$ cd packages/coding-agent && bun run test -- \
    test/suite/regressions/1370-rpc-internal-socket-mkdir.test.ts \
    test/rpc-host-lifecycle.test.ts \
    test/rpc-host-ensure.test.ts \
    test/suite/rpc-socket-ownership.test.ts

 Test Files  4 passed (4)
      Tests  63 passed (63)
   Duration  58.34s
```

`bunx biome check` on both changed files: `Checked 2 files in 31ms. No fixes applied.` A repo-wide `tsc --noEmit` in that sandbox reports 191 errors, all of them missing-module/implicit-`any` fallout of installing with `--ignore-scripts` under bun's isolated layout (128 are `Cannot find module 'vitest'`, in `packages/{pty,senpi-codemode,tui}`); **zero** errors are in either file this PR touches.

Windows caveat, stated plainly: validation ran on Linux and macOS. The direct-route subprocess test exercised the real POSIX route end to end; the win32 secret behaviour was exercised through the production supervisor with its process platform stubbed, up to child spawn. A native Windows run is not claimed.

## Rebase / non-blocking notes

- Rebased onto current `main`; `mergeable` is now `MERGEABLE`. `main` moved twice during this work, so the branch was rebased twice; both conflicts were changelog/tracker `.md` files only and were resolved by keeping **both** sides' entries (verified: every `main`-side changelog entry and tracker section is still present, with only this PR's entry added).
- The contributor's commits are preserved; the production change is a separate commit on top.

code-yeongyu (approved):
Both blocking findings resolved. runHostSupervisor now provisions the public socket secret before allocating the internal hop or spawning the child, so a provisioning failure leaves no scratch dir and no host process; an existing secret is reused rather than rotated, because resolveSocketTransportAddress derives the win32 pipe name from socket path AND secret, so rotating would move the endpoint away from the one the caller published. The isolated-helper test is replaced by a subprocess test that drives the real --internal-rpc-host-supervisor CLI route on a fresh profile through readiness, connect, handshake and get_protocol_info, plus two cases exercising the production win32 branch. 3 mutations, each RED then GREEN on restore, including one proving the subprocess test really drives the hidden CLI route. 63/63 tests pass on the pushed tree. Windows was not natively tested - POSIX end-to-end plus a platform-stubbed production branch is the stated ceiling. Approving.

PR #1421, opened 2026-09-07
fix(ai): normalize tool-result image data URLs

### LINKED ISSUE #1260: Tool-result image data URLs are double-prefixed for Codex Responses, producing invalid image_url
## Summary

Image-producing tools can return image content whose `data` field is already a `data:image/...;base64,...` URL. The OpenAI/Codex Responses converter unconditionally adds another data URL prefix, producing an invalid payload:

```text
data:image/png;base64,data:image/png;base64,...
```

Codex then rejects the request with:

```text
Invalid `input[214].output[1].image_url`. Expected a base64-encoded data URL with an image MIME type, but got an invalid base64-encoded value.
```

## Environment

- `omo-ai`: `5.0.0-0.beta.31`
- `@code-yeongyu/senpi`: `2026.8.31`
- `@code-yeongyu/senpi-ai`: `2026.8.31`
- Provider API: `openai-codex-responses`
- OS: Linux x64
- Node.js: 24.x

## Root-cause evidence

`convertToolResultOutput()` currently constructs every tool-result image as:

```ts
image_url: `data:${image.mimeType};base64,${image.data}`
```

This works when `image.data` is raw base64, but double-prefixes data URLs.

I reproduced this locally against the installed converter using the same valid 1x1 PNG payload in both representations:

- raw base64 input -> valid data URL
- `data:image/png;base64,<base64>` input -> `data:image/png;base64,data:image/png;base64,<base64>` -> invalid

The `input[*].output[*].image_url` parameter also identifies this as a tool-call output image rather than a normal user attachment.

## Expected behavior

The tool-result boundary should establish one canonical representation:

- accept raw base64 and add exactly one prefix; or
- detect a valid image data URL and preserve/normalize it; or
- reject non-canonical image content before it is persisted and replayed.

Malformed image data should not permanently poison subsequent turns.

## Suggested regression tests

1. Raw base64 tool-result images serialize to exactly one data URL prefix.
2. Already-prefixed data URLs do not become double-prefixed.
3. Malformed or empty image payloads are rejected or converted to a text placeholder before provider submission.
4. A failed image tool result does not wedge later plain-text turns.

## Related but not duplicate

- #302 concerns cumulative image bytes causing HTTP 413.
- openai/codex#16605 concerns empty clipboard image payloads.
- openai/codex#32287 concerns cumulative retained-image limits after compaction.

Altairpaca:
## Summary

Fixes #1260 by restoring the `ImageContent.data` contract at the shared message-normalization boundary before provider serialization.

Tool results occasionally contain an already-prefixed `data:<mime>;base64,...` string even though `ImageContent.data` is defined as raw base64. Responses adapters then add their own prefix and produce an invalid double-prefixed `image_url`.

## Root cause

`transformMessages()` normalized unsupported media and tool-call IDs but passed supported tool-result image payloads through unchanged. Provider adapters therefore had to trust that `ImageContent.data` was raw base64, and an already-prefixed value leaked into the Responses serializer.

## Changes

- Normalize tool-result media blocks in `transform-messages.ts`.
- If `data` starts with the exact canonical prefix for its declared MIME type (`data:${mimeType};base64,`), strip that prefix and keep raw base64 internally.
- Leave ordinary raw base64 byte-identical.
- Do not accept/normalize arbitrary or mismatched `data:` payloads.
- Add deterministic regression coverage for raw base64 and already-prefixed image data through `convertResponsesMessages()`.
- Add exact-nearest `packages/ai/src/changes.md` coverage for the shared normalization path.

## Scope / risk

The change is limited to tool-result media normalization. User attachments are unchanged. Provider serializers retain their existing raw-base64 contract and continue adding the wire prefix themselves.

## Validation / current sync

The branch is synchronized with upstream `main` at `464f1a54e8569cdfe8e4620583e19d594a6b96e9` (head `bdc33c87e57faeb1ca34719972707ef5248fc45d`). Current `main` still lacks tool-result data-URL normalization in `transform-messages.ts`, so #1260 remains live upstream. The semantic runtime diff remains exactly the production normalizer plus its focused regression; the third changed file is repository-policy tracker coverage.

PR-triggered CI, Changelog gate, and model-catalog workflows for this fork contribution may stop at `action_required` before jobs are created; that is an approval gate rather than executable test evidence. Any executable failure remains blocking.

Per `CONTRIBUTING.md`, release `CHANGELOG.md` entries remain maintainer-owned.

Closes #1260


Altairpaca:
Implementation is ready for review. The branch is current-main based and the diff is limited to the shared tool-result normalization seam plus the focused #1260 regression. All three fork workflows are currently `action_required` with no jobs executed; when convenient, please approve the workflow run. I’ll own any CI/review follow-up through merge.

PR #1426, opened 2026-09-07
fix(codemode): fail clearly when compiled runtime assets are missing

code-yeongyu:
## Incident
Desktop-app sessions on 2026-09-07 lost both eval kernels and, because `bash` is hidden whenever `eval` is registered, every shell: `Cannot find module '/$bunfs/root/inline-worker-entry.js'` and `python: can't open file '/$bunfs/root/prelude.py'`. The deployed runtime bundle had shipped without the `node_modules/@code-yeongyu/senpi-codemode` sidecar, and `resolveCodemodeRuntimeAsset` fell back to the compiled binary's virtual `$bunfs` path, which no Worker or python subprocess can open. Nothing named the sidecar, so the outage cost hours to diagnose.

## Change
- `kernels/shared/runtime-asset.ts`: new `requireCodemodeRuntimeAsset` never returns a Bun virtual path (`/$bunfs/`, `~BUN`, `%7EBUN`); it returns the module-relative asset when readable, else the sidecar beside `process.execPath`, else throws `CodemodeRuntimeAssetMissingError` whose message names the asset, the executable, and the expected sidecar file plus the fix ("Ship node_modules/@code-yeongyu/senpi-codemode next to the executable").
- js inline-worker / worker-startup, py transport, jl and rb runners route through the throwing resolver. `bundledBunSkillPath` keeps the non-throwing `resolveCodemodeRuntimeAsset` (logs once, returns undefined).
- `changes.md` entry.

## Tests
`packages/senpi-codemode/test/kernels/shared/runtime-asset.test.ts` (11 cases: local/sidecar/virtual-path resolution, the error's message fields, `isBunVirtualPath` table, propagation through the JS inline-worker and Python prelude resolvers).

Ran remotely on a Linux ascii box (bun 1.4.0 / node 24), never on the dev machine:
- RED (production change stashed, tests present): 10 failed | 1 passed.
- GREEN (focused file): 11 passed.
- Whole `packages/senpi-codemode` scope after `bun run build`: 107 files passed, 755 tests passed, 0 failed.

---
## Summary by cubic
Fixes the 2026-09-07 kernel outage caused by a missing codemode sidecar in deployed runtime bundles. Kernel asset resolution now fails with `CodemodeRuntimeAssetMissingError` instead of returning unopenable Bun virtual paths.

- `requireCodemodeRuntimeAsset` uses the local asset, then the sidecar beside the executable, then throws with the executable, expected sidecar path, and fix.
- Rejects Bun virtual paths (`/$bunfs/`, `~BUN`, `%7EBUN`) in the JS inline-worker, worker-startup, Python prelude, Julia, and Ruby resolvers; `bundledBunSkillPath` keeps the non-throwing resolver.
- Compiled binaries must ship `node_modules/@code-yeongyu/senpi-codemode` next to the executable.
- Adds 11 test cases covering local, sidecar, and virtual-path resolution plus error propagation.

<sup>Written for commit d11b259f5e7b826f86c8151701d31b464811685d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1426?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/senpi-codemode/changes.md
+++ b/packages/senpi-codemode/changes.md
@@ -1,5 +1,16 @@
 # senpi-codemode fork changes
 
+## 2026-09-07 - Fail clearly when compiled codemode assets are missing
+
+### What changed
+
+- Kernel runtime asset resolution now rejects Bun virtual paths and reports the missing codemode sidecar beside the executable, while skill contribution resolution remains non-throwing.
+
+### Why
+
+- Compiled binaries cannot pass embedded `/$bunfs` paths to workers or subprocesses; the actionable error identifies the expected sidecar asset and deployment fix.
+
+
 ## 2026-09-06 - Bun eval description steers away from Bun.spawnSync
 
 ### What changed
--- a/packages/senpi-codemode/src/kernels/jl/kernel.ts
+++ b/packages/senpi-codemode/src/kernels/jl/kernel.ts
@@ -1,6 +1,6 @@
 import { join } from "node:path";
 import type { BridgeConnectionConfig, KernelToHostMessage } from "../../bridge/protocol.ts";
-import { type CodemodeRuntimeAssetEnvironment, resolveCodemodeRuntimeAsset } from "../shared/runtime-asset.ts";
+import { type CodemodeRuntimeAssetEnvironment, requireCodemodeRuntimeAsset } from "../shared/runtime-asset.ts";
 import { SubprocessKernel, type SubprocessSpawn } from "../shared/subprocess-kernel.ts";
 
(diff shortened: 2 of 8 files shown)


code-yeongyu:

PR #1428, opened 2026-09-07
feat(ai): give the GPT-6 Astra series a 600,000 context window

code-yeongyu:
## Summary

Every GPT-6 Astra catalog entry now declares a 600,000-token context window, on every provider that ships the model.

senpi treats `contextWindow` as the prompt budget everywhere (compaction gates, usage meter, output clamp). #1427 made the whole GPT-5.x/GPT-6 family store the documented OpenAI input cap, which put the Astra series at 922,000 across all providers. This change gives the Astra series its own project default of 600,000 instead.

The generator stays the single source of truth: `applyGpt6AstraContextWindow` runs in the per-model normalization pass, after `applyOpenAiInputCap` and before provider grouping, so it reaches every catalog rather than only the hand-written OpenAI entries. Users who want a different budget still set it through model overrides.

## Changes

- `packages/ai/scripts/generate-models.ts`: `GPT_6_ASTRA_DEFAULT_CONTEXT_WINDOW` is 600,000, stamped onto every entry whose id carries `gpt-6-astra`.
- `packages/ai/src/providers/data/*.json`: regenerated; all 13 Astra entries across 7 provider catalogs move 922000 -> 600000.
- `packages/ai/test/gpt-6-astra-context-window.test.ts`: new regression walking every generated catalog, so a future regeneration cannot reintroduce a per-provider split.
- `packages/ai/test/gpt-6-astra-catalog.test.ts`, `packages/ai/test/openai-fast-models.test.ts`: existing Astra context assertions moved to the series value.

## Verification

Failing-first, then green, on a Linux box running bun 1.4.0:

- RED (before the generator change): the new suite listed every offending entry, e.g. `openai.json:openai-responses/gpt-6-astra=922000` and `openrouter.json:openai-completions/openai/gpt-6-astra=1050000`. Exit 1.
- GREEN (after regeneration): 2 passed.
- Full `packages/ai` suite: 2563 passed, 884 skipped, 0 test failures. The one failing file, `codex-apply-patch-wire-schema.test.ts`, cannot resolve `@earendil-works/pi-tui` because `packages/tui/dist` is not built in that environment; it fails identically without this change.
- `bunx tsx scripts/check-model-data.ts` reports `Generated model data is valid.`
- Diff scope: a structural before/after comparison of every regenerated catalog against `main` shows exactly 13 changed entries, all `gpt-6-astra*`, all `contextWindow` only. `gpt-5.6-sol` stays at 650,000, the GPT-5.x input caps from #1427 are untouched, and no cost or capability field moved.
- The input-cap contract from #1427 still holds: `openai-input-cap-catalog.test.ts` rejects documented totals (400,000 / 1,050,000) as `contextWindow`, and 600,000 is not one of them.

Rebased onto `main` after #1427 merged.


code-yeongyu:
## Question before merge — please answer in the PR body so it survives

Reducing `gpt-6-astra` / `gpt-6-astra-fast` from **922,000 to 600,000** across the provider catalogs needs its rationale recorded, because 922,000 is not an arbitrary number: it is the *measured provider input cap* that #1427 just corrected the catalog to, after real sessions on our side died at ~916k against a catalog that claimed 1.05M.

So which of these is this change?

- **(a) A deliberate operating margin** — the real cap is still ~922k, and 600k is chosen for safety, cost, latency, or retrieval quality.
- **(b) A correction of the cap itself** — the usable window is genuinely smaller than 922k.

This matters for two concrete reasons:

1. **Changelog wording.** (a) is a tuning change; (b) is a factual correction that partially supersedes #1427, which shipped in `2026.9.7` minutes ago. They should not be described the same way.
2. **Preventing a revert round-trip.** If this lands as a bare number change with no stated reason, a future contributor comparing the catalog against the provider's documented cap will read 600,000 as drift and "fix" it back to 922,000. That exact round-trip already happened once with the 1.05M values. An explicit sentence in the PR body is what stops it.

No objection to the change itself, and no need to block on this — the structural diff (13 rows, zero out-of-scope, `check-model-data` valid) looks clean. Just state (a) or (b) explicitly before merging.

code-yeongyu:
**(a) — a deliberate operating margin, not a correction of the cap.**

The provider limits are unchanged and #1427 read them correctly: OpenAI documents a 1,050,000-token window for GPT-6 Astra, and the Responses API rejects a prompt once it alone exceeds the window minus max output, which is the 922,000 figure #1427 stored. Nothing here claims that number is wrong.

600,000 is a project default in the same class as `gpt-5.6-sol`'s 650,000: a cost-and-behavior tier the catalog ships, deliberately below what the provider would still accept. `contextWindow` is the budget senpi gates on, so this is the point at which we choose to compact rather than the point at which the API refuses.

**So the two numbers are not competing measurements of one limit.** 922,000 answers "what will the provider still accept", 600,000 answers "how much of that do we spend by default". A future contributor who finds 600000 and raises it to 922000 to match the documented cap would be undoing the tier on purpose-built grounds, not fixing drift — the same way raising Sol's 650,000 to its own cap would be. Anyone who wants the full window sets it through model overrides.

That reasoning is now written into `packages/ai/changes.md` and the `[Unreleased]` CHANGELOG entry so the intent travels with the value.

PR #1441, opened 2026-09-07
fix(codemode): capture the last expression unless a top-level return exists (#1439)

### LINKED ISSUE #1439: JS eval kernel: any nested "return" in a cell silently discards the completion value
## Summary

In the JS eval kernel, a cell whose text contains the word `return` ANYWHERE — including inside a nested arrow function or helper — silently loses its completion value. The cell runs correctly; only the output disappears.

## Mechanism

`src/kernels/js/worker-indirect-eval.js:14`:

```js
if (/\breturn\b/u.test(persistentCode)) return `(async () => {\n${persistentCode}\n})()`;
```

The test runs over the WHOLE cell, so any occurrence of `return` causes the cell to be wrapped in an async IIFE. Inside that wrapper the final expression is no longer the completion value of the evaluated program, so nothing is captured. The intent is clearly to support top-level `return`, but the detection cannot distinguish a top-level `return` from one nested inside a helper.

## Reproduction (verified in a live kernel)

Loses output:
```js
const y = (() => { return 1; })();
"completion-value-should-appear"
```
→ `(no output)`

Identical semantics, no `return` token:
```js
const y = (() => 1)();
"completion-value-should-appear"
```
→ `"completion-value-should-appear"`

The only difference between the two cells is the word `return`.

## Impact

Any cell that defines a helper with an early return, a `try/catch` wrapper that returns, a `.map()` callback using `return`, or a `parallel()` thunk, silently produces no output. The failure is invisible: there is no warning, and the code itself worked. Callers reasonably conclude their cell produced nothing and re-run or restructure it.

This is a self-effacing defect — it degrades exactly the cells that are doing real work (control flow, error handling, batching), while trivial cells appear fine.

## Suggested direction

Detect a TOP-LEVEL `return` rather than any `return` token — e.g. parse and inspect the top-level statement list, or only wrap when a `return` appears at statement position outside any function body. Failing that, wrap but preserve the completion value (capture the last top-level expression explicitly rather than relying on the IIFE's implicit result).

## Workaround for callers

End such cells with an explicit `print(...)`/`console.log(...)` instead of relying on the completion value.

## Provenance

Root-caused by a peer session; reproduction above executed independently in a live kernel before filing. Present in the installed `@code-yeongyu/senpi-codemode` under both the global `senpi` install and its nested copy.

code-yeongyu:
## Summary

Fixes #1439.

`wrapUserCode` (`packages/senpi-codemode/src/kernels/js/worker-indirect-eval.js:14`) tested `/\breturn\b/u` over the whole cell source to decide whether the user had written a top-level `return`. Any `return` token anywhere — inside a nested arrow, a `.map()` callback, a try/catch helper, a string, or a comment — matched, so the wrapper skipped last-expression capture and the cell settled with no `valueRepr`. The tool then rendered `(no output)` even though the cell ran and its globals persisted. Every substantial cell (anything that defines a helper) hit this.

Measured on a live session (`01a07b11`, mengmotaHost): 16 of 20 JavaScript cells lost their value; each one contained a nested `return`. Duration, output size, and the final expression's shape were all ruled out (a 12 s sleep cell and a 2000-char value both print fine).

## What changed

- `scanTopLevelStatements` (formerly `findLastTopLevelStatementStart`) now also records whether a `return` token appears at brace/paren/bracket depth 0, outside literals and comments, and `wrapUserCode` consults that instead of the whole-source regex. A genuine top-level `return` still suppresses capture, exactly as documented.
- An identifier that follows `.` is treated as a property name, so `obj.return` no longer primes the scanner as the keyword (it used to start a regex-context/continuation as if `return` had been written).
- Eval truncation notices name the real cause. Before, any byte loss — including a pure column clamp — produced `truncatedBy: "bytes"` with no `maxBytes`, and the notice printed the output's own size as a limit (`(772B limit)`). The metadata now distinguishes `columns` (width + clamped line count), `bytes` (the actual cap, `DEFAULT_MAX_BYTES`), and `lines`; a notice without a known cap reports the dropped size instead of a limit.

## Tests

`packages/senpi-codemode/test/kernel-js-last-expression.test.ts`
- captures the last expression when a nested function contains return
- captures the last expression after a try/catch helper that returns
- captures the last expression when return appears only inside a string or comment
- treats a property named return as a plain member access
- keeps a genuine top-level return as the cell value

`packages/senpi-codemode/test/output/output-meta.test.ts`
- reports dropped bytes instead of a limit when no byte cap is known
- names the column clamp when lines were cut to a width

`packages/senpi-codemode/test/output/eval-output-collector-meta.test.ts` (new)
- attributes a width-clamped line to the column cap, not a byte limit
- reports the real byte cap when the tail buffer dropped output

RED captured on an x86_64 box before the fix (workspace built, bun 1.4.0 / node 24): 7 assertion failures — `expected undefined to be 42`, `expected undefined to be 6`, `'(772B limit)'` vs `'(1.2KB dropped)'`, and the two collector `truncatedBy` mismatches. GREEN after the fix: 3 files / 21 tests pass; the full `senpi-codemode` suite, root `biome check`, and root `tsc --noEmit` results are recorded in the PR comments.

## Notes

- `senpi-codemode` is fork-only (not present in the pinned `badlogic/pi-mono` tree), so no `changes.md` coverage applies; the package `CHANGELOG.md` carries the `[Unreleased]` entries.
- Workaround for hosts on older builds until this ships: end cells with an explicit `print(...)`, and never blindly re-run a blank cell that had side effects — query the resulting state instead.

---
## Summary by cubic
Fixes JavaScript eval cells in `senpi-codemode` losing their completion value whenever source contained a `return` inside a nested function, callback, try/catch helper, string, or comment: the wrapper previously matched any `return` and skipped last-expression capture, so cells settled with no value and rendered `(no output)`. Last-expression capture now runs unless there's a genuine top-level `return`, and a property named `return` is treated as member access. Also fixes eval output truncation notices to name the real cause — `columns`, `bytes`, or `lines` — instead of printing the output's own size as a byte limit.

- The statement scanner records `return` only at brace/paren/square depth 0, outside literals and comments, and treats an identifier after `.` as a property name.
- Truncation notices now distinguish width clamps (with the column cap and clamped line count), the actual byte cap, and line cuts; a notice without a known cap reports the dropped size.
- Includes a formatting fix in `packages/ai/test/gpt-6-astra-context-window.test.ts` to satisfy `biome check` without `--write`.

<sup>Written for commit 60f0c9e9fcd69b7471231a65bd87c3e180662101. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ai/test/gpt-6-astra-context-window.test.ts
+++ b/packages/ai/test/gpt-6-astra-context-window.test.ts
@@ -24,7 +24,10 @@ function collectAstraEntries(): AstraEntry[] {
 	for (const file of readdirSync(dataDirectory)
 		.filter((name) => name.endsWith(".json"))
 		.sort()) {
-		const parsed = JSON.parse(readFileSync(`${dataDirectory}${file}`, "utf8")) as Record<string, Record<string, CatalogEntry>>;
+		const parsed = JSON.parse(readFileSync(`${dataDirectory}${file}`, "utf8")) as Record<
+			string,
+			Record<string, CatalogEntry>
+		>;
 		for (const [api, models] of Object.entries(parsed)) {
 			if (!models || typeof models !== "object") continue;
 			for (const [id, model] of Object.entries(models)) {
--- a/packages/senpi-codemode/CHANGELOG.md
+++ b/packages/senpi-codemode/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### Fixed
 
+- JavaScript eval cells no longer lose their completion value when a nested function, callback, or try/catch helper contains `return`: the cell wrapper now skips last-expression capture only for a genuine top-level `return`, and a property named `return` no longer primes the statement scanner as the keyword (#1439).
+- Eval output truncation notices now name the real cause: a width-clamped line reports `N line(s) clamped to M columns (… dropped)`, a byte-capped tail reports the actual cap, and a notice never presents the output's own size as a limit.
+
 ### Removed
 
 ## [2026.9.7] - 2026-09-07
--- a/packages/senpi-codemode/src/kernels/js/worker-indirect-eval.js
+++ b/packages/senpi-codemode/src/kernels/js/worker-indirect-eval.js
@@ -11,7 +11,7 @@ export async function awaitMaybePromise(value) {
 
 export function wrapUserCode(code) {
 	const persistentCode = persistTopLevelDeclarations(code);
-	if (/\breturn\b/u.test(persistentCode)) return `(async () => {\n${persistentCode}\n})()`;
+	if (scanTopLevelStatements(persistentCode).hasTopLevelReturn) return `(async () => {\n${persistentCode}\n})()`;
 	return `(async () => {\n${captureLastExpression(persistentCode)}\n})()`;
 }
 
@@ -705,7 +705,7 @@ function skipBlockComment(code, start) {
 }
 
 function captureLastExpression(code) {
-	const start = findLastTopLevelStatementStart(code);
+	const start = scanTopLevelStatements(code).lastStatementStart;
 	const head = code.slice(0, start);
 	const tail = code.slice(start).trim();
 	if (!tail || isStatementOnly(tail)) return code;
@@ -726,7 +726,7 @@ function isStatementOnly(source) {
 
 const CONTROL_PAREN_KEYWORDS = new Set(["catch", "for", "if", "switch", "while", "with"]);
 
-function findLastTopLevelStatementStart(code) {
+function scanTopLevelStatements(code) {
 	let start = 0;
 	let round = 0;
 	let square = 0;
@@ -734,6 +734,7 @@ function findLastTopLevelStatementStart(code) {
 	let canStartRegex = true;
 	let lastSignificant = "";
 	let pendingControlParen = false;
+	let hasTopLevelReturn = false;
 	const controlParens = [];
 	for (let index = 0; index < code.length; index += 1) {
 		const char = code[index];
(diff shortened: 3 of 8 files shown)


code-yeongyu:
Remote verification (x86_64 ascii box `bx_822x9b3z`, bun 1.4.0 / node 24.18.1, workspace built with `bun run build:bun`), tree `60f0c9e9f` minus the hygiene commit:

- RED before the fix: 7 assertion failures across the three test files (`expected undefined to be 42`, `expected undefined to be 6`, `'(772B limit)'` vs `'(1.2KB dropped)'`, two collector `truncatedBy` mismatches).
- GREEN after the fix: `test/kernel-js-last-expression.test.ts` + `test/output/output-meta.test.ts` + `test/output/eval-output-collector-meta.test.ts` = 3 files / 21 tests passed.
- Full `packages/senpi-codemode` suite: 108 files passed, 1 skipped; 764 tests passed, 6 skipped; exit 0.
- Root `tsc --noEmit`: exit 0.
- `biome check --error-on-warnings` (without `--write`) on the package: clean.

Merging on observed remote green per the repository's merge policy.

PR #1458, opened 2026-09-07
fix(compaction): count retained images using token estimates

### LINKED ISSUE #1455: Deterministic compaction fallback counts image base64 bytes against the token budget
## Summary

The image-shape fix in #1412 is installed, but required compaction can still deadlock when the retained suffix contains realistically sized, well-formed tool-result images. The deterministic fallback treats the serialized JSON byte length, including image base64 payloads, as a lower bound on token usage:

```ts
Math.max(estimateTokens(message), Buffer.byteLength(serialized))
```

As a result, a valid image that the ordinary compaction estimator charges as 1,200 tokens can consume more than the entire retained token budget in this fallback. The prepared suffix is rejected with `retained-token-budget-exceeded`, and the user sees the generic error:

```text
Compaction rejected: deterministic compaction fallback cannot retain the prepared suffix
```

This is a residual problem in the same recovery path as code-yeongyu/oh-my-openagent#6871, not a recurrence of the old blanket `unsafe-retained-content` image rejection and not a missing update.

## Versions and related fixes

- Reproduced against installed Senpi **2026.9.7**, through OMO plugin **5.0.0-beta.46**, on macOS arm64 / Bun 1.4.
- Current upstream main at `b8fdd0a` still contains the serialized-byte floor.
- #1412 accepts valid tool-result image shapes and adds candidate diagnostics. Its positive regression uses a tiny 1x1 PNG, so it does not exercise realistic image payload sizing.
- The fix shipped in Senpi 2026.9.6 / OMO beta.45: https://github.com/code-yeongyu/oh-my-openagent/issues/6871#issuecomment-5561255473
- Source in the affected release: https://github.com/code-yeongyu/senpi/blob/v2026.9.7/packages/coding-agent/src/core/extensions/builtin/compaction/deterministic-fallback.ts#L189-L204

This is distinct from #1407 (resident-store read amplification while resuming) and #302 (provider HTTP request-body limits). Transport byte limits still matter; this report concerns mixing serialized bytes into a token-window budget.

## Observed incident and causal isolation

A native manual compaction started at 2026-09-07T10:30:53.229Z and was rejected at 10:32:54.445Z (121.216 seconds). The final event reports `disposition: rejected`, `rejectionCause: cancelled-by-extension`, and the generic error above. The original summarization exception is replaced by the fallback rejection, so the elapsed time is consistent with the watchdog but does not independently prove the precise upstream generator failure. Adjacent memory-reflection quota failures are not evidence that quota caused this compaction rejection.

An in-memory replay used a SHA-256-verified copy of the original pre-recovery transcript. No user transcript was edited and no provider was called. Holding the prepared retention boundary fixed:

| Input | Image safety | Result |
| --- | --- | --- |
| Original two retained images | Valid | `retained-token-budget-exceeded` |
| Only the image `data` strings replaced by a tiny valid PNG | Valid | Accepted at the same prepared boundary |
| Original image data restored | Valid | Budget rejection returns |

The effective context window was 1,000,000 with a 40,000 reserve: **960,000 retained-token budget**. In the replay, ordinary context estimation was **20,524 tokens**, while the fallback charged **1,155,991 units**. One tool result with two images alone was **2,767 estimated tokens** but **918,383 serialized bytes**. These are two different units, not two tokenizers disagreeing.

For recovery, a separately verified substantive summary was applied using native `ExtensionContext.applyCompaction(precomputed, { expectedRevision, reason: "manual" })`. This retained the exact original boundary, all 40 recent messages and both original images. Core revision/lifecycle and `_wouldCompactionOverflow` checks remained enabled. The resulting context was **22,754 estimated tokens**, and a real compaction entry was committed. All 2,456 original persisted records, 52 unfinished todos, and the active goal identity/objective/status were preserved. This was a session-recovery workaround, not an implementation fix.

Private conversation text, images, credentials, and local paths are deliberately not attached. The independent synthetic reproducer below demonstrates the same defect.

## Self-contained, offline reproduction

Save the script below as `repro.mjs` and run it with Bun against an installed Senpi package directory containing `dist/`:

```sh
bun repro.mjs /absolute/path/to/node_modules/@code-yeongyu/senpi
```

The script generates a deterministic, valid 512x512 RGBA PNG with PNG chunk checksums and a zlib-compressed IDAT. It does not use real screenshots, read a real session, write a session, or call a provider. The current assertions confirm the bug: large valid image rejected, tiny valid image accepted at the same boundary, original restored and rejected again.

<details>
<summary>Complete reproducer</summary>

```js
import assert from "node:assert/strict";
import { resolve } from "node:path";
import { pathToFileURL } from "node:url";
import { crc32, deflateSync } from "node:zlib";

const root = resolve(process.argv[2]);
const { createRequiredCompactionFallback } = await import(
  pathToFileURL(`${root}/dist/core/extensions/builtin/compaction/deterministic-fallback.js`).href
);
const { DEFAULT_COMPACTION_SETTINGS, estimateTokens } = await import(
  pathToFileURL(`${root}/dist/core/compaction/index.js`).href
);

function chunk(type, data) {
  const name = Buffer.from(type);
  const length = Buffer.alloc(4);
  length.writeUInt32BE(data.length);
  const checksum = Buffer.alloc(4);
  checksum.writeUInt32BE(crc32(Buffer.concat([name, data])));
  return Buffer.concat([length, name, data, checksum]);
}

const width = 512;
const pixels = Buffer.alloc((width * 4 + 1) * width);
let seed = 1;
for (let y = 0; y < width; y++) {
  for (let x = 1; x <= width * 4; x++) {
    seed = (Math.imul(seed, 1664525) + 1013904223) >>> 0;
    pixels[y * (width * 4 + 1) + x] = seed >>> 24;
  }
}
const header = Buffer.alloc(13);
header.writeUInt32BE(width, 0);
header.writeUInt32BE(width, 4);
header[8] = 8;
header[9] = 6;
const png = Buffer.concat([
  Buffer.from([137, 80, 78, 71, 13, 10, 26, 10]),
  chunk("IHDR", header),
  chunk("IDAT", deflateSync(pixels)),
  chunk("IEND", Buffer.alloc(0)),
]);
const tinyPng = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mP8/x8AAwMCAO+aZ1EAAAAASUVORK5CYII=";
const usage = {
  input: 1, output: 1, cacheRead: 0, cacheWrite: 0, totalTokens: 2,
  cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
};
const branch = [
  {
    type: "message", id: "user", parentId: null, timestamp: "2026-01-01T00:00:00Z",
    message: { role: "user", content: "Inspect the image.", timestamp: 1 },
  },
  {
    type: "message", id: "assistant", parentId: "user", timestamp: "2026-01-01T00:00:01Z",
    message: {
      role: "assistant", api: "openai-responses", provider: "openai", model: "fixture",
      content: [{ type: "toolCall", id: "read-image", name: "read", arguments: { path: "synthetic.png" } }],
      usage, stopReason: "toolUse", timestamp: 2,
    },
  },
  {
    type: "message", id: "result", parentId: "assistant", timestamp: "2026-01-01T00:00:02Z",
    message: {
      role: "toolResult", toolCallId: "read-image", toolName: "read", isError: false, timestamp: 3,
      content: [{ type: "image", mimeType: "image/png", data: png.toString("base64") }],
    },
  },
];
const preparation = {
  firstKeptEntryId: "assistant", tokensBefore: 10000, settings: DEFAULT_COMPACTION_SETTINGS,
};
function run(entries) {
  const diagnostics = {};
  const result = createRequiredCompactionFallback(
    preparation, 1000000, "summarization-timeout", {}, entries, diagnostics,
  );
  return { accepted: result !== undefined, firstKeptEntryId: result?.firstKeptEntryId, diagnostics };
}
const original = run(branch);
const small = structuredClone(branch);
small[2].message.content[0].data = tinyPng;
const reduced = run(small);
const restored = run(branch);
assert.equal(original.accepted, false);
assert.equal(original.diagnostics.candidateRejections[0].rejectionReason, "retained-token-budget-exceeded");
assert.equal(reduced.accepted, true);
assert.equal(reduced.firstKeptEntryId, preparation.firstKeptEntryId);
assert.equal(restored.accepted, false);
console.log(JSON.stringify({
  pngBytes: png.length,
  base64Bytes: branch[2].message.content[0].data.length,
  imageMessageTokenEstimate: estimateTokens(branch[2].message),
  imageMessageSerializedBytes: Buffer.byteLength(JSON.stringify(branch[2].message)),
  original, reduced, restored,
}, null, 2));
```

</details>

Observed on 2026.9.7, exit code 0:

```text
pngBytes: 1049476
base64Bytes: 1399304
imageMessageTokenEstimate: 1200
imageMessageSerializedBytes: 1399461
original.accepted: false
original first rejection: retained-token-budget-exceeded
reduced.accepted: true
reduced.firstKeptEntryId: assistant
restored.accepted: false
```

## Expected behavior and regression boundaries

A well-formed tool-result image must consume the existing image-token estimate rather than one token per serialized base64 byte. The fallback should retain the prepared suffix when its text/metadata and image-token charges fit the same effective token budget, preserving the original image data and atomic tool-call/result pairs.

The fix must continue to fail closed on malformed retained content, accessors/cycles/non-plain persisted values, invalid tool chains, and genuine retained-context overflow. It must not exempt arbitrary `data` fields or image-looking tool arguments/metadata from sizing, silently drop images, weaken transport byte protections, or make image-token costs disappear when surrounding text is large.

Coverage should include realistically sized images, image payload-size independence, image count/token accumulation, mixed text plus images, non-image large payloads, and a normal required-compaction path that persists the recovered checkpoint after a controlled summary failure. I will follow this report with a focused tested PR.

ayden94:
## Summary

Fixes #1455. Follow-up to #1412 / code-yeongyu/oh-my-openagent#6871.

The previous fix accepts well-formed tool-result images, but deterministic recovery still used serialized JSON byte length as a token floor. A realistic valid 512x512 PNG has a 1.4 MB Base64 payload while the existing image estimator charges 1,200 tokens. That made an otherwise admissible retained suffix fail after a summarizer error.

## Change

In the existing fallback sizing loop, after retained-content and bounded-value checks:

- Remove only the JSON-escaped payload bytes of recognized top-level tool-result image blocks from the serialized floor; retain their string delimiters and all envelope/metadata bytes.
- Add each image's existing `estimateTokens` cost to the remaining conservative byte floor. Keep the ordinary whole-message estimate as a lower bound.
- Keep original image data and the prepared boundary unchanged. Arbitrary `data` fields, image-looking tool arguments, extra image metadata, escaped text, malformed content, invalid tool chains, and genuine over-budget contexts are not exempted.

This does not change the shared estimator, transport byte limits, reserve scaling, fallback boundary search, or core compaction application. The production change is confined to the retained-message sizing loop.

## Regression coverage

The new fixture builds a deterministic, valid 512x512 PNG using zlib and PNG CRCs; no private screenshot or transcript is included.

- Realistic PNG accepted at the original prepared boundary without payload mutation.
- Same image-token cost for small and large valid PNG payloads.
- Multiple-image token accumulation still rejects genuine overflow.
- Image costs stay additive when a large serialized text floor dominates.
- Image-looking tool arguments and unrelated image metadata remain charged.
- Large escaped non-image text remains subject to the conservative bound.
- A real `AgentSession.compact()` with the builtin extension persists the checkpoint and original image after a controlled classified summarizer failure.

## Verification

- **RED:** the new large-PNG regression failed on the original code: expected a prepared compaction, received `undefined`.
- **GREEN:** **77 test files / 554 tests passed**, including the complete compaction directory and both new issue regression files.
- **Static validation:** `bun run check` passed, including Biome, repository metadata/lock checks, TypeScript, and browser smoke.
- **Build:** `bun run build` passed for the monorepo.
- **Independent local gate review:** APPROVE; a separate run of both new regression files and the existing deterministic-fallback safety tests passed all 38 tests.
- **Real-surface QA:** the repository's isolated `senpi-qa` RPC client and local mock server drive the actual source CLI with no real provider credentials. Cases cover a large valid image, two large images plus Korean text, and preserved rejection of an oversized non-image session at model admission. Success cases assert a persisted deterministic checkpoint, exact image content, original transcript prefix, and original boundary; the rejection case permits only the normal thinking-level metadata append and makes no provider request.
- **Original incident replay:** read-only replay of the SHA-256-verified pre-recovery snapshot now accepts the original prepared boundary on the first candidate without changing either original image. Conservative fallback estimate: 247,807 against a 960,000 budget. The private transcript is not attached.

Focused command:

```sh
CI=1 bun run --cwd packages/coding-agent test \
  test/compaction \
  test/suite/regressions/issue-1455-compaction-image-budget.test.ts \
  test/suite/regressions/issue-1455-required-image-compaction.test.ts
bun run check
bun run build
```

Local evidence is recorded under `local-ignore/qa-evidence/` and an external task evidence directory, not committed. The TypeScript LSP executable is not installed on the workstation; the full TypeScript checker and Biome both passed. No dependency, lockfile, model setting, or live user session was changed by this patch.

---
## Summary by cubic
Fixes deterministic compaction recovery rejecting valid tool-result images by charging image token estimates instead of serialized Base64 payload bytes against the token window.

A valid 512x512 PNG has a 1.4 MB Base64 payload but only costs 1,200 estimated tokens, so the previous conservative byte floor wrongly rejected an admissible retained suffix after a summarizer error.

- The fallback sizing loop now removes only the JSON-escaped payload bytes of validated top-level tool-result image blocks from the serialized floor, and adds each image's existing token estimate instead.
- Text, delimiters, image metadata, tool arguments, nested data fields, escaped non-image text, malformed content, and genuine over-budget contexts are still charged as before.
- The shared estimator, transport byte limits, reserve scaling, fallback boundary search, and core compaction application are unchanged.

**Regression coverage**

- Adds a deterministic 512x512 PNG fixture generated with zlib and PNG CRCs; no private screenshots or transcripts.
- Covers a realistic PNG retained at the prepared boundary without payload mutation, stable token cost across payload sizes, genuine overflow rejection, additive image costs when a text floor dominates, image-shaped arguments and duplicate metadata remaining charged, and large escaped non-image text staying bounded.
- The additive-cost case uses prose so the serialized byte floor (not the estimator's base64-run weighting) decides admission; it fails when the image token charge is removed.
- An end-to-end test drives a real `AgentSession.compact()` with the builtin extension and persists the checkpoint and original image after a classified summarizer failure.

<sup>Written for commit 1592b7d32a41e06e91e512822a85aba13d08f470. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1458?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (approved):
## Review: APPROVE

Thanks for the precise fix and the honest incident isolation, @ayden94. The change is correct and tightly scoped; I pushed one test-only fixup (`1592b7d32`) to the branch to close a coverage gap I found by mutation. Details below.

### What I verified

**Code.** The production hunk is confined to the retained-message sizing loop. The exemption only runs when `hasUnsafeRetainedContent` has already accepted the message (so `block.data` is guaranteed to be a non-empty string via `isWellFormedImageBlock`) and only for `toolResult` blocks with `type === "image"`. `JSON.stringify(block.data)` is byte-identical to the substring inside `JSON.stringify(message)`, so subtracting its length minus the two quote delimiters removes exactly the payload and keeps envelope, metadata and delimiter bytes charged. User-message images stay rejected by `hasTextOnlyContent`, so the toolResult-only scope is consistent with the existing safety policy. The byte-level guard for oversized requests is still `TRANSPORT_IMAGE_BUDGET_BYTES` (24 MiB, `messages.ts` / `elideOldImages`), so moving images out of the token unit does not open an unbounded request path. `estimateTokens({ ...message, content: [block] })` per block is a constant 1,200 in practice, but writing it that way keeps the fallback in lockstep with the shared estimator, which I prefer.

**Static.** `biome check --error-on-warnings` (no `--write`) on all changed files: clean. `tsc --noEmit` at the repo root: clean. `check-ts-relative-imports`: clean.

**Tests (Linux x86_64, bun 1.4.0 / node 24.20.0; vitest ran under Node).**
- GREEN with the PR's focused command (`test/compaction` + both new regression files): 77 files / 554 tests passed - matches the PR body.
- RED with the production file replaced by `origin/main`: 3 failed / 5 passed (the three positive cases fail, the negative guards keep passing) - as expected.
- Mutation "no payload subtraction" (`byteLength(serialized) + imageTokens`): 3 failed - caught.
- Mutation "drop `+ imageTokens`" (images cost nothing in the byte floor): **8 / 8 passed on the original branch** - not caught. See below.
- CI on the final sha `1592b7d32`: 14 / 14 jobs green, same job set as the latest main run, `Check and test` completed.

### Finding fixed in `1592b7d32` (test only)

`keeps image costs additive when the serialized text floor dominates` used `"x".repeat(10_000)`. That string matches `BASE64_RUN_RE` in `estimateTokens` (an unbroken alphanumeric run >= 512 chars is weighted 4x), so the ordinary estimate for the text alone was 10,000 tokens (+1,200 for the image) and dominated `Math.max`; the case was rejected by the estimator, not by the byte floor, so the `+ imageTokens` term was never exercised. The fixup uses prose so the serialized byte floor decides admission and derives the window from the text-only control plus headroom smaller than one image charge. With the same mutation applied, exactly that one assertion now fails; with the real code, 38 / 38 pass across both regression files and `required-compaction-deterministic-fallback.test.ts`.

`rejects genuinely excessive image-token accumulation` is likewise satisfied by the ordinary estimator alone (8 x 1,200 already exceeds the budget). It remains a valid "images are not free" guard, so I left it as is.

### Non-blocking notes
- `deterministic-fallback.ts` is now ~304 pure LOC (was ~295). Not this PR's job; the sizing loop is a natural extraction candidate for a follow-up.
- `changes.md` coverage was not strictly required here (`deterministic-fallback.ts` is fork-only, not in the `badlogic/pi-mono` pin), but the entry is accurate and welcome.

code-yeongyu:
Merged as `44251ea24`. Thanks @ayden94 - the isolation in #1455 (two different units, not two tokenizers disagreeing) made this easy to confirm, and the offline PNG reproducer was the right call.

One test-only fixup landed on the branch before merge (`1592b7d32`): the additive-cost case could not fail under a "`+ imageTokens` deleted" mutation because `"x".repeat(10_000)` trips the base64-run weighting in `estimateTokens`, so the ordinary estimate dominated `Math.max` and rejected the fixture on its own. It now uses prose and derives the window from the text-only control, and the same mutation fails exactly that assertion. Everything else is yours, unchanged.

Verification before merge: the real `origin/main` + PR merge tree ran green on a Linux x86_64 box (77 files / 554 tests, `test/compaction` plus both regression files), CI was 14/14 on the final sha, and `biome`/`tsc`/`check-ts-relative-imports` were clean locally.

PR #1464, opened 2026-09-07
test(credential-rotation): close the runtime stream before temp-dir teardown

### LINKED ISSUE #1457: test: intermittent ENOTEMPTY in afterEach temp-dir cleanup (two lanes, re-run passes)
Two lanes independently hit the same failure mode today (2026-09-07) while gating unrelated work, both correctly judged it not-their-change, and neither filed it — so it is unowned. Recording it with both observations.

## Observations
| lane | test scope | symptom |
|---|---|---|
| A | `model-runtime-credential-rotation` | `ENOTEMPTY` temp-dir race |
| B | coding-agent shard 3/3 | `ENOTEMPTY` removing `/tmp/combined-runtime-*`, attributed to an `afterEach` cleanup race |

Both passed on a plain re-run. Both were observed on a 4-vCPU ascii box; neither reproduced in CI, which is consistent with a concurrency-sensitive cleanup rather than a logic error.

## Shared mechanism (hypothesis, needs confirmation)
An `afterEach` hook removes a temp directory while something inside it is still being written or is still open, so `rm` fails with `ENOTEMPTY` — the directory is non-empty at the instant of removal. Higher parallelism and a slower/busier machine widen the window, which matches it appearing on a small box and not in CI.

## Why this is worth fixing rather than re-running
A test that fails intermittently and passes on re-run trains everyone to re-run reds instead of reading them. That is how a real regression gets waved through: the failure becomes indistinguishable from noise, and the suite quietly stops being able to fail for a class of defect. Two separate lanes already spent time triaging this today and reached the same dead end.

## Suggested direction
- Find the writer that outlives the hook (an unawaited write, a child process, a watcher, or a stream not closed) and await/close it before teardown, rather than making the removal more forceful.
- `rm -rf`-style force removal would hide the ordering bug instead of fixing it, and would leave the underlying lifetime issue in place for anything else sharing that directory.
- Namespace each test's temp dir per-test (not per-file) so one test's teardown can never race another's fixture.

## Verification bar
A fix should be provable by a loop that reproduces the failure on the pre-fix code — if the race cannot be observed failing, the fix cannot be shown to work. Reproduce under parallelism on a constrained machine, capture the `ENOTEMPTY`, then show the same loop clean after the change.

---
Filed by the coordinating session from two lanes' gating evidence. No owner assigned.

code-yeongyu:
Fixes #1457 (the ENOTEMPTY race; the interactive-host-runtime footer observation appended to the issue is unrelated and not addressed here).

## Mechanism

`ModelRuntime.stream` wraps the credential-rotation generator in a `lazyStream`. `AssistantMessageEventStream` reports completion **at the terminal event** (`push` sets `done`), so the test's `for await` loop exits while the rotation generator is still suspended after its final yield, with `onSuccess`'s pool-state persistence write (`credential-pool-state.json`, inside the test's temp dir, plus a proper-lockfile `.lock` create/release in the same directory) still pending. The test then runs `rmSync(dir, { recursive: true })` (in-test for `policy-only-runtime-*`/`combined-runtime-*`, or `afterEach` for `rotation-stream-*`) into the middle of that write cycle -> intermittent `ENOTEMPTY`. A loaded machine widens the scheduling window, which is why it appeared on 4-vCPU gating boxes and never in CI.

A second, smaller straggler: the tests called `registerNativeProvider` without awaiting it, leaving its refresh chain running into teardown. The sibling `model-runtime-registration-refresh.test.ts` already awaits it.

## Fix (test-only; no production behavior change)

- `collectClosedRuntimeEvents`: consume the runtime stream to its terminal event, then break — the lazy stream's cancellation handler awaits the inner generator's completion, so the pool-state write lands **before** teardown. This is the "close the stream before teardown" step; it also covers the error path (for-await teardown runs the same handler).
- `await runtime.registerNativeProvider(...)` in the three ModelRuntime tests.
- No force-removal, no retries, no sleeps. Temp dirs remain per-test (`mkdtemp` in `beforeEach` / in-test).

## Reproduction (failing-first, pre-fix `main`)

8-core linux x86_64 box, node v24.20.0 (CI-matched), bun 1.4.0 first in PATH, 6 parallel vitest processes over 6 copies of this test file x 40 rounds, 8 CPU spinners (load ~17). Positive control first: a clean run must print `Tests 17 passed`.

    FAIL test/repro-race-2.test.ts > credential rotation over a pooled provider > runtime admission runs an expired stored probe once
    Error: ENOTEMPTY, Directory not empty: /tmp/rotation-stream-bTkIlj '/tmp/rotation-stream-bTkIlj'
      at rmSync(dir, { recursive: true, force: true })   (afterEach site)

    FAIL test/repro-race-5.test.ts > credential rotation over a pooled provider > runtime admits policy-only credential slots
    Error: ENOTEMPTY, Directory not empty: /tmp/policy-only-runtime-zNMcxz '/tmp/policy-only-runtime-zNMcxz'
      at rmSync(dir, { recursive: true, force: true })   (in-test site, same shape as /tmp/combined-runtime-*)

4 independent hits within the first ~8 rounds per proc. Post-fix, the identical loop (same parameters, same box) runs clean; loop counts are posted below before merge.

---
## Summary by cubic
Fixes the intermittent ENOTEMPTY teardown race in the credential-rotation ModelRuntime tests by consuming the runtime stream to its terminal event before temp-dir cleanup, and awaiting provider registration.

- The stream's terminal event fires before the inner generator's final pool-state write completes; closing the stream after that event ensures the write lands before `rmSync` runs.
- Awaiting `registerNativeProvider` prevents its refresh chain from racing teardown.
- Test-only change; no production behavior affected.
- Reproduced under CPU oversubscription on 4-vCPU boxes, never in CI.

<sup>Written for commit 7b50db54d2192c40b8d5e440c4f13125467f788f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1464?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/test/model-runtime-credential-rotation.test.ts
+++ b/packages/coding-agent/test/model-runtime-credential-rotation.test.ts
@@ -1,7 +1,7 @@
 import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import type { AssistantMessage, AssistantMessageEvent } from "@earendil-works/pi-ai";
+import type { AssistantMessage, AssistantMessageEvent, AssistantMessageEventStream } from "@earendil-works/pi-ai";
 import { fauxAssistantMessage, fauxProvider } from "@earendil-works/pi-ai";
 import { rendezvousOrder } from "@earendil-works/pi-ai/auth/pool/select";
 import type { PooledCredential } from "@earendil-works/pi-ai/auth/pool/slots";
@@ -88,6 +88,26 @@ async function collect(source: AsyncGenerator<AssistantMessageEvent>): Promise<A
 	return seen;
 }
 
+/**
+ * Consumes a runtime event stream to its terminal event and then CLOSES it.
+ *
+ * The stream reports completion at the terminal event, but the credential
+ * rotation generator behind `ModelRuntime.stream` still has to run its final
+ * pool-state persistence write after yielding that event. Removing the
+ * runtime's temp dir while that write is in flight fails with ENOTEMPTY
+ * (issue #1457), so teardown must not race it: breaking out of the loop
+ * invokes the lazy stream's cancellation handler, which awaits the inner
+ * generator's completion before returning.
+ */
+async function collectClosedRuntimeEvents(stream: AssistantMessageEventStream): Promise<AssistantMessageEvent[]> {
+	const events: AssistantMessageEvent[] = [];
+	for await (const event of stream) {
+		events.push(event);
+		if (event.type === "done" || event.type === "error") break;
+	}
+	return events;
+}
+
 describe("credential rotation over a pooled provider", () => {
 	test("runtime admission runs an expired stored probe once", async () => {
 		const faux = fauxProvider();
@@ -99,7 +119,7 @@ describe("credential rotation over a pooled provider", () => {
 			agentDir: dir,
 			allowModelNetwork: false,
 		});
-		runtime.registerNativeProvider(faux.provider);
+		await runtime.registerNativeProvider(faux.provider);
 		await runtime.refresh({ allowNetwork: false, providers: ["faux"] });
 		const pool = (await (runtime as any).loadCredentialPool()).repository as CredentialSlotRepository;
 		await pool.mutateSlotState("faux", "stored", "default", () => ({
@@ -111,13 +131,9 @@ describe("credential rotation over a pooled provider", () => {
 			blockReason: "rate_limit",
 		}));
 		faux.setResponses([fauxAssistantMessage("probe")]);
-		const events: AssistantMessageEvent[] = [];
-		for await (const event of runtime.stream(
-			faux.getModel(),
-			{ messages: [], tools: [] },
-			{ sessionId: "runtime-probe" },
-		))
-			events.push(event);
+		const events = await collectClosedRuntimeEvents(
+			runtime.stream(faux.getModel(), { messages: [], tools: [] }, { sessionId: "runtime-probe" }),
+		);
 		expect(events.some((event) => event.type === "done")).toBe(true);
 		expect(faux.getCallLog()).toHaveLength(1);
 	});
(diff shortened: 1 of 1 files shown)


code-yeongyu:
Post-fix loop counts, as promised in the description (same box, same parameters as the pre-fix RED run).

## Post-fix verification (merge candidate 7b50db5)

Box: 8-core linux x86_64 (gorky), node v24.20.0 (CI-matched), bun 1.4.0 first in PATH, workspace built via build-all.mjs (RC=0). Fixed file on the box verified byte-identical to 7b50db5: sha256 013cde5a08c5a6ec63a11e0e1dbdce32fc756ae7dc74ec88ae4941cf19941aa3.

Positive control (run first, before any negative was trusted): `Tests 17 passed (17)` — a real tally or the loop aborts.

Race loop — identical parameters to the pre-fix run (6 parallel vitest processes over 6 copies of this test file, 8 CPU spinners, ANSI-stripped summaries, abort on any missing tally):

- **6 procs x 40 rounds = 240 vitest runs completed; 240/240 logs contain `Tests 102 passed (102)` (102 = 6 copies x 17 tests) — 4,320 executions of the three race-window runtime tests.**
- **ENOTEMPTY hits: 0. Failed tests: 0. Runs without a tally (NOPARSE): 0. Leftover /tmp/{rotation-stream,policy-only-runtime,combined-runtime}-* dirs after the loop: 0.**

Pre-fix, the same loop hit in every one of the 6 procs within rounds 2-16 (4 captured verbatim in the PR description). Post-fix it is clean through round 40 in all 6.

An independent verification run (separate tree, same box, same control-first discipline) also returned clean: 72/72 completed runs each tallying `Tests 17 passed (17)`, 0 ENOTEMPTY, under load ~34.

A pre-fix full `--shard=3/3` pass on the same box (the lane-B command surface) also ran: 1 unrelated load flake (`app-server-stdio` stdout timeout), no ENOTEMPTY — consistent with the race never reproducing under CI's 2-worker fork pool, which is exactly why it survived until the file-level loop.

PR #1490, opened 2026-09-08
fix(agent,goal): recover empty tool_use turns instead of stalling the goal

code-yeongyu:
## Problem

A provider can end a turn reporting the `tool_use` stop reason while the `tool_use` content block never arrives, producing an assistant message with **zero** `toolCall` blocks.

This was observed in production on 2026-09-08 (`claude-fable-5-1` through a proxy, during upstream overload). Token accounting on the incident message shows 1035 output tokens against only 299 reasoning tokens, so a substantial tool call was generated and lost in transit. The session had an active goal and silently froze mid-task.

Two layers then disagreed about what that message meant:

| Layer | Reading | Result |
|---|---|---|
| `packages/agent/src/agent-loop.ts` | zero tool calls → nothing to execute | emits `turn_end` + `agent_end`, turn "finished" |
| `goal/continuation.ts` `isContinuableStopReason` | stop reason is `toolUse`, not `stop`/`length` | refuses to queue a continuation |

The loop said the turn was over; the goal layer said it had not ended cleanly. Nothing was pending, nothing was running, and the goal stayed `active` forever. The user only noticed because the session went quiet.

## Fix

Three layers, each with a distinct responsibility, all agreeing on one definition of a coherent terminal state.

**1. Recover the lost tool call (`packages/agent/src/empty-assistant-recovery.ts`)**

A terminal message claiming `toolUse` with no tool call is structurally impossible, so it is treated like the empty-`stop` defect the file already handles: retry the request once, reusing the same closure so the retry reissues the identical model/context/options. If the retry is malformed too, the turn finalizes as `stopReason: "error"` with an `empty_tool_use_response_recovery` diagnostic rather than a silent clean end.

The wrapper deliberately stays scoped to models that already need stream recovery. Wrapping replaces the provider stream with a buffering proxy that does not carry the underlying stream's liveness surface (`trackLocalWork` / `hasPendingLocalWork`), which the loop's idle watchdog reads — wrapping unconditionally broke 7 timeout/abort tests. Universal coverage is provided by layer 2 instead.

**2. Normalize the contradiction (`packages/agent/src/assistant-terminal-state.ts`)**

`demoteToolUseWithoutToolCalls` is the exact inverse of the existing `promoteStopWithPendingToolCalls`, composed with it in the loop. It counts raw `toolCall` blocks so a cursor-exec-resolved call — a legitimately completed tool call — is never demoted. This applies to every model and needs no stream rewrapping.

**3. Resolve the deadlock (`goal/`)**

`isMalformedToolUseTurn` distinguishes the anomaly from a deliberate terminating-tool stop. At `agent_end` the anomaly routes through the existing `afterProviderFailure` / `providerRecovery` lane — infrastructure breakage, same class as a terminal provider error — and `lastTurnWasMalformedToolUse` is threaded into the verdict input so the immediate path admits only this case.

`isContinuableStopReason` is deliberately **not** widened to accept `toolUse`; that would wrongly admit deliberate terminating-tool stops. A turn a tool genuinely ended still waits for the user.

Every existing admission guard still bounds the recovery: single-flight latch, `GOAL_CONTINUATION_CAP` (8), `GOAL_REPETITION_HASH_STREAK` (3), and `GOAL_UNATTENDED_CONTINUATION_LIMIT` (150), which is the concrete bound if a provider returns the malformed shape persistently.

## Verification

Both new tests fail on `main` and pass here.

`packages/coding-agent/test/suite/regressions/empty-tool-use-goal-stall.test.ts` reproduces the incident end to end. Before the goal fix:

```
AssertionError: expected [] to have a length of 1 but got +0
 Tests  1 failed | 1 passed (2)
```

After:

```
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

`packages/agent/test/empty-tool-use-native-retry.test.ts` asserts the retry reissues with the **same model and context** (a call-count-only assertion would have passed an earlier broken revision where the retry factory was invoked with no arguments) and that a twice-malformed response terminates as `error` with the right diagnostic.

Full suites:

```
packages/agent          Test Files  43 passed (43)   Tests  555 passed | 1 skipped (556)
goal suites (11 files)  Test Files  11 passed (11)   Tests  199 passed (199)
```

Changelog gate:

```
changelog-gate: PASS - changes.md coverage complete (1 production path(s) covered);
changelog entry updated (packages/agent/CHANGELOG.md, packages/coding-agent/CHANGELOG.md)
```

Three pre-existing test files (`goal-modules`, `goal-monitor-continuation`, `issue-506-monitor-delayed-cap`) construct `GoalContinuationInput` directly and were updated for the new required field.

## Notes

- Both `changes.md` trackers are updated (`packages/agent/src/changes.md`, `goal/changes.md`).
- The upstream cause is a provider/proxy transport defect. This PR makes the client survive it rather than claiming to fix the provider.

---
## Summary by cubic
Fixes a stall where an active goal froze when a provider ended a turn reporting `tool_use` but lost the tool call block in transit. The agent now recovers that contradictory response, and the goal layer resumes through the existing provider-recovery path instead of reading it as a deliberate stop.

**Bug Fixes**
- Retries the request once with the same model and context when a terminal `toolUse` message has no tool call; a second malformed response ends the turn as `error` with an `empty_tool_use_response_recovery` diagnostic.
- Demotes the contradictory `toolUse` stop reason to `stop` for every model when no raw tool call block is present, stamping an `empty_tool_use_terminal_state` diagnostic so the demotion stays identifiable downstream; cursor-resolved calls (which are complete) are never demoted.
- `isMalformedToolUseTurn` accepts either the raw `toolUse` shape or the demoted `stop` carrying that diagnostic, so a plain clean stop is unaffected.
- Only wraps models that already use stream recovery, so the idle watchdog's liveness surface is preserved; a new test pins that unwrapped native models still get the demotion.
- Routes malformed tool-use turns through the provider-recovery lane for active goals; a turn a tool legitimately ended still waits for the user.
- All existing continuation bounds (single-flight, cap 8, repetition streak 3, unattended limit 150) still apply.

<sup>Written for commit bf17b5eb9ae009bdeb21dc3bf0a501be8dff2206. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- Recover a provider response that reports the `tool_use` stop reason while carrying no tool-call block. The turn is retried once on models that already use stream recovery, and the contradictory terminal state is demoted to a coherent stop for every model, so a lost tool call no longer ends the turn silently.
+
 ### Removed
 
 ## [2026.9.7-2] - 2026-09-07
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -16,6 +16,7 @@ import {
 } from "@earendil-works/pi-ai";
 import {
 	createTerminalFailureAssistantMessage,
+	demoteToolUseWithoutToolCalls,
 	isStreamIdleTimeoutError,
 	normalizeTerminalAssistantMessage,
 	promoteStopWithPendingToolCalls,
@@ -262,7 +263,7 @@ async function runLoop(
 				withEmptyAssistantRecovery(requestConfig.model, streamFunction),
 				isInitialProviderRequest ? config.timeoutMs : requestConfig.timeoutMs,
 			);
-			const message = promoteStopWithPendingToolCalls(streamed.message);
+			const message = demoteToolUseWithoutToolCalls(promoteStopWithPendingToolCalls(streamed.message));
 			const providerToolResults = streamed.providerToolResults;
 			newMessages.push(message);
 			const toolResults: ToolResultMessage[] = [];
--- a/packages/agent/src/assistant-terminal-state.ts
+++ b/packages/agent/src/assistant-terminal-state.ts
@@ -29,6 +29,12 @@ export class ProviderRetryWatchdogAbortError extends Error {
 	}
 }
 
+export function demoteToolUseWithoutToolCalls(message: AssistantMessage): AssistantMessage {
+	// Count raw blocks: cursor-resolved calls are legitimate completed tool calls and must not be demoted.
+	if (message.stopReason !== "toolUse" || message.content.some((block) => block.type === "toolCall")) return message;
+	return { ...message, stopReason: "stop" };
+}
+
 export function promoteStopWithPendingToolCalls(message: AssistantMessage): AssistantMessage {
 	if (message.stopReason !== "stop") return message;
 	if (!message.content.some((block) => block.type === "toolCall")) return message;
--- a/packages/agent/src/assistant-terminal-state.ts
+++ b/packages/agent/src/assistant-terminal-state.ts
@@ -29,10 +29,24 @@ export class ProviderRetryWatchdogAbortError extends Error {
 	}
 }
 
+/**
+ * Marks a terminal message whose `toolUse` stop reason was demoted because the provider sent no
+ * tool call. Demotion rewrites the stop reason, so this diagnostic is the only surviving evidence
+ * that the turn was malformed rather than a clean stop; downstream recovery keys on it.
+ */
+export const EMPTY_TOOL_USE_DEMOTION_DIAGNOSTIC = "empty_tool_use_terminal_state";
+
 export function demoteToolUseWithoutToolCalls(message: AssistantMessage): AssistantMessage {
 	// Count raw blocks: cursor-resolved calls are legitimate completed tool calls and must not be demoted.
 	if (message.stopReason !== "toolUse" || message.content.some((block) => block.type === "toolCall")) return message;
-	return { ...message, stopReason: "stop" };
+	return {
+		...message,
+		stopReason: "stop",
+		diagnostics: [
+			...(message.diagnostics ?? []),
+			{ type: EMPTY_TOOL_USE_DEMOTION_DIAGNOSTIC, timestamp: Date.now(), details: {} },
+		],
+	};
 }
 
 export function promoteStopWithPendingToolCalls(message: AssistantMessage): AssistantMessage {
--- a/packages/agent/src/changes.md
+++ b/packages/agent/src/changes.md
@@ -45,6 +45,27 @@
 
 # Changes
 
+## 2026-09-08 - Recover empty native tool-use responses
+
+### What changed
+
+- `packages/agent/src/empty-assistant-recovery.ts`: retry terminal native `toolUse` responses with no tool-call blocks once, then surface an error and telemetry diagnostic; preserve existing empty-stop gating.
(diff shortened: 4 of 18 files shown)


code-yeongyu:
## Review follow-up

Both findings from the gate review are addressed.

### BLOCKER — malformed provenance destroyed by the loop's demotion: **fixed**

Confirmed by probe: `demoteToolUseWithoutToolCalls` rewrote `stopReason` to `"stop"` at `agent-loop.ts:266` **before** the demoted message was pushed into `newMessages` (line 268), and every `agent_end` emits `newMessages`. So `isMalformedToolUseTurn` could never match on the live path.

Measured routing for the demoted shape, before the fix:
- `isMalformedToolUseTurn(demoted)` → `false` (stop reason already rewritten)
- `didTerminalProviderErrorEndTurn(...)` → `false` (requires `stopReason === "error"`)

so the turn fell through to `afterAgentEnd` rather than the intended `providerRecovery` lane.

Fix: the demotion now stamps an `empty_tool_use_terminal_state` diagnostic (exported as `EMPTY_TOOL_USE_DEMOTION_DIAGNOSTIC`), and `isMalformedToolUseTurn` accepts either the raw `toolUse` message or a demoted `stop` carrying that diagnostic. A plain clean stop is unaffected. After the fix the same probe gives `isMalformedToolUseTurn(demoted) === true` with `didTerminalProviderErrorEndTurn === false`, i.e. the anomaly now enters provider recovery on its own merits.

One correction to the finding's impact: because `didAgentEndCleanly` accepts `"stop"`, the goal did still receive a continuation through the ordinary lane, so this was a **lane-routing** defect rather than a renewed user-visible stall. It still needed fixing — the recovery guards and continuation accounting differ per lane — but the stall itself was closed by the original change.

### SHOULD-FIX — no coverage for the unwrapped native-model split: **fixed**

Added a case using a model id matching neither `/claude/i` nor `/kimi/i` with no tool-call format. It asserts `withEmptyAssistantRecovery` returns the stream function **identically** (`expect(recovered).toBe(streamFunction)`), that no retry fires, and that the loop's demotion still yields a coherent stop plus the diagnostic. This is the documented, intentional split: wrapping unconditionally replaces the provider stream and drops the `trackLocalWork`/`hasPendingLocalWork` liveness surface the idle watchdog reads, which broke 7 timeout/abort tests when attempted.

### Verification after both changes

```
packages/agent      Test Files  43 passed (43)   Tests  555 passed | 1 skipped (556)
goal suites         Test Files  10 passed (10)   Tests  146 passed (146)
empty-tool-use-goal-stall.test.ts                Tests  4 passed (4)
empty-tool-use-native-retry.test.ts              Tests  3 passed (3)
tsc --noEmit        NO TS ERRORS
biome check         clean
changelog-gate      PASS
```

On the remaining test-quality notes: the cursor-exec-resolved negative path is covered indirectly by the pre-existing `idle-after-completed-tools` / `cursor-exec-loop` suites (both green, and the demotion deliberately counts raw `toolCall` blocks so a cursor-resolved call is never demoted). A full end-to-end "persistently malformed provider exhausts the guards" test would need a live multi-turn harness; the guard bound is `GOAL_UNATTENDED_CONTINUATION_LIMIT` with the cap and repetition guards terminating earlier, and those bounds are already pinned by `issue-1139-goal-unattended-backstop` and `issue-447-goal-continuation`.

PR #1496, opened 2026-09-08
feat(auth): add immutable account display names

### LINKED ISSUE #1495: feat(auth): add immutable display names for multi-account OAuth slots
## Summary

Multiple authenticated OpenAI Codex and Claude SDK OAuth accounts need human-readable labels without renaming the operational slot identity. Today slot names participate in credential lookup, refresh, pins, failover, HRW affinity, and Claude session continuity, so renaming them is unsafe.

## Proposed contract

- Keep the existing slot `name` immutable.
- Persist optional `displayName` metadata with saved account slots.
- Expose secret-free account descriptors and a login completion receipt containing the committed slot ID.
- Provide an atomic rename/clear operation validating trimmed, non-empty, provider-local unique display names.
- Render `displayName (name)` while selectors, pins, refresh, failover, and affinity keep using `name`.
- Preserve legacy unnamed slots and avoid materializing environment credentials.

## User-facing surface

Use existing account command conventions, adding `rename <id> <display name...>` for the generic account command and provider-specific commands where supported. Login-time naming must be optional; cancellation leaves the successful login usable.

## Acceptance criteria

- OpenAI and Claude display-name round trips, validation failures do not mutate persisted state, and legacy slots still render their ID.
- Selection, pinning, OAuth refresh, failover, HRW affinity, health state, and Claude session bindings continue using the unchanged slot ID.
- Account/status/RPC/app-server output exposes only safe IDs and display names, never credentials or auth headers.
- Two persisted logins and Claude import are covered, including the existing envelope/append path.

I am preparing a focused contribution from a fork and will link it here.

eddieparc:
## Summary

Closes #1495.

Adds optional human-readable display names for saved OpenAI Codex and Claude SDK OAuth account slots while retaining each slot `name` as its immutable operational identity.

- optional post-login labels and `rename <id> <display name...>` / clear flows
- trimmed, provider-local unique labels; legacy slots stay name-only
- labels render as `displayName (name)`; pins, refresh, affinity, failover, health, and Claude continuity stay ID-based
- CLI, RPC, app-server, and account surfaces expose safe metadata only

## Verification

- RED: 26 focused failures before implementation
- GREEN: 30 focused regressions; related suites: 241 tests
- `bun run check` and `bun run build`
- isolated source CLI/RPC/TUI QA: `local-ignore/qa-evidence/20260908-account-display-names/REPORT.md`

## Residual risk

Live OAuth is intentionally excluded; QA uses synthetic isolated credentials. Standalone Claude `auth check` returns `provider_not_found` in the isolated source path, while safe descriptor and RPC account surfaces pass.

---
## Summary by cubic
Adds optional human-readable display names for saved OpenAI Codex and Claude SDK OAuth accounts while keeping each slot's `name` as its immutable operational identity. Account descriptors now include an optional `displayName`; listings and the footer render `displayName (name)` when set, while pins, refresh, failover, affinity, health, and Claude session continuity continue to use only the immutable `name`. Legacy accounts are unchanged and display their ID alone.

**New Features**
- `/account`, `/gpt-account`, `/claude-account`, and `/cursor-account` gain `rename <id> <display name...>` and `clear-name <id>`; environment accounts cannot be renamed, and rename refuses to promote a provider-managed sentinel flat credential (keyed on credential shape, not provider id).
- Login flows emit a committed-account receipt only after the credential write succeeds; add commands then offer optional post-login naming only for machine-generated slot IDs, and blank or cancelled input keeps the login usable.
- Display names are NFC-normalized, bounded at 32 terminal columns with internal whitespace collapsed, and render-unique per provider under a fold of case, Unicode compatibility forms, invisible code points, and Cyrillic lookalikes; rejected names write nothing and emit no event.
- The receipt is derived by comparing provider-owned envelopes against the locked current pool, so ambiguous envelopes produce no receipt.

**Bug Fixes**
- A bare `.` submitted on a session that already has messages no longer paints a user echo in the TUI; it stays the hidden manual-continue shortcut.
- The footer colours its right side from the structural runs that produced it, so a display name containing `)` or `:` no longer mutes the wrong span; wide labels are truncated at 24 columns so the account indicator stays visible.
- The Julia error-parity test waits for the kernel's `ready` frame before dispatching, so a cold boot reports which phase overran instead of a bare 40-second timeout.

<sup>Written for commit 7a0f115f7ee7a3992406c858138aec32215bf2f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1496?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (changes requested):
## Verdict

Requesting changes. The core identity discipline of this PR is sound and I could not find a credential leak: `CredentialAccountSummary`, `RpcProviderAccount` and `ProviderAccount` still carry only name/health, both transports project through the single sanitized seam (`src/modes/rpc/connection-handler.ts:1595`, `src/modes/app-server/server/account.ts:100`), `AccountLoginReceipt` is secret-free and is emitted only after the store write commits (`packages/ai/src/models.ts:639`), and pins/HRW affinity/refresh/failover/Claude continuity stay keyed on the immutable `name`. What is not ready is the presentation and validation boundary the feature is *made of*: the label is spliced into a footer segment that is re-parsed by regex, so every named account degrades the footer rendering; the documented uniqueness and length invariants are enforced with `trim`/`toLowerCase`/`String.length`, which does not hold for whitespace-, NFC/NFD- or width-differing names; the "legacy flat account" guard is keyed on one provider id while a sibling lane uses the identical sentinel shape; and the rename verb is exposed for every provider while only two lanes render it. Findings 1-4 below were reproduced by executing the verbatim PR-head copy of `packages/ai/src/auth/pool/slots.ts` under bun 1.3.14 on a throwaway remote Linux workspace (since deleted); the footer cases additionally use the two regexes copied verbatim from `footer.ts:85` and `footer.ts:90`. I did not run the repository test suite, so no claim below rests on it.

### 1. `modes/interactive/components/footer.ts:52` + `:56` — every named account breaks the footer's right-side colouring

`accountFooterSuffix` now returns `@displayName (name)`, and that string is interpolated *inside* the parenthesised provider segment at `footer.ts:240-246` (`(${state.model.provider}${accountSuffix}) `). `colorRightSide` re-parses that rendered text with `^\(([^)]+)\) (.*)$` (`footer.ts:85`); the label's own `)` makes the match fail, so the provider prefix is no longer muted and the whole segment is painted with the model accent. With a `:` in the display name the `^(.+):([^:]+)$` split at `footer.ts:90` cuts inside the label and the model id itself renders as the dim "thinking level" suffix. Reproduced:

```
3 unnamed: {"providerSegmentMuted":true,  "accent":"gpt-5-codex", "dim":":high"}
3 named  : {"providerSegmentMuted":false, "accent":"(openai-codex@Work account (second)) gpt-5-codex", "dim":":high"}
3 colon  : {"providerSegmentMuted":false, "accent":"(openai-codex@Work", "dim":": main (second)) gpt-5-codex"}
```

Fix: stop round-tripping structure through a rendered string — colour the provider/account/model segments from the values that produced them (pass them into `colorRightSide` instead of re-matching), or keep the label out of the parenthesised segment and render `@name` in the footer while `displayName (name)` stays in `/account list`. Add a footer test with a named account and with a `:` in the name.

### 2. `packages/ai/src/auth/pool/slots.ts:66-71` — "unique per provider (case-insensitive)" is not what the code enforces

Uniqueness compares `trim()`ed values with `toLowerCase()` only, so two accounts of one provider can hold labels that render identically. Reproduced (case-only duplicate is correctly rejected; the rest are accepted):

```
1 [control: case-only dup] rejected -> Display name is already used by another account for this provider.
1 [inner whitespace]  ACCEPTED -> ["Work account","Work  account"]
1 [NFD vs NFC]        ACCEPTED -> ["Café","Café"]
1 [cyrillic homoglyph] ACCEPTED -> ["Work","Вork"]
```

Consequence: `/account list`, `/gpt-account list` and the footer can show two indistinguishable labels, which defeats the only purpose of the metadata, and the guarantee stated in `modes/interactive/help-content.ts:79` and `docs/rpc.md:13` is false. Fix: normalise in `accountDisplayName` (`value.normalize("NFC")`, collapse internal whitespace runs) and compare with a case-fold of the normalised form; persist the normalised string so storage and comparison agree.

### 3. `packages/ai/src/auth/pool/slots.ts:41-42` — the 1-80 limit counts UTF-16 code units, and blank-looking names pass

`trimmed.length <= 80` is neither graphemes nor columns. Reproduced: 80 CJK code points are accepted and produce a 90-character / ~170-column label, while 41 emoji (82 code units) are rejected; names made only of non-whitespace invisible code points are accepted, as is a leading combining mark:

```
2 80 CJK code points accepted: true | code units: 80
2 label for that name is 90 chars / 170 terminal columns
2 41 emoji (82 code units) rejected: true | code points: 41
2 invisible/combining "ㅤ"    -> "ㅤ"        (U+3164 HANGUL FILLER)
2 invisible/combining "⠀⠀⠀" -> "⠀⠀⠀"    (U+2800 BRAILLE BLANK)
2 invisible/combining "́abc"  -> "́abc"      (leading U+0301)
```

Consequence: a legal 170-column label cannot fit the footer's provider segment, so `planFooterLayout` falls back to `right.minimal` (`footer.ts:243-246`) and the account indicator silently vanishes from the footer instead of being truncated; and an account can be labelled with something that renders as blank or as a mark glued onto the preceding character. Fix: bound the label by `visibleWidth` (the footer already imports it) at something the footer can hold — e.g. 32 columns — require at least one visibly-advancing character, reject leading combining marks, and truncate the label with an ellipsis at the footer boundary rather than dropping the segment.

### 4. `src/core/credential-accounts.ts:150` — the sentinel guard is keyed on a provider id, not on the credential shape

The guard that stops a Claude sentinel from being read as a legacy flat account is `provider === "claude-sdk-oauth" && !("accounts" in current …)`. `cursor-cli-oauth` defines the identical sentinel envelope (`src/core/extensions/builtin/cursor-cli-oauth/accounts.ts:21-28`) and gets the legacy-promotion path instead. Reproduced against the real `renameSlotDisplayName` with a cursor-shaped credential that has no `accounts` key:

```
4 cursor sentinel rename -> {"type":"oauth","access":"cursor-cli-oauth-managed","refresh":"cursor-cli-oauth-managed",
   "expires":4102444800000,"accounts":[{"name":"default","source":"login",
   "access":"cursor-cli-oauth-managed","refresh":"cursor-cli-oauth-managed","expires":...,"displayName":"My cursor"}]}
```

A cosmetic rename writes a fabricated `default` login slot whose `access`/`refresh` are the literal sentinel strings; `/account cursor-cli-oauth list` would then report it as a real saved login and `projectSlot` would hand those strings to the provider. To be fair to the change: I could not find a current writer that leaves a cursor credential without an `accounts` key (`emptyCredential()` always sets `accounts: []`), so this is a latent hole rather than a live bug — but the guard should be shape-based, not string-based. Fix: refuse promotion whenever `!Array.isArray(current.accounts)` and the flat oauth fields look like a provider sentinel (or invert to an explicit allowlist of lanes that documented flat legacy credentials, currently `openai-codex`).

### 5. `src/core/extensions/builtin/cursor-cli-oauth/account-command.ts:175` (and `:180-181`) — display names are write-only for the cursor lane

`/account <provider> rename …` accepts any provider string (`account/index.ts:54` → `credential-accounts.ts:141`) and `help-content.ts:77` documents it generically, but the cursor lane's own command still renders `account.name` for the listing, the "Pinned account" line and the "Affinity pick" line. A user who names a cursor account through the generic command sees the label in `/account cursor-cli-oauth list` and nowhere in `/cursor-account`. Fix: use `accountLabel` in the three cursor spots exactly as `claude-sdk-oauth/account-command.ts:148,153` now does, or reject rename for providers that do not render it.

### 6. `src/core/extensions/builtin/account-display-name.ts:37-41` — `/claude-account add` now asks for a name twice, and no test can see it

The Claude lane already prompts for the *slot id* on every login that has an existing account: `claude-sdk-oauth/oauth-login.ts:80-90` asks "Name for this account (existing: …)" and that prompt is relayed to the same `ctx.ui.input` (`provider-composer.ts:328` → `oauth-login-interaction.ts:71`). Immediately afterwards `promptAccountDisplayName` asks "Display name for account &lt;that same id&gt; (optional)". The new test cannot observe this: `1495-account-display-commands.test.ts:99-125` starts from an empty pool, so `promptAccountName` returns `"default"` without prompting and `expect(dialogs).toHaveLength(1)` (line 125) holds. Fix: prompt for a display name only when the committed id was machine-generated — `appendLoginSlot`'s `onAllocated` is the one place that knows this, so extend `AccountLoginReceipt` with e.g. `origin: "generated" | "provider"` and gate `promptAccountDisplayName` on it; then add a case with a pre-existing account so the two-dialog path is covered.

### 7. Verification gaps in the new tests

- `test/suite/regressions/1495-account-display-names.test.ts:82,85,135`: `await expect(...).rejects.toThrow()` with no matcher passes on *any* rejection, so the suite cannot distinguish "the label validator rejected this label" from "the sentinel guard rejected the credential" or a mistyped provider. Pin the messages (`/1-80 characters/`, `/already used/`, `/not found/`).
- `test/suite/regressions/1495-account-display-identity.test.ts:36`: `new CredentialSlotRepository()` takes the default agent-dir path and then writes a never-expiring `auth_error` block for `claude-sdk-oauth`/`second` (line 37) into that shared sidecar, and the assertion at line 47 compares against whatever that file holds. It is safe today only because `test/setup.ts` quarantines the agent dir; every sibling test passes an explicit temp path (`test/credential-accounts.test.ts:24`, `test/model-runtime-credential-rotation.test.ts:25`, `test/credential-pool-state-store.test.ts:26`). Pass a temp path here too.
- The wire contract the docs now promise (`docs/rpc.md:13`, `docs/app-server.md:55`) has no test: `test/suite/app-server-account.test.ts:132` still asserts only unnamed accounts, so a regression in the projection at `credential-accounts.ts:112-117` would ship silently on both transports. Add one named account to that fixture and to the RPC `get_provider_accounts` test.
- Nothing covers findings 1-3: no footer test with a named account, and no wide/NFC-NFD/invisible-name case anywhere in the four new files.

Nothing here questions the identity model — that part is well argued and well tested. Findings 1-3 are the blocking ones because they are exactly the behaviour a display-name feature exists to provide.

code-yeongyu:
All seven findings from the review are addressed on the rebased branch (head `5c68e3f7f`, force-pushed onto `feat/account-display-nicknames`; based on current `main` `e66c1553a`). Tests were executed remotely on Linux under bun 1.3.14; every fix carries mutation evidence below (the corresponding production line was removed, the test shown failing, the line restored, the test shown passing).

**1. Footer colouring broke on `)` / `:` in the label** — `footer.ts` no longer round-trips structure through the rendered string. `colorRightSide` now receives the runs that produced the segment — `(provider@account) ` muted, `⚡` warning, model id accent, `:thinking` dim — and clips each run to what the layout kept, instead of re-parsing with `^\(([^)]+)\) (.*)$` / `^(.+):([^:]+)$`. An account label is truncated with `…` at 24 columns so a wide label narrows the provider segment instead of pushing the plan onto `right.minimal` and silently dropping the account indicator. Commits: `83dfb8db2`, `763b76dd6`. Tests: `test/footer-account-label-coloring.test.ts` asserts the exact muted/accent/dim composition with a `Work: dev (b) (second)` label and the ellipsis truncation of a 32-column CJK label.
- Mutation (M1, paint every run accent): `FAIL ... keeps the provider prefix muted and the model accented when the label contains ) and :` → restored: `Tests 3 passed (3)`.
- Mutation (M2, remove the 24-column truncation): `FAIL ... truncates a wide label with an ellipsis instead of dropping the account segment` (`to contain '(openai-codex@'`) → restored: `Tests 3 passed (3)`.

**2. Uniqueness was not render-uniqueness** — `accountDisplayName` now stores the NFC-normalized, whitespace-run-collapsed form, and `renameSlotDisplayName` compares a fold of case + NFKC compatibility forms + invisible code points + Cyrillic lookalikes. `Work account` vs `Work  account`, NFC/NFD `Café`, fullwidth `Ｗork`, `Work` vs `Workㅤ` (U+3164 smuggled inside), and `Bork` vs Cyrillic-`Вork` are all rejected; distinct renderings (`Work account` vs `Workaccount`) still coexist. Commit: `83dfb8db2`. Tests: `packages/ai/test/account-display-names-unicode.test.ts` + the duplicate block in `1495-account-display-names.test.ts`.
- Mutation (M3, fold back to plain `toLowerCase`): `FAIL ... rejects whitespace, NFC/NFD, fullwidth, invisible and homoglyph duplicates` (`expected [Function] to throw an error`) → restored: `Tests 7 passed (7)`.

**3. Length counted UTF-16 units; blank-looking names passed** — the bound is now 32 terminal columns measured per grapheme cluster (`displayNameColumns`, `Intl.Segmenter` + wide-char/emoji rules), so 20 CJK code points (40 columns) are rejected while 16 emoji (32 columns) pass; a name must contain at least one visibly advancing character (U+3164, U+2800 rejected), may not begin with a combining mark (leading U+0301 rejected), and internal whitespace runs collapse before measuring. Help/docs (`help-content.ts`, `docs/rpc.md`, `docs/app-server.md`) state the real rules. Commits: `83dfb8db2`, `e654f8b23`.
- Mutation (M4, bound back to `.length <= 80`): `FAIL ... measures the limit in terminal columns, not UTF-16 code units` (`expected '中×20' to be undefined`) → restored: `Tests 7 passed (7)`.
- Mutation (M5, drop the blank check): `FAIL ... rejects names that render as blank or begin with a combining mark` (`expected 'ㅤ' to be undefined`) → restored: `Tests 7 passed (7)`.

**4. Sentinel guard keyed on a provider id** — `renameCredentialAccount` now refuses promotion on the credential *shape*: an OAuth credential with no `accounts` array whose identical `access`/`refresh` end in `-managed`. A cursor-shaped sentinel is no longer promotable into a fabricated `default` slot, and the guard no longer names any provider. Commit: `83dfb8db2`. Test: `refuses to promote a provider-managed sentinel flat credential for any provider lane` covers both lanes.
- Mutation (M6, revert to the `provider === "claude-sdk-oauth"` keying): `FAIL ... promise resolved "undefined" instead of rejecting` (the cursor sentinel rename succeeded, i.e. the fabrication reproduced) → restored: `Tests 12 passed (12)`.

**5. `/cursor-account` never rendered labels** — the listing, "Pinned account" and "Affinity pick" lines now render `accountLabel(...)` exactly like the Claude lane; `CursorCliAccountSlot` carries `displayName`. Commits: `83dfb8db2`.
- Mutation (M7, render `account.name` again): `FAIL ... renders display names the generic rename command can write` → restored: `Tests 18 passed (18)`.

**6. `/claude-account add` asked for a name twice** — `AccountLoginReceipt` gained `origin: "generated" | "provider"`, set by `appendLoginSlot`'s `onAllocated` (the one place that knows who chose the id), and `promptAccountDisplayName` offers the display-name dialog only for `origin === "generated"`. The Claude lane — whose envelope adapter names slots through its own prompt — never sees a second dialog; the OpenAI lane keeps the offer on every generated id. Help/docs updated to match. Commits: `83dfb8db2`, `e654f8b23`.
- Mutation (M8, drop the origin gate): `FAIL ... asks for the account name exactly once per login and never offers a display name` (`expected [...] to have a length of +0 but got 1` — the double prompt reproduced) → restored: `Tests 10 passed (10)`.

**7. Verification gaps** —
- Bare `rejects.toThrow()` are gone: rejections are pinned (`/1-32 terminal columns/`, `/already used/`, `/not found/`, `/not found|No stored credential/`).
- `1495-account-display-identity.test.ts` now constructs `new CredentialSlotRepository(<temp dir>/credential-pool-state.json)` like every sibling test, instead of the default agent-dir path.
- The wire contract is tested on both transports: `test/suite/app-server-account.test.ts` and the RPC `get_provider_accounts` case in `rpc-auth-and-connection-handler.test.ts` now assert `displayName: "Work: main (client)"` on a named account (label deliberately contains `)` and `:`).
- Findings 1–3 coverage: `packages/ai/test/account-display-names-unicode.test.ts` (CJK/emoji columns, NFC/NFD, fullwidth, Cyrillic homoglyph, U+3164/U+2800/leading-mark blanks) and `test/footer-account-label-coloring.test.ts` (footer rendering).
- Mutation (M9, drop the `displayName` projection in `summarizeCredentialAccounts`): `FAIL` in both `app-server-account.test.ts` and `rpc-auth-and-connection-handler.test.ts` (`to deeply equal` / `to match object`) → restored: `Tests 18 passed (18)`.
- (The temp-path repository change is test hygiene; there is no production line to mutate for it.)

**Suite results** (bun, remote Linux): `packages/ai`: 2691 passed, 1 suite fails pre-existing (`codex-apply-patch-wire-schema.test.ts` cannot resolve `@earendil-works/pi-tui` — reproduced identically with pristine `origin/main` `packages/ai` files, i.e. an unbuilt-dist artifact, not this change). `packages/coding-agent`: 10323 passed / 98 failed — the 98 are byte-for-byte the pristine-`origin/main` baseline failure set in the same environment (same 28 files, same per-file counts), so no regressions; all new and touched tests pass. `biome check --error-on-warnings .` and `tsc --noEmit` are clean for every touched path.

One extra fix found while re-running the suite: `footer-fast-mode-icon.test.ts` regressed because `planFooterLayout`'s right-truncation appends ANSI reset codes to `rightPlain` even with an empty ellipsis; run-clipping could then cut an escape mid-sequence. `colorRightSide` strips terminal sequences before walking runs (`763b76dd6`); that test now passes again.

code-yeongyu (approved):
All 7 findings resolved and mutation-proved (9 mutations, each RED then GREEN). The footer colours from the runs that produced the text instead of re-parsing it; uniqueness folds NFKC + case + invisibles + Cyrillic lookalikes; length is 32 terminal columns per grapheme with a visible-character requirement; the sentinel guard keys on credential shape rather than the provider id; cursor labels render; the double prompt is gated by receipt origin. Rebased onto d2f5f28fa with a file-by-file audit confirming no main-side content was dropped. packages/ai 2691 passed; coding-agent failing set identical to a pristine origin/main run in the same environment (zero regressions). Approving.

code-yeongyu:
## CI red on `dc0a0be6a`: root cause and fix

### Failing test

CI run [34498099869](https://github.com/code-yeongyu/senpi/actions/runs/34498099869), job `Test (workspaces + scripts)` (`102941627827`), step `Workspace tests (all but coding-agent)`:

```
packages/senpi-codemode/test/jl-error-parity.test.ts
  Julia error parity > Given an undefined variable when a cell runs then the exception type and name reach the result
  Error: Test timed out in 40000ms.
```

That annotation is the only test failure in the whole run; the step's plain-text log is cut off right after the `packages/ai` banner (the same truncation happens in green `main` runs, so `packages/ai` was a red herring — I re-ran its full suite and it is green, 278 files / 2705 tests).

### Root cause

`SubprocessKernel` arms a cell's `timeoutMs` when the run is **dispatched**, and dispatch waits for the interpreter's `ready` frame (`pumpRuns()` returns early while `!processReady`). The error-parity suite called `kernel.run(...)` straight after `JuliaKernel.start(...)`, so Julia's cold boot (interpreter startup plus `runner.jl`/`prelude.jl` evaluation) sat **outside every budget except the 40s test-level ceiling** — the cell budget had not started yet. Several Julia kernels boot concurrently across this package's suites (`jl-kernel.test.ts` starts five more), and on a loaded 4-core runner that boot overran the ceiling. The sibling suite already budgets Julia cold boot explicitly (`coldStartTimeoutMs = 60_000`, test timeout `120_000`, added in `5262475c9` / `8f426e7b2`); this file was the one that did not, so it is the one that lost the race.

The failure is not caused by this PR's production changes: nothing here touches `packages/senpi-codemode`, and the kernel boot path imports neither `packages/ai` nor `packages/coding-agent`.

### Fix

`test(senpi-codemode): wait for Julia kernel readiness in the error-parity suite` — the test now subscribes to the kernel's `ready` frame **before** starting it (`onMessage`, which `SubprocessRunQueue` forwards for `ready`) and awaits that signal under its own named deadline, then runs the assertion cell on the cold-start budget the sibling suite uses. The assertions on `UndefVarError` and `missing_var_xyz` are unchanged; nothing is skipped, relaxed or deleted.

### Evidence (reproduced on a Linux x86_64 box with Julia 1.11.6 installed)

Mutation proof of every assertion, against the exact committed bytes:

| step | production mutation | result |
| --- | --- | --- |
| baseline | none | `✓ ... 665ms` |
| assertion binds | `runner.jl`: `senpi_error` returns `"cell failed"` instead of `sprint(showerror, error)` | `× AssertionError: expected 'cell failed' to contain 'UndefVarError'` |
| restored | – | `✓ ... 622ms` |
| readiness wait binds | `runner.jl`: `senpi_emit(Dict("type" => "ready"))` withheld | `× Error: Julia kernel did not report ready within 60000ms` (60225ms) |
| same mutation, **old** test | – | `× Error: Test timed out in 40000ms.` (40022ms) — byte-identical to the CI failure |
| restored | – | `✓ ... 625ms` |

Suites (Julia on `PATH`, `npm ci --ignore-scripts && npm run build` first):

- CI's exact step 10 command: `node scripts/run-workspaces.mjs --if-present --workspace packages/agent --workspace packages/ai --workspace packages/pty --workspace packages/senpi-codemode --workspace packages/server --workspace packages/tui test` → exit 0 (agent 44/559, ai 278/2705, pty 7/66, senpi-codemode 116/806, server 7/51, tui pass).
- `packages/senpi-codemode` full suite pinned to 4 CPUs (`taskset -c 0-3`) → 116 files / 806 tests passed.
- `npm run check` → exit 0.
- Full `packages/coding-agent` suite → 1304 files passed, 1 pre-existing failure (`test/auth-storage.test.ts > keeps a coalesced reload alive while another credential reader is waiting`) that reproduces **5/5 on pristine `main` (00355479f)** on the same box and is green on GitHub runners — unrelated to this PR, reported separately.

### Also in this push: merged current `main`

`main` moved twice while this was in flight and the branch went `CONFLICTING`, which blocks CI from even producing a merge ref. Two merge commits resolve it:

- `packages/ai/src/auth/pool/slots.ts` (vs #1304): main's semantics win — a provider-owned pool now **merges** onto the value read under the credential lock (`mergeProvidedPool`) and the managed-sentinel repair algebra is kept — and this branch's optional `onAllocated` login receipt is re-applied on top. A provider-owned envelope still reports the single account it added, an absent pool still reports `default`, and a generated slot still reports its `login-N` name.
- `packages/coding-agent/src/core/changes.md`, `.../builtin/claude-sdk-oauth/changes.md`, `.../modes/rpc/changes.md`: both sides' tracker entries kept.

### Result

Head is now `7a0f115f7`, `mergeable: MERGEABLE` / `CLEAN`, and CI run [34506830141](https://github.com/code-yeongyu/senpi/actions/runs/34506830141) is fully green, including `Test (workspaces + scripts)` and the required `Check and test` gate.

code-yeongyu (approved):
All 7 findings resolved and mutation-proved. The footer colours from the runs that produced the text instead of re-parsing it; uniqueness folds NFKC + case + invisibles + Cyrillic lookalikes; length is measured in terminal columns per grapheme with a visible-character requirement; the sentinel guard keys on credential shape rather than the provider id; cursor labels render; the double prompt is gated by receipt origin. The CI red was traced to an unrelated Julia kernel cold-boot timeout in senpi-codemode (the suite armed its cell budget at dispatch while dispatch waits for the ready frame) and fixed by awaiting readiness under a named deadline - proven by withholding the ready frame, which reproduces CI's 40s timeout on the old test and the new named deadline at 60s. CI run 34506830141 fully green. Approving.

PR #1498, opened 2026-09-08
fix(claude-sdk-oauth): canonicalize eval summary continuity

### LINKED ISSUE #1472: [Bug]: eval summary clamping triggers assistant_rewritten with claude-sdk-oauth
## What happened?

With the built-in `eval` tool and `claude-sdk-oauth`, `eval.prepareArguments` clamps a summary longer than 80 characters by mutating the assistant tool-call arguments.

`AssistantCommitBoundary` hashes the provider-final assistant before that mutation and the committed assistant after it. The two summaries differ, so the turn is classified `assistant_rewritten` and continuity is invalidated. The next turn then forks or cold-flattens the session.

Live confirmation on an interactive root session (senpi 2026.9.6, `claude-opus-5`): I sent an 81-character summary, the session stored 80 characters ending in `...`, a `claude-sdk-oauth-binding` entry recorded `{"invalidated":true,"reason":"assistant_rewritten"}`, and the following turn reported `{"kind":"flatten","reason":"assistant_rewritten","deltaMessages":193,"payloadBytes":1209471}` with 562,740 cache-write tokens in that single request.

In one earlier long session, 11/11 `assistant_rewritten` events matched the same pattern: original summaries were 81-103 characters and every committed summary was exactly 80.

This is distinct from #691, which normalized volatile `thinking.startedAt` / `thinking.endedAt`. Here the differing field is `content[*].arguments.summary`.

## Steps to reproduce

Deterministic and offline, against the installed modules:

1. Import `AssistantCommitBoundary` from `.../claude-sdk-oauth/session-commit-boundary.js` and `clampEvalSummary` from `.../senpi-codemode/src/tool/eval-request.ts`.
2. Build a resident assistant message whose content is one `eval` tool call with this 81-character summary:

   `01234567890123456789012345678901234567890123456789012345678901234567890123456789X`

3. Call `captureProviderFinal(key, message)`.
4. Apply the clamp: `message.content[0].arguments.summary = clampEvalSummary(message.content[0].arguments.summary)`.
5. Call `commit(key, message, modelId)`.

Observed: `rewritten`. With a summary of 80 characters or fewer the same sequence returns `clean`.

Live variant: in an interactive `claude-sdk-oauth` session, have the model call `eval` once with the same 81-character summary, send one more message, then read the session JSONL.

Relevant paths:

- `packages/senpi-codemode/src/tool/eval-request.ts` (`clampEvalSummary`)
- `packages/senpi-codemode/src/tool/eval-tool.ts` (`prepareArguments`)
- `packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/session-commit-boundary.ts` (`assistantContentHash`)

## Expected behavior

Argument normalization should not invalidate Claude SDK session continuity. Either canonicalize the summary before `captureProviderFinal`, or apply the same normalization inside `assistantContentHash`. The reproduction above should return `clean`, and the next turn should stay incremental when nothing semantic changed.

## Version

- `senpi`: 2026.9.6
- `omo-ai`: 5.0.0-0.beta.45
- Provider: `claude-sdk-oauth`
- Models observed: `claude-opus-5`, `claude-fable-5-1`
- OS: Windows 11 x64

Altairpaca:
## Summary

Fixes #1472.

`eval.prepareArguments` normalizes run summaries before schema validation. For summaries longer than the schema limit, the provider-final assistant can therefore differ from the committed assistant only by harness-owned normalization. `AssistantCommitBoundary` previously fingerprinted the raw tool arguments and classified that normalization as `assistant_rewritten`, invalidating Claude SDK continuity and making the next turn fork or cold-flatten.

This patch canonicalizes the effective `eval` run summary at the continuity fingerprint boundary while keeping every other tool and argument fail-closed.

## Maintainer review follow-up

Addressed the requested deduplication: continuity no longer carries a second implementation of the eval summary clamp.

- `clampEvalSummary` has one implementation owned by `@code-yeongyu/senpi-codemode`.
- `eval-request.ts` continues to export that same helper, and `types.ts` continues to export the same `EVAL_SUMMARY_MAX_LENGTH` constant.
- The narrow runtime leaf is published as `@code-yeongyu/senpi-codemode/eval-summary` so compiled Node code does not statically import a `.ts` file from the source-only package under `node_modules`.
- `AssistantCommitBoundary` imports that canonical helper and preserves the eval-only, `peek`/`stop`, string-summary, and fail-closed guards.
- The regression now drives the committed summary through the real eval tool `prepareArguments` path instead of hand-writing an 80-character expected value.

## Changes

- Canonicalize only `eval` run `summary` values for `assistantContentHash` using the tool-owned normalizer.
- Leave `peek` / `stop` arguments untouched.
- Keep text, thinking, tool id/name, all non-summary eval arguments, and every other tool argument rewrite-sensitive.
- Add a focused #1472 regression using real `eval.prepareArguments`, plus negative controls for real summary and code rewrites.
- Expose the codemode normalizer through a runtime-safe ESM leaf while preserving its existing TypeScript exports and single limit constant.

## Independent runtime evidence

The issue has an independent macOS arm64 reproduction on `omo-ai 5.0.0-0.beta.51` / bundled Senpi `2026.9.9-2` / `claude-sdk-oauth/claude-fable-5-1`: the committed eval summary was exactly 80 characters ending in `...`, the binding invalidation reason was `assistant_rewritten`, and the following request flattened 72 delta messages into a 943,217-byte payload. See #1472 for the measured details.

## Verification status

Focused regression:

`packages/coding-agent/test/suite/regressions/1472-claude-sdk-oauth-eval-summary-clamp.test.ts`

Final head `fc00aceef3be605d6a349afa62eff499d74a3a4a` is a single commit on current upstream `main` `c68ec4d3c0a2dc5fd105da69b2ab4d4e898e7e4d`.

Pre-submit audit checked:

- final diff scope: 7 expected files, no unrelated main drift;
- one clamp implementation / one summary-length constant;
- real `prepareArguments` is used by the continuity regression;
- source-only codemode packaging includes the `.mjs` runtime leaf and `.d.mts` declaration;
- Node ESM + TypeScript conditional-export resolution and `npm pack --dry-run` were exercised in an isolated package fixture.

This environment cannot execute the repository's installed dependency graph, so I am not claiming a local full-suite green run. The PR-triggered CI, Changelog gate, and binary-parity workflows were created for this head but are currently `action_required` pending repository approval; any executable type/test failure remains blocking evidence.

## Risk

Low and bounded to the continuity fingerprint seam. Two provider/committed tool calls compare equal only when their `eval` run summaries normalize to the same effective value that the harness executes. Semantic summary changes within the limit and changes to other eval arguments remain `assistant_rewritten`.

No release changelog entry is included in this contributor patch; repository-side changelog policy/label handling remains with the maintainer workflow.


Altairpaca:
Rebased/synchronized onto current `main` (`464f1a54e8569cdfe8e4620583e19d594a6b96e9`); the PR still reduces to one production file plus the focused regression and remains mergeable. Issue #1472 now also has an independent macOS arm64 reproduction matching this exact 80-character clamp -> `assistant_rewritten` boundary, including a subsequent 943,217-byte flatten request. I updated the PR body with that evidence and kept the negative-control scope explicit. Current CI/Changelog runs are `action_required` before jobs execute, so there is no failing test result to address yet. When convenient, a workflow approval/review pass would unblock the next evidence step; I’ll own any executable CI or review follow-up.

code-yeongyu:
Reviewed against a field report of the same symptom (a Claude session re-sending the whole conversation every turn, hundreds of KB per turn). Your diagnosis is right and the fix is in the right place: the harness-owned clamp lands after the provider-final boundary was hashed, so the fingerprint has to compare the *effective* tool input. The three tests are the right shape too — the clamp is clean, a real summary rewrite is still `rewritten`, and a change to another argument is still `rewritten`.

**One change requested before merge: do not duplicate the clamp — import it.**

The normalization you re-implemented in `semanticToolCallArguments` is already exported by the package that owns it, and that package is already a dependency of `packages/coding-agent`:

- `@code-yeongyu/senpi-codemode` → `src/tool/eval-request.ts` exports `clampEvalSummary(value: unknown): string | undefined`, with exactly the semantics your copy reproduces: `trim()`, collapse `\s+` (unicode), empty becomes undefined, otherwise truncate to `EVAL_SUMMARY_MAX_LENGTH - "...".length` and append the ellipsis.
- `EVAL_SUMMARY_MAX_LENGTH = 80` is exported from that package's `src/tool/types.ts`.
- `packages/coding-agent/package.json` already declares `@code-yeongyu/senpi-codemode`.

As written, the invariant now lives in two places across a package boundary with nothing binding them. If the eval tool ever changes its limit or its ellipsis, continuity silently starts classifying clean turns as rewritten again — and the regression test added here would still pass, because it hardcodes the same two numbers on both sides of the assertion. That is the same failure class this PR is fixing, one release later.

Concretely: call `clampEvalSummary` (and delete the local `EVAL_SUMMARY_MAX_LENGTH` / `ELLIPSIS` constants and the normalization body), keep the `eval`-only and `peek`/`stop` guards and the fail-closed behavior for every other tool exactly as you have them. Then one of the tests should drive the real tool's `prepareArguments` output rather than a hand-written 80-character string, so the two sides cannot drift without a red test.

If you would rather not extend this PR, say so and I will land the import swap as a small follow-up right after this merges — I do not want a review round trip to hold a correct fix. The behavior you shipped is what we want either way.

Altairpaca:
Addressed the requested review change in `fc00aceef3be605d6a349afa62eff499d74a3a4a`.

- Removed the duplicated continuity-side clamp and now call the codemode-owned `clampEvalSummary`.
- Preserved the existing eval-only, `peek`/`stop`, string-summary and fail-closed guards.
- Changed the regression to derive the committed summary through the real eval tool `prepareArguments` path rather than a hand-written 80-character value.
- Kept the source-only package safe for the compiled Node path by exposing the canonical helper as a narrow `.mjs` leaf with a `.d.mts` declaration; `eval-request.ts` re-exports the same helper and `types.ts` re-exports the same limit, so there is still one implementation and one constant.
- Rebased/squashed the PR to one commit on current `main` (`c68ec4d3c0a2dc5fd105da69b2ab4d4e898e7e4d`).

I also checked the final 7-file diff and exercised the package export with isolated Node ESM / TypeScript resolution plus `npm pack --dry-run`. I am not claiming a repository-suite green run: the new CI, Changelog gate and binary-parity runs are all currently `action_required` pending repository approval.

One repository-side gate detail: this GitHub App connection cannot apply the `no-changelog` label (403). I left release changelog handling unchanged rather than manufacturing a contributor-side release entry.

code-yeongyu:
Thanks for this — the diagnosis is right and the offline reproduction in #1472 is what made the root cause obvious. I used both as the starting point.

I'm taking a different fix in #1959 and will close this one once that merges, so here is the reasoning in full rather than a bare decline.

This patch canonicalizes the `eval` run summary **inside** `assistantContentHash`. That makes the fingerprint agree for `eval`, but the defect is not `eval`-shaped. `prepareArguments` is documented as a normalizer, and the real problem is that some shims normalize by mutating the object they were handed and returning that same reference. `eval` is one; `prepareEditArguments` in `packages/agent/src/harness/tools/edit.ts` does the same thing with `args.edits` today, and any shim added later can do it again. The caller's identity guard cannot catch it, because `prepared === toolCall.arguments` is only true for a shim that returned its input — exactly the shim that already mutated it.

So #1959 fixes the caller: `prepareArguments` now runs against `structuredClone(args)`. Three consequences that matter here:

1. Every shim is covered, including `edit` and anything future, instead of the one tool that happened to be reported.
2. The continuity fingerprint keeps hashing tool arguments verbatim. Nothing is relaxed for a specific tool, so a genuine semantic edit — including one past the display limit — still classifies as `assistant_rewritten`. oh-my-openagent#8424 argued the same point independently: prepare an execution copy, keep the provider's arguments verbatim, and never normalize authoritative history.
3. No new published export surface. This patch adds `@code-yeongyu/senpi-codemode/eval-summary` with a hand-written `.mjs` + `.d.mts` pair so compiled code can reach the clamp; the caller-side fix needs none of that.

There is also precedent for the mechanism: `validateToolArguments` in `packages/ai/src/utils/validation.ts` already does `structuredClone(toolCall.arguments)` for the same reason, so the cost profile is established rather than new.

Your regression intent is preserved — #1959 keeps a test for the over-limit summary committing `clean`, plus negative controls for a real summary rewrite and a real code rewrite — but it drives one assistant message through the actual order (capture provider-final → real `prepareAgentToolCall` → commit) instead of comparing two hand-written copies, so it fails if the mechanism regresses anywhere along that path.

PR #1521, opened 2026-09-09
fix(goal): stop automatic recovery after terminal policy rejection

### LINKED ISSUE #1520: Goal recovery repeats terminal Codex safety rejections until the continuation cap
## What happened?

Senpi's builtin goal recovery automatically repeats a non-retryable Codex safety rejection until the continuation cap is exhausted:

```text
Codex error: This request was blocked by our safety systems. Reason: Potentially unintended activity.
```

Two observed bursts each contained nine error responses and eight hidden `goal-continuation` messages (51.604 s and 50.010 s). Ordinary retry classification reports `retryable=false`, but `continueGoalAfterAgentEnd()` routes the error into `afterProviderFailure()`, which stages `providerRecovery` after `agent_settled`. Recovery admission does not distinguish a policy rejection from a recoverable infrastructure failure.

## Steps to reproduce

1. Keep a builtin goal active.
2. Have the provider return the error above with `stopReason: "error"` and no further provider retry (`willRetry: false`).
3. Deliver `agent_end` and `agent_settled`.
4. Observe a hidden continuation and another provider call. Repeating the lifecycle reaches eight continuations before `continuation cap reached`.

An isolated reproduction using the installed provider parser and real goal routing/admission reproduced nine calls and eight continuations; HTTP and persistence were substituted. This is a builtin-extension defect, not a claim about extension-free core behavior.

## Expected behavior

Preserve the goal and stop automatic recovery for terminal policy rejections. Retain recovery for genuinely recoverable failures. I am preparing a regression test and fix.

The initial rejection came from the Codex backend; local records do not identify the offending input or the backend classifier rationale.

## Version

Senpi `2026.9.8`, distributed through `omo-ai 5.0.0-0.beta.49`, macOS arm64. Latest source inspected: `78b220bf8`.

rlaope:
## Summary

Stop an active goal from automatically resubmitting a request after a terminal policy rejection. The reported Codex error was classified as non-retryable by ordinary retry handling, but goal recovery still submitted eight hidden continuations before hitting its cap. This change preserves the goal and blocks that recovery path immediately.

Closes #1520.

## Behavior and implementation

- Recognize existing structured classifier refusal/sensitive stops and the exact Codex backend safety diagnostic, only when the retry owner reports `willRetry: false`.
- Persist the active goal as blocked before provider/system recovery routing; synchronize the monitor to clear pending recovery and timers.
- Keep the goal ID, objective, and continuation counts. An explicit goal resume remains available; unrelated input does not automatically restart this policy-blocked goal.
- Preserve infrastructure recovery, explicit retry ownership, and ordinary assistant text that discusses a safety error.

The change is confined to two builtin goal modules, their change tracker, lifecycle tests, and a reusable offline CLI QA fixture/driver. It does not attempt to change the provider's safety decision.

## Verification

- **Failing first:** seven policy assertions failed before the patch; the same fourteen targeted cases passed afterward.
- **Related regression suite:** 127 tests passed.
- **Lead-run real source CLI:** `node .agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-qa.mjs --self-test` passed. The policy scenario made the goal-creation call plus the rejected call, then **zero continuations**. The infrastructure control made **one continuation** and completed.
- Both CLI sandboxes and child clients were cleaned; the real authentication file was unchanged. The provider is an in-process faux fixture, with no external requests.
- Workspace `npm run build`, TypeScript, scoped Biome, and `bun run check` exited 0.
- Full coding-agent suite: 10,251 passed, four lifecycle tests failed, 39 existing skips. Re-running all four affected files without file parallelism passed all 93 tests. The full concurrent suite is **not** claimed green.

Sanitized local evidence is under `local-ignore/qa-evidence/20260909-terminal-policy-goal-recovery/`: `red.txt`, `green.txt`, `lead-cli/result.json`, `lead-build.txt`, `lead-check.txt`, `lead-package-tests-built.txt`, and `lead-failed-files-isolated.txt`. The committed QA driver makes the principal behavior reproducible.

## Risks and limits

The original provider response retained only a generic safety diagnostic, not the backend classifier rationale; the offending input remains unknown. This patch fixes repeated requests, not the initial server decision. OMO's separate plan-continuation hooks need their companion guard when those extensions are active.

`bun run check` autoformatted two untouched baseline files; those unrelated formatting changes were restored. LSP tooling is unavailable locally, so compiler and Biome checks are reported instead. No dependencies or lockfiles changed.

AI-assisted implementation and report; local evidence and production diff were independently inspected by the lead coding session.

---
## Summary by cubic
Stops active goals from auto-recovering after terminal policy rejections. Previously, a Codex safety block or classifier refusal let goal recovery queue up to eight hidden continuations; now the goal is persisted as blocked before recovery routing, and that recovery path is canceled. Closes #1520.

**Bug Fixes**
- Detects structured refusals and sensitive stops plus the exact Codex safety diagnostic, only when the retry owner reports `willRetry: false`.
- Trusts the unstructured Codex diagnostic only on the `openai-codex-responses` API; other providers or gateways emitting the same sentence keep infrastructure recovery.
- Pins that identity gate to the `OPENAI_CODEX_MODELS` catalog, so renaming the API id there fails the test suite instead of silently restoring the recovery loop.
- Preserves the goal ID, objective, and continuation counts; only an explicit goal resume restarts a policy-blocked goal.
- Leaves infrastructure recovery, explicit retry ownership, and ordinary assistant text about safety errors unchanged.

<sup>Written for commit b01734280ff60113b1d7c050a8548a2a1b0a73fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1521?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-fixture.ts
+++ b/.agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-fixture.ts
@@ -7,8 +7,13 @@ import type { ExtensionAPI, ExtensionContext } from "../../../../../packages/cod
 export default function policyRecoveryFixture(pi: ExtensionAPI): void {
 	const scenario = process.env.SENPI_QA_POLICY_CASE;
 	if (scenario !== "policy" && scenario !== "infrastructure") throw new Error("Unknown policy QA scenario");
+	// The faux provider stamps its own `api` onto every response, so the policy
+	// lane must register under the Codex API id: that identity is exactly what the
+	// predicate requires before trusting the unstructured diagnostic. The
+	// infrastructure lane keeps a non-Codex api, which also proves the identity
+	// gate does not swallow ordinary provider failures.
 	const faux = fauxProvider({
-		api: "faux-policy-qa",
+		api: scenario === "policy" ? "openai-codex-responses" : "faux-policy-qa",
 		provider: "faux-policy-qa",
 		models: [{ id: "policy-qa", contextWindow: 1_000_000, maxTokens: 4_096 }],
 	});
--- a/.agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-fixture.ts
+++ b/.agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-fixture.ts
@@ -1,0 +1,57 @@
+import { fauxAssistantMessage, fauxProvider, fauxToolCall } from "../../../../../packages/ai/src/providers/faux.ts";
+import { readGoal } from "../../../../../packages/coding-agent/src/core/extensions/builtin/goal/store.ts";
+import { goalStoreRef } from "../../../../../packages/coding-agent/src/core/extensions/builtin/goal/store-ref.ts";
+import type { ExtensionAPI, ExtensionContext } from "../../../../../packages/coding-agent/src/core/extensions/types.ts";
+
+// Only loaded explicitly by goal-policy-rejection-qa.mjs in its isolated CLI.
+export default function policyRecoveryFixture(pi: ExtensionAPI): void {
+	const scenario = process.env.SENPI_QA_POLICY_CASE;
+	if (scenario !== "policy" && scenario !== "infrastructure") throw new Error("Unknown policy QA scenario");
+	const faux = fauxProvider({
+		api: "faux-policy-qa",
+		provider: "faux-policy-qa",
+		models: [{ id: "policy-qa", contextWindow: 1_000_000, maxTokens: 4_096 }],
+	});
+	faux.setResponses([
+		fauxAssistantMessage(fauxToolCall("create_goal", { objective: "QA preserve unfinished work" }, { id: "qa-create" }), {
+			stopReason: "toolUse",
+		}),
+		fauxAssistantMessage("", {
+			stopReason: "error",
+			errorMessage: scenario === "policy"
+				? "Codex error: This request was blocked by our safety systems. Reason: Potentially unintended activity."
+				: "upstream connection closed",
+		}),
+		fauxAssistantMessage(fauxToolCall("update_goal", { status: "complete" }, { id: "qa-complete" }), {
+			stopReason: "toolUse",
+		}),
+		fauxAssistantMessage("QA_RECOVERY_COMPLETE"),
+	]);
+	pi.registerProvider(faux.provider);
+	let context: ExtensionContext | undefined;
+	let createdGoalId: string | undefined;
+	const endings: Array<{ stopReason?: string; errorMessage?: string; willRetry?: boolean }> = [];
+	pi.on("session_start", (_event, ctx) => { context = ctx; });
+	pi.on("agent_end", async (event, ctx) => {
+		context = ctx;
+		const goal = await readGoal(goalStoreRef(ctx.sessionManager, ctx.cwd));
+		createdGoalId ??= goal?.id;
+		const lastAssistant = event.messages.findLast((message) => message.role === "assistant");
+		if (lastAssistant?.role === "assistant") {
+			endings.push({ stopReason: lastAssistant.stopReason, errorMessage: lastAssistant.errorMessage, willRetry: event.willRetry });
+		}
+	});
+	pi.rpc.handle("qa.policy.snapshot", async () => {
+		if (context === undefined) throw new Error("QA session has not started");
+		const branch = context.sessionManager.getBranch();
+		return {
+			scenario,
+			calls: faux.state.callCount,
+			createdGoalId,
+			goal: await readGoal(goalStoreRef(context.sessionManager, context.cwd)),
+			endings,
+			continuations: branch.filter((entry) => entry.type === "custom_message" && entry.customType === "goal-continuation").length,
+			pending: context.hasPendingMessages(),
+		};
+	});
+}
--- a/.agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-qa.mjs
+++ b/.agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-qa.mjs
@@ -1,0 +1,82 @@
+#!/usr/bin/env node
+// #1520: zero-token source-CLI QA. No server, real credentials, or provider calls.
+// node .agents/skills/senpi-qa/scripts/scenarios/goal-policy-rejection-qa.mjs --self-test
+// Optional: --evidence <directory>. Captures receipts even on assertion failure.
+import assert from "node:assert/strict";
+import { mkdirSync, writeFileSync } from "node:fs";
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { guardRealAuth, installCleanupHooks, makeSandbox, repoRoot } from "../lib/common.mjs";
+import { hermeticEnv } from "../lib/mock-loop-support.mjs";
+import { TargetRpcClient } from "../lib/target-rpc-client.mjs";
+
+const root = repoRoot();
+const evidenceIndex = process.argv.indexOf("--evidence");
+if (evidenceIndex >= 0 && !process.argv[evidenceIndex + 1]) throw new Error("--evidence needs a directory");
+const out = resolve(evidenceIndex >= 0 ? process.argv[evidenceIndex + 1] : join(root,
+	"local-ignore/qa-evidence/20260909-terminal-policy-goal-recovery/cli"));
+const fixture = join(dirname(fileURLToPath(import.meta.url)), "goal-policy-rejection-fixture.ts");
+const guard = guardRealAuth();
+installCleanupHooks();
+mkdirSync(out, { recursive: true });
+
+async function runScenario(scenario) {
+	const box = makeSandbox(`goal-policy-${scenario}`);
+	writeFileSync(join(box.agentDir, "settings.json"), JSON.stringify({ retry: { enabled: false }, compaction: { enabled: false } }));
+	const client = new TargetRpcClient({
+		targetRoot: root,
+		cwd: box.cwd,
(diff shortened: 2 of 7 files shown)


code-yeongyu (changes requested):
The change addresses the reported recovery loop, but I cannot approve it because the new unstructured diagnostic path does not actually identify Codex. It treats a provider-agnostic string as a Codex policy decision and therefore changes recovery behavior for providers and gateways that happen to emit the same message. That is a compatibility regression in the central goal recovery path, and the new tests do not cover it.

## Blocking findings

1. **`packages/coding-agent/src/core/extensions/builtin/goal/terminal-provider-error.ts:21-22`** — `CODEX_POLICY_ERROR_PATTERN` matches the diagnostic with an optional `Codex error: ` prefix, while `didTerminalPolicyRejectionEndTurn()` applies it to every assistant message without checking `message.api`, `message.provider`, or any other Codex identity (`:214-221`). As a result, a non-Codex provider or an intermediary that returns `This request was blocked by our safety systems. Reason: ...` with `stopReason: "error"` and `willRetry: false` is persisted as `blocked` by `agent-end-continuation.ts:16-24` instead of entering the existing provider/system recovery path. This strands an otherwise recoverable goal and contradicts the PR's stated promise to preserve infrastructure recovery while claiming to match the exact Codex diagnostic. Scope the unstructured matcher to the actual Codex API/provider identifier used by the Codex adapter (or replace it with a provider-owned structured signal), and add a regression case using a different provider/API with the same text that asserts one recovery continuation and an active goal.

## Non-blocking notes

The structured classifier-refusal path is appropriately gated by `willRetry === false`, and the lifecycle tests exercise settlement admission rather than only testing the predicate. However, the new suite covers only the faux provider's messages and does not expose the provider-identity collision above.

## What is good

The change persists the blocked state before settlement recovery admission, synchronizes the monitor to cancel staged work, and keeps the goal identity and counters intact. The linked issue's observed eight-continuation loop is directly targeted without removing the separate infrastructure recovery behavior.

rlaope:
Fixed in `ef24b95ad`. The finding was correct: the diagnostic was matched on every assistant message, so any provider or gateway emitting the same sentence was persisted as `blocked` instead of taking the provider/system recovery path.

`didTerminalPolicyRejectionEndTurn()` now trusts the unstructured Codex diagnostic only when `api === "openai-codex-responses"` — the value the real transcripts carry (verified against the reported session: 1,720 `openai-codex-responses` messages alongside 516 `anthropic-messages`). Structured classifier refusals stay provider-independent, since they carry their own policy details rather than a free-text sentence.

Two test-side details are worth flagging, because they are why the gap was invisible:

- `fauxAssistantMessage` hardcodes `api` and ignores an override, so the unit fixtures needed the identity applied to the returned message.
- The faux provider stamps its own `api` onto every response via `cloneMessage`, so the live QA lane had to register the policy provider under the Codex API id. My first attempt set it per-message and the CLI run silently continued recovering — which is exactly the collision you described, reproduced from the other side.

The infrastructure lane deliberately keeps a non-Codex api, so it now doubles as proof that the identity gate does not swallow ordinary provider failures.

New coverage: two cross-provider cases (`anthropic-messages`, `openai-completions`) asserting the same diagnostic still yields one continuation and an `active` goal. Captured RED first (2 failed), then green.

Verification on `ef24b95ad`:

- `goal-policy-rejection` + 5 adjacent goal suites: 129 passed
- real CLI QA: policy `calls=2 continuations=0 goal=blocked`; infrastructure `calls=4 continuations=1 goal=complete`; both sandboxes removed, real auth unchanged
- `bun run check`: clean

The tracker entry in `goal/changes.md` now states the identity restriction explicitly.

code-yeongyu:
All findings from the `CHANGES_REQUESTED` review are resolved on `d002d21b1`, rebased onto current `origin/main` (`d2f5f28fa`) with both of @rlaope's commits preserved.

## Finding 1 (blocking) — `terminal-provider-error.ts:21-22`: the unstructured matcher was not scoped to a Codex identity

Resolved by `55f148039` (@rlaope) and hardened by `d002d21b1`.

- `isCodexPolicyRejection()` now returns `false` unless `message.api === "openai-codex-responses"`, the api id every shipped Codex model carries (`packages/ai/src/providers/data/openai-codex.json`, `openaiCodexProvider()`, and `lazy.ts`/`faux.ts` stamp `model.api` onto error messages). A different provider or an intermediary emitting `This request was blocked by our safety systems. Reason: ...` with `stopReason: "error"` / `willRetry: false` now falls through to `didTerminalProviderErrorEndTurn()` and keeps the existing provider/system recovery path instead of being persisted as `blocked`. Structured classifier refusals stay provider-independent through `isClassifierRefusal()`, since they carry their own policy details.
- The regression cases the review asked for are in the suite: *"recovers instead of blocking when anthropic-messages reports the same diagnostic"* and *"...when an unknown gateway (`openai-completions`) reports the same diagnostic"*, both asserting exactly one `goal-continuation` message and `status: "active"` after `agent_end` + two `agent_settled` rounds.
- Hardening in `d002d21b1`: the gate's api id was a literal copy of a string that `packages/ai` owns, and nothing enforced the two staying equal — renaming that api id would have silently disarmed the #1520 guard while every hardcoded case in the suite stayed green. The suite now drives the same `agent_end` → `agent_settled` lifecycle once per api id in `OPENAI_CODEX_MODELS` and asserts the goal is blocked with no continuation.

## Non-blocking note — the suite only covered the faux provider's messages

Covered by the same two commits: the file is now 17 cases spanning the Codex api, structured refusal/sensitive stops, the Anthropic policy text, four infrastructure-recovery lanes, explicit retry ownership, ordinary assistant text, the monitor backstop with system-abort provenance, the two provider-identity-collision lanes, and the shipped-catalog identity pin.

## Mutation evidence

Every assertion that carries the fix was proved by removing the production line, observing the failure, restoring it, and observing the pass. All runs on a Linux x86_64 box (bun 1.3.14, node 24.20.0) after a fresh `bun install`, from the repository root:

```
bun run --cwd packages/coding-agent test -- test/suite/goal-policy-rejection.test.ts
```

| mutation | result |
|---|---|
| baseline at `55f148039` | `Tests 16 passed (16)` |
| delete `if (message.api !== CODEX_RESPONSES_API) return false;` (`terminal-provider-error.ts`) | `Tests 2 failed \| 14 passed` — exactly the two collision cases: `expected [] to have a length of 1 but got +0` (`goal-policy-rejection.test.ts:157`) |
| restore | `Tests 16 passed (16)` |
| delete the whole `didTerminalPolicyRejectionEndTurn` branch (`agent-end-continuation.ts:19-28`) | `Tests 7 failed \| 9 passed` — the six blocking cases plus the backstop case (`expected { status: 'active' } to match { status: 'blocked' }`) |
| restore | `Tests 16 passed (16)` |
| baseline at `d002d21b1` | `Tests 17 passed (17)` |
| rename `"openai-codex-responses"` → `"openai-codex-responses-v2"` in `packages/ai/src/providers/data/openai-codex.json` (simulated catalog rename) | `Tests 1 failed \| 16 passed` — only the new pin fails: `blocks the diagnostic on the shipped Codex catalog api openai-codex-responses-v2`, `expected [ { message: … } ] to have a length of +0 but got 1`. The 16 hardcoded cases stay green, which is precisely the silent regression the pin now prevents. |
| restore | `Tests 17 passed (17)` |

## Test results

```
# goal surface, 36 files
bun run --cwd packages/coding-agent test -- test/suite/goal- test/suite/app-server-goal-wire.test.ts \
  test/suite/app-server-thread-goal.test.ts test/suite/loop-guard-goal-isolation.test.ts
→ Test Files 36 passed (36) | Tests 365 passed (365)

# full coding-agent package, after `bun run build` at the repository root
bun run --cwd packages/coding-agent test
→ Test Files 2 failed | 1298 passed | 5 skipped (1305)
→ Tests 2 failed | 10498 passed | 40 skipped (10540)

# static validation
bunx tsc --noEmit                     → 0
bunx biome check <changed files> --error-on-warnings → 0
bun run check:ts-imports              → 0
```

The two full-suite failures are pre-existing and unrelated to this change (they touch neither the goal builtin nor `packages/ai`), and reproduce on a pristine `origin/main` (`d2f5f28fa`) checkout on the same box:

- `test/auth-storage.test.ts > keeps a coalesced reload alive while another credential reader is waiting` — fails on `origin/main` as well.
- `test/cursor-cli-oauth/settings.test.ts > re-reads settings from disk on every load` — mtime-granularity dependent: 4/5 runs fail on this branch and 4/5 fail on `origin/main`.

The PR description's "four lifecycle tests failed" no longer reproduces: with the workspace built first, the full package suite reports no goal or lifecycle failure. The earlier `dist`-related failures in this environment were the `workspace-build-prerequisite` guard firing before `bun run build` had been run, not test regressions.

code-yeongyu (approved):
Findings resolved. isCodexPolicyRejection now requires the Codex responses api id, so a provider-agnostic refusal no longer ends a goal as a Codex policy rejection and keeps provider/system recovery. The api id was a literal copy of a string packages/ai owns with nothing keeping them equal - a rename there would have silently disarmed the #1520 guard - so it is pinned against the shipped catalog. 3 mutations, each RED then GREEN on restore, including renaming the api id in openai-codex.json which fails only the new pin. Goal surface 36 files / 365 tests pass; full package 10498 passed with 2 pre-existing mtime-granularity failures that reproduce identically on pristine main. CHANGELOG entry added for the gate. Approving.

PR #1528, opened 2026-09-10
fix(coding-agent): record model switches the session refuses

### LINKED ISSUE #1526: Model switch rejected by the context-window guard leaves no record in the session or logs
## Summary

A manual `/model` switch to a model with a smaller context window is rejected before anything is recorded, so the session keeps using the old model and no evidence of the attempt survives anywhere. The user sees "I changed the model and nothing happened", and afterwards there is no log line to explain why.

This bites hardest in exactly the situation where switching matters most: an account-level `429` storm, where the user is trying to escape to a different provider.

Note this is about context-window size, not model capability: the target here is a perfectly good model that simply advertises a smaller window than the live transcript needs.

## What happened

A long-running session on `anthropic/claude-fable-5-1` started receiving account-level rate-limit errors:

```
429 {"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed your account's rate limit. Please try again later."}} (retry-after-ms: 14754000)
```

The user opened the model selector and picked `openai-codex/gpt-6-astra`. The TUI returned to the editor and the session continued on the previous model. Selecting `anthropic/claude-opus-5` a few minutes later did work. Later, after resuming the session, selecting `gpt-6-astra` worked too, which made the earlier failure look non-deterministic.

## Root cause

`AgentSession._setModel()` validates before it records:

```ts
private async _setModel(model: Model<Api>, updateGlobalDefaults: boolean) {
  this.assertModelUsable(model, this._getDownswitchLiveContextTokens(model)); // throws here
  if (!(await this._modelRuntime.checkAuth(model.provider))) { ... }
  return await this._switchActiveModel(model, { appendSessionEntry: true, ... });
}
```

`assertModelUsable()` throws `ModelUsabilityBudgetError` when the live context does not fit the target model. In this session the last recorded assistant turn reports `totalTokens: 769591`, against these advertised context windows:

| Model | Context window | Switch outcome |
| --- | --- | --- |
| `anthropic/claude-fable-5-1` (source) | 1M | n/a |
| `anthropic/claude-opus-5` | 1M | succeeded |
| `openai-codex/gpt-6-astra` | 600K | rejected, unrecorded |

The switch is therefore rejected on the first statement. Because the throw happens before `_switchActiveModel({ appendSessionEntry: true })`, the `model_change` entry is never appended. `claude-opus-5` cleared the same check, which is why that one switch succeeded, and why the switch also succeeded after a resume, when the live context was small again.

The budget rejection itself is a deliberate guard and is not the complaint. The two problems are what it does to observability and to the user's only escape hatch.

The thrown error itself is well built. `ModelUsabilityBudgetError` already names the target model, the context window, the shortfall, the requirement, the live context, the full reserve breakdown, and the remedy ("Compact the session, then revalidate and retry the model switch"). Nothing about the message needs improving — the problem is entirely where it goes.

## Problem 1: the rejection leaves no trace

The only surface for the failure is the TUI catch in `selectModelFromUi()`:

```ts
} catch (error) {
  this.showError(error instanceof Error ? error.message : String(error));
}
```

That is a transient status message and nothing else. Checked after the fact:

- session JSONL: no `model_change`, and no error entry for the attempt
- `logs/session.log`: only `provider_error` 429 lines
- `logs/fallback.log`: only `no_chain` decisions

During a 429 storm the status line is being repainted constantly, so the one-shot message is easy to miss entirely. Reconstructing the incident afterwards is impossible from artifacts: the attempt can only be inferred from the *absence* of a `model_change` entry, which is indistinguishable from "the user never tried".

## Problem 2: the guard blocks the escape hatch it is needed for

When a provider rate-limits an account, moving to another provider is the remedy. But a long session is exactly the one whose live context exceeds a smaller target window, so the switch that would resolve the incident is the one that gets refused. The user is left on the rate-limited model with no in-product path forward except restarting or resuming the session.

Compaction would make the target model usable, but the switch path does not offer or perform it; it only throws.

## Expected

1. A rejected model switch is recorded durably, with the reason and the numbers behind it. A session-level entry and/or a `logs` line, not only a transient toast.
2. The remedy the error already names is reachable from where the refusal happens, so a rate-limited session is not a dead end.

## Suggested direction

Two independent changes; the first is the one that matters.

**Record the rejection.** `_setModel()` is the natural seam: catch around the two guards, append a durable record, and rethrow unchanged. `ModelUsabilityBudgetError.projection` already carries every field worth recording (`model`, `contextWindow`, `liveContextTokens`, `requiredTokens`, `shortfallTokens`, `admission`), so this needs no new computation — only a sink. A session entry carrying the attempted model and a rejected outcome (either a `model_change` variant or a sibling entry type) makes "did the user try, and why did it not take?" answerable from artifacts alone; a `logs/session.log` line covers the same question for non-TUI surfaces. The auth guard below it deserves the same treatment, since it fails just as invisibly.

**Make the remedy reachable.** The `switch` admission message already tells the user to compact and retry. Since `selectModelFromUi()` can distinguish this error by type, it can offer that compaction inline and re-attempt the switch on confirmation, instead of ending at a toast. Compaction already exists and already operates on this session's transcript; this only invokes it at the point of refusal.

Deliberately not suggested: auto-compacting on switch without asking. That rewrites the transcript as a side effect of a selection the user may abandon, so it should stay an explicit confirmation.

Also worth noting for scope: the same budget guard is already special-cased inside the retry path (`agent-session.ts`, the `tryFallback` wrapper swallows `ModelUsabilityBudgetError` and returns `false`), which is a second place a budget refusal disappears without a record.

## Environment

- omo `5.0.0-0.beta.49`, senpi `2026.9.8`, macOS arm64
- Session context at the time: ~769K tokens; target `openai-codex/gpt-6-astra` context window: 600K
- All three providers (`anthropic`, `openai-codex`, `og`) were authenticated and answered a probe request successfully, so this is not an auth failure

## AI assistance disclosure

This report was drafted with a coding agent from local session logs and source inspection of the installed and current-`main` senpi trees. The observed values above (token counts, timestamps, absent log entries) were read from artifacts on disk; no provider-side behavior is claimed.

rlaope:
## Summary

- A model switch refused by the context-window or auth guard now leaves a durable record instead of vanishing: `_setModel` appends a `model_change_rejected` session entry, emits a matching event, and rethrows the original error unchanged.
- The same refusal now uses the switch wording, so it carries the remedy it already knows ("Compact the session, then revalidate and retry the model switch") instead of claiming the model "cannot start".

Before this change an attempted-and-rejected switch was indistinguishable from a switch the user never made. That is not hypothetical: during an account-level `429` storm a user selected another provider, the guard refused it, the one-shot toast was buried under repeated provider errors, and the session kept running on the rate-limited model. Afterwards nothing on disk explained it — no `model_change`, no error entry, nothing in `logs/session.log` or `logs/fallback.log`. Reconstructing the incident was only possible by inferring from the *absence* of an entry.

Closes #1526.

## Changes

- `packages/coding-agent/src/core/agent-session.ts`: `_setModel` wraps its two guards; `_recordRejectedModelChange` appends the entry and emits `model_change_rejected` with the projection numbers (`contextWindow`, `liveContextTokens`, `requiredTokens`, `shortfallTokens`, `safetyMarginProfile`) plus the guard's own text. The `assertModelUsable` call declares `admission: "switch"`.
- `packages/coding-agent/src/core/session-manager.ts`: adds `ModelChangeRejectedEntry` (in the `SessionEntry` union) and `appendModelChangeRejected()`, mirroring the existing `appendModelChange()`.
- `packages/coding-agent/src/core/changes.md`: tracker entry.
- `packages/coding-agent/test/suite/agent-session-model-switch-rejection.test.ts`: new regression suite.
- `packages/coding-agent/test/suite/model-usability-budget.test.ts`: the one existing expectation that pinned the old cold-start sentence for this path, updated to the switch contract.

Two design notes worth reviewer attention:

**Why `detail` and not `message`.** The entry's text field is `detail` because `SessionMessageEntry.message` is an object; naming it `message` collapsed the `SessionEntry` discriminated union and broke unrelated call sites at the type level. Caught by `tsc`, not guessed.

**Why `admission: "switch"` is safe.** `admission` selects wording and one `resume`-only branch; `start` and `switch` compute the identical budget. The default infers `"switch"` only when `liveContextTokens > 0`, which is false on this path whenever the target's usable context is not smaller than the current model's — hence the wrong wording on exactly the switches a user makes to escape a rate limit.

No renderer was added to the TUI: `selectModelFromUi` already catches and shows this error, so an event handler there would double-report the same failure.

## QA & Evidence

- **What was tested:** `bun run --cwd packages/coding-agent test test/suite/agent-session-model-switch-rejection.test.ts`
  **Observed result:** RED before the fix — 2 failed / 2 passed, failing precisely on the missing session entry and the missing event. GREEN after — 4 passed.
  **Artifact:** `local-ignore/qa-evidence/20260909-model-switch-rejection/{red.txt,green.txt}`
  **Why sufficient:** the two failures are the exact contract this PR introduces; the two that passed before and after guard against a false positive (unchanged active model, and no rejection recorded on a successful switch).

- **What was tested:** six adjacent model/session suites (`agent-session-model-extension`, `gpt-model-switch-{characterization,lifecycle-characterization,rpc-set-model}`, `extension-session-model-api`, `model-fallback-command`).
  **Observed result:** 41 passed.
  **Artifact:** `local-ignore/qa-evidence/20260909-model-switch-rejection/adjacent.txt`
  **Why sufficient:** these cover the switch, cycle, fallback and RPC paths that share the modified `_setModel` and the extended `SessionEntry` union.

- **What was tested:** `bun run check`, then `npm run build` and the full `packages/coding-agent` suite.
  **Observed result:** check clean. Suite: 10,243 passed, 2 failed. Both were re-run in isolation and pass (`Test Files 3 passed | Tests 36 passed`): `rpc-socket-host` is pre-existing port contention under parallel files, and `model-usability-budget` is the expectation this PR updates.
  **Artifact:** `local-ignore/qa-evidence/20260909-model-switch-rejection/{check.txt,full-suite.txt,recheck.txt}`
  **Why sufficient:** it establishes the new entry type does not disturb the rest of the session/entry surface, and separates the one intended contract change from unrelated flake.

## Risks & Residuals

- **New persisted entry type.** `model_change_rejected` joins the `SessionEntry` union, so consumers switching exhaustively over entries see a new member. Mitigated: the union is type-checked repo-wide and `tsc --noEmit` is clean; readers that ignore unknown types are unaffected.
- **Changed refusal wording on the `setModel` path.** Intended, and the only test that pinned the old sentence is updated in this PR. Accepted.
- **`liveContextTokens` can legitimately be `0`** in the recorded entry, because `_getDownswitchLiveContextTokens` reports 0 when the target's usable context is not smaller. The test asserts the field's presence and type rather than a non-zero value, so it locks behavior instead of harness geometry.
- **Not addressed here:** the retry path's `tryFallback` wrapper swallows the same `ModelUsabilityBudgetError` and returns `false`, which is a second place a budget refusal disappears without a record. Called out in #1526; left out of this diff to keep it reviewable.

## Automated Checks

```bash
bun run check
bun run --cwd packages/coding-agent test
```

## Related Issues

Closes #1526

---
## Summary by cubic
Records refused model switches as `model_change_rejected` entries instead of dropping them, preserving the original error while making context-budget and auth failures auditable. Favorite cycles now validate before committing a model, global default, or related change event, so a rejected target cannot become the active or future-session model. Addresses #1526.

**Changes**
- A shared guard path covers `setModel`, post-`model_select` revalidation, and favorite-cycle checks, recording the reason, detail, and budget projection values while emitting a matching event.
- The bookkeeping entry stays out of model context, survives Claude SDK OAuth resume, crosses RPC validation without inflating status snapshots, and renders and searches in `/tree`.
- Admission wording now includes the compaction remedy only when the session has context to compact.
- `SessionEntry` consumers must handle the new `model_change_rejected` member.

**Validation**
- Targeted suites passed 87 tests and `bun run check` is clean; the full suite reports 98 pre-existing failures from missing `dist/` artifacts and socket-port contention.

<sup>Written for commit a708cb3dfd5b272ab1b2be58652f37b603b75d98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1528?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -12,6 +12,20 @@
 
 - A bare `.` submitted on a session that already has messages no longer renders as a user message in the TUI. It stays the manual-continue shortcut the session delivers as a hidden continuation, so nothing is painted for it while idle or while steering an active turn; a `.` on an empty session and a `.` carrying image attachments remain ordinary user input ([#1569](https://github.com/code-yeongyu/senpi/issues/1569))
 
+- A model switch the session refuses is now recorded instead of vanishing: every
+  guard (`setModel`/`setSessionModel`, the revalidation after `model_select`, and
+  favorite cycling) appends a `model_change_rejected` session entry carrying the
+  budget projection numbers and emits a matching event, and the refusal keeps its
+  compaction remedy only when there is context to compact. A refused favorite
+  cycle no longer records a `model_change` or writes the global default for a
+  model that never ran, so the session and every new session stay on the model
+  that is actually active. The new entry is bookkeeping everywhere it is
+  consumed: it never reaches the model, keeps Claude SDK OAuth continuity across
+  resume, renders and searches in `/tree`, crosses the RPC
+  `append_session_entry` seam, and stays out of RPC status snapshots
+  ([#1528](https://github.com/code-yeongyu/senpi/pull/1528) by
+  [@rlaope](https://github.com/rlaope))
+
 - Claude SDK OAuth reattach no longer closes a healthy resumed query when
   normal completed-request cleanup aborts the request controller; the
   initialization cancellation listener is detached once initialization settles,
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -4867,7 +4867,7 @@ export class AgentSession {
 	}
 
 	/**
-	 * #1526: both guards below reject before `_switchActiveModel` appends its
+	 * #1526: every switch guard rejects before `_switchActiveModel` appends its
 	 * `model_change`, so a refused switch used to leave no entry, no event, and
 	 * no log line - indistinguishable from a switch the user never attempted.
 	 * Record the refusal, then rethrow the original error unchanged.
@@ -4895,19 +4895,41 @@ export class AgentSession {
 		this._emit({ type: "model_change_rejected", model, reason, detail, ...numbers });
 	}
 
+	/**
+	 * The admission only selects wording, and the switch wording's remedy
+	 * ("Compact the session, ...") is executable only when there is context to
+	 * compact. `_setModel` is also reachable from `session_start` (the
+	 * `recommended-models` builtin switches the model there), so derive the
+	 * admission from the branch instead of hardcoding a switch: with conversation
+	 * context this is a switch, on an empty session it is still a cold start.
+	 */
+	private _modelSwitchAdmission(): ModelUsabilityAdmission {
+		return this.sessionManager.hasContextMessages() ? "switch" : "start";
+	}
+
+	/**
+	 * The single guard seam for a model switch (#1526): every site that can
+	 * refuse a switch - the `_setModel` pre-flight, the post-`model_select`
+	 * revalidation in `_switchActiveModel`, and both cycle guards - records the
+	 * refusal and rethrows the original error unchanged, so no refusal is
+	 * observable on one path and invisible on its sibling.
+	 */
+	private _assertModelUsableForSwitch(model: Model<Api>, liveContextTokens: number): void {
+		try {
+			this.assertModelUsable(model, liveContextTokens, { admission: this._modelSwitchAdmission() });
+		} catch (error) {
+			this._recordRejectedModelChange(model, error);
+			throw error;
+		}
+	}
+
 	private async _setModel(
 		model: Model<Api>,
 		updateGlobalDefaults: boolean,
 	): Promise<SystemPromptChangeEvent | undefined> {
-		try {
-			// Classify as a switch, not a cold start: only the "switch" admission
-			// names the remedy ("Compact the session, then revalidate and retry the
-			// model switch"), which is the whole point of the message here.
-			this.assertModelUsable(model, this._getDownswitchLiveContextTokens(model), { admission: "switch" });
-			if (!(await this._modelRuntime.checkAuth(model.provider))) {
-				throw new Error(`No API key for ${model.provider}/${model.id}`);
-			}
-		} catch (error) {
+		this._assertModelUsableForSwitch(model, this._getDownswitchLiveContextTokens(model));
+		if (!(await this._modelRuntime.checkAuth(model.provider))) {
+			const error = new Error(`No API key for ${model.provider}/${model.id}`);
 			this._recordRejectedModelChange(model, error);
 			throw error;
 		}
@@ -5000,7 +5022,7 @@ export class AgentSession {
 			const systemPromptChange = opts.emitModelSelect
 				? await this._emitModelSelect(model, previousModel, opts.modelSelectSource)
 				: undefined;
-			this.assertModelUsable(model, liveContextTokens);
+			this._assertModelUsableForSwitch(model, liveContextTokens);
 			if (opts.appendSessionEntry) {
 				this.sessionManager.appendModelChange(
 					model.provider,
@@ -5106,7 +5128,10 @@ export class AgentSession {
 			const alternatives = favoriteModels.filter((entry) => !modelsAreEqual(entry.model, currentModel));
 			const onlyAlternative = alternatives.length === 1 ? alternatives[0] : undefined;
 			if (onlyAlternative) {
-				this.assertModelUsable(onlyAlternative.model, this._getDownswitchLiveContextTokens(onlyAlternative.model));
+				this._assertModelUsableForSwitch(
+					onlyAlternative.model,
+					this._getDownswitchLiveContextTokens(onlyAlternative.model),
+				);
 			}
 			return {
 				model: currentModel,
@@ -5117,7 +5142,7 @@ export class AgentSession {
 		}
 		const next = favoriteModels[selectedIndex];
 		const liveContextTokens = this._getDownswitchLiveContextTokens(next.model);
-		this.assertModelUsable(next.model, liveContextTokens);
+		this._assertModelUsableForSwitch(next.model, liveContextTokens);
 		const invalidatesCompaction =
 			this._modelSelectionChangesContext(currentModel, next.model) ||
 			currentModel?.provider !== next.model.provider ||
@@ -5130,8 +5155,6 @@ export class AgentSession {
 		const thinking = this._getThinkingForModelSwitch(next.model, next.thinkingLevel, next.thinkingSelection);
 
 		this.agent.state.model = next.model;
-		this.sessionManager.appendModelChange(next.model.provider, next.model.id);
-		this.settingsManager.setDefaultModelAndProvider(next.model.provider, next.model.id);
 		const previousTier = this._currentServiceTier;
 		const previousFastMode = this.isFastModeActive();
 		this._currentServiceTier = this._resolveServiceTier(next.model, next.serviceTier);
@@ -5139,19 +5162,25 @@ export class AgentSession {
 		// Apply thinking level and provenance from the favorite projection or remembered preference.
 		this._setThinkingLevel(thinking.level, false, thinking.selection);
 
-		// Post-switch, same contract as _switchActiveModel: the level in force AFTER the cycle.
-		this._emit({
-			type: "model_changed",
-			model: next.model,
-			thinkingLevel: this.thinkingLevel,
-			source: "cycle",
-		});
-		this._emitServiceTierChangeIfNeeded(previousTier, previousFastMode);
-
 		const previousSystemPrompt = this.agent.state.systemPrompt;
 		try {
 			const systemPromptChange = await this._emitModelSelect(next.model, currentModel, "cycle");
-			this.assertModelUsable(next.model, liveContextTokens);
+			// #1526: the `model_select` hook may have grown the system prompt, so the
+			// switch is not decided yet. Nothing durable - no `model_change`, no global
+			// default - may be written before this guard accepts, or a refused cycle
+			// would resume on a model that never ran (the ordering `_switchActiveModel`
+			// already uses).
+			this._assertModelUsableForSwitch(next.model, liveContextTokens);
+			this.sessionManager.appendModelChange(next.model.provider, next.model.id);
+			this.settingsManager.setDefaultModelAndProvider(next.model.provider, next.model.id);
+			// Post-switch, same contract as _switchActiveModel: the level in force AFTER the cycle.
+			this._emit({
+				type: "model_changed",
+				model: next.model,
+				thinkingLevel: this.thinkingLevel,
+				source: "cycle",
+			});
+			this._emitServiceTierChangeIfNeeded(previousTier, previousFastMode);
 
 			const cycleResult: ModelCycleResult = {
 				model: next.model,
@@ -5167,6 +5196,7 @@ export class AgentSession {
 			if (currentModel) this.agent.state.model = currentModel;
 			else delete (this.agent.state as { model?: Model<Api> }).model;
 			this.agent.state.systemPrompt = previousSystemPrompt;
+			this._currentServiceTier = previousTier;
 			throw error;
 		}
 	}
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -492,6 +492,18 @@ export type AgentSessionEvent =
 			thinkingLevel: ThinkingLevel;
 			source: ModelSelectSource;
 	  }
+	/** A switch the session refused; recorded so the attempt survives (#1526). */
+	| {
+			type: "model_change_rejected";
+			model: Model<any>;
+			reason: "context-budget" | "auth";
+			detail: string;
+			contextWindow?: number;
+			liveContextTokens?: number;
+			requiredTokens?: number;
+			shortfallTokens?: number;
+			safetyMarginProfile?: string;
+	  }
 	| {
 			type: "model_change_skipped";
 			model: Model<any>;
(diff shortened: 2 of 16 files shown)


code-yeongyu (changes requested):
**Verdict.** The recording mechanism itself is sound (the new suite is green — `vitest --run test/suite/agent-session-model-switch-rejection.test.ts` → 4 passed — and repo-wide `tsc --noEmit` is clean on 8154140), but the PR does not establish the invariant it claims to close #1526 with. On sibling paths of the *same* public API a refusal still vanishes exactly as before, and on one of them the session records a switch that never happened, which is strictly worse than a missing record: the resumed session and the global default end up on a model that was refused and never answered. Findings 1-3 and 4 are reproduced with runnable evidence against this head, not read off the diff. Also, `core/changes.md` carries ~12 unrelated blank-line deletions that add diff noise.

**1. `packages/coding-agent/src/core/agent-session.ts:5123-5124` — a refused *cycle* records a `model_change` that never happened (recorded model diverges from the model that ran).** `_cycleFavoriteModel` appends the entry and writes the global default *before* the post-`model_select` guard at `:5144`; the catch at `:5157` rolls back only `agent.state`. Reproduced with the fixture from `test/suite/model-usability-review.test.ts:69-90` (cycle rejects with `ModelUsabilityBudgetError`, live model stays `current`): `sessionManager.getEntries()` → `[..., {type:"model_change", modelId:"target"}]`, `sessionManager.buildSessionContext().model` → `{provider:"faux", modelId:"target"}`, `settingsManager.getDefaultModel()` → `"target"`, while `session.model?.id` is `"current"`. Consequence: after a refused cycle the session resumes on the refused model (`session-manager.ts:475-505` treats the last `model_change` as the authoritative explicit selection) and every new session inherits it as the default. Fix: move `appendModelChange` + `setDefaultModelAndProvider` to *after* the `assertModelUsable` at `:5144` — the ordering `_switchActiveModel` already uses (`:4993` asserts, `:4995` appends) — and add a regression test asserting no `model_change` entry and an unchanged default after a refused cycle.

**2. `packages/coding-agent/src/core/agent-session.ts:4993` — a `setModel` refusal from the second guard is still invisible, and still uses the cold-start wording.** The try/catch added at `:4894-:4903` wraps only the pre-flight pair; the guard that re-validates after `_emitModelSelect` may have grown the system prompt is outside it. Reproduced with the fixture from `test/suite/model-usability-review.test.ts:93-113` (`await session.setModel(target)` rejects): `sessionManager.getEntries()` → `[]`, `eventsOfType("model_change_rejected")` → `0`, and the message is `Model "faux/target" cannot start: context window 100000 tokens is 87463 tokens short of the 187463-token minimum (...)` — no entry, no event, no compaction remedy, i.e. all three symptoms #1526 describes, on the very method this PR instruments. Fix: funnel every guard site (`:4896`, `:4993`, `:5110`, `:5144`) through one `_assertModelUsableForSwitch(model, liveContextTokens)` helper that calls `_recordRejectedModelChange` and rethrows, instead of wrapping only the pre-flight pair.

**3. `packages/coding-agent/src/core/extensions/builtin/claude-sdk-oauth/session-binding.ts:138-145` — the new entry type is missing from `LEDGER_ONLY_ENTRY_TYPES`, so recording a refusal destroys Claude SDK OAuth continuity on resume.** `isLedgerOnlyEntry` (`:147`) returns false for `model_change_rejected`, so `bindingFromStoredBranch` (`:99-111`) refuses to restore. Measured with the fixtures of `test/claude-sdk-oauth-binding-anchor.test.ts`: branch `[marker, assistant, model_change]` → binding restored (`sdk-1`); branch `[marker, assistant, model_change_rejected]` → `undefined`. Via `session-registry-wiring.ts:81-84`, `undefined` deletes the stored sidecar on `session_start`. Consequence: a user who attempts a switch, gets refused, and later resumes loses SDK session continuity — fresh upstream session, full context re-send, prompt-cache loss — caused by an entry the model never sees. Fix: add `"model_change_rejected"` to the set next to `"model_change"`, plus a case in the anchor suite (mirroring `:153-171`).

**4. `packages/coding-agent/src/core/session-manager.ts:1103-1112` + `test/suite/agent-session-model-switch-rejection.test.ts:1-105` — durability is neither tested nor true in the motivating scenario.** `_persist` buffers every entry until the branch contains an assistant message. Re-running the PR's own fixture with `createHarness({persistSession: true, ...})`: after the refusal, `existsSync(sessionManager.getSessionFile())` is `false` — nothing on disk. All four new tests use `SessionManager.inMemory()` (`test/suite/harness.ts:135-137`) and assert `getBranch()`, so none of them can observe this. Consequence: in the exact incident the description cites (a 429 storm before any assistant reply lands), the post-mortem still finds no session file and the record dies with the process. Fix: add a `persistSession: true` case asserting the `model_change_rejected` line is in the JSONL after the session flushes and that `SessionManager.open` reads it back; if a bookkeeping-only session is deliberately left unflushed, state that limitation on `ModelChangeRejectedEntry` and in #1526.

**5. Consumers of the widened `SessionEntry` union were not updated, so the record is unreadable where it matters.** `src/modes/rpc/rpc-input-validation.ts:63-65` (`default: return false`) makes an `append_session_entry` carrying the new entry fail with "append_session_entry entry is malformed", so it cannot cross the seam `src/modes/rpc/rpc-client.ts:494` exists for. `src/modes/interactive/components/tree-selector.ts:859-861` falls to `default: result = ""` and `:359-365` does not list the type as a settings entry, so in `/tree` the entry appears in the default view as a blank row, with no search text either (`:601-611`). `src/modes/rpc/connection-handler.ts:224` still counts only `model_change`/`thinking_level_change` as bookkeeping, so a session whose sole content is a refused switch now ships its full entry list in every status snapshot. Consequence: the "reconstruct the incident afterwards" use case in the description does not work in the one entry browser the product ships. Fix: add the render + search + `isSettingsEntry` cases, the bookkeeping exclusion, and a validator branch (`typeof value.provider === "string" && typeof value.modelId === "string" && typeof value.detail === "string"`).

**6. `test/suite/agent-session-model-switch-rejection.test.ts:93-104` — the "successful switch" case cannot fail in the way it claims, and half the new discriminant is untested.** `harness.getModel()` returns `faux-roomy`, which is already the active model, so `expect(harness.session.model?.id).toBe(roomy.id)` at `:103` holds before `setModel` is called and the case cannot detect a spurious rejection on a real cross-model switch (it only appends a `model_change`, verified). The `reason: "auth"` branch (`agent-session.ts:4868`) is never exercised, though it is one line: `createHarness({withConfiguredAuth: false})` + `setModel(roomy)` yields `{type:"model_change_rejected", reason:"auth"}` (verified). Fix: switch to the *other* admissible model in that case, and add an auth-refusal case asserting `reason` and `detail`.

**7. (minor) `packages/coding-agent/src/core/agent-session.ts:4896` — hardcoding `admission: "switch"` is not safe on the cold-start caller the design note overlooks.** `_setModel` is reachable from `session_start`: `src/core/extensions/builtin/recommended-models/index.ts:136` calls `pi.setModel(target.model)` inside its `session_start` handler (through `src/core/extensions/loader.ts:585`). On an empty session a refusal now reads "Compact the session, then revalidate and retry the model switch" — a remedy the user cannot execute, since there is nothing to compact. Fix: derive the admission from session state (any message entry on the branch → `"switch"`, else `"start"`), which still gives the rate-limit case the wording and remedy this PR wants. Also consider asserting `error.projection.admission === "switch"` in `test/suite/model-usability-budget.test.ts:90-95` rather than pinning the whole sentence, so the test fails for the reason it is about.

*Evidence:* worktree at 8154140, `bun`/`vitest 4.1.11` on macOS arm64; commands run: `vitest --run test/suite/agent-session-model-switch-rejection.test.ts` (4 passed), `tsc --noEmit` at the repo root (clean), and four throwaway probe suites (deleted) reproducing findings 1-4, 6 with the fixtures named above.

code-yeongyu:
Every finding from the review is addressed on the rebased head `b0bd7f174` (branch force-pushed with lease after rebasing onto current `origin/main`, `e66c1553a`). The fix-up commit is `b0bd7f17` ("fix(coding-agent): make every refused model switch observable and inert"); the original commit was rebased unchanged beneath it. Per finding:

**1. Refused cycle records a `model_change` that never happened.** `_cycleFavoriteModel` no longer appends its `model_change`, writes the global default, or emits `model_changed`/`service_tier_changed` until the post-`model_select` guard accepts — exactly the ordering `_switchActiveModel` uses — and its catch restores the service tier along with the model and prompt. Pinned by the new "favorite cycle refused after `model_select`" case: no `model_change` entry, `buildSessionContext().model` not the target, default unchanged, `model_changed` never emitted. Mutation: moving `appendModelChange` + `setDefaultModelAndProvider` back before the guard fails exactly that test (RED) and passes again on restore (GREEN).

**2. `setModel` refusal at the second guard is invisible and cold-worded.** All four guard sites (`_setModel` pre-flight, `_switchActiveModel` post-`model_select`, both cycle guards) funnel through one `_assertModelUsableForSwitch(model, liveContextTokens)` that calls `_recordRejectedModelChange` and rethrows unchanged. The `model-usability-review.test.ts:93-113` fixture now yields one `model_change_rejected` entry + event with the switch wording ("cannot switch … Compact the session"). Mutations: reverting the `_switchActiveModel` site to plain `assertModelUsable` fails the post-`model_select` case; deleting `_recordRejectedModelChange` from the shared seam fails seven cases; each restored to green.

**3. `LEDGER_ONLY_ENTRY_TYPES`.** `"model_change_rejected"` added next to `"model_change"`, pinned by a new anchor-suite case mirroring the `model_change` one (`[marker, assistant, model_change_rejected]` → binding restored). Mutation: removing the set entry fails exactly that case, restore → green. The entry also never reaches the LLM context (`buildSessionContext().messages` length asserted unchanged), which is what makes the classification safe.

**4. Consumers taught the type.** RPC validator (`provider`/`modelId`/`detail` are strings), `/tree` render (`[model rejected: <id> (<reason>)]`), search text, `isSettingsEntry`, and the `connection-handler` bookkeeping exclusion. Three separate tree-selector tests pin render / search / default-hidden, one validator test pins accept-and-reject-malformed, and the RPC-state case pins that a session whose only content is a refused switch still omits `entries` from its status snapshot (with a control showing a real user message turns it back on). Each of the four production edits was mutation-proved individually.

**5. Verification gaps.** Two `persistSession: true` cases: past the first assistant reply the `model_change_rejected` line is on disk and `SessionManager.open` reads it back with `detail` intact; before any assistant reply the file legitimately does not exist yet (documented on `ModelChangeRejectedEntry` — the shared pre-assistant `_persist` buffering contract, not a special case) and the record flushes with the first reply, which the second case asserts. The `reason: "auth"` branch is covered by a `withConfiguredAuth: false` case asserting entry + event with `reason` and `detail`. The success case now switches `faux-roomy` → `faux-spacious`, a genuinely different admissible model, and asserts the `model_change` entry, the active model, and the default. Mutations for both the auth seam and the `appendModelChange` on success were killed.

**6. Success case and auth branch (file-level).** — same evidence as 5.

**7. `admission: "switch"` hardcoded on a cold-start caller.** The admission is now derived from the branch (`sessionManager.hasContextMessages()`), so the `recommended-models` `session_start` path gets `"start"` wording and a populated session gets the switch wording with its compaction remedy. `model-usability-budget.test.ts` returns to its original sentence and additionally pins `error.projection.admission === "start"`; the new suite pins `"switch"` for a session with context. Mutations both ways (`"switch"` hardcoded fails the budget test, `"start"` hardcoded fails three new cases) were killed.

Also per the verdict note: `core/changes.md` is back to a pure addition — the ~12 unrelated blank-line deletions are gone. New tracker entries were added for the rpc, interactive, and claude-sdk-oauth surfaces (`scripts/check-pr-changelog.mjs --base e66c1553a` → PASS), and a `CHANGELOG.md` `[Unreleased]` Fixed entry was added so the changelog gate passes without the `no-changelog` label.

**Evidence** (run on a disposable worktree, not the shared checkout): `bun run test -- test/suite/agent-session-model-switch-rejection.test.ts` → 10 passed; `test/tree-selector.test.ts test/rpc-input-validation.test.ts test/claude-sdk-oauth-binding-anchor.test.ts test/suite/model-usability-budget.test.ts test/suite/model-usability-review.test.ts test/suite/agent-session-model-extension.test.ts` → 87 passed; `bun run check` → clean (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, platform-lock, `tsc --noEmit`, browser-smoke). Full coding-agent suite on the head: 10298 passed / 99 failed — the failing set is byte-identical to a run of **unmodified origin/main** in the same environment (same 28 files: missing `dist/` build artifacts, qa-port bind contention, rpc worker host timeouts), i.e. pre-existing environment failures, not regressions. 14 mutations were run end to end (mutate → RED → restore → GREEN); the log is in the task report.

code-yeongyu (approved):
All findings from review 5167950734... resolved: the refused cycle now records nothing durable, the guard sites funnel through one recording seam, model_change_rejected is in LEDGER_ONLY_ENTRY_TYPES, and the RPC validator / tree selector / bookkeeping consumers were taught the type. Verified by 14 mutations (each RED then GREEN on restore) and a full-suite run whose failing set is byte-identical to unmodified origin/main. Approving.

code-yeongyu (approved):
Findings resolved and mutation-proved: the refused cycle records nothing durable, all four guard sites funnel through one recording seam, model_change_rejected is in LEDGER_ONLY_ENTRY_TYPES, and the RPC validator / tree selector / bookkeeping consumers were taught the type. 14 mutations each RED then GREEN on restore; full-suite failing set byte-identical to unmodified origin/main. Approving.

PR #1540, opened 2026-09-10
fix(publish): stage a Bun-compile-safe css-tree into the tarball

code-yeongyu:
## What changed

- `scripts/prepare-bun-compile-assets.mjs` becomes a module plus CLI entry and owns ONE implementation of the portable css-tree data inlining, exported as `inlineCssTreeCompileData(nodeModulesRoot)`. The jsdom rewrites (`patchJsdomBinaryLookups`) and the imagegen skill copy (`stageImageGenSkill`) are exported beside it; the script keeps its path, its behaviour and its summary line.
- `copyPublishDependencies()` in `scripts/prepare-senpi-bundled-workspaces.mjs` calls the inliner on the staged runtime, so both `publish.mjs` and `local-release.mjs` produce a compile-safe tarball from the single staging entry.

## Why

css-tree resolves `data/patch.json`, the three mdn-data dictionaries and its own `package.json` through `createRequire(import.meta.url)` at module scope. Bun's compiled filesystem serves no dynamic require, so every consumer that compiles this engine dies on the first webfetch HTML conversion (`format: markdown`/`text`):

```
Cannot find module '../data/patch.json'
Require stack:
- /$bunfs/root/<binary>
```

The published `@code-yeongyu/senpi@2026.9.9-2` tarball still ships that un-inlined css-tree — `tar -xzOf <tgz> package/node_modules/css-tree/lib/data-patch.js` starts with `import { createRequire } from 'module';`. The inlining existed only in `build:binary` / `build-binaries.sh`, `scripts/` is not in the published `files`, and `publish.mjs` packs with `--ignore-scripts`, so `prepublishOnly` never ran at publish time. Downstream consumers had to patch the installed engine themselves.

**Why jsdom is deliberately left alone:** `patchJsdomBinaryLookups` hardcodes `../../node_modules/jsdom/.../xhr-sync-worker.js` under `typeof Bun !== "undefined"`, which is correct only inside our standalone binary layout. Shipping it would hand a plain Bun consumer of the npm package a wrong path, so it stays binary-only. A counterfactual confirmed jsdom's own `__dirname` stylesheet read and `require.resolve` survive `bun build --compile` untouched.

## Evidence

Scenario 1 — unit, `node --test` (RED before the implementation, GREEN after; the inliner was stubbed to keep RED assertion-shaped):

```
✖ inlines css-tree's data so a compiled binary never resolves it at runtime
✖ leaves an already prepared css-tree byte-identical
✖ fails loud when css-tree's data module matches no known shape
✖ stages a Bun-compile-safe css-tree without touching the installed source tree
ℹ pass 7  ℹ fail 4        →  after: ℹ pass 11  ℹ fail 0
```

Mutation pass (each new assertion has a mutation that kills it): serializer → `{}` fails the data-equality and staging cases; removing the drift guard fails only the fail-loud case; removing the staging call fails only the wiring case; restore → 11/0.

Scenario 2 — the real artifact. Full `prepareSenpiBundledWorkspaces()` then `npm pack` (240 bundled packages, 34 MB, 25208 files):

```
packed css-tree/lib/data-patch.js  →  const patch = {          (was: import { createRequire } ...)
packed css-tree/lib/data.js        →  import patch from './data-patch.js'; const mdnAtrules = {
packed css-tree/lib/version.js     →  export const version = "3.2.1";
createRequire left in packed css-tree  →  none
packed jsdom computed-style.js     →  fs.readFileSync(path.resolve(__dirname, ...))   (unchanged)
packed jsdom XMLHttpRequest-impl   →  require.resolve("./xhr-sync-worker.js")         (unchanged)
```

The developer's `node_modules/css-tree` is byte-identical before and after staging — only the staged copy is rewritten.

Scenario 3 — end-user proof, same published dist, only difference is this inliner. Install the tarball into a scratch dir, `bun build --compile` a probe importing `dist/core/extensions/builtin/webfetch/webfetch/content.js`, run it on fixture HTML, with no downstream patch hook anywhere:

```
before (pristine 2026.9.9-2):  error: Cannot find module '../data/patch.json' from '/$bunfs/root/probe'
after  (+ inlineCssTreeCompileData):  MARKDOWN_OK 1877 / TEXT_OK 1875
```

Scenario 4 — regression: `npm run test:scripts` → `tests 207, pass 207, fail 0`, with the pre-existing jsdom and cwd-independence cases untouched.

## Notes

- The module/CLI entry check compares real paths: macOS `TMPDIR` is a symlink (`/var/folders` → `/private/var/folders`), and comparing `argv[1]` to `import.meta.url` directly would silently skip `main()` — which is exactly how the existing "resolves the repository root independently of the caller's working directory" test caught it.
- Inlining leaves css-tree's `createRequire` import unused, so it is dropped once nothing calls it; a shipped artifact should carry no dead resolver.
- Unrelated pre-existing finding: `node scripts/audit-changes-md.mjs` reports `packages/client/src/unix.ts` as uncovered by its nearest tracker (introduced by an upstream sync, not by this PR).

---
## Summary by cubic
Fixes the published tarball so `css-tree` works when consumers compile it with Bun. `css-tree` resolves its data through `createRequire` at module scope, which Bun's compiled filesystem can't serve, so every consumer died on the first webfetch HTML conversion; the tarball now carries the inlined data and the developer's installed dependency is never rewritten.

- `scripts/prepare-bun-compile-assets.mjs` is now a module plus CLI entry and exports the inlining as `inlineCssTreeCompileData()`.
- `copyPublishDependencies()` stages the inlined `css-tree`, so `publish.mjs` and `local-release.mjs` both produce a compile-safe tarball.
- The jsdom rewrites stay binary-only because their worker path is only correct inside the standalone binary layout.
- A `css-tree` data shape that no longer matches the known pattern fails loud instead of shipping a binary that breaks at runtime.
- The unused `createRequire` import is dropped from the shipped `css-tree` files.

<sup>Written for commit 36bcb3f1e691e0b1812a9219b8cdd4779034c612. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:

PR #1595, opened 2026-09-11
perf(coding-agent): batch evicted resume hydration

### LINKED ISSUE #1407: Resuming image-heavy sessions stalls the TUI; evicted strings cause repeated full-history reads
### What happened?
Reopening one image-heavy session causes severe UI lag. Its JSONL is 117,277,295 bytes with 81 embedded images; base64 payloads account for 96.16%. It has no compaction entries. The private transcript and images are not attached.

An extension-free core benchmark confirms read amplification in `getEntries() -> _materializeEntry() -> loadEntriesFromFile()` when the 64 MiB resident-string cache evicts entries:

| Synthetic payload | First getEntries() | Full-history reads |
| --- | ---: | ---: |
| 64 MiB | 0.56 ms | 0 |
| 80 MiB | 1,087.81 ms | 16 |

The 80 MiB case processes 1.34 GB through the reader. An immediate repeat is memoized, but appending one text entry causes 16 reads again (1,046.29 ms). These are **in-memory synthetic timings, not disk or UI timings**; the share of real UI lag attributable to this path remains unmeasured. Related but different payload: #648.

### Steps to reproduce
Open/resume a session containing many embedded tool-result images. For an isolated core reproduction, run the command below from the installed `@code-yeongyu/senpi` package directory. It imports the core directly without loading extensions and uses an in-memory filesystem fixture; it does not touch a real session or call a model. The fixture uses synthetic image strings to exercise storage, not PNG rendering. A literal `pi -ne` UI comparison has **not** been performed.

<details>
<summary>Extension-free reproduction command</summary>

```bash
python3 - <<'PY'
import subprocess
source = r'''
import fs from 'node:fs';
import { syncBuiltinESMExports } from 'node:module';
import { performance } from 'node:perf_hooks';
const { SessionManager } = await import('./dist/core/session-manager.js');
const path = '/__senpi_synthetic_read_only__.jsonl';
const saved = Object.fromEntries(['existsSync','openSync','readSync','closeSync'].map(k => [k,fs[k]]));
let fixture, offset = 0, opens = 0, bytes = 0;
const fd = 987654321;
fs.existsSync = p => p === path || saved.existsSync(p);
fs.openSync = (p, ...args) => { if (p !== path) return saved.openSync(p,...args); if (args[0] !== 'r') throw Error('write forbidden'); offset = 0; opens++; return fd; };
fs.readSync = (f,b,o,n,p) => { if (f !== fd) return saved.readSync(f,b,o,n,p); const count = Math.min(n, fixture.length-offset); fixture.copy(b,o,offset,offset+count); offset += count; bytes += count; return count; };
fs.closeSync = f => { if (f !== fd) return saved.closeSync(f); };
syncBuiltinESMExports();
try {
  for (const count of [64,80]) {
    const header = { type:'session', version:3, id:'synthetic', timestamp:'2026-09-01T00:00:00.000Z', cwd:'/tmp' };
    const entries = [header, ...Array.from({length:count}, (_,i) => ({type:'message',id:String(i),parentId:i?String(i-1):null,timestamp:header.timestamp,message:{role:'user',timestamp:0,content:[{type:'image',mimeType:'image/png',data:'A'.repeat(1024*1024)}]}}))];
    fixture = Buffer.from(entries.map(e=>JSON.stringify(e)).join('\n')+'\n');
    const manager = new SessionManager('/tmp','',path,false,undefined,entries);
    const measure = () => { opens=0; bytes=0; const start=performance.now(); const result=manager.getEntries(); return {ms:Math.round((performance.now()-start)*100)/100,fullFileReads:opens,bytesRead:bytes,entries:result.length,complete:result.slice(0,count).every(e=>e.message.content[0].data.length===1024*1024)}; };
    const cold=measure();
    const cached=measure();
    manager.appendEntry({type:'message',id:'new',parentId:String(count-1),timestamp:header.timestamp,message:{role:'user',timestamp:0,content:[{type:'text',text:'synthetic'}]}});
    const afterAppend=measure();
    console.log(JSON.stringify({count,fileBytes:fixture.length,resident:manager.getResidentStoreStats(),cold,cached,afterAppend}));
  }
} finally { Object.assign(fs,saved); syncBuiltinESMExports(); }
'''
result = subprocess.run(['node','--input-type=module'], input=source, text=True, timeout=90)
raise SystemExit(result.returncode)
PY
```

</details>

### Expected behavior
Opening large sessions should remain responsive. Restoring evicted payloads should not reparse the entire history separately for each entry.

### Version
`@code-yeongyu/senpi 2026.9.5`, via `omo-ai 5.0.0-0.beta.43`; Node.js 26.8.1; macOS 26.6.2; Apple M1 Max (arm64); Ghostty.

*AI-assisted drafting and diagnostic analysis. The UI symptom is user-reported; the benchmark results were measured locally. This disclosure is in the body rather than a follow-up comment to avoid triggering GitHub Actions.*

code-yeongyu:
## Summary

Fixes #1407 by removing repeated full-history JSONL parsing when /resume materializes entries whose large strings were evicted from the bounded resident cache.

The ordered materialization pass now performs one authoritative history load and preserves branch order, cached view identity, and message-entry position tracking. The real resume QA driver also accepts both the current pi v and legacy senpi v startup banners.

## Measured evidence

- Current-main baseline: 104.29 ms end-to-end median / 199.42 ms p95 for the 5,000-message benchmark; 5,000 entries/messages; digest 451e6fba64db501f6faf09d1ae231f24e63edef36ca9f2f51c6470dae096c35.
- Frozen fixed tree: 91.96 ms median / 100.99 ms p95; 5,000 entries/messages; identical digest. Median is 11.8% lower and p95 is 49.4% lower on the ordinary control flow.
- Issue #1407 fixture: seven evicted 1 MiB entries changed from 490 sentinel parses and 388.25 ms median in the mutation control to 70 parses and 55.70 ms median on the fix (85.7% lower).
- RED -> GREEN: the resident-mirror regression fails on the old per-entry path and passes with the batch helper; the mutation restored the old behavior and failed again.

## QA

- Real hermetic PTY /resume with 5,000 messages: final marker, xterm.js grid assertion, valid PNG, and cleanup sequence passed; submit-to-final-marker was 192.18 ms.
- Focused tests: resident-mirror 7/7; resume-load regression 4/4.
- Root check passed after integrating the latest main; the initial check failure was the pre-existing prompt fixture contract fixed by PR #1592.
- CPU profile identified session JSON parsing, resident-store transformation, compact-entry recovery, and render-signature work as the source-owned baseline costs.

Detailed raw artifacts are held locally and available to reviewers on request. No credentials or external provider calls were used.


Altairpaca (commented):
Reviewed against the #1407 diagnosis and the current #1595 head. The batching boundary looks correct to me: resident entries remain resident, missing entry IDs are collected before a single ordered full-history materialization, `getEntries()` / `getBranch()` preserve their ordering/cache semantics, and message-position bookkeeping is updated after materialization rather than during repeated fallback reads.

The maintainer-side evidence is also aligned with the actual failure mode: the RED→GREEN regression, mutation control showing the repeated parse amplification collapsing from ~490 parses to ~70 in the synthetic case, the 5,000-message transcript/digest equivalence check, the hermetic PTY `/resume` validation, and green CI/changelog checks cover the important performance and semantic risks.

I did not find a blocking issue in the current diff. I would keep this PR narrow and avoid adding unrelated client/provider caching work; from my review side it looks ready for maintainer merge/release handling. I have not independently rerun the full suite, so this is a source/evidence review rather than a separate execution claim.

PR #1605, opened 2026-09-11
feat(ai): add the Devin Cascade model transport

### LINKED ISSUE #1604: Devin Cascade model transport is missing: the OAuth credential cannot be used
## Summary
senpi has the Devin OAuth login (#1601) but nothing to spend the credential on: there is no Cascade transport, so a signed-in user still cannot select a Devin model. This issue tracks porting Devin's model transport so Devin models actually stream in senpi.

## Expected (ideal state)
- A native `devin-agent` API speaking Cascade's Connect protocol: server-streaming `GetChatMessage`, gzipped request frame, 5-byte framed responses, JSON trailer.
- Cascade deltas mapped onto senpi's assistant event protocol: text, thinking (with signatures), tool calls, usage accounting, stop reasons, typed errors, and clean abort.
- A `devin` provider bound to the merged Devin OAuth credential, with a public model seed and credential-scoped `GetCliModelConfigs` discovery that never publishes an empty catalog.
- Registered as a first-class api id and builtin provider, generated protobuf code produced the same way as the Cursor schema.

## Actual
No `devin-agent` api, no Cascade framing, no Devin provider, no discovery.

## Scope
In: vendored schema subset + generated code, the api implementation, framing, request building, delta mapping, discovery, provider factory, registry wiring, tests.
Out: Devin-specific agent features beyond chat streaming (model assignment RPC, account usage panels).

## Verification
Offline vitest coverage driving real protobuf frames over a local HTTP server for the happy path, stop-reason mapping, HTTP failures, abort, discovery normalization and discovery failure, plus an end-to-end stream drive against a stub Cascade edge.

code-yeongyu:
## Summary
Implements the Devin (Cognition) Cascade model transport, so the Devin OAuth credential merged in #1602 can finally be spent: a signed-in user can select a Devin model and stream real completions through the native `devin-agent` Connect/protobuf protocol. Ported from Devin's released-CLI behaviour and adapted to senpi's provider-neutral architecture, exactly the way `cursor-agent` was.

## Changes
- **Schema** - `packages/ai/proto/devin/cascade.proto` vendors the Cascade subset the transport needs, generated to `src/api/devin-agent/gen/cascade_pb.ts` with protoc-gen-es v2.13.0 and the existing enum transform. Every field number matches upstream, so unknown upstream fields round-trip instead of corrupting a message.
- **Transport** - `devin-agent.ts` runs one server-streaming `GetChatMessage` call per turn; `frames.ts` owns Connect framing (5-byte prefix, gzip flag, trailer flag, 64 MiB payload cap); `request.ts` builds the request (Cascade has no system role, so the system prompt rides the top-level `prompt` field and history becomes flat entries whose `source` carries the role, with deterministic message ids); `stream-state.ts` maps implicit Cascade deltas onto senpi's block protocol and stop-reason vocabulary.
- **Catalog** - `discovery.ts` performs credential-scoped `GetCliModelConfigs` discovery and returns undefined on failure or an empty roster, so a working provider is never replaced by an empty catalog; `providers/devin.ts` binds the merged Devin OAuth flow and ships the public SWE seed.
- **Wiring** - the `devin-agent` api id, its options type, the lazy Node-only boundary, the builtin api-registry entry and the builtin provider registration.

## QA & Evidence
- **RED first** for each layer: the wire, stream and provider suites were captured failing (missing modules) before any production file existed.
- **GREEN**: 17 new tests pass - framing incl. the cap guard, Cascade role mapping, gzip round-trip, streamed text/thinking/tool-call/usage mapping, stop-reason mapping, HTTP failure, abort, discovery normalization, discovery failure and empty-roster fallback, provider seed and registry wiring.
- **Mutation proof (these tests can fail)**: dropping the session-token prefix kills 3 wire tests; publishing an empty discovery result kills the seed-retention test; mapping `FUNCTION_CALL` to stop kills the toolUse test; mapping `MAX_TOKENS` to stop kills the length test. Reverting each restores green.
  The first mutation round exposed a real gap - toolUse was inferred from the presence of a tool-call block, so a broken stop-reason mapping could not fail any test. Two tests that exercise stop reasons *without* a tool-call block were added, and both now fail under their mutation.
- **Typecheck**: `tsgo -p packages/ai/tsconfig.build.json --noEmit` clean, plus the repo pre-commit gate (biome, repo-wide `tsc --noEmit`, browser smoke).
- **Full suite**: `packages/ai` 282 files / 2731 tests pass, 0 failures.
- **Real-surface drive**: the real provider + a credential in the OAuth credential-store shape streamed end to end against a stub Cascade edge over real HTTP with real gzipped protobuf frames. Observed: `POST /exa.api_server_pb.ApiServerService/GetChatMessage` with `connect-content-encoding: gzip` and the `devin-session-token$` bearer, request carrying the system prompt / model uid / tool definitions / mapped history roles, streamed thinking and text deltas, a tool call materialised with parsed arguments, usage 1234/88/200/10 and terminal `done reason=toolUse`. Stub closed, no leftovers. No real Devin endpoint and no real token were used.
- Every new file is under the 250 pure-LOC ceiling (largest: stream-state 129).

## Scope
Chat streaming and catalog only. Devin's model-assignment RPC and account-usage panels stay out of this PR.

## Related Issues
Fixes #1604

---
## Summary by cubic
Adds the Devin (Cognition) Cascade model transport so a signed-in Devin OAuth user can select a Devin model and stream completions through the native `devin-agent` Connect/protobuf protocol (fixes #1604).

- Vendors the Cascade protobuf subset with field numbers matching upstream, so unknown fields round-trip instead of corrupting a message.
- Runs one server-streaming `GetChatMessage` call per turn, with gzip framing, a 64 MiB frame cap, and the HTTP status/headers forwarded via `onResponse`.
- Maps streaming deltas for text, thinking, tool calls and usage onto senpi's block and stop-reason protocol, keeping a truncated tool-call turn as `length` instead of `toolUse`.
- Ends a Cascade error stop with an error event instead of a normal done.
- Adds session-token-authenticated model discovery that falls back to the bundled SWE seed when discovery fails or returns empty.
- Registers the `devin-agent` api id and the Devin provider with the existing Devin OAuth flow.

<sup>Written for commit 6a4c64c345569084aafcf805ab34ee3af216ddae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
**Self-review round — three real defects found in my own code and fixed, each with a mutation-proven regression test.**

1. **A Cascade error stop was terminating as a `done` event.** `mapStopReason` can yield `error` (ERROR / CONTENT_FILTER), but the terminal push was `{ type: "done", reason: output.stopReason as "stop" | "length" | "toolUse" }` — the `as` cast laundered an invalid reason, so a failed turn would have been consumed as a successful assistant message. It now terminates as an `error` event with a message, and the cast is replaced by an exhaustive narrowing.
2. **A truncated turn carrying a tool call was reported as `toolUse`.** The unconditional override meant a server-reported `length` lost to inference, advertising a truncated (possibly half-serialized) tool call as complete. Only the default `stop` is upgraded now.
3. **Model discovery was unauthenticated.** `fetchDevinModels` put the session token only in the protobuf `Metadata` and sent no `authorization` header, so Cascade would reject it and every signed-in account would silently degrade to the static seed. It now sends the same `Bearer devin-session-token$…` header as the chat call.

Also: real response headers are forwarded to `onResponse` instead of `{}`, and two `as unknown as Model<…>` double casts were removed (they were never needed — the `Model` shape already matched).

**State now**: 20 tests on this surface, full `packages/ai` suite 282 files / 2734 tests pass, typecheck and the repo pre-commit gate clean, real-surface drive re-verified PASS after the changes.
Every fix above was mutation-proven: restoring the old behaviour fails exactly its new test, and reverting restores green.

PR #1610, opened 2026-09-11
fix(imagegen): enable native images for Codex OAuth

DescodeNet:
## Summary

- Enable the existing native image-generation surface for official OpenAI Codex
  OAuth endpoints while preserving the `openai-codex-responses` transport.
- Keep proxy defaults conservative and preserve explicit capability opt-in,
  opt-out, and the global disable flag.
- Contribute the bundled image skill when native generation is available,
  without requiring a separate image API key.

## Reproduction

1. Log in to the OpenAI Codex provider with ChatGPT OAuth.
2. Select `openai-codex/gpt-6-astra` without changing its API or endpoint.
3. Request an image using the built-in image-generation capability.

Before this change, the native gate rejected `openai-codex-responses` before
checking its endpoint. With a compatible gateway configured, the client
`generate_image` path could choose that gateway instead of the existing OAuth
session. Without client image credentials, the bundled image skill was also
absent.

The fix stays in the two existing image builtins. It does not change provider
authentication, credential precedence, request transport, image persistence, or
model catalogs. Azure and unrelated API behavior remain unchanged.

## Verification

- New regressions were run before the production patch: 7 failed, 17 passed.
- Focused image suites after the patch: 10 files, 150 tests passed.
- Independent original-versus-patched gate probe: original failed the native
  capability assertion; patched passed with the same Codex model.
- `bun run check`: passed.
- `bun run build`: passed.
- Isolated CLI smoke: 8/8 checks passed.
- RPC `get_state`: successful.
- Real CLI mock tool loop with `openai-responses`: 4/4 checks passed.
- Live OpenAI OAuth smoke through the patched automatic extension path:
  the outgoing native tool requested `gpt-image-2.5-sunburst`, HTTP 200,
  one generated PNG (1254 x 1254, 1,434,514 bytes), and the real credential file
  was unchanged. The observer only inspected the payload; it did not inject or
  modify the image tool.

Broader validation did not pass completely on this macOS machine:

- Root scripts: 206/207 passed. The unchanged issue-446 packaging test fails
  when reading `packed.files` after `npm pack --json`; reproduced in the
  original upstream clone and with direct Node execution.
- Workspace suites: 11/12 passed. Coding-agent passed 10629 tests, with five
  `ENOTEMPTY` RPC cleanup failures and one 1 ms cold-storage timestamp assertion.
  The unchanged upstream clone reproduced the RPC cleanup failure class.
  The timestamp test passed in that isolated baseline run and remains an
  unisolated full-suite timing failure.

These failures are outside the changed image paths. No tests were removed,
skipped, or weakened.

Independent gate review: approved with no blocking findings. The reviewer ran
135 tests across nine relevant files, non-writing Biome checks, and root
typechecking; independently reproduced the base/candidate gate difference; and
validated the live PNG and automatic-injection receipts.

## Scope

Regression coverage includes official endpoints, wrong APIs, malformed and
lookalike URLs, explicit capability overrides, OAuth-only skill discovery,
native/client mutual exclusion, and switching away from native generation.
The two existing no-client-credential/native-capable matrix rows now expect the
bundled guidance to be available.

No dependency, lockfile, release changelog, or generated-file changes.

---
## Summary by cubic
Enables native image generation for official OpenAI Codex OAuth sessions without requiring a separate image API key. The gate previously rejected the `openai-codex-responses` transport, so these sessions fell back to the client `generate_image` tool and never got the bundled image skill.

**Bug Fixes**
- `openai-codex-responses` on the exact `chatgpt.com` host now passes the native image gate without changing the model's transport or base URL.
- The bundled `gpt-image-gen` skill is contributed whenever native generation is active, even without client image credentials.
- Proxies, lookalikes, malformed URLs, explicit opt-out, and the global disable flag keep their prior behavior.

**Verification**
- Added gate unit tests and a Codex lifecycle suite; updated arbitration truth-table rows for native-without-credentials.
- No dependency, lockfile, or generated-file changes; the changelog records the fix.

<sup>Written for commit d837a4d6cf3cad28d3f12c3d3635d97c87cec657. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1610?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


DescodeNet:
Updated for the changelog gate and synced with `main`.

- `docs(changelog)`: added the `[Unreleased] > Fixed` entry in `packages/coding-agent/CHANGELOG.md`. The required `Changelog gate` was failing with `runtime source changed without a CHANGELOG.md entry` for `imagegen/index.ts` and `openai-image-gen/gate.ts`; `scripts/check-pr-changelog.mjs` now returns PASS locally against `main`.
- `merge`: synced with `code-yeongyu/main@7be1375c`. The branch was behind the `2026.9.16-3` release cut, so the changelog edit conflicted. `CHANGELOG.md` was the only conflict and the entry now sits in the current `[Unreleased] > Fixed`. The diff against `main` is still the original 8 files plus that one line.

Note: `CONTRIBUTING.md` says not to edit `CHANGELOG.md`, but the required gate asks for either the entry or the `no-changelog` label, which I can't apply myself. Happy to drop the changelog commit if you would rather label it instead.

`Changelog gate` and `Check and test` are still blocked on `workflows awaiting approval` (first-time contributor). Could you approve the workflow runs when you get a chance?

Issue #1603, opened 2026-09-11
cursor: aggregate history cap deletes whole turns even with zero tool results, leaving ~6K tokens of a 1M-token window


DevNewbie1826:
### What happened?

`truncateToolResultBodies` in `packages/coding-agent/src/core/agent-session.ts` runs on every request for `cursor` / `cursor-cli-oauth`. Alongside the per-result 2000-char cap from #1043, it enforces an aggregate `CURSOR_TOOL_RESULT_MAX_BYTES = 50_000` budget and, when history still doesn't fit, deletes the **oldest whole turns**.

The per-result cap looks right and I'm not asking to touch it. The aggregate part has three problems:

- It fires on conversations with **zero tool results**. The repro below never calls a tool, and the previous turn is still dropped — that's outside what #1043 was about.
- Deletion is silent and lossy: no summary, no marker in the transcript, nothing in the session log. The model simply stops knowing earlier turns, and a reload can't recover them.
- The budget is a fixed byte constant unrelated to the model's context window. On `cursor/kimi-k3` (`contextWindow: 1048576`) the retained history measures ~25 KB ≈ 6.2K tokens, ~0.6% of the window. Compaction never gets a chance, because its 80%-of-window threshold can't be reached once history is already capped this low.

### Steps to reproduce

Validated with `-ne` (no extensions loaded).

```bash
FILLER=$(python3 -c "print('The quick brown fox jumps over the lazy dog. '*700)")

DEBUG_CURSOR=2 senpi -ne --model cursor/kimi-k3 --session-dir /tmp/x \
  -p "$FILLER
Remember this codeword: BANANA7. Reply with just: OK"

DEBUG_CURSOR=2 senpi -ne --model cursor/kimi-k3 --session-dir /tmp/x --continue \
  -p "What codeword did I ask you to remember? Reply with just the codeword."
```

Observed:

```
turn 1 -> OK                                      builtRunRequest {"bytes":30237,"tools":17}
turn 2 -> I don't have a codeword from earlier.   builtRunRequest {"bytes":245,"tools":17}
```

Control — same two commands, but a short first message instead of the 30 KB one:

```
turn 2 -> BANANA7                                 builtRunRequest {"bytes":348,"tools":17}
```

So the turn is evicted purely because of its size, with no tool result involved.

### Expected behavior

Turn 2 recalls the codeword.

[...]

PR #1614, opened 2026-09-11
fix(coding-agent): size Cursor admission from the server-reported context limit

DevNewbie1826:
Issue: #1603

## The defect

Cursor admission enforced a hardcoded 50,000-byte cap by deleting the oldest complete turns, destroying conversation history that no later stage can recover: a 551-message transcript collapsed to 4 messages, and a conversation containing zero tool results lost its 30 KB first turn outright. Model ceilings also came from a static catalog guess (`kimi-k3` recorded 1,048,576 while every live server checkpoint reported 200,000), and the measurement gating the cap double-counted Cursor's display copies, so even a 410,000-character history (~102,500 tokens) that fit the model input lost its sentinel against the 800,000-byte budget.

One follow-up defect remained under Cursor's in-call wire-id rotation: checkpoint limits are recorded against the rotated wire id, while admission can only name the base conversation id (`options.conversationId ?? options.sessionId`), so after a rotation the recorded limit became unreachable and admission silently fell back to the fixed 50,000-byte cap. A real-provider rotation probe reproduced it: the limit was recorded as 200,000 tokens, yet installed admission still used 50,000 bytes, reduced 3 messages to 1, and deleted the sentinel.

A second follow-up defect survived review in the recorder and its call site: `recordCursorConversationContextLimit` treated every non-positive value as "no information" and returned early, and `onConversationCheckpoint` passed `checkpoint.tokenDetails?.maxTokens ?? 0`, so a checkpoint with no token details and a checkpoint that explicitly reported `maxTokens: 0` were indistinguishable. A conversation that had reported 200,000 therefore kept that limit when the server later reported an explicit zero, and admission stayed at 800,000 bytes instead of falling back to exactly the 50,000-byte bootstrap cap (real-provider probe: actual 200000, expected undefined).

## Measured evidence

- **551-message transcript collapsed to 4 messages** under the fixed 50,000-byte aggregate cap.
- **Zero-tool-call repro lost a 30 KB first turn** -- nothing in the conversation was a `toolResult` for the cap to bound.
- **`kimi-k3` recorded `contextWindow` 1,048,576 vs server-reported `maxTokens` 200,000** on every live checkpoint (`claude-opus-4-7`: 1,000,000 recorded vs 300,000 reported).
- **The 410,000-character counter-case:** ordinary English plus a sentinel measured 102,507 tokens under the repo estimator and 820,247 serialized bytes across the root + display copies against the new 800,000-byte cap; admission returned 1 of 3 messages and deleted the sentinel even though the model input fit. It is committed as a regression test asserting the model-input measurement stays under 800,000 bytes while the full wire exceeds it.

## What changed

- `packages/ai/src/cursor/conversation-context-limit.ts` (new): records the server-reported `ConversationTokenDetails.maxTokens` from the existing `onConversationCheckpoint` callback, keyed by the session/conversation pair, and forgets it on session teardown. A checkpoint for a different conversation replaces the entry, and a zero-valued bootstrap checkpoint can never inherit or retain another conversation's value.
- `packages/ai/src/api/cursor-agent/measure.ts`: new `measureCursorModelInputSerializedBytes` counts only the `rootPromptMessagesJson` blobs Cursor builds the model prompt from, excluding the `turns[]` display copies. `measureCursorHistorySerializedBytes` is unchanged and still bounds what goes on the wire (#1043).
- `packages/coding-agent/src/core/agent-session.ts`: `resolveCursorAdmissionMaxBytes()` converts the conversation-scoped reported limit to serialized bytes with the repo-wide 4-chars-per-token estimate, floored at `CURSOR_TOOL_RESULT_MAX_BYTES`. `truncateToolResultBodies` gates whole-turn dropping on the model-input measurement once the budget exceeds the legacy cap; the fixed cap keeps its wire-level gate, and the per-result character cap / tool-body blanking passes are unchanged.
- `packages/ai/src/cursor/conversation-context-limit.ts` + `packages/ai/src/api/cursor-agent.ts`: every attempt publishes its base-to-live-wire mapping through `setCursorActiveConversationWire`, and `getCursorConversationContextLimit` resolves the base identity to the published wire and returns only that wire's own record. A rotation therefore keeps the live wire's limit reachable, while a rotated or brand-new wire contributes no budget until its own positive checkpoint reports; the mapping is cleared on session teardown. Recording stays keyed by the wire that reported the limit.
- `packages/ai/src/cursor/conversation-context-limit.ts` + `packages/ai/src/api/cursor-agent.ts`: an explicit non-positive report now deletes the conversation's recorded limit (restoring the exact 50,000-byte bootstrap fallback), while a checkpoint that carried no `tokenDetails` reaches the recorder as `undefined` and preserves the existing record (`onConversationCheckpoint` passes `checkpoint.tokenDetails?.maxTokens` through instead of `?? 0`). The misleading unit test that passed a literal `0` while claiming the checkpoint "omits" the limit was corrected to pass `undefined` and keep its preservation assertion; new explicit-positive-then-explicit-zero coverage asserts the record drops for both direct and published-wire lookups.

## Guarantee

The first checkpoint of every conversation reports `maxTokens` 0, so **the first request of every conversation admits against exactly 50,000 bytes**. A replacement conversation opened under a session whose previous conversation reported 200,000 bootstraps at 50,000 until its own checkpoint reports.

## Verification

Combined verification on this branch (base `origin/main` 2f3c9150f):

- `npx tsc --noEmit` -> exit 0.
- `bun run --cwd packages/ai test` -> 284 files passed / 26 skipped (310), 2757 tests passed / 884 skipped (3641); exit 0. New tests: 1603 regression 11/11; conversation limit + rotation stream + model-input measurement 26/26.
- `bun run --cwd packages/coding-agent test` after `npm run build` at this head -> 76 failing tests / 10547 passed against 75 failures in the pristine `origin/main` baseline capture. Every current-only failure is a known environment class: four `ENOTEMPTY` temp-directory teardowns and one millisecond timestamp race in `app-server-thread-handlers-archive.test.ts`; the documented QA-port-exhaustion and dead-inherited-supervisor-PID failures are also present in the baseline -- zero failures touch Cursor admission. Refreshing the build also clears the two stale-dist `workspace-build-prerequisite` failures that a pre-commit dist otherwise reports (8/8 now pass).
- Rotation follow-up: `cursor-conversation-rotation-stream.test.ts` passes a real-provider probe where an in-call rotation has the rotated wire record 200,000 tokens and `getCursorConversationContextLimit(sessionId, sessionId)` resolves it; `1603-cursor-history-budget.test.ts` 11/11 keeps all 3 messages and the sentinel at the post-rotation 800,000-byte budget, admits at exactly 50,000 between the rotation and the rotated wire's own report, and drops back to exactly 50,000 after an explicit zero both directly and through the installed transform; `1043-cursor-toolresult-truncate.test.ts` 24/24 unchanged.
- All 20 tracked changelog paths are hash-identical to `origin/main`; no `CHANGELOG.md` was touched.

---
## Summary by cubic
Sizes Cursor admission from the server-reported conversation context limit instead of a fixed 50,000-byte cap, and measures only the actual model input, so long conversations no longer lose their oldest turns (#1603).

- Records Cursor's checkpoint-reported `maxTokens` per conversation and reads it back at admission, falling back to the 50,000-byte cap until a positive report arrives and when a checkpoint reports an explicit zero (a checkpoint with no token details preserves the earlier record).
- Gates whole-turn dropping on `measureCursorModelInputSerializedBytes`, which counts only the `rootPromptMessagesJson` blobs Cursor builds its prompt from and excludes the `turns[]` display copies; the full-wire measurement still bounds what goes on the wire (#1043).
- Keeps the recorded limit reachable across Cursor's in-call wire-id rotation by publishing the live wire per base conversation id; a rotated wire contributes no budget until its own positive checkpoint.
- Exports `getCursorConversationContextLimit` and the new measurement from `packages/ai/src/index.ts` so `packages/coding-agent` can reach them; the changelog entry stays in the Unreleased section.

<sup>Written for commit 6204406f637e81719054f7050a44ec27228e670e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1614?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Thanks for this work — the measurements (kimi-k3 reporting 200K against the 1M catalog window, the turns[] double-count, the wire-rotation reachability gap) are exactly right, and `measureCursorModelInputSerializedBytes` is the correct thing to budget against.

Heads-up on direction: I am taking #1603 to a stricter target than this PR, and will supersede it with a branch that builds on those findings:

- The 50 KB bootstrap goes away entirely. Because the recorded limit here is in-memory per session, every fresh process (including the issue's own two-`-p` repro) still admits its first request against 50 KB and still deletes the 30 KB first turn; the new regression test in this PR asserts that deletion as expected behaviour. The replacement sizes the first request from the catalog window and from a limit persisted per model id (`<agentDir>/cursor-context-limits.json`), so a later process starts with the server's real ceiling.
- Whole-turn deletion (pass 3) is removed rather than re-gated. Over-budget history after tool-body blanking is admitted and routed through the existing Cursor overflow -> compaction path, and every blanking / over-budget admission is logged.
- Keying the limit by model id instead of by session/conversation also removes the need for the wire-rotation publish/resolve maps.

I will reuse the model-input measurement from this branch with attribution, link the superseding PR here, and leave this PR to you to close once it lands. If you see a hole in the model-id keying (a case where the same model id reports different `maxTokens` for different conversations), please say so on #1603.

code-yeongyu:
Superseding PR is up: #1624. It reuses the model-input measurement from this branch (with attribution) and replaces the per-conversation in-memory limit with a per-model persisted store; the 50 KB bootstrap and the turn-deleting pass are gone. Please review there if you have time.

code-yeongyu:
#1624 has merged (`6a072f599`) and #1603 is closed, so this PR is superseded. Thank you for the investigation and the model-input measurement, which is now in `packages/ai/src/api/cursor-agent/measure.ts` with attribution. Feel free to close this one.

PR #1630, opened 2026-09-12
feat(coding-agent): support Notification hook event and fire it for ask-user settlements

Indosaram:
## Summary

Supports the `Notification` hook event in builtin hooks v1 and fires it for ask-user question settlements, so question timeouts and answers can trigger user-defined hook commands (e.g. desktop or mobile push on `ask-user-timeout`).

- `Notification` moves from unsupported-known to supported hook events with a dedicated input wire (`message`, `kind`, `title`, `notification_source`, `request_id`, `status`).
- Matcher ignores the matcher string (fires for all Notification handlers, like Stop); output accepts `additionalContext` and rejects decisions with an `unsupported_field` diagnostic.
- ask-user fires a best-effort Notification on every non-cancelled settlement without delaying the tool result or the follow-up user message: `kind` is `ask-user-timeout` on timeout, `ask-user-settled` otherwise.

## Test plan

- New `packages/coding-agent/test/suite/hooks-notification.test.ts`: schema, matcher, output-parser, and end-to-end ask-user timeout delivery (4 tests).
- Existing suites green: hooks-schema, hooks-matcher, hooks-output-parser, hooks-dispatcher, hooks-lifecycle, ask-user-extension, ask-user-pending (54 tests total across 8 files).
- `tsc --noEmit` clean for `packages/coding-agent/src` (remaining errors are pre-existing worktree dependency gaps in unrelated packages).

Created by muse spark 1.3 contributor-free

---
## Summary by cubic
Adds a `Notification` hook event to builtin hooks v1 and fires it for every non-cancelled ask-user settlement, so timeouts and answers can trigger user-defined hook commands like desktop or mobile push notifications.

- The Notification input wire carries `message`, `kind`, and optional `title`, `notification_source`, `request_id`, and `status`.
- Matchers are ignored; it fires for every handler, and output accepts `additionalContext` while rejecting decisions with an `unsupported_field` diagnostic.
- Live and resumed settlements emit `kind: "ask-user-timeout"` on timeout and `kind: "ask-user-settled"` otherwise, best-effort and without delaying the tool result or the follow-up user message.
- Config and trust are read asynchronously; disabling or excluding hooks prevents execution, and no handlers means no trust I/O.

<sup>Written for commit 4bf65774d107b6cba2c37bd869dd2b4632d2462c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1630?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Added
 
-- Added the `Notification` hook event to builtin hooks v1 and fire it for ask-user question settlements, so question timeouts and answers can trigger user-defined hook commands (for example desktop or mobile push on `ask-user-timeout`).
+- Added the `Notification` hook event to builtin hooks v1 for live and resumed ask-user question settlements, so timeouts and answers can trigger trusted user-defined commands while hooks is enabled (for example desktop or mobile push on `ask-user-timeout`). Notification matchers are ignored; cancellation does not notify, and asynchronous configuration/trust reads and command execution do not delay settlement.
 
 ### Changed
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Added
 
+- Added the `Notification` hook event to builtin hooks v1 and fire it for ask-user question settlements, so question timeouts and answers can trigger user-defined hook commands (for example desktop or mobile push on `ask-user-timeout`).
+
 ### Changed
 
 ### Fixed
--- a/packages/coding-agent/src/changes.md
+++ b/packages/coding-agent/src/changes.md
@@ -1,5 +1,26 @@
 # changes
 
+## 2026-09-12 - Support the Notification hook event and fire it for ask-user settlements
+
+### What changed
+
+- `packages/coding-agent/src/core/extensions/builtin/hooks/types.ts`: `Notification` moves from `UNSUPPORTED_KNOWN_HOOK_EVENTS` to `SUPPORTED_HOOK_EVENTS`, and `HookInputWire` gains a `Notification` variant carrying `message`, `kind`, optional `title`, `notification_source`, `request_id`, and `status`.
+- `packages/coding-agent/src/core/extensions/builtin/hooks/matcher.ts`, `dispatcher.ts`, `output-parser.ts`, `lifecycle-adapter.ts`: `Notification` dispatches like the other non-blocking lifecycle events (matcher ignored, block-only aggregation, `additionalContext` accepted, decisions rejected with an `unsupported_field` diagnostic). New `buildNotificationHookInput` / `dispatchNotificationHookEvent` / `notificationResultDetails` helpers mirror the SessionStart path.
+- `packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts` (new) + `tool.ts`: every non-cancelled question settlement fires a best-effort `Notification` hook (`kind` `ask-user-timeout` on timeout, `ask-user-settled` otherwise) without delaying the tool result or the follow-up user message. Hook `additionalContext` is recorded through the standard lifecycle result path.
+- `packages/coding-agent/test/suite/hooks-notification.test.ts` (new): schema, matcher, output-parser, and end-to-end ask-user timeout coverage.
+
+### Why
+
+- Question timeouts previously arrived only as framed user messages, so there was no hook surface for notifying on them (for example desktop or mobile push on `ask-user-timeout`).
+
+### Why an extension could not handle it
+
+- Hook event registration and the ask-user settlement path are both owned by in-tree builtins; an outside extension cannot add a supported hook event or observe the timeout delivery.
+
+### Expected merge conflict zones
+
+- `SUPPORTED_HOOK_EVENTS` / `UNSUPPORTED_KNOWN_HOOK_EVENTS` in `hooks/types.ts` and any upstream change that adds a `Notification` event with different semantics.
+
 ## 2026-09-12 - Remove the client transcript/remote-session island, keep the dist `./client` export
 
 ### What changed
--- a/packages/coding-agent/src/changes.md
+++ b/packages/coding-agent/src/changes.md
@@ -4,10 +4,10 @@
 
 ### What changed
 
-- `packages/coding-agent/src/core/extensions/builtin/hooks/types.ts`: `Notification` moves from `UNSUPPORTED_KNOWN_HOOK_EVENTS` to `SUPPORTED_HOOK_EVENTS`, and `HookInputWire` gains a `Notification` variant carrying `message`, `kind`, optional `title`, `notification_source`, `request_id`, and `status`.
+- `packages/coding-agent/src/core/extensions/builtin/hooks/types.ts`: `Notification` moves from `UNSUPPORTED_KNOWN_HOOK_EVENTS` to `SUPPORTED_HOOK_EVENTS`, and `HookInputWire` gains a `Notification` variant carrying `message`, `kind`, optional `title`, `notification_source`, `request_id`, `status`, and `transcript_path`.
 - `packages/coding-agent/src/core/extensions/builtin/hooks/matcher.ts`, `dispatcher.ts`, `output-parser.ts`, `lifecycle-adapter.ts`: `Notification` dispatches like the other non-blocking lifecycle events (matcher ignored, block-only aggregation, `additionalContext` accepted, decisions rejected with an `unsupported_field` diagnostic). New `buildNotificationHookInput` / `dispatchNotificationHookEvent` / `notificationResultDetails` helpers mirror the SessionStart path.
-- `packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts` (new) + `tool.ts`: every non-cancelled question settlement fires a best-effort `Notification` hook (`kind` `ask-user-timeout` on timeout, `ask-user-settled` otherwise) without delaying the tool result or the follow-up user message. Hook `additionalContext` is recorded through the standard lifecycle result path.
-- `packages/coding-agent/test/suite/hooks-notification.test.ts` (new): schema, matcher, output-parser, and end-to-end ask-user timeout coverage.
+- `packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts`, `tool.ts`, and `resume.ts`: live and resumed non-cancelled settlements publish `ask-user:settled`. The active hooks builtin owns Notification execution (`kind` `ask-user-timeout` on timeout, `ask-user-settled` otherwise); disabling or excluding hooks prevents execution. Configuration and atomic trust snapshots are read asynchronously, and no Notification handlers means no trust I/O. Only validated hook `additionalContext` is recorded.
+- `packages/coding-agent/test/suite/hooks-notification*.test.ts`: schema/output parsing, real trusted command dispatch with ignored matchers, disabled/excluded builtin activation through the resource loader, registered-tool blocking/async answers and authoritative fake-clock timeouts, late UI responses, cancellation, concurrent IDs, resume/reload exactly-once delivery, gated preparation/command completion, rejected-output recording, and typed persistent/in-memory payload fields.
 
 ### Why
 
@@ -15,7 +15,7 @@
 
 ### Why an extension could not handle it
 
-- Hook event registration and the ask-user settlement path are both owned by in-tree builtins; an outside extension cannot add a supported hook event or observe the timeout delivery.
+- The supported hook wire event and authoritative ask-user settlement publication are owned by in-tree builtins. Extensions can observe the settlement event, but adding the Notification wire contract requires core changes.
 
 ### Expected merge conflict zones
 
--- a/packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts
@@ -13,7 +13,7 @@ import { formatResultText } from "./format.ts";
 import type { QuestionRequest, QuestionResponse } from "./schema.ts";
 
 export async function emitAskUserNotification(
-	pi: ExtensionAPI,
+	pi: Pick<ExtensionAPI, "sendMessage">,
 	ctx: ExtensionContext,
 	request: QuestionRequest,
 	response: QuestionResponse,
--- a/packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts
+++ b/packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts
@@ -1,0 +1,89 @@
+import { existsSync, readFileSync } from "node:fs";
+import type { ExtensionAPI, ExtensionContext } from "../../types.ts";
+import { loadHookConfigSources } from "../hooks/config-loader.ts";
+import {
+	buildNotificationHookInput,
+	dispatchNotificationHookEvent,
+	notificationResultDetails,
+	recordLifecycleHookResult,
+} from "../hooks/lifecycle-adapter.ts";
+import { emptyHookTrustState } from "../hooks/trust.ts";
+import { FileHookStateStorage } from "../hooks/trust-storage.ts";
+import { formatResultText } from "./format.ts";
+import type { QuestionRequest, QuestionResponse } from "./schema.ts";
+
+export async function emitAskUserNotification(
+	pi: ExtensionAPI,
+	ctx: ExtensionContext,
+	request: QuestionRequest,
+	response: QuestionResponse,
+	variant: "codex" | "claude",
+): Promise<void> {
+	let handlers: Awaited<ReturnType<typeof loadHookConfigSources>>["executableHandlers"] = [];
+	let trust = emptyHookTrustState();
+	try {
+		const sources = ctx.getLoadedHookSources?.() ?? {
+			agentDir: ctx.cwd,
+			cwd: ctx.cwd,
+			globalHookSourcePaths: [],
+			globalHooksPath: `${ctx.cwd}/hooks.json`,
+			preSessionHookSourcePaths: [],
+			projectHookSourcePaths: [],
+			projectHooksPath: `${ctx.cwd}/.senpi/hooks.json`,
+			runtimeHookSourcePaths: [],
+		};
+		const parsed = loadHookConfigSources({
+			agentDir: sources.agentDir,
+			cwd: sources.cwd,
+			fileSystem: {
+				readTextFile(path) {
+					return existsSync(path) ? readFileSync(path, "utf-8") : undefined;
+				},
+			},
+			globalHookSourcePaths: sources.globalHookSourcePaths,
+			globalHooksPath: sources.globalHooksPath,
+			globalSettingsHooks: sources.globalSettingsHooks,
+			preSessionHookSourcePaths: sources.preSessionHookSourcePaths,
+			projectHookSourcePaths: sources.projectHookSourcePaths,
+			projectHooksPath: sources.projectHooksPath,
+			projectSettingsHooks: sources.projectSettingsHooks,
+			runtimeHookSourcePaths: sources.runtimeHookSourcePaths,
+		});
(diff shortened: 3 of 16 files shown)


Indosaram:
Created by muse spark 1.3 contributor-free

code-yeongyu (dismissed):
## Verdict: request changes

Reviewed commit `09f1549216801d90cbdd7c74daf7b418891bd37b`. **7 findings: 0 blockers, 3 majors, 4 minors.** The green test plan does not establish that this feature works through its advertised runtime path.

## Major

### R1. The actual dispatch path drops the advertised `ask-user-timeout` matcher

**Location:** `packages/coding-agent/src/core/extensions/builtin/hooks/lifecycle-adapter.ts:106-110` (selection at `:234`; compare `hooks/matcher.ts:58-62`).

`dispatchNotificationHookEvent()` calls the lifecycle dispatcher with `matcherInputs: ["Notification"]`. That dispatcher applies `matchesLifecycleMatcher()` **before** calling the shared dispatcher. Consequently, a trusted Notification handler with `matcher: "ask-user-timeout"` is filtered out. The new `matcher.ts` branch that ignores Notification matchers is never reached for that handler. This breaks the concrete timeout configuration shown in the new schema test, as well as the PR's explicit "matcher ignored" contract.

**Reproduced:** the direct matcher selects one trusted handler; `dispatchNotificationHookEvent()` with exactly that handler returns `undefined`. Removing the matcher executes the real child command and returns its context. This is not a trust failure.

**Required change:** make Notification selection use the shared Notification matcher semantics without the lifecycle prefilter, while preserving trust/disabled-handler checks. Add a real dispatch regression with a nonmatching matcher string, not just a `matchingHookHandlers()` unit test. Under the stated contract, even an invalid regex string must not become a filter or matcher diagnostic for Notification.

### R2. Disabling the hooks builtin does not disable these hook commands

**Location:** `packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts:35-66` (called from `ask-user/tool.ts:54,211`).

The ask-user builtin independently reloads hook configuration/trust and launches hook commands. It does not depend on the hooks builtin being loaded. This bypasses `enabledBuiltinExtensions` / `disabledBuiltinExtensions`, whose effective activation gate is in `packages/coding-agent/src/core/resource-loader.ts:1159-1174`. A previously trusted shell command still executes after the user explicitly disables hooks. Per-command trust does not substitute for the builtin's off switch.

**Reproduced through the real loader and registered tool:** settings contained `enabledBuiltinExtensions: ["ask-user"]` and `disabledBuiltinExtensions: ["hooks"]`; `DefaultResourceLoader` loaded `<builtin:ask-user>` and did not load `<builtin:hooks>`. Executing the registered `ask_user_question` tool nevertheless ran a trusted global Notification command, wrote its JSON stdin to disk, and appended its `additionalContext` to the session.

**Required change:** route settlement notifications to a listener owned by the active hooks builtin instead of executing hook infrastructure directly from ask-user. With hooks disabled or excluded by the builtin allowlist, no Notification command may run. Retain the existing per-handler trust gate and add a real resource-loader/tool regression for both activation settings.

### R3. Resumed/reloaded questions never emit settlement notifications

**Location:** `packages/coding-agent/src/core/extensions/builtin/ask-user/tool.ts:54,209-211`; missed settlement path: `packages/coding-agent/src/core/extensions/builtin/ask-user/resume.ts:98-100`.

The new calls cover only questions started by `createAskUserTool()`. On resume/reload, `resumeDanglingQuestion()` reopens an unresolved question directly through `ctx.ui.question()` and delivers its response through its own `deliver()` function. It never traverses either new call site. A question that times out or is answered after being reopened therefore has no Notification, despite the stated coverage of every non-cancelled question settlement.

**Reproduced:** seeded a dangling `ask_user_question` assistant tool call, resumed it with a question bridge returning `status: "timed_out"`, and awaited the resume function. The framed user-message sink ran once; neither hook-source access nor notification delivery was invoked.

**Required change:** connect resumed question responses to the same settlement-notification path as live questions. Preserve the original request ID, exactly-once delivery, and the cancelled-settlement exclusion. Cover resumed timeout and answer responses; do not fix this by narrowing the changelog promise to newly created questions.

## Minor

### R4. Notification result handling restores context rejected by the output parser

**Location:** `packages/coding-agent/src/core/extensions/builtin/hooks/lifecycle-adapter.ts:116-117` (reparsing at `:304-307,389-392`); consumer: `ask-user/notify.ts:83-84`.

The new result helper delegates to the lifecycle helper that reparses `summary.run.stdout`, rather than using the validated `summary.output`. For example, stdout `{"hookSpecificOutput":[],"additionalContext":"rejected-context"}` is rejected by `parseHookOutput()` with `invalid_event_config`, and its parsed output has no context. `notificationResultDetails()` nevertheless returns `contexts: ["rejected-context"]`, which the emitter records as model-visible context. The Notification parser and its runtime consumer therefore disagree about which output is accepted.

**Reproduced with a real command:** the dispatch summary had no `output.additionalContext` and did contain the invalid-event diagnostic; the Notification result details restored the rejected string.

**Required change:** derive Notification context from the validated dispatch summaries, not the raw-stdout lifecycle reparser. Keep the diagnostics, but do not apply rejected event-specific output. Add a dispatch-to-recording regression for malformed and mismatched `hookSpecificOutput`.

### R5. Fire-and-forget still synchronously blocks settlement on config/trust I/O

**Location:** `packages/coding-agent/src/core/extensions/builtin/ask-user/notify.ts:35-55`; blocking tool call site: `ask-user/tool.ts:209-213`.

Calling an `async` function with `void` does not defer its synchronous prefix. All source reads and both trust reads execute before this function reaches its first `await`, and the blocking tool calls it before returning its result. `FileHookStateStorage.read()` can enter `acquireHookStateLockSync()`, whose contention retry busy-waits (`hooks/trust-storage.ts:120-139`). This occurs even when the parsed configuration contains no Notification handlers.

**Reproduced:** with no hook handlers, missing trust snapshots, and contended global/project trust lock directories, the call to `emitAskUserNotification()` blocked synchronously for **359.8 ms before returning its promise**. The `void` at the tool call site does not remove that delay, and the event loop is blocked during it.

**Required change:** keep blocking config/trust work off the settlement path: process the notification with non-blocking config/trust reads, and return before trust I/O when there are no Notification handlers. Preserve trust verification; do not solve latency by executing without it. Add a regression that proves tool/user-message settlement is independent of notification preparation and command completion.

### R6. The claimed end-to-end timeout test never executes the ask-user tool

**Location:** `packages/coding-agent/test/suite/hooks-notification.test.ts:115-135`; inaccurate coverage claim: `packages/coding-agent/src/changes.md:10`.

The test creates a question UI mock but never calls it. It calls `emitAskUserNotification()` directly with a handcrafted timeout response. It therefore does not test either new `tool.ts` call site, the pending-question timer, blocking versus asynchronous settlement, cancellation suppression, or exactly-once delivery. Its hook configuration also omits a matcher, so the actual matcher-routing defect is hidden. Assertions cover only five wire fields and never inspect the collected hook result messages.

**Required change:** exercise the registered tool and its real settlement machinery. Cover blocking and asynchronous timeout/answer delivery, cancelled settlement, concurrent request-ID isolation, and the activation/resume regressions above. Assert machine-consumed fields including `notification_source`, request/session IDs, status/kind, title, and optional transcript path, plus exactly-once hook execution/result recording. Use fake time for the question timer and pre-subscribed, bounded completion signals for child commands; no sleep/poll-based tests. Correct the tracker coverage description to match the tests actually shipped.

### R7. The declared Notification wire omits the transcript field the builder emits

**Location:** `packages/coding-agent/src/core/extensions/builtin/hooks/types.ts:198-207`; emitted field: `hooks/lifecycle-adapter.ts:95`.

`buildNotificationHookInput()` emits `transcript_path` when a session file exists, but the Notification member of `HookInputWire` does not declare it. The conditional object spread happens to evade excess-property checking, so the project typecheck is green while a correctly narrowed typed consumer cannot read the actual wire field.

**Reproduced compiler error:** accessing `input.transcript_path` where `input` is `Extract<HookInputWire, { event: "Notification" }>` fails with `TS2339: Property 'transcript_path' does not exist`.

**Required change:** declare `readonly transcript_path?: string` on the Notification wire and cover both persistent-session and in-memory-session payloads without casts that conceal the contract mismatch.

## Verification and counterchecks

- Read the full PR diff and surrounding ask-user, loader, hook matching/dispatch, trust, command-runner, output-parser, lifecycle, extension-runtime, and session-message paths.
- Ran the PR's eight stated suites in one invocation: **54/54 tests passed**. Also ran five temporary adversarial probes: **5/5 passed while asserting the defective observations above**, including real child processes and the real builtin activation loader. These probes were review artifacts, not changes to the PR.
- Root `tsgo --noEmit` passed; focused Biome checks and language-server diagnostics for all eight changed TypeScript files passed. The separate typed-consumer probe intentionally failed with TS2339 as reported in R7.
- A package build was attempted but was blocked by missing workspace dependency declarations (`TS2307` for packages including `@earendil-works/pi-ai` and `@earendil-works/pi-tui`). A complete production build/CLI smoke is **not verified**; that environment limitation is not counted as a PR defect.
- Strongest counterchecks: a no-matcher trusted handler really executes, while an untrusted handler does not. The lifecycle prefilter checks original trust before removing a matcher, so matcher stripping alone is **not** an untrusted-command execution bypass. Question fields are serialized to JSON stdin by `command-runner.ts:107`, not interpolated into the configured shell command. The command-execution finding is specifically the ignored builtin off switch, not a claim of arbitrary command injection.
- The review uses the PR's explicitly stated matcher-ignore contract. The Unreleased changelog placement is correct; the inaccurate behavior/coverage claims are addressed by R1, R3, R5, and R6 rather than separate cosmetic findings.

**Do not merge until the three major defects and the four concrete contract/test defects above are addressed.**

code-yeongyu:
Addressed all seven findings from review https://github.com/code-yeongyu/senpi/pull/1630#pullrequestreview-5186801982 in **4bf65774d107b6cba2c37bd869dd2b4632d2462c**, pushed directly to this PR's contributor branch. Ready for final review; not merged.

All regression test names below are in `packages/coding-agent/test/suite/hooks-notification-review.test.ts` unless noted.

| Finding | Fix commit | Change and proving test |
| --- | --- | --- |
| R1 | `4bf65774d` | Notification bypasses the lifecycle matcher prefilter while retaining original trust and enabled-state checks. `R1 dispatch ignores matcher %s but preserves trust and disabled checks` executes real commands with timeout, unrelated, and invalid-regex matchers. |
| R2 | `4bf65774d` | Ask-user only publishes settlements; a listener registered by the active hooks builtin owns execution. `R2 %s hooks do not prepare or execute notifications from registered tools` covers disabled hooks and allowlist exclusion using the real resource loader. |
| R3 | `4bf65774d` | Resume/reload publishes through the same notification path, retaining the original request ID and excluding cancelled responses. `R3 resumed $status retains original ID and excludes cancellation` covers answers, timeouts, cancellation, and repeated reload. |
| R4 | `4bf65774d` | Notification recording consumes validated `summary.output`, not reparsed stdout, and retains diagnostics. `R4 records diagnostics without rejected context for %j` dispatches real commands with malformed and mismatched event-specific output and checks recording. |
| R5 | `4bf65774d` | Configuration and atomic trust snapshots are read asynchronously; no Notification handlers means no trust reads. `R5 settlement waitForAnswer=%s completes while trust preparation is suspended`, `R5 settlement waitForAnswer=%s does not await command completion`, and `R5 no Notification handlers skips trust reads even with contended lock directories` prove settlement independence without bypassing trust. |
| R6 | `4bf65774d` | Replaced the fabricated-emitter timeout test with registered-tool execution. Added blocking/async answers, authoritative fake-clock timeouts, late UI responses, cancellation, concurrent request isolation, activation/resume, wire-field and exactly-once recording assertions. Tests are the `R6 registered tool answer...`, `R6 authoritative timeout...`, `R6 cancellation...`, and `R6 concurrent async requests...` cases, plus `fires Notification with the timed-out question payload through the ask-user tool` in `hooks-notification.test.ts`. Updated the change tracker and changelog to match shipped behavior and coverage. |
| R7 | `4bf65774d` | Declared `readonly transcript_path?: string`. `R7 typed payload declares optional transcript_path %s` covers persistent and in-memory payloads through a narrowed wire consumer, without a cast hiding the contract. |

Verification:
- RED regression run: 14 failures before fixes.
- One expanded related-suite invocation: **243 passed, 1 existing Windows-only skip, 34 suites passed**, using `vitest run test/suite/hooks-*.test.ts test/suite/ask-user-*.test.ts --maxWorkers=2`.
- Root `tsgo --noEmit`, Biome on all changed TypeScript files, language-server diagnostics on all changed TypeScript files, complete root `npm run build`, and built CLI `--version`: passed. The local build's initial missing-workspace declaration problem was resolved with worktree-local workspace links; no dependency or lockfile changes were needed.
- **CI is green:** https://github.com/code-yeongyu/senpi/actions/runs/34700915869. All three full coding-agent shards, static checks/build, workspace/script tests, platform jobs, changelog/security checks, and review claim gate passed.
- CI attempt 1 had an unrelated Windows RPC lifecycle failure: `does not exit while a client is attached, then exits after it detaches` failed with `connect ENOENT` on its named pipe before opening a session. The unchanged test uses a 600 ms idle-exit window. After inspecting that path, reran failed jobs only; attempt 2 passed. No test was deleted, skipped, or relaxed.
- The existing review-claim sweep released the stale claim after confirming its owner had already submitted the formal request-changes review. No review was approved or dismissed by this fix pass.

Indosaram:
Review findings addressed in 4bf65774d. Independent verification on macOS (arm64) at that commit:

- R1 — `selectLifecycleHandlers` no longer applies the lifecycle matcher prefilter to `Notification`, so a handler whose matcher names a kind (`ask-user-timeout`) or carries an invalid regex still runs and produces no matcher diagnostic.
- R2 — `ask-user` only publishes `ask-user:settled` on `pi.events`; the hooks builtin owns config loading, trust, and command execution, so an inactive hooks builtin runs nothing.
- R3 — resumed questions notify through the same path (answered, timed out, orphaned-after-restart), cancellation excluded.
- R4 — Notification contexts are taken from validated parser output (`summary.output.additionalContext`), not a second read of raw stdout.
- R5 — `loadHookConfigSourcesAsync` + `FileHookStateStorage.readAsync` keep config and trust reads off the settlement path, and dispatch returns before any trust I/O when no `Notification` handler is configured.
- R6 — coverage runs through the registered `ask_user_question` tool: concurrent request-id isolation, exactly-once settlement, resumed statuses, settlement completing while trust preparation is suspended, and lock-contended trust directories.
- R7 — the `Notification` wire member declares `transcript_path`.

Verification run: `hooks-notification`, `hooks-notification-review`, `ask-user-resume`, `hooks-builtin-extension`, `hooks-lifecycle`, `hooks-loader`, `hooks-trust-storage-aba`, `hooks-trust-storage-errors` — 8 files / 52 tests passed; `biome check --error-on-warnings` clean over `src` and `test`; `check:ts-imports` clean; `tsc --noEmit` reports no `packages/coding-agent` errors. CI and the Changelog gate are green on this head; the only red check is the review claim gate, which needs the claimer to submit their review to release the `in-review` label.

Created by muse spark 1.3 contributor-free

code-yeongyu (approved):
## Verdict: approve

Reviewed the current head `4bf65774d` independently against the hooks/ask-user subsystem (loader activation gate, extension runtime event bus + subscription invalidation, lazy `ExtensionContext` getters, trust storage, config loader, lifecycle/dispatcher/matcher/output-parser, pending-question timer, resume path). All seven findings from the prior review are fixed at the root, each with a regression that asserts machine-consumed behavior and would fail if the defect returned.

### Per-finding verification

- **R1 (matcher prefilter)** - `selectLifecycleHandlers` skips `matchesLifecycleMatcher` for `Notification` while still running `listHookTrustRecords` on the original handler before the matcher is stripped, so trust/disabled gating is preserved. Regression dispatches a real child command with matcher `ask-user-timeout`, `something-else`, and the invalid regex `[`: one summary, zero diagnostics; untrusted and `enabled:false` states return `undefined`.
- **R2 (bypassed builtin activation)** - `ask-user/notify.ts` is now publish-only on `pi.events`; the hooks builtin owns the `ask-user:settled` listener, config/trust reads, dispatch, and recording. Subscriptions are tracked by the extension runtime and torn down on invalidation. Regression goes through `DefaultResourceLoader` with `disabledBuiltinExtensions:["hooks"]` and `enabledBuiltinExtensions:["ask-user"]` against the registered tool: no `getLoadedHookSources` call, no stdin payload, no hook message.
- **R3 (resume/reload never notified)** - `resumeDanglingQuestion` funnels every response (answer, timeout, orphaned) through one `deliver` that emits with the original tool-call ID and `dangling.variant`; cancelled returns before both the user message and the emit. Regression seeds a dangling `ask_user_question`, emits `session_start` resume then reload, and asserts the UI was called once, exactly one payload with `request_id:"resumed"`, and none for cancelled.
- **R4 (raw-stdout reparse)** - `notificationResultDetails` reads validated `summary.output.additionalContext` only; diagnostics keep `result.diagnostics` + command-failure diagnostics. Regression covers `hookSpecificOutput: []` and mismatched `hookEventName:"Stop"`: no context, diagnostics present, and the recorded `senpi.hook` message never contains the rejected string.
- **R5 (synchronous trust I/O on the settlement path)** - Emitter is synchronous and trivial; the listener captures session identity synchronously, then `await loadHookConfigSourcesAsync` (fs/promises), returns before any trust read when there are no Notification handlers, and uses the new `FileHookStateStorage.readAsync` (atomic snapshot, missing/invalid parses to empty state, so it fails closed rather than weakening trust). Regressions gate `readAsync` and `runCommandHook` behind unresolved promises and prove the tool result / user-message boundary completes first, that the sync `read()` is never called, and that a contended `.lock` directory with no handlers triggers no trust read at all.
- **R6 (test never ran the tool)** - Both suites now execute the registered tool through the real harness: blocking and async answers, an authoritative fake-clock (`setTimeout`/`clearTimeout`/`Date`) 30-minute timeout that beats a late UI settlement, cancellation suppression, concurrent request-ID isolation, and exactly-once recording. Completion is observed via pre-subscribed `message_end` signals with a bounded ceiling; no sleeps or polling.
- **R7 (wire type)** - `transcript_path?: string` is declared on the Notification wire member; regression narrows `Extract<HookInputWire, { event: "Notification" }>` without casts and checks both presence and absence via `Object.hasOwn`.

### Local verification on the head SHA

- `vitest --run` over 34 hooks/ask-user suites: 243 passed, 1 skipped (Windows-only). The new review suite passed three consecutive runs.
- Root `tsgo --noEmit`: clean. `biome check --error-on-warnings` on all 14 changed TS files: clean.
- CI: all required checks green on `4bf65774d` (run 34700915869, attempt 2). Attempt 1 failed only in `RPC named pipes (Windows)` on the RPC host lifecycle step, which this PR does not touch; it passed on rerun with no code change. The stale "Review claim gate" failure in the rollup is superseded by the later successful run of the same gate.

### Non-blocking notes (no action required before merge)

1. `resume.ts` now also suppresses the framed "The user dismissed the question." user message for a cancelled resumed question (previously it was sent and triggered a turn). This aligns resume with the documented live async contract in `tool.ts#deliverAnswer`, and I agree with it, but it is not mentioned in `changes.md`/CHANGELOG. Worth a one-line note in a follow-up.
2. The dependency now points hooks -> ask-user (`formatResultText`, `ASK_USER_SETTLED_EVENT`). A future tidy-up could have ask-user publish a neutral payload (`kind`, `message`, `title`, `source`, `requestId`, `status`) so hooks does not import ask-user internals and other builtins can publish notifications through the same channel.
3. `tool.ts` keeps `void emitAskUserNotification(...)` on a now-synchronous function and a redundant `!== "cancelled"` guard before it (the emitter already checks). Cosmetic.
4. The listener includes config-parse diagnostics in the recorded hook message when at least one Notification handler exists; other lifecycle paths surface those only via `/hooks`. Harmless, just slightly inconsistent.

PR #1634, opened 2026-09-12
fix(coding-agent): clear the ask-user own-answer editor when advancing to the next question

Indosaram:
## Summary

When a multi-question ask-user dialog is answered via the own-answer ("Type your own answer...") editor, the typed text leaked into the next question:

- `commitOwnAnswer()` stored the text in `AskUserQuestionState` but never cleared `ownAnswerInput`.
- The Enter path runs commit + `state.advance()` and keeps focus in own-answer, so the next question's editor rendered with the previous question's text still in it (the user's report: the entered content "remains entered on the next question").
- Pressing Enter again committed that stale text as the next question's own answer, and `emitProgress()` had already advertised it as the next question's draft answer (the user's report: it "remains in the choices as the custom answer").

Fix: `commitOwnAnswer()` clears `ownAnswerInput` after storing the value. Revisiting a tab is unchanged — `openOwnAnswer()` still reloads the question's saved text from `AskUserQuestionState.textFor()`, so committed answers stay editable per question.

## Changes

- `packages/coding-agent/src/modes/interactive/components/ask-user-question.ts`: clear the own-answer editor after commit.
- `packages/coding-agent/test/suite/ask-user-question-component.test.ts`: regressions for the empty editor on the next question, no stale `onProgress` answer, and saved-answer reload on revisit.
- `packages/coding-agent/src/changes.md`: tracker entry.

## Verification

- Regression test fails before the fix (render contains the previous question's text on the next question) and passes after.
- Full `ask-user` test surface: 15 test files, 119 tests, all passing.
- Root `bun run check` (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, platform-lock, `tsc --noEmit`, browser smoke): exit 0.

## Risk

- `commitOwnAnswer()` also runs on the ctrl+Enter submit path; clearing the editor there is inert because the dialog is already finishing. No behavioral change for single-question dialogs or the comment editor (dialog-scoped by design).

---
## Summary by cubic
Fixes the ask-user own-answer editor so typed text no longer carries over to the next question in multi-question dialogs, and pressing Enter again no longer submits the previous answer as the next question's own answer.

- `commitOwnAnswer()` now clears the editor after storing the value.
- Revisiting a question still reloads its saved answer, so committed answers remain editable per question.
- Adds a changelog entry for the fix.

<sup>Written for commit 3f6871e221a27608cf79d9d81e3fac1b078c5961. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1634?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Fixed
 
+- Fixed the ask-user question dialog carrying a committed own-answer into the next question: after answering a question with typed text, the next question's editor no longer shows the previous answer's text and pressing Enter again no longer submits it as the next question's own answer.
+
 ### Removed
 
 ## [2026.9.12-3] - 2026-09-12
--- a/packages/coding-agent/src/changes.md
+++ b/packages/coding-agent/src/changes.md
@@ -1,5 +1,25 @@
 # changes
 
+## 2026-09-12 - Clear the ask-user own-answer editor when advancing to the next question
+
+### What changed
+
+- `packages/coding-agent/src/modes/interactive/components/ask-user-question.ts`: `commitOwnAnswer()` clears `ownAnswerInput` after committing the typed text to the active question. Previously the editor kept the committed value, so the Enter path (commit + `advance()`) left focus in own-answer on the next question with the previous question's text still in the editor, and pressing Enter again committed that stale text as the next question's own answer (it also reached `onProgress` drafts through `emitProgress`).
+- Revisiting a tab is unchanged: `openOwnAnswer()` still reloads the question's saved text from `AskUserQuestionState.textFor()`, so committed answers stay editable per question.
+- `packages/coding-agent/test/suite/ask-user-question-component.test.ts`: regression coverage for the empty editor on the next question, no stale `onProgress` answer for the next question, and the saved answer reloading when the tab is revisited.
+
+### Why
+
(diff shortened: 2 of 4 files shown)


code-yeongyu (approved):
Reviewed the diff and the two new component tests against  /  on main.\n\nThe change is correct and safe:  now resets  after the text is stored in , so the Enter path (commit + ) can no longer carry Q1's text into Q2's editor or submit it as Q2's own answer;  still reloads the saved text per question, which the revisit test pins. CI is green (3/3 coding-agent shards, static checks, changelog gate).\n\nOne note for the record, not a blocker: the underlying reason the editor is still open on the next question is that  never resets  to  when it moves to the next question, so after this fix Q2 still opens with an empty own-answer editor focused instead of the option list. I am landing that root-cause fix (plus the other own-answer/Submit-tab focus exits) in a follow-up PR on top of this one. Thanks for the fix and the regression tests!

code-yeongyu:
Follow-up landed: #1642 (merged as f5bc9ba54) resets the focus to the option list on advance and adds the editor/Submit-tab navigation exits on top of this fix. Tracking issue #1641.

Issue #1656, opened 2026-09-13
Standalone binary and dependency diet: --splitting, Node-only provider registration, webfetch DOM swap, jiti/codemode single-ship, dependency hygiene


code-yeongyu:
## Summary
Track one release-oriented dependency-diet effort for standalone binaries and the published package.

## Reproduction
Build the standalone binaries and inspect their entry graphs and published tarball. Exercise Bedrock, Cursor-agent, and Devin-agent from an embedded `dist/cli.js` consumer, and exercise webfetch and codemode in a compiled binary.

## Expected
Standalone builds avoid duplicating the shared RPC session-worker graph, provider modules resolve when first used in every isolate, webfetch uses a compile-safe lightweight DOM, jiti and codemode are shipped once, and release artifacts omit unnecessary sourcemaps and unused dependencies.

## Actual
Compiled standalone binaries bundle each entry point separately, so the RPC session worker duplicates approximately 21 MiB of the main graph. Bedrock/Cursor-agent/Devin-agent use variable-specifier lazy imports while only the senpi binary's own startup registers static overrides, so consumers embedding `dist/cli.js` can get `Cannot find module` at first provider use; Devin has no override setter. Webfetch pulls jsdom plus css-tree, approximately 8 MiB of a 21 MiB entry bundle, and css-tree needs compile-time patching inside `$bunfs`. jiti is statically imported in compiled binaries. Codemode ships twice (bundled factory plus on-disk sidecar). The published package includes 19 MiB of `dist/**/*.map`.

## Evidence
Bun documentation states that multiple entrypoints are bundled separately; `--splitting` is supported with `--compile` and was verified with the production SessionWorkerClient. The startup-only static override is in `src/bun/runtime-setup.ts`. The dependency audit found exact-pinned packages behind: `@anthropic-ai/sdk 0.125.0`, `claude-agent-sdk 0.3.270`, `bedrock 3.1131.0`, `@google/genai 2.22.0`, `@bufbuild/protobuf 2.15.0`, `typebox 1.3.30`, and `zod 4.6.4`; `glob`, `@opentelemetry/api`, and `canvas` are unused.

## Root cause
The release build treats entrypoints and runtime/provider registration as binary-local concerns, while several providers and runtime assets are discovered through variable-specifier imports. Heavy DOM and runtime tooling remains in the default graph, and packaging does not consistently enforce single-shipping or sourcemap exclusion.

## Scope + acceptance criteria
[...]

PR #1658, opened 2026-09-13
build(binary): share standalone entry graphs with splitting

code-yeongyu:
## What and why

Enable `--splitting` immediately after `--compile` in both release commands and `build:binary`, retaining `--minify`, `--keep-names`, autoload isolation, and all four entrypoints. Add the missing session-worker entry to the package command. No worker-entry define semantics, environment sentinels, or runtime code changes.

- Add parsed-argv contracts for the release and package commands.
- Characterize split and unsplit relocated production worker clients: two distinct live workers, SharedArrayBuffer round-trips, and clean native exits. Observers are armed before messages; waits have bounded deadlines and no sleeps.
- Pin the compiled-worker workflow to Bun 1.4.2 and widen graph-input path filters. Ubuntu, macOS, and Windows remain required matrix legs.
- Document splitting support and update the nearest fork trackers.

## Evidence

| Check | Result |
| --- | --- |
| RED on base `cb31c35ce` | 9 pass / 4 expected failures: three splitting flags and the package session-worker entry |
| GREEN contract + existing packaging tests | 18 pass |
| Full scripts suite | 249 pass, 0 fail |
| Split/unsplit real-client relocation | Both pass in three consecutive runs on macOS arm64, Bun 1.4.2 |
| Full static check | Pass; four existing informational lint notices |
| Isolated npm/Bun/pnpm install + build | All three pass |
| Same-script base binary | 107,470,512 bytes |
| Split binary | 83,618,064 bytes |
| Reduction | **23,852,448 bytes / 22.75 MiB**, exceeding the 14 MiB gate |
| Relocated release smoke + `--version` | Exit 0 |
| Relocated multi-session RPC | Two correctly routed local-model answers; same-path reattachment; malformed request rejected; clean exit 0 |
| Mutation: remove Windows command splitting | Exactly one contract fails; 13 pass after restoration |

Acceptance commands ran remotely; commit-hook static and package-manager validators were also delegated remotely without disabling the hooks. The secondary Linux mesh host was offline, so that separate host run is unavailable; the Linux CI matrix remains enabled.

The first manual QA attempt observed the user-message event instead of the assistant-message event. The observer was corrected, the failed receipt retained, and the real binary scenario passed; no production fix was needed. The existing script-level LSP setup lacks Bun declarations, while the repository static check passes. No prose assertions were added.

Build/test/CI/documentation-only change; release-note exemption does not waive the updated fork trackers.

## Final CI and cleanup

Compiled-worker run **34751382116 is green on Ubuntu, macOS and Windows**. The only red checks are `RPC named pipes (Windows)` and its aggregate `Check and test` in run 34751382143. The active-turn lifecycle test fails with `connect ENOENT` on its named pipe (26 pass, 1 fail).

Both exact-base main run **34748558133** and newer-main run **34751189154** pass that job; an identical base failure is not established. The failed job uses the unchanged workspace `build` command and Node source host, already with Bun 1.4.2. It never invokes this patch's standalone build commands or compiled-worker tests. This is classified as an unresolved external-path lifecycle risk, not a reproduced base defect. No blind rerun or unrelated fix was made.

A deliberate missing-start mutation also proves both worker strategies fail at their bounded observation deadline rather than hanging; the original test bytes were restored. The remote task namespace was removed successfully. The offline secondary host received no task files or processes.

Refs #1656

Plan: .omo/plans/omo-dependency-diet.md


code-yeongyu:
CI attribution: the compiled-worker matrix is green on all three platforms. Broader RPC named pipes (Windows), run 34751382143/job 103708405568, fails the active-turn lifecycle test with connect ENOENT; Check and test is the downstream aggregate failure. Exact-base main run 34748558133 and newer-main run 34751189154 both pass that job. The failed path is unchanged: ci.yml already pins Bun 1.4.2, runs npm run build (not build:binary), and the test launches src/modes/rpc/host-lifecycle.ts through Node. None of the changed standalone commands or worker-compile tests execute in that job. Classified external to this patch by execution-path contrast; the exact lifecycle race is not proven and has not been reproduced on base. No blind retries or unrelated changes.

PR #1662, opened 2026-09-13
fix(rpc): record connection edges before Windows named-pipe idle exit

code-yeongyu:
## Summary

Windows named-pipe RPC hosts no longer drop the public endpoint between `ensureHost()` readiness and the first real client, and the remaining turn-active lifecycle waiter no longer uses a nested 15s clock inside a 60s test.

## Root causes (product)

1. Authenticated connection edges (including short readiness probes) can fall entirely between idle ticks. The supervisor now records attach/detach and turn start/settle immediately so the idle window measures continuous inactivity.
2. Post-readiness lock release and state writes could outlast a transient host's idle window on a loaded Windows runner, so the named pipe was gone before `ensureHost()` returned (`connect ENOENT`). The authenticated readiness connection now stays attached through the endpoint lock's commit and close.

## Residual CI failure (test waiter, not ENOENT)

On head `3fbd2e673`, [run 34823696518](https://github.com/code-yeongyu/senpi/actions/runs/34823696518) attempts 1–2 passed. Attempt 3 ([job 103917291025](https://github.com/code-yeongyu/senpi/actions/runs/34823696518/jobs/103917291025)) failed:

`does not exit while a turn is active even with no connections; exits after the turn settles` → `Timed out waiting for RPC record` at `JsonlPeer.waitFor` (default 15s). 30 passed / 1 failed / 3 skipped.

That stack is the record waiter, not `connect ENOENT`. The test never reached the no-connection idle assertion. Same SHA passed on attempts 1, 2, and 4, so this is a nested 15s budget vs loaded-runner session/prompt/`agent_start` latency, not a third product idle-race.

Fix (test-only): arm `JsonlPeer.waitFor` before scanning buffered records; mark sibling waiter timeouts handled; include seen `{type,id,command}` on timeout; bind open/prompt/`agent_start` waits to the case's existing 60s test budget; restore the mock API key for the held-model turn. No product idle-window change, no retries, no skips.

## Head

`231546b4a` on `origin/main` `5fb5d638d` (includes the waiter fix `f04ecdc1a`).

## Proof still required on this head

- Three consecutive green `RPC named pipes (Windows)` executions on `231546b4a` (earlier-head streaks do not count).
- Combined lifecycle/ensure/transport/daemon suites on the macOS proof host (pending; not claimed here).

Refs #1656

Plan: .omo/plans/omo-dependency-diet.md


code-yeongyu:
Handoff caveat: main advanced again to 5d8c339fc (PR #1665) after the requested rebase onto 792433525 and the three successful CI executions on head 4b7ba6bd8. GitHub now reports this PR CONFLICTING / DIRTY. All recorded acceptance results apply to 4b7ba6bd8 against 792433525; they do not claim compatibility with the subsequently advanced main. No merge was performed.

PR #1669, opened 2026-09-13
fix(coding-agent): stop /btw duplicating its question and add an explicit off switch

realsigridjin:
## Summary

Two `/btw` UX fixes reported by users of the interactive TUI, kept to the smallest diff that removes both problems:

1. **The question rendered twice.** Typing `/btw <question>` painted a `/btw <question>` user bubble in the transcript *and* the side-question panel already showed `btw: <question>`. The bubble stayed for the whole side-query stream, then vanished.
2. **No way to turn the panel off.** The settled footer only said "(dismisses on next message)"; Escape was matched against the raw `\x1b` byte only (so it silently did nothing under the kitty keyboard protocol - Ghostty, kitty, WezTerm), and Escape also interrupts a streaming main turn.

## Root causes

- `interactive-mode.ts` `setupEditorSubmitHandler` painted an optimistic user echo for extension commands, but `AgentSession.prompt()` reports `promptDisposition("handled")` only after the command handler resolves, and a command never becomes a canonical user message. Any long-running command shows a phantom bubble; `/btw` awaits its whole stream so it was the visible case.
- `builtin/btw/index.ts` compared `onTerminalInput` data with `"\x1b"` instead of pi-tui's `matchesKey(data, "escape")`, which also accepts kitty CSI-u and modifyOtherKeys encodings. Bare `/btw` only printed usage.

## Changes

- `src/modes/interactive/interactive-mode.ts`: both `isExtensionCommand` branches dispatch `session.prompt(text)` with no optimistic echo (the same shape `handleFollowUp` already used; upstream pi dispatches commands there without an echo).
- `src/core/extensions/builtin/btw/index.ts`: Escape matched via `matchesKey`; bare `/btw` dismisses the active panel (or cancels an in-flight side query) instead of printing usage. Usage hint stays when nothing is active.
- `src/core/extensions/builtin/btw/panel.ts`: footers read `answering… (/btw or Esc to cancel)` and `(/btw or Esc to dismiss; clears on next message)`.
- Trackers: `src/modes/interactive/changes.md`, `builtin/btw/changes.md`; `CHANGELOG.md` [Unreleased] > Fixed.

Disabling the command entirely was already possible via `settings.json` `"disabledBuiltinExtensions": ["btw"]`; this PR does not change that.

## Tests (RED -> GREEN, faux provider, zero tokens)

- `test/suite/optimistic-pending-user-echo.test.ts`: new case - an extension command submitted through the real Enter handler records `["prompt:/btw hi"]` with no `render:` entry (RED on main: `["render:/btw hi", "prompt:/btw hi"]`).
- `test/suite/btw-side-query.test.ts`: new cases - settled panel dismisses on raw *and* kitty CSI-u (`\x1b[27u`) Escape (kitty case RED on main); bare `/btw` dismisses an active panel with no usage notify and no provider call (RED on main); bare `/btw` with nothing active still shows usage.
- `bun run --cwd packages/coding-agent test test/suite/optimistic-pending-user-echo.test.ts test/suite/btw-side-query.test.ts test/suite/regressions/488-btw-runtime-provider.test.ts test/suite/manual-continue-tui-echo.test.ts` -> 4 files / 39 tests green. `npm run check` green.

## Real-CLI QA (senpi-qa: tmux 120x34 pty, isolated sandbox, fake model server streaming 30 chunks x 200 ms, no real credentials)

Driver: `/btw what is 2+2` -> capture mid-stream -> capture settled -> bare `/btw` -> `/btw what is 3+3` -> settled -> Escape. 17/17 checks pass on this branch; the fake server saw exactly 2 requests (bare `/btw` made none); tmux session, server, and sandbox torn down; real auth file hash unchanged.

**Before (origin/main 92db3be54, same driver, mid-stream)** - note the `/btw what is 2+2` bubble above the panel:

```
 senpi v2026.9.13-2
 esc interrupt · ctrl+c/ctrl+d clear/exit · / commands · ! bash · ctrl+o more
 Press ctrl+o to show full startup help and loaded resources.
 Pi can explain its own features and look up its docs. Ask it how to use or extend Pi.
 Tip: Use shift+tab to cycle the model's thinking level.
 ↳ Want the full story on any tip? Ask about it — the give-me-tips skill has the tour.
 Settings: settings.json (JSON)
 /btw what is 2+2
────────
 btw: what is 2+2
 The
 answering… (Esc to cancel)
────────
────────
❯
────────
```

**After (this branch, mid-stream)**:

```
 senpi v2026.9.13-2
 esc interrupt · ctrl+c/ctrl+d clear/exit · / commands · ! bash · ctrl+o more
 Press ctrl+o to show full startup help and loaded resources.
 Pi can explain its own features and look up its docs. Ask it how to use or extend Pi.
 Tip: Use shift+tab to cycle the model's thinking level.
 ↳ Want the full story on any tip? Ask about it — the give-me-tips skill has the tour.
 Settings: settings.json (JSON)
────────
 btw: what is 2+2
 answering… (/btw or Esc to cancel)
────────
────────
❯
────────
```

**After (settled)** - then bare `/btw` and Escape each removed the panel:

```
 senpi v2026.9.13-2
 esc interrupt · ctrl+c/ctrl+d clear/exit · / commands · ! bash · ctrl+o more
 Press ctrl+o to show full startup help and loaded resources.
 Pi can explain its own features and look up its docs. Ask it how to use or extend Pi.
 Tip: Use shift+tab to cycle the model's thinking level.
 ↳ Want the full story on any tip? Ask about it — the give-me-tips skill has the tour.
 Settings: settings.json (JSON)
────────
 btw: what is 2+2
 The answer is four. Two plus two equals four, which is the sum you asked about in this side question.
 (/btw or Esc to dismiss; clears on next message)
────────
────────
❯
────────
```

On main the same driver timed out waiting for bare `/btw` to dismiss the panel (it printed usage instead).

---
## Summary by cubic
Fixes `/btw` rendering its question twice in the interactive TUI and gives the side panel an explicit off switch. Previously the question stayed as a user bubble for the whole side-query stream, and the panel could only clear on the next message; now it appears only in the panel and can be dismissed with bare `/btw` or Escape.

## Bug Fixes
- Extension commands no longer show a temporary user bubble in the transcript.
- Bare `/btw` dismisses the active panel or cancels an in-flight side query instead of showing usage; the usage hint remains when nothing is active.
- Bare `/btw` dismisses without interrupting the main turn; Escape keeps its existing interrupt behavior.
- Escape now works under the kitty keyboard protocol in Ghostty, kitty, and WezTerm, and the panel footers name `/btw` and Esc.
- Regression tests cover the duplicate echo and kitty CSI-u Escape.

<sup>Written for commit 3425221435fa698710205adc0e9a7125866d84d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1669?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (dismissed):
Thanks for the clean, well-scoped PR — the Enter-path fix and the Escape matcher are solid, and the new tests are deterministic. Two findings from review, one of them blocking:

**P2 (request changes) — Alt+Enter during a streaming main turn still duplicates the /btw question.**
The fix covers `setupEditorSubmitHandler` (interactive-mode.ts:4805-4812), but `handleFollowUp()`'s streaming branch (~interactive-mode.ts:6489-6494) still calls `beginUserEcho(text, images)` before `session.prompt(text)`. The comment there asserts extension commands never reach that branch, but they do: with the main turn streaming (not compacting), Alt+Enter on `/btw <q>` paints the optimistic bubble *and* the side panel shows `btw: <q>` — the exact duplication this PR removes on the Enter path. Reproduced with the real handler + session and gated provider responses. Suggest dispatching extension commands (no echo) before that streaming branch, mirroring the Enter-path guard, and extending `optimistic-pending-user-echo.test.ts` to cover Alt+Enter while streaming.

**P3 — Kitty-protocol Escape key *releases* also dismiss the panel.**
`btw/index.ts:85` matches `CSI 27;1:3u` (key-up) because terminal input listeners run ahead of the TUI's release filter. Normally the press has already dismissed the panel, but a release whose press was consumed elsewhere can cancel the underlying side query unexpectedly. Recommend ignoring `isKeyRelease(data)` in the matcher and adding a negative release-event test.

Non-blocking test notes:
- `btw-side-query.test.ts:455` only dismisses a settled panel; an event-gated in-flight cancellation case would pin the advertised behavior (the replacement-query test is a good deterministic pattern).
- `btw-side-query.test.ts:480` pins the exact usage-sentence prose; asserting warning delivery + zero provider calls would be less brittle.

Verified: no functional dependency on the removed echo (commands return as handled, panel-less commands like /websearch notify independently), main-turn Escape interruption intact, dismissal state machine sound, all 32 tests in the changed files pass.

PR #1673, opened 2026-09-14
fix(coding-agent): Enter on multi-select option now toggles selection (#8249)

code-yeongyu:
## Summary

Fixes #8249 — pressing Enter on a highlighted multi-select option silently skipped selection, then advanced past the question, causing "You have not answered all questions" on submit.

## Root cause

In `activateHighlighted()`, the guard `!confirm || !state.activeQuestion.multiSelect` prevented `activateOption()` from being called when Enter (`confirm=true`) was pressed on a multi-select question. The handler still called `advance()`, walking past the unanswered question.

Secondary: `setOwnAnswer()` with empty text unconditionally deleted existing option selections via `this.selected.delete(questionId)`, so entering and immediately leaving the own-answer row wiped a prior choice.

### `ask-user-question-keys.ts`
- `activateOption()` is now always called, regardless of `confirm` / `multiSelect`
- For multi-select + Enter: toggles the highlighted option and **stays on the question** (user uses Tab or navigates to Submit when done)
- Single-select + Enter behavior unchanged (select + advance/submit)

### `ask-user-question-state.ts`
- `setOwnAnswer()` with empty text is now a no-op (early return) — existing option selections survive an empty own-answer commit

## Testing

Headless state/key verification confirms:
- Enter on multi-select row 0 → selects Option A, stays in options focus
- Second Enter on same row → deselects (toggle), stays in options
- Space behavior unchanged (toggle without advance)
- Single-select Enter → selects + advances to next question
- Empty `setOwnAnswer()` no longer clears prior selections
- Non-empty `setOwnAnswer()` still replaces option selections with typed text

---
## Summary by cubic
Fixes #8249 — pressing Enter on a highlighted multi-select option now toggles it instead of silently skipping it and advancing past the question. The question stays open until you Tab to Submit, and the hints now say Enter toggles and Tab goes to Submit.

- Empty typed answers no longer clear existing multi-select selections.
- Single-select Enter still selects and advances.
- Adds a `coding-agent` changelog entry for the fix.

<sup>Written for commit 01863bd5ec73ed399d93127b57226b14516ca8e5. Summary will update on new commits.</sup>

[...]

PR #1674, opened 2026-09-14
fix(interactive): keep the ask-user answer when Enter confirms a question

sanguneo:
## What

Confirming an ask-user question with **Enter** could submit nothing.

On a `multiSelect` question, `activateHighlighted()` skipped `activateOption()` whenever `confirm` was true, but still ran `state.advance()`. The highlighted row was never toggled while the overlay walked on to the Submit tab, so the user saw their choice "taken" and then got `! You have not answered all questions` on submit.

A second path lost an answer the same way: `setOwnAnswer()` deleted the question's selections before checking the typed text, so committing an empty own-answer editor with Enter erased an option the user had already picked. The Up/Down/Tab paths already avoid this through `saveTypedOwnAnswer()`; the Enter and Ctrl+Enter paths did not.

## Repro (before this change)

1. Ask a single `multiSelect: true` question. The overlay opens with row 1 highlighted.
2. Press `Enter` (the hint line reads `enter next`).
3. The overlay moves to Submit and the review row shows `unanswered`.
4. Press `Enter` to submit → `! You have not answered all questions`, submit blocked.

Same shape for the second path: pick option 1 with `space`, move down to `Type your own answer...`, press `Enter` to open it and `Enter` again on the empty editor. The option selection is gone.

## How

- `ask-user-question-keys.ts`: Enter stays pure navigation only for a multi-select question that already holds an answer. On an empty one it selects the highlighted row first, so confirming can never advance with nothing recorded. Single-select behavior is unchanged, and `space` still toggles extra choices without touching the existing ones.
- `ask-user-question-state.ts`: `setOwnAnswer()` drops the stored text for an empty editor but keeps selections. Non-empty text still replaces the selection as before, and `clearAnswer()` (backspace on the options view) remains the explicit way to clear an answer.
- `changes.md` entry added per the fork contract.

## Tests

`packages/coding-agent/test/suite/ask-user-question-component.test.ts`:

- `selects the highlighted row when Enter confirms a multi-select question with nothing chosen`
- `keeps a selected option when the own-answer editor is committed empty`

Both fail on `main` and pass with this change.

`clears the own-answer editor when advancing to the next question` was adjusted: its third step used to press Enter on the options row of Q2, which now selects an option instead of opening the editor. It now opens Q2's own-answer editor explicitly, so it still proves the original intent (Q1's text never leaks into Q2's answer).

## Verification

```
bun run --cwd packages/coding-agent test test/suite/ask-user-question-component.test.ts \
  test/suite/ask-user-question-own-answer-focus.test.ts \
  test/suite/ask-user-question-submit-focus.test.ts \
  test/suite/ask-user-keyboard-model.test.ts \
  test/suite/ask-user-question-reachability.test.ts
```

56 of 57 tests pass. The one failure, `ask-user-keyboard-model.test.ts > alt+enter sends a bound reply as an ordinary follow-up`, fails identically on a clean `main` checkout of this Windows machine, so it is pre-existing and unrelated to this change.

`bun run check` passes (also verified through the pre-commit hook on the commit itself).

The changelog gate asked for it, so this PR adds an entry under `## [Unreleased] / ### Fixed` in `packages/coding-agent/CHANGELOG.md` using the external-contribution format. Maintainers should reword or drop it if they prefer to curate it themselves. The `changes.md` tracker entry is included.

Reported downstream as https://github.com/code-yeongyu/oh-my-openagent/issues/8249.

---
## Summary by cubic
Fixes two paths where confirming an ask-user question with Enter could lose the user's answer.

**Bug Fixes**
- Enter on an empty multi-select question now selects the highlighted row before advancing; previously it advanced to Submit and left the question unanswered.
- Committing an empty own-answer editor with Enter now keeps existing selections instead of erasing them.
- Single-select questions, space toggling, and backspace clearing behave as before.
- Adds tests covering both Enter paths and a changelog entry.

<sup>Written for commit 9b3025024e862c0608f842f95dc7ebbd93122c15. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1674?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Closing as a duplicate: the same root fix landed via #1673, merged as 5598c08e9bd1b32dff3911d8872297226699d3bc (Enter on a multi-select option toggles without advancing; empty own-answer commits preserve existing selections). The analysis in this PR matched the shipped fix — thank you for the contribution.

PR #1700, opened 2026-09-14
fix(codemode): bound eval-cell, tool-call, and display retention (fixes #1695)

### LINKED ISSUE #1695: bug(codemode): eval cells, tool-call queue, and display buffers are retained for the session lifetime (unbounded memory growth)
## Summary
Three unbounded-retention defects in packages/senpi-codemode make every long-lived session's JS heap grow without limit:

1. src/tool/detached-cell-manager.ts — #cells entries are never removed once a cell settles. The only #cells.delete is at line 84 (same-id reuse), and dispose() (lines 186-194) stops detached cells but never clears the map. Every eval cell's input, live result, terminal result, and onKill closure (capturing the whole CellExecution) is retained for the session lifetime.
2. src/kernels/js/context-manager.ts:216 — every tool-call kernel message is pushed to #pendingToolCalls when no waiter is registered. The only consumer is nextToolCall() (line 83), which has no caller on the normal production path (git grep: only tests and the subprocess-kernel wrapper). Full tool arguments (write/edit bodies, task prompts) accumulate forever.
3. src/tool/image.ts — #displayImages (line 95) and the JSON display list grow without any byte or count cap; compaction accounting (compaction.ts:408-411) counts message content but not these details, so compaction never relieves the heap.

## Context
Root-caused during a profiling investigation of a long-lived macOS host (M4 Pro, 64 GB) running ~28 concurrent senpi/omo sessions for several days. Symptoms: load average ~150 with only ~50% CPU utilization (threads blocked on swap I/O), 20 GB swap in use, idle sessions each burning 2-8% CPU sustained, and idle session heaps growing to multiple GB (one 4-day-idle session: 648 MB resident JSC heap + 1.4 GB swapped, 3.3 GB peak footprint). All file:line references are against main (b19491745). Raw profiling evidence held locally.

## Expected (ideal state)
Settled cells are removed from the registry (terminal snapshots remain available via the notification/artifact path, or a small bounded LRU); the tool-call fallback queue is bounded or absent on paths that cannot consume it; display buffers have byte/count caps with oldest-first eviction and are included in compaction accounting.

## Acceptance criteria
- Unit tests prove: after N settled eval cells the registry size stays bounded; the tool-call queue does not grow on the normal path; display buffers evict beyond the cap.
- A soak running repeated eval cells shows heap delta plateau instead of linear growth.

## Related
- #1461 (compaction/session memory growth — same theme, different surface)

code-yeongyu:
## Summary
Failing-first contracts + fix for #1695: unbounded retention in senpi-codemode (settled eval cells, unconsumed JS tool-call queue, per-cell display buffers).

## Plan
- [x] RED: retention-bound contracts (this commit)
- [ ] GREEN: bounded terminal snapshots, tool-call queue cap + interrupt clear, display caps
- [ ] Remote vitest evidence (mengmotaMac)

Fixes #1695

---
## Summary by cubic
Bounds three unbounded retentions in `senpi-codemode` that let long-lived sessions retain settled eval cells, pending tool calls, and display payloads for the session lifetime (fixes #1695).

- Settled detached eval cells move from the live registry into a 32-entry terminal snapshot LRU; `peek`, `stop`, and `waitForTerminal` stay answerable for recent cells, and evicted cell IDs can be reused.
- The JS kernel's pending tool-call queue is capped at 256 (drop-oldest) and cleared on interrupt/reset/close/crash; the subprocess queue now uses the same cap.
- Per-cell display buffers cap at 8 images, 24 MiB of base64 image data, and 64 JSON outputs; overflow is noted in the output sink.
- Adds `eval-retention-bounds.test.ts` covering all three bounds, red on `main` and green here.

<sup>Written for commit cf5bac54d245570ea3d3f37c6f831bbfa2c3f41d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1700?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Remote RED→GREEN evidence (Apple Silicon runner)

**RED** @ `6f05d77c` (contracts on pre-fix source):
```
FAIL test/eval-retention-bounds.test.ts — Test Files 1 failed (1), Tests 6 failed | 1 passed (7)
  ✕ caps terminal snapshots at the LRU bound
  ✕ clears snapshots when a settled cell is disposed
  ✕ drops the oldest pending tool call past the bound
  ✕ clears pending tool calls on interrupt
  ✕ caps retained display image bytes per cell
  ✕ caps retained JSON display outputs per cell
```

**GREEN** @ `b53ec9dfe` (branch tip, same runner):
```
Test Files 1 passed (1), Tests 7 passed (7)
```

Command (both runs): `bunx vitest run test/eval-retention-bounds.test.ts` after `bun install && bun run build` on a clean worktree.

PR #1703, opened 2026-09-14
fix(codemode): retire a cell's children when the cell settles

### LINKED ISSUE #1697: bug(codemode): eval cell child processes are abandoned at cell end (clear without kill, no process-group kill, Bun.$ untracked)
## Summary
packages/senpi-codemode/src/kernels/js/worker-runtime.js:

- Line 45: the cell finally block runs #children.clear() WITHOUT killing anything. Any child process still running at cell end (a dev server, a watcher, a build) loses its only owner and is reparented to launchd.
- interrupt() kills each tracked child directly but never the process group, so grandchildren survive.
- Bun.$ subprocesses are never tracked (worker-shell-capture.js), and on hard timeout worker.terminate() drops every reference — the most likely source of the orphaned vite / dev-runner / vp trees (up to 3.1 GB each, fully compressed = dead weight) found on the profiled host.

Related lifecycle gaps in the same run:
- pty session kill (packages/pty session.ts:185-203): after the first SIGTERM, kill() is a no-op — no SIGKILL escalation; registry grace expiry only marks 'stopping' then deletes the entry (registry.ts:120-134, 200-213), abandoning the process.
- coding-agent tools/bash.ts:170-181: on shell exit the pid is untracked immediately while descendants in its process group survive.

## Context
Root-caused during a profiling investigation of a long-lived macOS host (M4 Pro, 64 GB) running ~28 concurrent senpi/omo sessions for several days. Symptoms: load average ~150 with ~50% CPU utilization (threads blocked on swap I/O), 20 GB swap in use, 61 ppid=1 orphan processes (~2.1 GB, including vite dev servers up to 3.1 GB compressed, dev-runner and vp trees), and 12-14 zombies whose parents are live TUI main processes. All file:line references are against main (b19491745). Raw profiling evidence held locally.

## Expected (ideal state)
At cell end (settle, timeout, or interrupt), tracked children are terminated — process-group TERM then KILL after a grace — or explicitly handed off to a background registry by user intent (detached shell). Bun.$ children are tracked. PTY and bash tool sessions escalate SIGTERM to SIGKILL and keep group ownership until the last descendant exits.

## Acceptance criteria
- Tests: a cell spawning a sleeper (direct and via Bun.$) leaves zero live descendants after settle/timeout/interrupt; pty kill escalates to SIGKILL when TERM is ignored; no ppid=1 survivors in the test tree.
- Zombie reaping verified: no defunct children accumulate under the session process.

## Related
- #1507 (shared RPC host ps-child zombies — same zombie theme, different owner)

L4XB:
Refs #1697 — the first bullet of it, measured and fixed. Scope is deliberately narrower than the whole issue; what is left is listed at the bottom.

## The leak

`JsWorkerRuntime.run`'s finally block ran `#children.clear()` and signalled nothing. A child still running at that point had lost its only owner: the cell was over, nothing would await it again, and it was reparented to init.

Reproduced against the real kernel under Bun 1.4.2, with a cell that returns without awaiting its child:

```
cell result: {"ok":true,"valueRepr":"\"done\"","durationMs":2}
child pid: 98670   ALIVE AFTER CELL END: true
```

After the change, the same cell:

```
child pid: 2457    ALIVE AFTER CELL END: false
```

## The change

`#terminateChildren()` sends SIGTERM to every tracked child that is still running, waits up to one second, then sends SIGKILL — and only then clears the set. `run`'s finally awaits it, so "the cell settled" now means "its children are gone", which is what the package's own invariant already claims for the other two exits (*Timeout and abort cleanup retires child work before ownership is released*). Normal settle was the one exit that did not.

`interrupt()` goes through the same path. Its SIGTERM is still sent synchronously, so the interrupt-latency budget in `js-kernel-interrupt-bun.test.ts` is unaffected, but a child that ignores TERM is now escalated rather than left behind.

**`detached: true` is honoured.** The issue asks for children to be terminated "or explicitly handed off to a background registry by user intent (detached shell)". `capturedSpawn` now passes the spawn options to `onChild`, and a child spawned detached is not tracked and never signalled — that is the cell asking for a process that outlives it.

## Verification

Four new cells in `test/js-kernel-cell-end-children-bun.test.ts`, built on the existing bun-driver harness from `js-kernel-interrupt-bun.test.ts`. Four mutations, all killed:

| mutation | result |
| --- | --- |
| back to `#children.clear()` without signalling (the reported bug) | 2 cells fail |
| no SIGKILL escalation | the TERM-ignoring cell fails |
| detached children are killed too | the detached cell fails |
| grace stretched past the settle budget | the TERM-ignoring cell fails |

One mutation survived and I am not going to pad the suite to hide it: **signalling a child that has already exited**. Dropping the `exitCode !== null || signalCode !== null` guard changes nothing observable — Bun's `kill()` on an exited subprocess is a no-op — so the guard is there for clarity, not for behaviour, and no assertion can earn it.

A note on the TERM-ignoring fixture, because it caught me out first: `sh -c "trap '' TERM; sleep 30"` only ignores TERM once the shell has run the `trap`. Killed a few milliseconds after spawn it dies on TERM alone, and the escalation is never exercised. The cell sleeps 300 ms before returning so the trap is installed; without that pause the mutation "no SIGKILL escalation" survives.

`vitest run --root packages/senpi-codemode`: 398 passed / 10 failed on this branch, 394 passed / 10 failed on its base — same ten failures before and after, none in the kernels (they are collection failures in this environment). `tsc --noEmit -p packages/senpi-codemode` reports the same 1310 pre-existing errors with and without the commit, none in the touched files. `biome check` is clean and the repo's pre-commit `npm run check` passed.

## Not in this PR

The issue lists more than this one line, and the rest needs decisions I should not make on your behalf:

- **Process-group kill.** Killing the group requires spawning children detached in the first place, which changes signal delivery and terminal-control semantics for every cell. A child spawned without `detached` shares the worker's group, so `kill(-pid)` there would take the worker with it. Grandchildren therefore still survive this fix.
- **`Bun.$` tracking.** `ShellPromise` does not expose its subprocess, so tracking it needs a different mechanism than `onChild`.
- **pty SIGTERM → SIGKILL escalation** (`packages/pty`) and **`coding-agent tools/bash.ts` pid untracking** — different owners, and each deserves its own fix and its own test.

Happy to follow up on any of those if you tell me which shape you want.

---

---
## Summary by cubic
Fixes #1697's first bullet: when a cell settles normally, its still-running children were dropped from tracking and reparented to init, leaking long-lived processes. The settle path now sends SIGTERM, escalates to SIGKILL after a one-second grace, and only then clears the child set; `detached: true` children are exempt. `interrupt()` uses the same path, so TERM-ignoring children now get escalated instead of being left behind.

**Not in this PR**

- Grandchildren can still survive; process-group kill is left out.
- `Bun.$` children, the `pty` package, and `coding-agent tools/bash.ts` are separate follow-ups.

<sup>Written for commit 7d38c0235e5507ee36bd29e01d4a3261acbf7b13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1703?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Thanks @L4XB — your fix and its four mutation-killed tests are the foundation here. I've carried your commit forward (cherry-picked with authorship preserved) into #1713, which extends it to the rest of #1697: whole-child-tree termination (grandchildren, not just the tracked child), `node:child_process` coverage, host-side retirement when a blocked worker is abandoned, and the Python close/host-loss group sweeps. Superseding this PR by #1713 so the issue lands as one coherent change. Really appreciate the careful RED/GREEN work.

PR #1721, opened 2026-09-15
fix(coding-agent): zero-spawn watchdog ppid fallback (fixes #1507)

### LINKED ISSUE #1507: [Bug]: shared RPC host (--mode rpc --multi) never reaps its ps children - 9,386 zombies filled the macOS process table and every spawn on the machine failed with EAGAIN
## Incident

2026-09-09 ~01:00 KST on mengmotaHost (macOS 26.4.1, bun 1.4.2, global `omo-ai@5.0.0-0.beta.48` = senpi 2026.9.7-2): every process on the machine started failing with `EAGAIN: resource temporarily unavailable` on `posix_spawn` - shells, `gh`, `ps`, the omo eval kernels, even `/usr/bin/true`. The kernel process table held **10,981 entries, 9,750 of them zombies**.

Forensics were taken without spawning anything (bun:ffi over `sysctl KERN_PROC_ALL` / libproc):

- 9,386 of the zombies had `p_comm = ps`; the rest were 334 Chromium helpers from unrelated browser sessions.
- All `ps` zombies had one of six parents, every one of them a senpi shared RPC host from the global install:
  `/Users/yeongyu/.bun/bin/bun ~/.bun/install/global/node_modules/@code-yeongyu/senpi/dist/cli.js --extension ~/.bun/install/global/node_modules/omo-ai/plugin --mode rpc --multi`
  with 2097, 2097, 1324, 1322, 1319 and 1227 zombie children respectively (parent chains: `bun <- bun <- bun <- bun`, i.e. omo TUI sessions' senpi-task hosts).
- The desktop's socket hosts (`omo --mode rpc --multi-session --listen`, ~/omo-desktop-runtime) had **0** zombies each, so the leak is specific to the `--mode rpc --multi` host process.
- The only `ps` invocation that shape reaches inside that host is the omo plugin's process-identity probe `/bin/ps -o lstart= -p <pid>` (`execFile` from `node:child_process`; used by memory locks, reflection/facts reservations and lock recovery, so it runs continuously). A 30x `execFile("/bin/ps")` micro-repro from a plain bun 1.4.0 main thread and from a `Worker` thread reaps normally (1 transient zombie), so the unreaped state depends on something the `--multi` host does (SIGCHLD handling, worker admission introduced by #1499's predecessor, or the 1.4.2 runtime) - not yet isolated.

Recovery: `kill(pid, SIGTERM)` on the six hosts via ffi; the zombies were reparented to launchd and reaped within seconds (10,981 -> 1,578 processes), and `posix_spawn` worked again. The owning omo sessions respawn their host on demand.

## Why this matters

A single long-lived TUI session with senpi-task takes the whole machine down after a few hours: the host never `waitpid`s its `ps` children, and macOS `kern.maxproc` is shared by every user process. Nothing else on the box can fork once the table is full, which is exactly the failure mode that hit this session (monitors dying with exit 128, PTY allocation failing).

## Next

1. Reproduce on a live `--mode rpc --multi` host: count zombie children per minute (`sysctl KERN_PROC_ALL` parse, or `ps -o stat= -p`) while the omo plugin's memory workers run; identify whether the identity probe or another spawner leaks, and whether the leak needs bun 1.4.2.
2. Fix at the source (reap or never leave a subprocess handle unobserved; a `SIGCHLD`/`waitpid` sweep in the host as defence in depth).
3. Regression guard: a host-level test that spawns N short children through the same path and asserts zero zombies after they exit.

A watcher is running on the affected machine (5-minute `sysctl` scan, alerts at 50+ zombies per parent) so a recurrence is caught before the table fills again.

code-yeongyu:
## What

Fixes #1507.

A long-lived shared RPC host (`--mode rpc --multi`) spawned `ps -o lstart= -p <supervisor>` every 250ms through the watchdog's ppid fallback. On runtimes whose `execFile` does not reap, those children accumulated as zombies — 9,386 measured, after which every spawn on the machine failed with EAGAIN.

## Change

`watchPpid` drops the identity probe entirely. A dead supervisor is reaped by its own parent and this process is then reparented, so `kill(pid, 0)` plus the `process.ppid` comparison — both free syscalls — detect the loss with zero child processes.

Two failure modes die with the probe:

- the zombie accumulation itself (no more `ps` children), and
- the watchdog firing after three consecutive probe failures against a **live** supervisor (an observability gap shutting the host down).

Detection timing: supervisor death is observed at reparenting instead of during the brief zombie window; the fire reasons and the fd path are unchanged.

## Contracts

`test/suite/regressions/issue-1507-watchdog-ppid-no-spawn.test.ts` (RED on `main`, GREEN here):

- zero process-identity probes while the supervisor is alive (5s of ticks)
- still fires when the supervisor pid is gone

## Evidence

Remote RED→GREEN on an Apple Silicon runner: posted below.

---
## Summary by cubic
Fixes #1507 by removing the `ps` identity probe from the RPC host watchdog's ppid fallback, which ran `ps -o lstart=` every 250ms against the live supervisor. On runtimes that don't reap children those probes accumulated as zombies (9,386 measured) until every spawn failed with EAGAIN; the fallback now uses only the free `kill(pid, 0)` syscall plus the `process.ppid` comparison, since a dead supervisor is reaped by its own parent and the host is then reparented.

**Bug Fixes**
- Removes the false-shutdown path where three failed identity probes against a live supervisor fired the watchdog.
- Supervisor death is now detected at reparenting instead of during the zombie window; fire reasons and the fd path are unchanged.
- Adds a regression test verifying zero identity probes while the supervisor is alive and that a gone supervisor still fires.

<sup>Written for commit 694e5b2fa725d50c8a71fb72d11722d80cea5b3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1721?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Remote RED→GREEN evidence (Apple Silicon runner)

**RED** (tip contracts + `main` source): `Tests 1 failed | 1 passed (2)` — the zero-probe contract fails (the live-supervisor path probes every tick); the gone-supervisor guard passes.

**GREEN** @ branch tip (same runner): `Tests 2 passed (2)`.

Command (both runs): `bunx vitest run test/suite/regressions/issue-1507-watchdog-ppid-no-spawn.test.ts` after `bun install && bun run build` on a clean worktree.

PR #1726, opened 2026-09-15
feat(coding-agent): recoverable resident-string backing and idle materialized-view release

Indosaram:
## Summary

- Give the resident string store a recoverable on-disk backing so budget eviction no longer loses data, and release the memoized materialized session views at the `agent_idle` settle boundary so idle sessions stop pinning the full session text in memory.
- Persisted sessions back evicted strings with `<sessionDir>/resident-blobs/<sessionId>/` (`--no-session` never writes blobs). `materialize` hydrates evicted strings from the backing **transiently** — without re-entering the resident cache, so a bulk read cannot refill the budget. The session file stays authoritative: a missing or corrupt blob falls back to the existing batched JSONL recovery unchanged.
- Without a backing directory eviction is now disabled instead of lossy (previously a standalone store over budget either lost strings on eviction or kept everything resident; in-memory sessions over 64 MiB served unreadable sentinel tokens).

## Behavior and scope

- `session-resident-store.ts`: write-before-evict (temp file + rename), transient hydration, LRU refresh on read, `clear()` wipes the blob cache, `stats()` gains `evictedCount`/`evictedBytes`.
- `session-manager.ts`: lazy per-session blob dir wiring + `dropMaterializedCaches()` (`entriesCache`, `branchCache`, `compactEntriesCache`).
- `agent-session.ts`: `_emitAgentIdleAfterDeferredTurns()` calls `dropMaterializedCaches()` right before emitting `agent_idle`.
- Public `ToolResultMessage`/`AgentToolResult` shapes and every existing consumer contract unchanged; recovery misses still route through the existing `loadEntriesFromFile` batched path.

Measured context (macOS, `footprint`): an idle long session held ~1.1 GB dirty JS heap — the store budget bounded only its own map while the materialized views re-pinned every hydrated string. This PR bounds idle memory to the resident budget plus the blob metadata.

## Verification

- `bun run --cwd packages/coding-agent test test/session-resident-store.test.ts test/session-manager/materialize-cache.test.ts test/session-manager/session-mirror-budget.test.ts` — 24 passed (new store suite: budget eviction + hydration, LRU order, unwritable-backing fail-safe, `clear()` cleanup, surrogate pairs; `session-mirror-budget` updated: evicted recovery now asserted via blob hydration with `loadCount === 0`, standalone budget asserted with a backing directory).
- Root `bun run check` (biome, pinned deps, ts imports, shrinkwrap/install locks, `tsc --noEmit`, browser smoke) — clean.

## Rollback

Single commit; revert restores the previous budget-only store and view retention.

---
## Summary by cubic
Gives the resident string store a recoverable on-disk backing so budget eviction no longer drops large tool-result strings, and releases materialized session views at `agent_idle` so idle sessions stop pinning the full session text in memory.

Persisted sessions write evicted blobs under `<sessionDir>/resident-blobs/<sessionId>/`; stores without a writable blob directory keep strings resident, and missing or corrupt blobs fall back to the existing batched JSONL recovery.

- Evicted strings are written atomically (temp file + rename) and hydrated transiently without re-entering the resident cache, so a bulk read cannot refill the budget; reads refresh LRU order.
- Blob names are SHA-256 hashes of the text, so two processes sharing a session directory can only collide on identical bytes; corrupt envelopes are deleted on read so the next eviction rewrites them.
- Compaction spills resident strings to the blob backing instead of clearing it, keeping post-compaction reads on the per-string hydration path.
- `agent_idle` releases the entry, branch, and compact-entry caches and tokenizes `agent.state.messages` in place only when the store has evicted something; every out-of-turn reader rehydrates in place before returning.
- Branching materializes entries before clearing so sentinel tokens can't be written into the branched JSONL.
- Switching, branching, reopening, `newSession()`, `clear()`, and `dispose()` remove the session's blob directory; when several live managers share one session file, the last one to dispose clears it. `dispose()` also unregisters the session writer.
- `stats()` reports `evictedCount`/`evictedBytes`; `--no-session` never writes blobs; JSON transformation throws for BigInt values to match `JSON.stringify`.

<sup>Written for commit d4b1802030fe7bf296cd73de4a3ea4cc0fd333bc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1726?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

## Maintainer follow-up (2026-09-16)

This branch now carries the whole stack: #1729's commit is included, so merging this PR lands both. A review of the two PRs found six defects that are fixed here; the tracking issue is #1746.

| # | Defect at `32618a6c` | Fix |
| --- | --- | --- |
| A | `spillResident()` deleted from `strings` but not from the `idsByText` reverse index, whose keys are the spilled strings — spilling released nothing | The reverse index is gone; the id IS the content hash |
| B | The removed BigInt guard let a bigint reach `seen.add(value)` on a `WeakSet` → `TypeError: WeakSet values must be objects` | `TypeError("Do not know how to serialize a BigInt")` restored, matching `JSON.stringify` |
| C | `get messages()` returned the tokenized runtime array between `agent_idle` and the next turn, so idle-time readers saw `\0senpi-resident-string:v1:<id>` | A latched `_runtimeMessages()` hydrates in place behind the getter, `prepareNextTurnWithContext`, and every out-of-turn token estimator; `test/suite/harness.ts` is byte-identical to `main` again |
| D | Per-instance counter ids: a hydrated string re-externalized at the next idle minted a new id and a new `.blob`, and nothing removed the directory on teardown or reopen | Ids are `sha256(text)`, so the same text always maps to the same file; `_writeBlob` skips an existing blob; `SessionManager.dispose()` (called from `AgentSession.dispose()`) removes the directory and unregisters the writer; `_setSessionFile` clears a stale directory on reopen |
| E | Two instances over one session directory both started at id `0` and overwrote each other's `N.blob`, hydrating the wrong text | Content addressing makes a name collision a byte-for-byte match |
| F | The idle release ran on every settle, so settled-time `getEntries()` consumers re-materialized the whole history each turn even when nothing had been evicted | The release runs only when `evictedCount > 0` (a resident string is shared with the store, so tokenizing it frees nothing) |

A corrupt blob is also deleted on read now, so the next eviction of that content writes a readable one instead of skipping over a broken file.

Each fix landed as a RED test commit followed by its fix commit. New coverage: five cases in `test/session-resident-store.test.ts` (spill round-trip, shared-directory isolation, no churn on re-externalize, corrupt-blob repair, BigInt), three in `test/suite/resident-idle-release.test.ts` (token-free `session.messages` after idle, release skipped under budget, release taken over budget), two in `test/session-manager/session-mirror-budget.test.ts` (dispose removes the directory, reopen clears a stale one).

Real-surface check against the built `dist` `SessionManager` (70 x 1 MiB entries, persisted): blob names are 64-hex content hashes, re-reads add no files, all 70 payloads hydrate intact, a foreign stale blob is cleared on reopen while the session's own hashes are rewritten, `dispose()` and `newSession()` both remove the directory — 10/10 assertions. The same probe scored 7/10 on `32618a6c` (numeric names, stale blobs surviving reopen, no `dispose`).

The four CHANGELOG lines had merged into the released `[2026.9.16]` section during the `main` sync; they are back under `[Unreleased]` and rewritten for the final behavior.


Indosaram:
@sisyphus-dev-ai please take a look here

PR #1729, opened 2026-09-15
feat(coding-agent): spill resident strings to the blob backing across compaction

Indosaram:
## Summary

Stacked on #1726 (merge that first). Keeps compact-context recovery O(string) across the compaction boundary instead of falling back to a full session-JSONL parse.

- `session-resident-store.ts`: new `spillResident()` — writes every resident string to the blob backing and empties the map while **keeping the backing itself**, unlike `clear()` which wipes both.
- `session-manager.ts`: `_trimMirrorAfterCompaction()` spills instead of clearing, so strings referenced by the retained mirror keep hydrating from the backing after compaction.
- Session switches keep `clear()`: the backing directory is scoped to the session id, so wiping it there is correct.

## Behavior and scope

- No public API or persisted-format change; `spillResident()` is store-internal surface used by the mirror-trim path.
- Disk growth: blobs for pruned pre-compaction entries remain until the next session switch (bounded by the resident budget); the JSONL stays authoritative throughout.
- Follows #1726's hydration contract: reads never re-enter the resident cache, so bulk reads cannot refill the budget.

## Verification

- `bun run --cwd packages/coding-agent test test/session-resident-store.test.ts test/session-manager/materialize-cache.test.ts test/session-manager/session-mirror-budget.test.ts` — 26 passed (new: spill persists + empties + keeps backing for hydration, no-backing spill no-op; `session-mirror-budget` compact-context recovery updated from one batched JSONL reload (`loadCount === 1`) to pure blob hydration (`loadCount === 0`)).
- Root `bun run check` — clean.

## Rollback

Single commit on top of #1726; revert restores the previous clear-and-reparse compaction path.

---
## Summary by cubic
Adds a per-session blob backing for the resident string store so large strings evicted over the 64 MiB budget — and those spilled across compaction — hydrate from small blob files instead of re-parsing the whole session JSONL.

- Evicted strings are written to `<sessionDir>/resident-blobs/<sessionId>` and hydrate on read; with no backing directory (`--no-session`) eviction stays disabled so no string is dropped unreadably.
- Compaction spills resident strings to the backing and empties the map while keeping it; `clear()` still wipes both on session switches.
- Idle sessions release memoized materialized views and tokenize runtime message state at `agent_idle`; the next turn re-materializes both before any read.
- Blobs are validated JSON envelopes so corrupt files fall back to JSONL recovery, and branched sessions materialize before the previous backing is removed so sentinel tokens never reach the new session file.
- No public API or persisted-format change; the JSONL stays authoritative and the blob backing is a disposable cache.

<sup>Written for commit 32618a6c8ddd0e66c12f29354bb9f330487affdc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1729?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
The fixes for this stack landed on #1726's branch, and `32618a6c` (this PR's head) is an ancestor of that branch, so merging #1726 merges this commit too and GitHub will mark this PR merged.

Review of both PRs (tracked in #1746) found that `spillResident()` deleted from `strings` but not from the `idsByText` reverse index, whose keys are the spilled strings themselves — so spilling released no memory and this PR's stated goal was not reached. The fix removes the reverse index entirely by making the blob id the SHA-256 of the text: the same text always resolves to the same token and the same file, which also ends the per-idle blob churn and the cross-process `N.blob` overwrite. `spillResident()` itself keeps its contract and now genuinely empties the map.

Evidence: `test/session-resident-store.test.ts` gained "re-externalizing a spilled string makes it resident again with the same token" (RED at `32618a6c`: `expected +0 to be 1`, GREEN after), plus shared-directory isolation, no-churn, and corrupt-blob repair cases. Details in #1746 and in #1726's maintainer section.

code-yeongyu:
Merged with #1726 as `62aa59e4f4cbfca5f15e1c6015dce2306946b8e7`. This PR's head `32618a6c` is an ancestor of `main`, so the spill-across-compaction commit shipped exactly as written; the defect it carried (`spillResident()` leaving the `idsByText` keys behind, so spilling released nothing) is fixed on the same branch by content-addressing the blob ids, which removed the reverse index entirely. Details and evidence: #1746.

PR #1773, opened 2026-09-17
fix(coding-agent): retry compaction summarization without the reasoning-effort override after an empty stop

Indosaram:
## Why

Incident 2026-09-17: a `(mahoquot) z-ai/glm-5.3-flash` session surfaced, repeatedly:

```
Compaction rejected: summarization response contained no text (stopReason: stop)
```

The compaction summarizer deliberately pins `reasoningEffort` (first non-null of
low/medium/high) to keep compaction cheap. On a cline-backed OpenAI-completions
relay, that combination — explicit reasoning effort + large tool-bearing
summarization prompt — made the upstream complete "normally" with **zero text**:
the gateway (Generic account, verbatim SSE passthrough) recorded HTTP 200 with
`completion=7` tokens (role prelude only), `reasoning=0`, 162 times, while the
same model answered ordinary agent traffic (no effort pin) without issue. The
empty response burned the full overflow-retry budget (one burst shrank the
input 97k→112k across 19 attempts, all empty) before surfacing as a terminal
`empty-summary` failure.

The one request-level lever the summarizer owns is the reasoning-effort
override, so an empty stop now earns exactly one retry without it — mirroring
the existing bare-tool-call retry (`toolChoice: "none"`).

## What changed

- `speculative.ts`: an empty summary with `stopReason: "stop"` spends one retry
  of the same request with `omitReasoningOptions: true` before the terminal
  `SummaryGenerationError("empty-summary")` throw. Persistent emptiness keeps
  the existing contract (typed error → deterministic fallback classification).
- `speculative-summary.ts`: `generateSummaryMessage` accepts
  `omitReasoningOptions` and skips `summarizationReasoningOptions` when set, so
  the retried request carries no reasoning-effort override.
- `compaction/changes.md`: fork-tracker entry (what/why/conflict zones).
- New policy test: `test/compaction/summarization-empty-stop-retry.test.ts` —
  (1) empty stop → exactly one retry without the effort override, retry summary
  used; (2) persistent empty stop → one retry spent, then typed
  `empty-summary`/`stopReason: stop`; (3) non-reasoning model → retry still
  applies, no effort override ever offered.

## Validation

- `bun run --cwd packages/coding-agent test test/compaction/summarization-empty-stop-retry.test.ts test/compaction/summarization-tooluse-retry.test.ts test/compaction/speculative-compaction.test.ts` → 36/36 green (existing empty-summary and toolUse-retry contracts unaffected; faux queue exhaustion still yields `stopReason: "error"`, which this retry deliberately does not touch).
- `biome check` clean on all changed files.
- `tsc --noEmit`: no errors under `packages/coding-agent`. (Unrelated local-env note: a partial `node_modules` copy in my worktree lacks the new `web-tree-sitter` dep from #1772, producing 5 module-resolution errors under `packages/agent` — CI installs properly.)
- Pre-existing on `main`, unrelated to this PR: root `bun run check` currently fails at the biome stage on `test/steering-tool-context.test.ts` (`lint/complexity/useLiteralKeys` warnings).

## Notes

- The retry is bounded (one extra attempt, outer loop like the toolUse retry),
  so a relay that always returns empty stop costs one additional call before
  the deterministic fallback — no new unbounded loop, no breaker interaction
  change.

---
## Summary by cubic
Fixes compaction summarization failing on OpenAI-completions relays that return an empty `stopReason: "stop"` when the request pins an explicit reasoning effort. The empty answer used to exhaust the overflow-retry budget and surface as a terminal `empty-summary` error; now it earns one retry without the reasoning-effort override before throwing.

- Mirrors the existing bare-tool-call retry and is bounded to one extra attempt; the retry only runs when the first attempt pinned a reasoning override, so non-reasoning models keep the single-call terminal contract.
- Adds `omitReasoningOptions` to `generateSummaryMessage` so the retried request skips `summarizationReasoningOptions`.
- Persistent empty stops still throw `SummaryGenerationError` and use the deterministic fallback.
- Aligns existing retry-contract tests, adds faux-provider and openai-completions wire coverage, and records the fix in the changelog.

<sup>Written for commit 14a1573a0ccb168cc0024e46649dfc250c672275. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1773?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu (changes requested):
Thanks for the incident write-up and the bounded retry — the shape (one extra attempt, outer-loop `continue`, typed error preserved) is right. Requesting changes on four points; I will push the fixes to this branch myself (maintainer edit) so the PR keeps your commits and lands today.

### Blocking

**1. The retry fires even when no reasoning override was ever applied** (`speculative.ts:437`)

`summarizationReasoningOptions()` returns `{}` for `model.reasoning === false` and for every api family outside `anthropic-messages` / `openai-*` (`speculative-summary.ts:47-62`). For those models "retry without the override" is a byte-identical replay of a large summarization request, which the PR's own rationale ("the one request-level lever the summarizer owns") does not justify. The tell is in the diff: `blocking-compaction-review-hardening.test.ts` and `issue-7952-compaction-recovery.test.ts` both use `registerFauxProvider()` whose default model is non-reasoning (`faux.ts:521`), and both had to be bumped from 1 to 2 provider calls. The bare-tool-call retry this mirrors is gated on `summarizationToolsOffered`; this one needs the same gate: spend the retry only when the first attempt actually carried an override. With that gate the two existing contracts go back to one call and their edits should be reverted; test case 3 ("non-reasoning → retry still applies") inverts to "non-reasoning → one call, terminal empty-summary".

**2. The mechanism described to users is not the one the code can produce** (CHANGELOG.md, `changes.md`, PR body)

"burned the whole overflow-retry budget (97k→112k across 19 attempts)" cannot happen on this path: an empty text response throws `SummaryGenerationError("empty-summary")` at the first response (`speculative.ts:441`), `isContextOverflow()` is false for an empty 200 so the overflow branch (`:387`) never runs, and `isRetryableSummaryAttempt` rejects `SummaryGenerationError` (see the 2026-08-31 entry in `changes.md`). A growing 97k→112k input across 19 attempts is the session re-triggering compaction over time — `pre_prompt` before each prompt, the idle warm-up plus its `MAX_IDLE_WARMUP_RETRIES`, and the breaker re-opening every `COOLDOWN_MS` — each paying one full summarization request. The exact text you quoted, `Compaction rejected: summarization response contained no text`, is the `index.ts:734` cancel for a non-required route (`isRequiredCompactionFallbackReason` = manual|threshold|overflow), which is why the deterministic checkpoint did not rescue it. Please describe that chain; the CHANGELOG line is user-facing.

**3. Anthropic scope is unstated** (`speculative-summary.ts:48`)

For `anthropic-messages` reasoning models the override is `{ thinkingEnabled: false }`, so omitting it on the retry re-enables the provider's default thinking on a near-window-size prompt — the opposite of the "compaction must be fast" contract documented right above `summarizationReasoningOptions`. The code applies to every wire family while the comment and the tests say "OpenAI-completions relays". Either restrict the retry to the effort-pin families or pin the Anthropic behavior explicitly (test + comment). I am going with "gate on override present, any family" and pinning Anthropic, since the retry is bounded by `deadline.attemptBudgetMs` and a last-ditch request that differs is still better than the deterministic checkpoint.

**4. Interaction with the bare-tool-call retry is untested**

`toolUseRetrySpent` and `reasoningOverrideRetrySpent` both persist across `continue`, so a bare tool call → `toolChoice: "none"` retry → empty stop → third request must carry `toolChoice: "none"` and no effort override, and persistent emptiness after that must be terminal at three calls. Add that case.

### Notes (non-blocking)

- No wire-level proof that the retried request actually differs on the wire: the faux provider records `options`, not the serialized `reasoning_effort`. I will add a local-HTTP `openai-completions` test (real `ModelRuntime.registerProvider` + `createAgentSession` + `session.compact()`) asserting `reasoning_effort` is present on attempt 1 and absent on attempt 2, because that is the claim the incident rests on.
- Nothing logs that the override retry was spent, so the hypothesis cannot be confirmed in the field. `SpeculativeCompactionContext` has no logger, so this is a follow-up, not a change here.
- `const modelRegistry = Object.create(null) as …; if (modelRegistry) { … }` — the `if` is always true; copied from sibling tests, so leaving it.

code-yeongyu:
Pushed three commits on top of yours that close the four blocking points from my review (`dbdfba0fe6` RED contract, `8a426d95f8` fix, `14a1573a0c` wire test). What changed and how each point was verified:

**1. Retry gated on a pinned override** — `speculative-summary.ts` exports `hasSummarizationReasoningOverride(model)` (non-empty `summarizationReasoningOptions`), `speculative.ts` computes `reasoningOverrideOffered` next to `summarizationToolsOffered` and the empty-stop branch requires it. The two existing one-call contracts (`blocking-compaction-review-hardening.test.ts`, `issue-7952-compaction-recovery.test.ts`) are back to their base versions and pass unchanged.
- RED on your head `27cb902`: `summarization-empty-stop-retry.test.ts` case "non-reasoning summarizer … no retry is spent" → `AssertionError: expected undefined to be 'SummaryGenerationError'` (the loop replayed the request and consumed the trap response). 1 failed / 5 passed.
- GREEN on `8a426d95f8`: 6/6.

**2. Accurate mechanism in the user-facing text** — CHANGELOG entry and the `changes.md` section now say what the code does: the empty stop throws at the first response, never enters the overflow branch, and the repeats came from routes re-triggering compaction (`pre_prompt`, idle warm-up retries, breaker cooldown), with the non-required-route `Compaction rejected:` cancel explained. The CHANGELOG line passes the prose linter used for release text (score 0, no high hits).

**3. Anthropic pinned** — new case: `anthropic-messages` reasoning summarizer, first call carries `thinkingEnabled: false`, the retry carries neither `thinkingEnabled` nor `reasoningEffort`, retry summary used. Code comment states the consequence (provider-default thinking on the retry, clamped by the compaction deadline).

**4. Tool-call interaction pinned** — two new cases: bare tool call → `toolChoice: "none"` retry (still `reasoningEffort: low`) → empty stop → third request has `toolChoice: "none"` and no effort, summary used; and the same prefix with persistent empty stops is terminal at exactly three calls.

**Wire-level proof (new `summarization-empty-stop-retry-wire.test.ts`)** — real `ModelRuntime.registerProvider` (`openai-completions`, `reasoning: true`) + `createAgentSession` + two prompts + `session.compact()` against a local HTTP relay stub that answers a summarization request only when `reasoning_effort` is absent. Asserts request #1 body has `reasoning_effort: "low"` and `tools`, request #2 body has no `reasoning_effort`, and the relay's text is the applied compaction entry. Mutation check: with `omitReasoningOptions` ignored in `generateSummaryMessage`, the test fails at the `"reasoning_effort" in body` assertion; restored → passes.

**Gates** (run on a designated ARM64 compute host, `CI=1`, Bun 1.4.2): the 10-file compaction regression set (`summarization-empty-stop-retry`, `-wire`, `summarization-tooluse-retry`, `speculative-compaction`, `required-compaction-deterministic-fallback`, `blocking-compaction-review-hardening`, `issue-7952-compaction-recovery`, `summarization-turn-order`, `overflow-retry-bound`, `summarization-body-too-large`) → 10 files / 104 tests passed; `biome check` on the changed `.ts` files → exit 0; root `tsc --noEmit` → exit 0. CI on `14a1573a0c` is the merge gate.

Follow-ups I am not taking in this PR: nothing logs that the override retry was spent (`SpeculativeCompactionContext` has no logger), and `speculative.ts` is well past the size where the summarization attempt loop should live in its own module.

code-yeongyu (approved):
All four blocking points are closed on 14a1573a0c (gate on a pinned override, accurate mechanism in CHANGELOG/tracker, Anthropic and tool-call interaction pinned, wire-level proof). Evidence: https://github.com/code-yeongyu/senpi/pull/1773#issuecomment-5710705282. CI green on the head.

PR #1786, opened 2026-09-17
fix(eval): reject cell declarations that would replace kernel globals

### LINKED ISSUE #1784: codemode: JS eval kernel silently replaces platform globals (fetch, URL, ...) when a cell redeclares them — session wedged until reset
## Summary

The JS eval kernel's persistence transform hoists top-level cell declarations onto `globalThis` without checking for collisions. A cell that declares `const fetch = ...` (or `let`/`var`, or a destructured binding like `const [fetch] = ...`) silently **replaces the platform global**. Every later cell in the session then fails with confusing `TypeError`s surfaced far from the cause, and the only recovery is a full kernel reset that loses all session state. Observed twice in real sessions: `URL` (2026-08-30) and `fetch` (2026-09-17, via destructured `const [which, remote, fetch] = await Promise.all(...)`).

## Reproduction

In any senpi session, js kernel:

```js
// cell 1
const fetch = "oops";
```

```js
// cell 2
typeof fetch; // "string" — the native fetch is gone for the rest of the session
await fetch("https://example.com"); // TypeError: fetch is not a function
```

Any later cell that touches `fetch` (directly or through a library, e.g. an SDK whose `fetchFn` resolves to the poisoned global) fails until the kernel is reset.

## Expected (ideal state)

A cell whose top-level declaration would overwrite an existing global (platform builtins like `fetch`, `URL`, `crypto`, or kernel prelude globals like `print`, `tool`) is **rejected before execution** with a clear error that names the identifier and the remedy (rename the binding, or assign `globalThis.fetch` explicitly if replacing is truly intended). The platform global survives; no silent wedge; no forced reset.

## Actual

`wrapUserCode` -> `persistTopLevelDeclarations` -> `rewriteDeclaration` collects binding names and emits `globalThis["name"] = ...` assignments unconditionally (`packages/senpi-codemode/src/kernels/js/worker-indirect-eval.js`). There is no collision check, so the poison is silent and the failure surfaces arbitrarily far from the guilty cell.

## Root cause

`rewriteDeclaration` in `worker-indirect-eval.js` rewrites every top-level `const`/`let`/`var` binding (plain and destructured) into `globalThis["<name>"] = ...` for cross-cell persistence. It never asks whether `<name>` already exists as a real global.

## Scope + acceptance criteria

- Guard fires for plain and destructured `const`/`let`/`var` top-level bindings that collide with a global that existed at kernel boot (platform + prelude), detected via a lazy `Object.getOwnPropertyNames(globalThis)` snapshot taken at first transform.
- Cell-created globals stay re-declarable across cells (a later cell may re-`const` a name an earlier cell created).
- Explicit `globalThis.x = ...` assignments in cell code are untouched (deliberate escape hatch / repair path).
- Error message names the offending identifier and the rename remedy.
- New regression tests in `test/kernel-js-persistence.test.ts` (guard rejects, native global intact in the next cell, re-declaration of cell-created names still works); full `senpi-codemode` suite and root `check` stay green.

## Notes

- `worker-indirect-eval.js` is 851 LOC (pre-existing size smell). The guard lands as a new small module (`worker-shadow-guard.js`) with only an import + call site in the big file; a full split of the lexer file is deliberately out of scope for this focused fix.

code-yeongyu:
## Summary

Fixes #1784 — the JS eval kernel's persistence transform hoisted every top-level `const`/`let`/`var` binding (plain or destructured) onto `globalThis` without a collision check. A cell declaring `const fetch = ...` (or `const [fetch] = ...`) silently replaced the platform global for every later cell; sessions wedged with confusing `TypeError`s far from the cause and only a full kernel reset recovered. Observed in real sessions with `URL` and `fetch`.

## Fix

- New `src/kernels/js/worker-shadow-guard.js`: lazily snapshots `Object.getOwnPropertyNames(globalThis)` at the first guard call — the first cell's transform, when prelude globals exist but no cell-created global does — and reports the first binding that would replace a protected global. Freezing at that moment protects platform + prelude globals forever while keeping cell-created names legal to re-declare.
- `worker-indirect-eval.js` `rewriteDeclaration` consults the guard before emitting `globalThis[...]` assignments; a colliding declaration fails the cell before execution with an error naming the identifier, the rename remedy, and the explicit `globalThis.<name>` escape hatch (deliberate replacements and the repair path stay possible).
- Error surfaces through the normal cell-error path (`result.ok === false`, message intact) — same as a syntax error today.

## Evidence (remote linux/x86_64 runner, node v24.20.0, bun 1.4.2)

RED (before the fix, same tests): `kernel-js-persistence.test.ts` exit 1 — 4 failed / 22 passed; the 4 new guard tests fail with `expected true to be false` on `poison.result.ok` (the shadowing cell silently succeeds today = the defect).

GREEN (after): same file exit 0, 26/26. Sibling `kernel-js-last-expression.test.ts` 13/13.

Full package suite: exit 0 — 132 test files passed (1 skipped), 922 tests passed (28 skipped, pre-existing capability gaps); zero failures, so no existing cell in the suite collides with the guard.

Real-surface QA (`scripts/qa-js-shadow-guard.ts`, real `JavaScriptKernel`, runtime=node): `QA-SHADOW-GUARD PASS` — cell `const fetch = "shadowed"` fails with

> eval cell declares top-level `fetch`, but a global with that name already exists in the JS kernel. Persisting it would replace that global for every later cell. Rename the binding (for example `fetchLocal`), or assign globalThis.fetch explicitly if replacing it is truly intended.

and the next cell still sees `typeof fetch === "function"`; destructured `const [URL] = ...` is rejected; a cell-created global stays re-declarable (`const qaShadowReuse = 1` then `= 2` → 2).

Root `bun run check`: green (result posted below).

## Notes

- `worker-indirect-eval.js` carries a pre-existing size smell (851 LOC); the guard lands as a new small module with only an import + call site in the big file. A full split of the lexer is deliberately out of scope here.
- Prompt change considered and declined: the guard error is self-healing (the model reads it, renames, retries), so no per-session token cost or prompt-snapshot churn.
- `changes.md` and `CHANGELOG.md [Unreleased]` entries are in this commit.

---
## Summary by cubic
Fixes #1784 by rejecting JS eval cell declarations that would replace existing kernel globals. Previously a cell like `const fetch = "shadowed"` silently overwrote `fetch` for every later cell and wedged the session; now the cell fails before execution with an error naming the binding and offering rename or explicit `globalThis` alternatives.

**Bug Fixes**
- A lazy snapshot of worker globals is taken at the first guard call, so cell-created globals remain re-declarable across cells.
- Explicit `globalThis.<name>` assignments still work as the deliberate override path.
- Adds test and QA-script coverage for plain, `let`, `var`, destructured, and prelude-global collisions.

<sup>Written for commit 0eb6a7f5ac959c510f1059ccf739060b7e6aac2b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1786?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Gate receipts at HEAD `0eb6a7f5ac` (run on a designated linux/x86_64 runner, node v24.20.0, bun 1.4.2; base `886f5371bd`):

```
# RED (tests only, before the fix)
kernel-js-persistence.test.ts: exit 1 — 4 failed / 22 passed
  (each failure: "expected true to be false" on poison.result.ok — the shadowing cell silently succeeds)

# GREEN (fix applied)
kernel-js-persistence.test.ts: exit 0 — 26 passed (26)
kernel-js-last-expression.test.ts: exit 0 — 13 passed (13)

# Full package suite
vitest run: exit 0 — Test Files 132 passed | 1 skipped (133); Tests 922 passed | 28 skipped (950)

# Real-surface QA (real JavaScriptKernel, runtime=node)
bunx tsx scripts/qa-js-shadow-guard.ts: exit 0 — QA-SHADOW-GUARD PASS
  poison cell `const fetch = "shadowed"` -> ok:false with the guard message naming `fetch` + rename remedy
  next cell `typeof fetch` -> "function"; destructured `const [URL] = ...` rejected;
  cell-created global re-declaration -> 2

# Root static gate
bun install --frozen-lockfile: exit 0 (442 packages)
bun run check: exit 0 (biome + tsgo + lock checks, zero TS errors)
```

code-yeongyu:
Fixed by #1786, merged as `08ae84b25b9ae07d0478a335b1eb58ed7d79f1f7` (`fix(eval): reject cell declarations that would replace kernel globals`, commit `0eb6a7f5ac`).

Evidence (full receipts on the PR):

- RED: new guard tests failed on the unfixed transform — `kernel-js-persistence.test.ts` exit 1, 4 failed / 22 passed, each failure `expected true to be false` on the shadowing cell's `result.ok` (silent poison = the defect).
- GREEN: same file exit 0, 26/26; sibling `kernel-js-last-expression.test.ts` 13/13.
- Full `senpi-codemode` suite: exit 0 — 132 test files passed (1 skipped), 922 tests passed (28 pre-existing capability skips); zero failures.
- Real-surface QA (`scripts/qa-js-shadow-guard.ts`, real `JavaScriptKernel`, runtime=node): `QA-SHADOW-GUARD PASS` — `const fetch = "shadowed"` now fails the cell with `eval cell declares top-level ` + "`fetch`" + `, but a global with that name already exists in the JS kernel. Persisting it would replace that global for every later cell. Rename the binding (for example ` + "`fetchLocal`" + `), or assign globalThis.fetch explicitly if replacing it is truly intended.` and the next cell still sees `typeof fetch === "function"`; destructured `const [URL] = ...` rejected; cell-created globals stay re-declarable.
- Root `bun run check`: exit 0 (biome + tsgo + lock checks).
- PR CI: full matrix green (Static checks, Test shards, Windows legs, Changelog gate, Review claim gate).

Behavior contract: top-level `const`/`let`/`var` declarations (plain or destructured) that name a global present at kernel boot (platform or prelude) are rejected before execution with a rename remedy; cell-created globals remain re-declarable; explicit `globalThis.<name> = ...` remains the deliberate escape hatch.

PR #1798, opened 2026-09-17
fix(scripts): make the CI format gate able to fail (fixes #1443)

### LINKED ISSUE #1443: CI runs biome with --write, so a formatting regression can never fail the Static checks job
## Summary

The root `check` script — which the CI **Static checks** job runs as `npm run check` — invokes Biome with `--write`:

```json
"check": "biome check --write --error-on-warnings . && ..."
```

In CI, `--write` **reformats the offending files inside the runner and exits 0**. The rewrite is never committed, so the drift stays in the repository while the job reports success. A formatting regression therefore cannot fail CI.

## Measured

Against `origin/main`'s copy of `packages/ai/test/gpt-6-astra-context-window.test.ts`:

| invocation | result |
|---|---|
| `biome check --error-on-warnings .` (no `--write`) | **rc=1** — reports the violation |
| `biome check --write --error-on-warnings .` (the CI form) | **rc=0**, and rewrites the file (4 insertions / 1 deletion) |

So that file is genuinely unformatted on `main` today, and CI has been green on it throughout.

## Why this matters

This is a gate that cannot fail. Its purpose is to reject unformatted code; instead it silently repairs a copy that is thrown away when the runner exits. Two consequences:

1. **Formatting drift accumulates in the repo unnoticed** — nothing ever reports it.
2. **Local and CI disagree.** A contributor running the documented `npm run check` locally gets their working tree rewritten (convenient), while a contributor running `biome check` without `--write` sees a failure CI never shows. That mismatch cost a session real time today: a lane observed the local failure, reasonably inferred it was the cause of a red CI aggregate, and reported it as such — it was not, because the CI form cannot fail that way.

## Suggested direction

Separate the two intentions rather than overloading one script:

- CI should run a **verify** form with no `--write` (e.g. `biome check --error-on-warnings .`), so drift fails the job.
- Keep a **fix** form (`--write`) as a separate script for local use, e.g. `check:fix`.

Whatever the split, the CI path must be able to fail on unformatted input — otherwise the check is decorative.

## Note

The immediate drift in `gpt-6-astra-context-window.test.ts` is being reformatted separately as hygiene. That does not address this issue: the next unformatted file will be equally invisible.

MoerAI:
Fixes #1443

## Root cause

The root `check` script led with `biome check --write --error-on-warnings .`, and the CI **Static checks** job runs exactly `npm run check`. With `--write`, biome reformats the offending files inside the runner and exits 0. The rewrite is discarded when the runner exits, so the drift stays in the repository while the job reports success — a gate that cannot fail.

`releasability.yml` had already worked around this by hand-inlining a read-only biome step plus a verbatim copy of the remaining check sub-scripts, with a comment naming the hazard. That copy had since drifted from the real chain: it was missing `check:entry-graphs` and `check:claude-sdk-platform-lock`.

Root `AGENTS.md` also states the local check and CI run the same commands and must stay in sync. With autofix in the shared script they disagreed by construction: a contributor running `biome check` without `--write` saw a failure CI could never show.

## Change

- `package.json`: `check` now runs `biome check --error-on-warnings .` (read-only). New `check:fix` keeps the autofix form (`biome check --write --error-on-warnings . && npm run check`) for local use.
- `.husky/pre-commit`: calls `npm run check:fix`, so the hook's autofix-then-verify behavior is unchanged.
- `.github/workflows/releasability.yml`: drops the inlined duplicate and calls `npm run check` directly, removing the drift source.
- `changes.md` / `.husky/changes.md`: tracker entries for the two upstream-tracked paths.

No dependency, lockfile, or release-version changes.

## Test

`scripts/ci-format-gate.test.mjs` (runs under `npm run test:scripts`) parses `.github/workflows/ci.yml`, resolves the script the Static checks job actually invokes, and asserts:

1. that script's biome invocation carries no `--write`/`--fix`;
2. running it against a drifted file inside a biome-included path exits 1 and leaves the file byte-identical;
3. an explicit local autofix script still exists.

It derives the flags from the live script string rather than hardcoding them, and resolves the biome binary from `node_modules` instead of a package-manager runner, so it behaves the same under npm, bun and pnpm.

### RED (before the fix)

```
$ node --test scripts/ci-format-gate.test.mjs
AssertionError: CI biome invocation rewrites files and exits 0 on drift: biome check --write --error-on-warnings .
AssertionError: expected drift to fail. stdout: Checked 1 file in 33ms. Fixed 1 file.
AssertionError: expected a check:fix script for local autofix
ℹ tests 3
ℹ pass 0
ℹ fail 3
```

### GREEN (after the fix)

```
$ node --test scripts/ci-format-gate.test.mjs
✔ runs a biome form that reports drift instead of rewriting it
✔ fails on unformatted input with the CI biome invocation
✔ keeps an explicit autofix script for local use
ℹ tests 3
ℹ pass 3
ℹ fail 0
```

## Verification on the real command

A deliberately unformatted file was placed under a biome-included path and the actual CI command was run:

```
$ npm run check
EXIT=1
  × Some errors were emitted while running checks.
# fixture afterwards: export const qaDrift   =    1;   (unchanged)
```

The same fixture under the previous form, for contrast:

```
$ biome check --write --error-on-warnings <fixture>
Checked 1 file in 38ms. Fixed 1 file.
EXIT=0
# fixture afterwards: export const qaDrift = 1;        (silently rewritten)
```

Clean-tree `npm run check` exits 0 on current `main` content (4262 files, "No fixes applied", 6 pre-existing infos that `--error-on-warnings` does not escalate), so making CI read-only does not turn `main` red.

## QA

CLI smoke via `.agents/skills/senpi-qa/scripts/cli-smoke.mjs --self-test` — 8/8 passed, including the credential-isolation assertion that the real `auth.json` is unchanged:

```
[PASS] --help prints usage with flags — code=0
[PASS] --version prints a version — 2026.9.17-2
[PASS] --list-models lists built-in models offline (no API) — lines=1681
[PASS] unknown option is reported, not silently ignored — code=1
[PASS] real auth unchanged
cli-smoke.mjs --self-test: 8/8 passed
```

This change touches build/CI tooling only — no source under the release-managed runtime packages — so the CLI channel is the applicable surface. The pre-commit hook exercised the new `check:fix` path when this commit was created.

## Note on the local full-suite run

`npm test` fails on this machine at `test:scripts` with 6 failures in `scripts/probe-native-grep.test.mjs`, all `Internal Error: cargo metadata failed to run`. These are pre-existing and environmental: `cargo`/`rustc` are not installed here, the file is byte-identical to `main`, and a pristine `main` checkout (d35d45d1c) with none of these changes reproduces the identical 6 failures. All other 288 script tests pass, including the new gate test. CI has the Rust toolchain and will run the full suite.

---
## Summary by cubic
Fixes #1443 by making the root `check` script run Biome read-only, so the CI format gate can actually fail. `check` used `biome check --write`, which silently rewrote format drift in the CI runner and exited 0; now it reports drift and exits non-zero.

**Changes**

- `check` now runs `biome check --error-on-warnings .`; the new `check:fix` script keeps the autofix form for local use.
- The pre-commit hook runs `check:fix`, preserving its existing autofix-then-verify behavior.
- `releasability.yml` now calls `npm run check` directly, removing an inlined duplicate that lacked `check:entry-graphs` and `check:claude-sdk-platform-lock`.
- Adds `scripts/ci-format-gate.test.mjs` to enforce that the CI biome invocation cannot autofix, that drift makes it fail, and that an explicit local autofix script exists.

<sup>Written for commit cb1600925fa2db4d25b268988511fb8b3b3f722f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Rebased onto current main as #1858 (head includes #1855, which formatted the session-worker-requests drift that made this PR's Static checks fail — that red was the gate working).

Landing #1858 with Co-authored-by credit. This PR can close as superseded once #1858 merges.

PR #1816, opened 2026-09-18
feat(ai): add B.AI provider

Momentum96:
## Summary
- add native `bai` API-key auth and credential-scoped `/v1/models` discovery
- generate official B.AI metadata for 47 classified chat models and route Responses, Messages, or Chat Completions per family
- normalize missing root `type: "object"` only on B.AI function schemas, including union-root `workpool`
- migrate cached entries by model ID and document setup, pricing, and image-only exclusions

## Verification
- `bun run check`
- focused AI/provider/model-data tests: 56 passed, 1 opt-in skipped
- coding-agent resolver and goal wake tests: 133 passed
- `PI_ENABLE_BAI_LIVE=1 ... test/bai-live.test.ts`: passed against all available B.AI wire families
- isolated source CLI: `update --models` succeeded and `bai/gpt-5.6-luna` returned `OK`; real auth unchanged
- CLI smoke 8/8; Responses mock tool loop 4/4

## Local full-suite note
The full workspace run exposed one existing compaction characterization assertion (`4012 <= 4000`) on this Node 26 host. Removing B.AI registration and rerunning the focused case produced the identical failure. The B.AI-caused default-model failure was fixed and its 55-test resolver suite is green. CI remains the authoritative Node 24 full-suite result.

---
## Summary by cubic
Adds a built-in `bai` provider so B.AI models can be used with `BAI_API_KEY` or stored credentials. Previously unavailable, B.AI now supports credential-scoped discovery with metadata for 56 documented chat models.

**Behavior**
- GPT and DeepSeek use OpenAI Responses, Claude uses Anthropic Messages, and other supported families use Chat Completions.
- Discovery accepts both dot-version and hyphenated model IDs; failed responses do not replace the existing catalog.
- Fresh catalogs start empty until a credentialed refresh succeeds, while cached entries are remapped by model ID.
- B.AI Responses requests merge union-root function schemas into object schemas without mutating caller payloads.
- Image-only IDs such as `gpt-image-2` stay out of the chat catalog, and `bai` defaults to `gpt-5.6-sol`.
- Generated prices are standard reference values; B.AI’s billing record remains authoritative.

**Migration**
- Set `BAI_API_KEY` or run `/login bai`, then refresh with `senpi update --models`.

<sup>Written for commit 8feaf42543cf3a6954e22a7df6806ff98a0f382c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1816?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/ai/src/providers/bai-stream.ts
```diff
@@ -0,0 +1,60 @@
+import type { ProviderStreams, SimpleStreamOptions, StreamOptions } from "../types.ts";
+
+function isJsonObject(value: unknown): value is Record<string, unknown> {
+	return value !== null && typeof value === "object" && !Array.isArray(value);
+}
+
+function normalizeFunctionTool(tool: unknown): unknown {
+	if (!isJsonObject(tool) || tool.type !== "function") return tool;
+
+	if (isJsonObject(tool.parameters) && tool.parameters.type === undefined) {
```
code-yeongyu: **B1 (blocker).** `{ type: "object", ...tool.parameters }` on a union root yields `{ type: "object", anyOf: [...] }` with no top-level `properties`/`required`. On the Responses path this is exactly what reaches the wire (`api/openai-responses-shared.ts:587` -> `api/constrained-sampling.ts:129` is a pass-through), so the tool is advertised to the model as taking no arguments — see the comment at `api/anthropic-messages.ts:2517-2519`. Use `resolveRootObjectSchema` / `normalizeToolParametersForOpenAICompat` from `utils/tool-schema-compat.ts`, which merge branch `properties`/`required` up into the root.

packages/ai/src/providers/bai-stream.ts
```diff
@@ -0,0 +1,60 @@
...
+		return {
+			...tool,
+			parameters: {
+				type: "object",
+				...tool.parameters,
+			},
+		};
+	}
+
+	if (!isJsonObject(tool.function)) return tool;
```
code-yeongyu: **B2 (blocker).** Unreachable. `api/openai-completions.ts:497` re-runs `normalizeRequestToolSchemas` **after** `onPayload`, which re-normalizes `tool.function.parameters` through `normalizeToolParametersForOpenAICompat` (`:1814-1833`) and overwrites this. The Anthropic Messages path never matches either — its tools are `input_schema` (`api/anthropic-messages.ts:2543`), already root-collapsed at `:2521`. Drop the dead branches and correct the claim in `packages/ai/src/changes.md`.

packages/ai/test/bai-tool-schema.test.ts
```diff
@@ -0,0 +1,82 @@
...
+						type: "object",
+						properties: { path: { type: "string" } },
+					},
+				},
+			],
+		};
+
+		const normalized = normalizeBaiRequestPayload(payload);
+
+		expect(normalized).toEqual({
```
code-yeongyu: **B3 (blocker).** The expectation is re-derived from the input (`{ type: "object", ...payload.tools[0].parameters }`), so it restates the implementation instead of pinning a contract, and it inspects `normalizeBaiRequestPayload` in isolation — the only place the result survives. It stays green through both B1 and B2. Assert the payload that actually reaches the transport: `baiProvider().streamSimple(model, ctx, { fetch: capture })` with a union-root tool, per wire API.

packages/ai/scripts/generate-models-bai.ts
```diff
@@ -0,0 +1,82 @@
...
+ * - https://docs.b.ai/llmservice/pricing-and-usage/
+ */
+export function getBaiModels(): Model<BaiApi>[] {
+	return Object.entries(baiMetadata).map(([id, metadata]) => ({
+		id,
+		name: formatBaiModelName(id),
+		api: metadata.api,
+		provider: "bai",
+		baseUrl: metadata.api === "anthropic-messages" ? BAI_ANTHROPIC_BASE_URL : BAI_BASE_URL,
+		reasoning: metadata.thinkingLevelMap !== undefined,
```
code-yeongyu, line 89: **B4 (blocker).** `reasoning` is derived here from this file's own `thinkingLevelMap`, but later id-keyed steps in `generate-models.ts` inject a map without re-deriving it. Result in `src/providers/data/bai.json`: `claude-opus-4.6`, `claude-opus-4.8`, `claude-sonnet-4.6` and `gpt-5.5-instant` ship `reasoning: false` next to a populated `thinkingLevelMap`, so `api/openai-responses.ts:486` and `api/anthropic-messages.ts:1956,2104` skip the reasoning payload and those four models lose thinking entirely.

packages/ai/src/providers/bai.ts
```diff
@@ -0,0 +1,95 @@
...
+			authorization: `Bearer ${apiKey}`,
+		},
+		signal: context.signal,
+	});
+	if (!response.ok) {
+		throw new Error(`Could not load B.AI model catalog: ${response.status}`);
+	}
+
+	const payload = (await response.json()) as BaiModelsResponse;
+	if (!Array.isArray(payload.data)) {
```
code-yeongyu, line 87: **N2 (note).** The documented List Models body is `{ object, success, data }`. Only `data` is checked, so a `success: false` response with an empty or partial `data` array is accepted as a valid catalog rather than surfaced as a refresh error.

code-yeongyu (changes requested):
## REQUEST CHANGES — reviewed against the official B.AI spec (https://docs.b.ai/llmservice/api/)

Thanks for this — the shape is right: one key, three wire protocols, credential-scoped `/v1/models`, a generated catalog. Auth resolution is correct (`envApiKeyAuth` → `resolveRefreshCredential` means the documented `export BAI_API_KEY=… && senpi update --models` flow really works), the Anthropic base-URL split resolves to `https://api.b.ai/v1/messages`, and the spec's "DeepSeek must not receive a web search tool" rule is already satisfied because `sanitizeUnsupportedNativeTools` strips `web_search_preview` on custom Responses endpoints (`packages/ai/src/api/openai-responses.ts:142-148,254`).

The tool-schema layer is the problem. It is the PR's central justification, and as written it is unreachable on two of the three protocols and actively harmful on the third.

---

### B1 — blocker: union-root tool schemas reach B.AI Responses without `properties`/`required`, so the tool is advertised as taking no arguments

`bai-stream.ts:10-16` repairs a union root by spreading a `type` in front of it:

```ts
parameters: { type: "object", ...tool.parameters }   // -> { type: "object", anyOf: [...] }
```

The Responses tool builder passes `tool.parameters` through untouched — `packages/ai/src/api/openai-responses-shared.ts:587` calls `getJsonSchemaToolParameters`, which is a straight pass-through (`packages/ai/src/api/constrained-sampling.ts:129-131`). So that shallow object **is** what goes on the wire.

A root union carries no top-level `properties` or `required`. This repo already says, twice, what that costs:

- `packages/ai/src/api/anthropic-messages.ts:2517-2519` — *"A root union carries no top-level properties, so reading them directly would advertise the tool to the model as taking no arguments at all."*
- `packages/ai/src/utils/tool-schema-compat.ts` (`resolveRootObjectSchema` doc) — *"a root union carries neither, so they would otherwise describe the tool to the model as taking no arguments."*

So the 400 goes away and `workpool` becomes a tool the model believes has no parameters. That is worse than the error it replaces, because it fails silently at runtime.

The correct merge already exists and is already used by the other two protocols: `mergeRootObjectUnion` / `ensureRootObjectSchema` in `packages/ai/src/utils/tool-schema-compat.ts`, exposed as `resolveRootObjectSchema` and `normalizeToolParametersForOpenAICompat`. Its doc comment names the exact B.AI error you are chasing:

> OpenAI-compatible gateways reject a root that only carries `anyOf`/`oneOf` with `tools.function.parameters.type is required and must be "object"`.

**Fix:** normalize through that helper instead of the ad-hoc spread. The root cause is not B.AI-specific — the Responses tool builder is the only one of the three that skips root normalization — so fixing it there (guarded by compat, as the other providers do) is preferable to a provider-local patch, and every strict Responses gateway gets it.

### B2 — blocker: two thirds of `normalizeBaiRequestPayload` is dead code, and `changes.md` claims otherwise

- **Chat Completions:** `packages/ai/src/api/openai-completions.ts:493-497` runs `onPayload`, then immediately runs `normalizeRequestToolSchemas`, which re-normalizes `tool.function.parameters` through `normalizeToolParametersForOpenAICompat` (`:1814-1833`). Whatever `bai-stream.ts:20-33` produced is overwritten before the request is sent.
- **Anthropic Messages:** tools there are `input_schema`, not `parameters` (`packages/ai/src/api/anthropic-messages.ts:2543`), and the root is already collapsed by `resolveRootObjectSchema` at `:2521`. `normalizeBaiRequestPayload` looks for `parameters` and never matches.

`packages/ai/src/changes.md` states the repair is needed across Responses / Messages / Chat Completions "including union-root `workpool`". Only Responses is real. Please delete the unreachable branches and correct the tracker rather than shipping a claim the code does not support.

### B3 — blocker: `bai-tool-schema.test.ts` cannot fail for the regression it names

`packages/ai/test/bai-tool-schema.test.ts:38-52` derives its expectation from the input:

```ts
parameters: { type: "object", ...payload.tools[0].parameters }
```

That restates the implementation rather than pinning a contract, and it tests `normalizeBaiRequestPayload` in isolation — the one place where the result is still intact. It would stay green through both B1 and B2. Please assert the payload that actually reaches the transport: drive `baiProvider().streamSimple(model, ctx, { fetch: capture })` with a union-root tool and assert the request body contains the merged `properties`/`required`, for each of the three wire APIs.

### B4 — blocker: four models ship `reasoning: false` next to a populated `thinkingLevelMap`, so their thinking levels are unreachable

In `packages/ai/src/providers/data/bai.json`:

| model | `reasoning` | `thinkingLevelMap` |
|---|---|---|
| `bai/claude-opus-4.6` | `false` | `{ max }` |
| `bai/claude-opus-4.8` | `false` | `{ xhigh, max }` |
| `bai/claude-sonnet-4.6` | `false` | `{ max }` |
| `bai/gpt-5.5-instant` | `false` | `{ off, xhigh }` |

`generate-models-bai.ts:68` derives `reasoning` from the source file's own `thinkingLevelMap`, but later id-keyed steps in `generate-models.ts` inject a map without re-deriving `reasoning` — in all four cases `bai-models.json` has no map and the generated shard does.

Consequence, in code: `packages/ai/src/api/openai-responses.ts:486` (`if (model.reasoning)`) and `packages/ai/src/api/anthropic-messages.ts:1956,2104` skip the reasoning/thinking payload entirely, and `openai-responses-shared.ts:285` also drops those models back to the `system` role. Compare `anthropic/claude-opus-4-6`, which is `reasoning: true` for the same underlying model. Derive `reasoning` after the shared overrides, or assert the invariant in a test.

### B5 — blocker: no `CHANGELOG.md` `[Unreleased]` entry

`.github/workflows/changelog-gate.yml` requires a `[Unreleased]` entry (or the `no-changelog` label) for user-facing PRs. This PR adds four `changes.md` trackers and no changelog entry. A new user-facing provider is exactly what that gate is for.

### B6 — blocker: CI has never run on this PR

At `75fa5b29`, `CI`, `Changelog gate`, `Publish Model Catalog` and `Read summary binary parity` are all `action_required` with 0s runtime. Only the review-claim gate, cubic and GitGuardian have reported. Every verification claim in the description is local-only, on a Node 26 host, against a run the description itself reports as failing. Nothing here has been validated on the Node 24 CI matrix.

---

### Notes (non-blocking)

- **N1** `packages/ai/scripts/bai-models.json` is described as "B.AI's published standard metadata", but several values are rewritten downstream — e.g. `gpt-5.6-sol` 1,050,000 → 922,000 through the global context cap at `generate-models.ts:417`. Worth saying so in the file header, otherwise the next maintainer will trust the numbers as shipped.
- **N2** The spec's List Models response carries `success` alongside `data`. `bai.ts:59` only checks `data`, so a `success: false` body with an empty or partial `data` array is accepted as a valid catalog.
- **N3** `bai.ts:48` sends only `Authorization: Bearer`. Fine — the spec calls `Authorization` and `x-api-key` equivalent and recommends `Authorization` for OpenAI-compatible clients.
- **N4** `GET /v1/balance` is documented and unused. Optional follow-up if we ever want quota display.
- **N5** The DeepSeek/web-search rule holds today only because B.AI models do not set `compat.supportsWebSearchPreview`. Worth a comment so nobody enables that flag for the GPT families and silently breaks DeepSeek.

I'm taking the fixes on this branch myself so it can land today.

code-yeongyu:
Every blocker from my review is fixed on this branch; pushed as `8feaf425`.

**B1 — union-root tool schemas.** Captured at the transport before and after. A `workpool` tool, pre-fix, on `POST https://api.b.ai/v1/responses`:

```
workpool schema keys: type,anyOf
root type:      "object"
properties:     undefined
required:       undefined
```

After routing the repair through `normalizeToolParametersForOpenAICompat()`:

```
workpool schema keys: type,properties,required
root type:      "object"
properties:     ["op","name","pool_id"]
required:       ["op"]
```

**B2 — dead branches.** The same capture showed Chat Completions and Messages already sending `properties:["op","name","pool_id"]` with no B.AI transform at all, so only the Responses API is wrapped now. `packages/ai/src/changes.md` says so instead of claiming all three.

**B3 — test integrity.** `bai-tool-schema.test.ts` now drives `baiProvider().streamSimple()` with a capturing `fetch` and asserts the request body per wire API. Mutation-proved: restoring the old shallow spread fails it with `expected [] to deeply equal [ 'name', 'op', 'pool_id' ]`, and the other two protocols stay green under that mutation, which is the second half of the B2 evidence. It also pins the documented output-token field per endpoint (`max_output_tokens` / `max_completion_tokens` / `max_tokens`).

**B4 — reasoning consistency.** `generate-models.ts` re-derives `reasoning` for B.AI models after the shared thinking-level passes. The regenerated shard has zero models with a populated `thinkingLevelMap` next to `reasoning: false`, down from four.

**B5/B6 — gates.** `[Unreleased]` entries added to `packages/ai` and `packages/coding-agent`; `node scripts/check-pr-changelog.mjs` now reports `PASS - changes.md coverage complete (7 production path(s) covered); changelog entry updated`. The four workflows that had been sitting at `action_required` since the branch opened are approved and running.

**N2** — a `success: false` model list now fails the refresh instead of publishing an empty catalog.

---

Two things the review did not yet cover, found while checking the catalog against every per-model page reachable from `docs.b.ai/sitemap.xml`:

**Model ID aliases.** B.AI states dot-version IDs on most model pages (`gpt-5.6-sol`) but hyphenated ones for Claude (`claude-fable-5-1`), and its Claude Code guide says both aliases are accepted. Which spelling `GET /v1/models` returns is not something I can verify without a key, and `catalogById.get(id)` drops anything that does not match exactly — so on the hyphenated answer all 11 Claude entries would silently vanish. Discovery now indexes both spellings.

**Catalog data.** 44 of 47 entries matched the published pricing table exactly. The three that did not: Hy3 carried promotional rather than standard prices and was off ~7.5x in every field; GLM-5.3 and MiMo-V2.5 priced cache writes at zero. Hy3, GLM-5.2 and Kimi K3 asserted output limits their pages contradict; GPT-5.4 is documented as 272K by default, expandable only through explicit configuration; GPT-5.5 Instant shares its sibling's 1,050,000-token window. Nine models B.AI documents as active were absent, so an entitled key saw nothing for them — DeepSeek V3.2 / V4 Flash / V4 Flash Vision Exp, Grok 4.5 / 4.6, Kimi K2.5, Muse Spark 1.3, Qwen3.6 27B, Qwen3.7 Max. The catalog is now the 56 models B.AI lists as active, `gpt-image-2` still excluded.

Local gates on `8feaf425`: `bun run check` exit 0, `packages/ai` 2884 tests passed / 0 failed across 298 files. One suite failed to import on first run because `packages/tui/dist` is absent in a fresh worktree; it passes after `bun run --cwd packages/tui build`, and it is unrelated to this change.

Nice work on the provider shape — the auth resolution, the Anthropic base-URL split, and the credential-scoped catalog were all right, and the `sanitizeUnsupportedNativeTools` path already satisfies B.AI's "no web search tool for DeepSeek" rule for free.

code-yeongyu (approved):
Approving. Every blocker from my earlier review is resolved on `8feaf425`, and the branch's first real CI run is green.

- **B1** union-root schemas now reach B.AI Responses as a merged object schema — captured at the transport, `properties: undefined` -> `["op","name","pool_id"]`.
- **B2** the two unreachable normalization branches are gone and `packages/ai/src/changes.md` no longer claims them.
- **B3** the schema test drives `baiProvider().streamSimple()` with a capturing `fetch` per wire API; mutation-proved against the old implementation.
- **B4** the regenerated shard has zero models carrying a `thinkingLevelMap` next to `reasoning: false`.
- **B5** `[Unreleased]` entries added; the Changelog gate passes.
- **B6** the four workflows that had never run are approved and green.
- **N2** a `success: false` model list fails the refresh instead of publishing an empty catalog.

Plus the two findings from the catalog audit: discovery now resolves both the dot-version and hyphenated spelling of every model ID, and the catalog matches B.AI's published pages and pricing table at 56 active models.

End-to-end discovery proof, env-only credential against a local B.AI stub:

```
before refresh: 0 models
request path: /v1/models | authorization: Bearer <BAI_API_KEY> (no stored credential)
after refresh: 3 models
  gpt-5.6-sol        api=openai-responses     ctx=922000
  claude-fable-5.1   api=anthropic-messages   ctx=1000000   (discovered as claude-fable-5-1)
  deepseek-v4-flash  api=openai-responses     ctx=1000000
image-only id excluded: true
```

Thanks for the provider — the auth resolution, the Anthropic base-URL split, and the credential-scoped catalog design were all sound.

packages/ai/src/providers/bai-stream.ts
code-yeongyu: Fixed in `3bddac6852`: the union root is merged through `normalizeToolParametersForOpenAICompat()`, so the Responses payload now carries `properties`/`required` instead of a bare `type` in front of `anyOf`. Wire capture is in the PR comment.

packages/ai/src/providers/bai-stream.ts
code-yeongyu: Fixed in `3bddac6852`: the Chat Completions and Messages branches are removed and only the Responses API is wrapped. `packages/ai/src/changes.md` now states that the other two paths normalize their own roots.

packages/ai/test/bai-tool-schema.test.ts
code-yeongyu: Fixed in `3bddac6852`: the test now captures the request that reaches the transport for each wire API. Mutation-proved — restoring the old shallow spread fails it with `expected [] to deeply equal [ 'name', 'op', 'pool_id' ]`.

packages/ai/scripts/generate-models-bai.ts
code-yeongyu: Fixed in `0f08521bbf`: `generate-models.ts` re-derives `reasoning` for B.AI models after the shared thinking-level passes. The regenerated shard has zero models carrying a populated `thinkingLevelMap` next to `reasoning: false`.

packages/ai/src/providers/bai.ts
code-yeongyu: Fixed in `3bddac6852`: a `success: false` model list now fails the refresh with the server's message instead of being accepted as a valid catalog.

PR #1827, opened 2026-09-18
feat(prompt-preset): route Kimi K2.8 Preview through the Kimi K2.7 prompt

code-yeongyu:
Fixes #1826

## What

Moonshot upgraded Kimi Code's rolling `kimi-for-coding` model id to **K2.8 Preview in place** on 2026-09-11, so every Kimi Code session has been served by K2.8 while resolving to no prompt preset at all - no `kimi` dialect, no workstation dialect, no tuning. This adds a `kimi-k2-8` preset and points the rolling ids at the preset their published model version implies.

- `kimi-k2-code.ts` (new) holds the K2.7 tuning as `buildKimiK2CodePrompt(options, modelName)`. `kimi-k2-7.ts` and the new `kimi-k2-8.ts` are thin aliases over it (the `glm-5.ts` / `glm-5-{2,3}.ts` shape), so the two prompts are byte-identical apart from the model name they announce. K2.8 is an efficiency and context upgrade inside the same K2 coding family, not a new prompting contract.
- `presets.ts` gains `hasKimiK28Signal` / `isKimiK28Model`, dispatched ahead of `kimi-k2-7`. Both Kimi matchers also accept Kimi Code's rolling product ids by exact match: `kimi-for-coding` -> `kimi-k2-8`, `kimi-for-coding-highspeed` -> `kimi-k2-7` (still K2.7 Code HighSpeed per the same table).
- `"kimi-k2-8"` joins `PromptPresetName` / `VALID_PRESETS`, `docs/settings.md`, and both `AGENTS.md` inventories.

## Source

The Kimi Code model table is the evidence for both mappings: <https://www.kimi.com/code/docs/en/kimi-code/models.html> (checked 2026-09-18), and the rollout note ("The Model ID is unchanged - still `kimi-for-coding`"): <https://www.kimi.com/code/docs/en/kimi-code/whats-new.html>.

## Tests

New `test/suite/prompt-presets-kimi-k2-8.test.ts`:

- every K2.8 id shape (`kimi-for-coding`, `kimi-k2.8`, `kimi-k2.8-code`, `kimi-k2p8-code`, `moonshotai/kimi-k2.8-code`, `moonshotai/kimi-k2.8:thinking`, `@cf/...`, `accounts/fireworks/models/kimi-k2p8-code`) resolves to `kimi-k2-8` with a Kimi prompt;
- a catalog row that carries the version only in its display name (`Kimi K2.8 Preview`) resolves too;
[...]

Issue #1826, opened 2026-09-18
prompt-preset: Kimi K2.8 Preview has no preset, so Kimi Code sessions fall back to the default prompt


code-yeongyu:
## Summary

Moonshot rolled **Kimi K2.8 Preview** out across Kimi Code on 2026-09-11 and upgraded the `kimi-for-coding` model id **in place** - same id, no client configuration change. senpi's `prompt-preset` builtin ships presets for `kimi-k2-6`, `kimi-k2-7` and `kimi-k3`, but nothing matches K2.8, so every K2.8 session silently falls back to the generic dynamic prompt and loses the Kimi tuning (execution-tooling stance in the `kimi` dialect, `workstationDialect: "kimi"`, the restrained outcome-first tuning).

The same model table also shows `kimi-for-coding-highspeed` is **K2.7 Code HighSpeed**, and it misses the K2.7 preset today for the same reason: the Kimi Code ids are rolling product ids, not version-tagged ones.

## Reproduction

```ts
import { resolvePresetName } from "./packages/coding-agent/src/core/extensions/builtin/prompt-preset/presets.ts";

const settings = { promptPreset: "auto" } as const;
resolvePresetName({ id: "kimi-for-coding", provider: "kimi-coding" }, settings);
// => undefined   (expected: "kimi-k2-8")
resolvePresetName({ id: "kimi-for-coding-highspeed", provider: "kimi-coding" }, settings);
// => undefined   (expected: "kimi-k2-7")
resolvePresetName({ id: "moonshotai/kimi-k2.8", provider: "openrouter" }, settings);
// => undefined   (expected: "kimi-k2-8")
```

## Expected (ideal state)

1. A `kimi-k2-8` preset exists and renders **the K2.7 prompt**: K2.8 Preview is the in-place successor in the same K2 coding family, so the tuning is shared verbatim and only the model name differs.
2. Auto-detection resolves K2.8 from both the version-tagged id shapes (`kimi-k2.8`, `kimi-k2p8`, `moonshotai/kimi-k2.8:thinking`, `@cf/...`, `accounts/fireworks/models/kimi-k2p8-code`, display name `Kimi K2.8 Preview`) and Kimi Code's rolling `kimi-for-coding` id.
3. `kimi-for-coding-highspeed` resolves to `kimi-k2-7`, matching the published model table.
4. `promptPreset: "kimi-k2-8"` is selectable from settings and documented in `docs/settings.md`.
[...]

PR #1829, opened 2026-09-18
fix(ai): let a fork-owned provider survive a catalog regeneration

### LINKED ISSUE #1825: Release is blocked: a fork-shipped provider (kimi-coding) drops out of the regenerated model catalog
## Summary

Every `publish-npm` dispatch on green `main` fails in `npm run check`, after the release regenerates the model catalogs. The cause is upstream data, not a code change: models.dev no longer describes `kimi-coding`, and the fork still ships that provider.

## What four consecutive dispatches showed

1. The prune deleted `packages/ai/src/providers/kimi-coding.models.ts` while `kimi-coding.ts` imports it -> `TS2307`. Fixed in #1823.
2. The kept shard was then rejected by the aggregator/shard equality gate as `extra`. Fixed in #1824.
3. Now the regenerated aggregate no longer contains the provider at all, so the type derived from it loses the `"kimi-coding"` literal:

```
packages/ai/test/stream.test.ts(1056,25): error TS2345:
  Argument of type '"kimi-coding"' is not assignable to parameter of type
  '"azure-openai-responses" | "google-vertex" | ... 30 more ... | "venice"'
```

About a dozen packages/ai tests plus `packages/coding-agent/test/model-resolver.test.ts` fail this way.

## Why ordinary CI is green

CI type-checks the committed catalog; only the release regenerates it first. `packages/ai/scripts/model-shards.ts` documents exactly this blind spot.

## Expected

A provider the fork ships survives a generation that upstream no longer describes - shard, data, aggregate entry and provider-id literal - the way `devin` is handled, or the fork's tests stop depending on the generated union for such a provider.

## Acceptance

- `publish-npm` on a green main reaches the publish step without a catalog-related type error.
- A test fails when a fork-shipped provider would drop out of the generated aggregate.

Found while releasing unrelated RPC-host work; #1823 and #1824 are already merged and are prerequisites of this one.

code-yeongyu:
Fixes #1825.

## Why the release keeps failing

`publish-npm` regenerates the model catalogs before it type-checks. models.dev no longer describes `kimi-coding`, so the run emits neither its shard nor its `data/kimi-coding.json`, the provider drops out of the generated `MODELS`, and the fifteen call sites that name it stop type-checking. Ordinary CI never sees it because it type-checks the committed catalog instead of regenerating one.

Two earlier fixes removed the first two layers - #1823 stopped the prune from deleting a shard a provider module imports, #1824 stopped the aggregator gate from calling that shard `extra`. This is the third and last: the provider still vanished from the aggregate itself.

## What this does

Makes `kimi-coding` fork-owned the way `devin` already is:

- the shard is hand-written with inline values (5 models, one API) instead of importing a data file a generation run will not produce, and is listed in `FORK_OWNED_MODEL_SHARDS`;
- `data/kimi-coding.json` and its manifest entry are gone, since `data/` is generator-owned territory - the manifest was rewritten by the repo's own `createModelDataManifest`, not by hand;
- because such a provider can never be a key of `MODELS`, `providers/all.ts` gains a `FORK_OWNED_CATALOGS` seam that `getBuiltinModel`, `getBuiltinModels` and `getBuiltinProviders` read alongside the generated ones.

**No test file needed changing.** The tests were right; the catalog type had quietly narrowed under them.

## Evidence

```
tsgo --noEmit           15 errors -> 0
check:model-data        Generated model data is valid. (exit 0)
model-shards + model-data-validation + generate-models-strict   24 passed
fork-owned-catalogs.test.ts                                      4 passed
```

RED verified by restoring only main's `all.ts` under the new test: 3 of 4 fail, and pass again with the seam in place.

## Note for the release

This unblocks the catalog step. The new test also fails if a future change puts a fork-owned provider back into the generated aggregate or drops it from the catalog API, so the next regeneration cannot repeat this quietly.

---
## Summary by cubic
Fixes the release job's catalog regeneration dropping `kimi-coding` from the generated `MODELS` type, which broke type-checking at fifteen call sites. The provider is now fork-owned like `devin`, so it survives regeneration and stays served by the catalog API.

- The shard is hand-written with inline model values and listed in `FORK_OWNED_MODEL_SHARDS` so generation leaves it alone.
- `data/kimi-coding.json` and its manifest entry are removed; `getBuiltinModel`, `getBuiltinModels`, and `getBuiltinProviders` read fork-owned providers from a new seam in `providers/all.ts`.
- A new test fails if a fork-owned provider returns to the generated aggregate or drops out of the catalog API, and the imported-shard test now targets whichever generated shard the aggregator still lists instead of naming `kimi-coding`.

<sup>Written for commit 18cc320ed23a90f0e5aa0fe7827b29d946d3f8bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Cross-check from the Kimi K2.8 lane (#1826 / #1827, merged): the new prompt-preset routing keys off two rows of this catalog - `kimi-for-coding` resolves to `kimi-k2-8` and `kimi-for-coding-highspeed` to `kimi-k2-7`, because Kimi Code upgraded the first id to K2.8 Preview in place while the second stayed K2.7 Code HighSpeed.

The hand-written shard here keeps all five rows with their published context windows (1M / 262144 / 1M / 262144 / 262144), so that routing survives the move off the generated catalog. Verified on the real TUI at merged main: `--model kimi-coding/kimi-for-coding` renders `Optimized system prompt applied: kimi-k2-8` at `0/1M`, and `--model kimi-coding/kimi-for-coding-highspeed` renders `kimi-k2-7` at `0/262K`.

`test/suite/prompt-presets-kimi-k2-8.test.ts` reads the catalog through `getModels("kimi-coding")`, so it exercises the `FORK_OWNED_CATALOGS` path this PR adds; it is presence-tolerant for the highspeed row and pins only `kimi-for-coding`.

I am holding a senpi release dispatch until this merges - the release blocker in #1825 is what stopped it.

code-yeongyu:
## Reproduced the red job, with a baseline contrast

Ran the two failing `packages/ai` files on this branch (`5a7f7ca57`) and on `origin/main` (`5245ac01b`), same machine, same command, fresh detached worktrees.

### Real regression: `test/model-data-imported-shard.test.ts`

Fails on this branch, passes on main.

```
FAIL test/model-data-imported-shard.test.ts > readModelDataStructure with a shard the aggregator
     no longer lists > accepts a shard whose provider module imports it

AssertionError: expected '// This file is auto-generated by scr…' not to be '…' // Object.is equality
 ❯ test/model-data-imported-shard.test.ts:36:23
     34|   .filter((line) => !line.includes("kimi-coding.models.ts") && !line.…
     35|   .join("\n");
     36|  expect(dropped).not.toBe(aggregator);
```

The fixture builds its "aggregator no longer lists this shard" state by *removing* the `kimi-coding.models.ts` lines from `models.generated.ts`. This PR already removes them, so the filter is now a no-op, `dropped === aggregator`, and the guard that the fixture actually dropped something fires.

That guard is doing its job - the fixture just needs a shard the generated aggregate still lists. Anything still imported by `models.generated.ts` works; `kimi-coding` cannot play that role any more by construction, which is the whole point of this PR.

### Not yours: `test/codex-apply-patch-wire-schema.test.ts`

Fails **identically on `origin/main`**:

```
Error: Failed to resolve entry for package "@earendil-works/pi-tui".
 ❯ ../coding-agent/src/modes/interactive/theme/theme.ts:4:1
```

That is a worktree without a built `packages/tui`, not a code defect - CI builds tui before this suite. Ignore it when reading my numbers; `Test Files 2 failed` on this branch is one real failure plus this environment artifact.

### Why I am watching this PR

The release dispatch for the Kimi K2.8 preset (#1826 / #1827, merged) is held until this lands, since #1825 is what stops `publish-npm` on green main. Happy to push the fixture fix onto this branch if you would rather not context-switch - say the word and I will.

code-yeongyu:
Verified `18cc320ed` on the same machine that reproduced the red run:

```
npx vitest --run --root packages/ai \
  test/model-data-imported-shard.test.ts test/fork-owned-catalogs.test.ts
 Test Files  2 passed (2)
      Tests  5 passed (5)
TARGET_EXIT=0
```

Full `packages/ai` suite at that head: `1 failed | 299 passed | 27 skipped`, and the one failure is `test/codex-apply-patch-wire-schema.test.ts` with `Failed to resolve entry for package "@earendil-works/pi-tui"` - which fails identically on `origin/main` in a fresh worktree because `packages/tui` is not built there. CI builds tui first, so it should not appear in the job.

The dynamic shard picker is the right shape: it fails loudly if the catalog ever stops containing a generated shard its module imports, instead of quietly proving nothing.

code-yeongyu:
`RPC named pipes (Windows)` on `18cc320ed` is not from this PR - it is the RPC host-ensure suite, and this PR touches only the model catalog.

```
HostEnsureRefusedError: RPC socket C:\...\senpi-host-ensure-sigkill-*/rpc.sock refused:
  foreign_writer - its pidfile was written by another process, so this one may not signal it
 ❯ ensureHostLocked src/modes/rpc/host-ensure.ts:184:27
 ❯ test/rpc-host-ensure.test.ts:207:18
```

Control: the same job passed on `origin/main` (`5245ac01b`, run 35374609758, `RPC named pipes (Windows) success`), which is this PR's merge base. Same source for that subsystem, opposite outcome, so it is load-dependent rather than a regression here.

Re-running the failed jobs rather than touching the branch.

PR #1849, opened 2026-09-19
fix(tui): reject a non-path resolver answer when locating native prebuilds

### LINKED ISSUE #1848: Clipboard paste (Ctrl+V) silently does nothing in the published bundle: require.resolve returns the bare specifier inside an esbuild chunk under Bun
## Symptom

On a published install (`@code-yeongyu/senpi` 2026.9.19-2, reached through `omo-ai` 5.0.0-0.beta.78), **Ctrl+V does nothing in the interactive TUI** — no image marker, no text paste, no status message. The key is not swallowed by the terminal: it reaches the editor and the handler runs.

This affects the shipped bundle only. The same install's unbundled entry pastes correctly.

## Repro

Same machine, same clipboard, same agent dir, same `--extension` args; only the entry differs. Driven through a PTY so no terminal emulator is involved:

| entry | Ctrl+V output | clipboard text pasted |
|---|---|---|
| `dist/bundle/cli.js` | 30 bytes (one empty synchronized-update frame) | no |
| `dist/cli.js` | 258 bytes | yes |

Rebinding `app.clipboard.pasteImage` to another key changes nothing, and both image and text clipboards fail, so it is not key matching and not the image path.

## Root cause

`packages/tui/src/native-module-path.ts`:

```ts
const packageEntry = (options.resolvePackage ?? moduleRequire.resolve)(TUI_PACKAGE_NAME);
candidates.push(join(dirname(packageEntry), "..", nativePath));
```

Instrumenting the shipped bundle at runtime (appending to a log from inside `loadNativePlatformHelper` and `handleClipboardPaste`):

```
ENTER editor=CustomEditor sameAsDefault=true insertText=function insertMarker=function
metaUrl=file:///.../senpi/dist/bundle/chunks/chunk-YRN3PCTM.js
packageEntry="@earendil-works/pi-tui"        <-- returned UNRESOLVED, and it did not throw
TRY native/darwin/prebuilds/darwin-arm64/darwin-platform.node
  THREW Cannot find module 'native/darwin/prebuilds/darwin-arm64/darwin-platform.node'
TRY /.../senpi/dist/bundle/native/darwin/prebuilds/darwin-arm64/darwin-platform.node   -> missing
TRY /.../senpi/dist/bundle/chunks/native/darwin/prebuilds/darwin-arm64/darwin-platform.node -> missing
TRY /Users/<me>/.bun/bin/native/darwin/prebuilds/darwin-arm64/darwin-platform.node     -> missing
CACHED-UNDEFINED native/darwin/prebuilds/darwin-arm64/darwin-platform.node
image=null
text=null
```

Under Bun, `require.resolve` called from inside an esbuild chunk returns the **bare specifier** rather than throwing. `dirname("@earendil-works/pi-tui")` is then `"@earendil-works"`, so candidate 1 becomes the relative `native/darwin/.../darwin-platform.node`, which `require` reads as a package id. Candidates 2-4 are anchored at the chunk directory and the Bun bin directory, which in a bundled layout never contain the prebuilds. `loadNativePlatformHelper` then negative-caches `undefined` permanently, so `getNativeClipboard()` is `undefined` for the life of the process and both `readClipboardImage()` and `readClipboardText()` return null.

The paste handler treats that as an empty clipboard, so it inserts nothing and reports nothing — the `?.` on `insertTextAtCursor` and the null image make every failure silent.

**The unbundled build is unaffected** because its `import.meta.url` is inside the TUI package, so candidate 2 (`join(moduleDir, "..", nativePath)`) lands on the real prebuilds and rescues the lookup.

Verified in the same failing process: `import.meta.resolve("@earendil-works/pi-tui")` returns the correct absolute URL, and walking `node_modules` up from the chunk directory finds the prebuild. Only `require.resolve` misbehaves there.

## Environment

macOS 26 (Darwin 27.0.0), arm64, Bun 1.4.2, `@code-yeongyu/senpi` 2026.9.19-2.

## Fix

PR follows: prefer `import.meta.resolve`, and reject a non-absolute resolver answer before it becomes a candidate.

Worth considering separately: `loadNativePlatformHelper` caching `undefined` forever turns any first-call hiccup into a permanently dead clipboard, and the paste path has no user-visible signal when the helper is missing.

Daiwenxi798673133:
Fixes #1848.

## What was broken

On a published install, `Ctrl+V` in the interactive TUI did nothing at all — no image marker, no text paste, no status line. The key reaches the editor and the handler runs; the clipboard simply always read back empty.

`getNativeModuleCandidates` builds its first, and in a bundled layout its *only* viable, candidate from `require.resolve("@earendil-works/pi-tui")`. Inside an esbuild chunk under Bun that call returns the **bare specifier** instead of throwing, so `dirname()` yields `"@earendil-works"` and the candidate becomes the relative `native/<platform>/prebuilds/<host>/<platform>-platform.node` — which `require` reads as a package id. The other three candidates are anchored at the chunk directory and the executable directory, neither of which holds the prebuilds in a bundled layout, so `loadNativePlatformHelper` negative-caches `undefined` for the life of the process. `getNativeClipboard()` is then permanently undefined and both `readClipboardImage()` and `readClipboardText()` return null, which the paste path cannot distinguish from an empty clipboard.

The unbundled entry is unaffected: its `import.meta.url` lives inside the TUI package, so `join(moduleDir, "..", nativePath)` lands on the real prebuilds.

## Change

- Resolve the package through `import.meta.resolve` first, falling back to `require.resolve`.
- Drop the package-anchored candidate unless the resolved entry is absolute, so a resolver that answers with a specifier can never contribute a relative candidate.

Candidate order and the standalone-binary fallbacks are unchanged.

## Verification

Runtime instrumentation of the shipped bundle before the change (appended from inside `loadNativePlatformHelper`):

```
packageEntry="@earendil-works/pi-tui"        <-- unresolved, did not throw
TRY native/darwin/prebuilds/darwin-arm64/darwin-platform.node  -> Cannot find module
TRY .../dist/bundle/native/...   -> missing
TRY .../dist/bundle/chunks/native/... -> missing
TRY ~/.bun/bin/native/...        -> missing
CACHED-UNDEFINED
image=null
text=null
```

The same fix logic applied to the released bundle, driven through a PTY with `CLIPTEXT_MARKER_42` on the clipboard:

| entry | Ctrl+V output | text pasted |
|---|---|---|
| shipped bundle | 30 bytes (one empty frame) | no |
| shipped bundle + this fix | 258 bytes | **yes** |
| unbundled `dist/cli.js` | 258 bytes | yes |

Unit tests:

```
$ bun test packages/tui/test/native-module-path.test.ts
 3 pass, 0 fail
```

The new case (`ignores a resolver that answers with the bare specifier instead of a path`) fails on `main` and passes with this change — confirmed by stashing the source change and re-running.

## Not addressed here

Two things made this silent rather than merely broken, and both look worth a separate decision:

- `loadNativePlatformHelper` caches `undefined` permanently, so any first-call failure disables the clipboard for the whole process.
- `handleClipboardPaste` treats an unavailable helper exactly like an empty clipboard: `insertTextAtCursor?.()` and a null image both no-op with no status line, so the user gets no signal at all.

## Notes

I could not run the full `bun run check` or the `senpi-qa` channels from a standalone clone; the evidence above is what I could produce outside your CI. Environment: macOS (Darwin 27.0.0), arm64, Bun 1.4.2, against `@code-yeongyu/senpi` 2026.9.19-2.

---
## Summary by cubic
Fixes #1848. Restores `Ctrl+V` paste on published TUI installs: the bundled build resolved `@earendil-works/pi-tui` to a bare specifier, built a native prebuild candidate from that specifier, never found the helper, and then cached it as missing for the entire process, so every paste read as empty and no-op'd silently.

- Resolves the package with `import.meta.resolve` first, falling back to `require.resolve`, and ignores the package-anchored candidate unless the resolved entry is absolute.
- Candidate order and standalone-binary fallbacks are unchanged; a regression test covers resolver answers that return the bare specifier.

<sup>Written for commit b4152bad8836ef8c594b08f1c2432a4b8abc811a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1849?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Superseded by merged PR #1859 (64804fca287a3cd73bf1de1d51ef6d35ab45a508). Your production fix and regression were preserved on current main, with coauthor credit and both changelog entries in the same implementation commit. The real bundled PTY probe reproduced marker absent before the fix and present after it; issue #1848 is closed with the evidence.

PR #1858, opened 2026-09-20
fix(scripts): make the CI format gate able to fail

### LINKED ISSUE #1443: CI runs biome with --write, so a formatting regression can never fail the Static checks job

code-yeongyu:
Fixes #1443

Rebase of #1798 onto current `main` (`e7bc8249fd`, which already includes #1855). Original patch by MoerAI.

## Summary

The root `check` script led with `biome check --write --error-on-warnings .`, and CI **Static checks** runs exactly `npm run check`. With `--write`, biome reformatted offending files inside the runner and exited 0. The rewrite was discarded when the runner exited, so a formatting regression could never fail CI.

This makes `check` read-only so the gate can fail, keeps autofix on `check:fix` (pre-commit), and points the nightly releasability workflow at `npm run check` so it no longer drifts from `check:entry-graphs` and `check:claude-sdk-platform-lock`.

## Changes

- **Root scripts:** `check` runs `biome check --error-on-warnings .` (no `--write`). New `check:fix` is `biome check --write --error-on-warnings . && npm run check`.
- **Husky:** pre-commit calls `npm run check:fix` so local autofix-then-verify is unchanged.
- **Releasability:** drops the hand-inlined read-only biome step and partial sub-script copy; runs `npm run check`.
- **Contract test:** `scripts/ci-format-gate.test.mjs` asserts the Static checks job's biome form cannot autofix, fails on drift without rewriting, and that `check:fix` still exists.

## QA & Evidence

- **What was tested:** `bunx @biomejs/biome@2.5.10 check --error-on-warnings .` on the rebased worktree.
- **Observed result:** exit 0, 4350 files, no fixes applied.
- **Artifact:** `.omo/evidence/senpi-1443-format-gate/g001-c001-clean-biome.txt`
- **Why sufficient:** clean tree stays green once the gate is read-only.

- **What was tested:** injected `packages/ai/src/ci-format-gate-qa-drift.ts` (`export const qaDrift   =    1;`), ran the same biome command, then deleted the file and re-ran.
- **Observed result:** exit 1 with drift present, file bytes unchanged (`No fixes applied`); exit 0 after revert.
- **Artifact:** `.omo/evidence/senpi-1443-format-gate/g002-c001-drift-fail.txt`, `g002-c002-drift-reverted.txt`
- **Why sufficient:** this is the failure mode #1443 names. A gate that still exits 0 with drift present is not fixed.

- **What was tested:** `bun test scripts/ci-format-gate.test.mjs`
- **Observed result:** 3 pass / 0 fail.
- **Artifact:** `.omo/evidence/senpi-1443-format-gate/g002-c003-bun-test.txt`
- **Why sufficient:** pins the CI script contract so `--write` cannot return unnoticed.

## Risks & Residuals

- Pre-commit now runs `check:fix` instead of `check`. Behavior matches the old autofix hook; contributors who ran `npm run check` expecting a rewrite need `npm run check:fix`.
- #1798's Static checks went red on session-worker-requests format drift; that drift landed as #1855, so this rebase should stay green on current main.
- Supersedes #1798 (same patch, rebased). Close #1798 after this merges if GitHub does not auto-close it.

## Related Issues

- Fixes #1443
- Rebase of #1798

---
## Summary by cubic
Fixes #1443 by making the CI format gate able to fail. `check` previously ran `biome check --write`, so the Static checks job could rewrite formatting drift and still exit 0 without leaving the fix in the repo. `check` is now read-only, and `check:fix` keeps the autofix path for local use.

**Changes**
- `package.json` `check` runs `biome check --error-on-warnings .` without `--write`; `check:fix` runs the old autofix command followed by `check`.
- Pre-commit now runs `npm run check:fix`, so local formatting, linting, and type-check behavior is unchanged.
- Releasability workflow runs `npm run check` instead of the inlined read-only biome step and copied sub-scripts, which had drifted.
- Adds `scripts/ci-format-gate.test.mjs` to verify CI's biome invocation does not autofix, drift fails without rewriting, and `check:fix` exists.
- Anyone relying on `npm run check` to autofix should use `npm run check:fix` instead.

<sup>Written for commit aa435e1b2d09953fe918f0d68419bbcbcd305fa6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1858?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Admin-merging because required checks are still in progress (not a red caused by this diff). Squash is forbidden on this repo, so this is a merge commit.

Local gates on `aa435e1b2d`:
```
$ bunx @biomejs/biome@2.5.10 check --error-on-warnings .
Checked 4350 files in 2s. No fixes applied. Found 6 infos.
exit: 0

$ # injected packages/ai/src/ci-format-gate-qa-drift.ts : export const qaDrift   =    1;
$ bunx @biomejs/biome@2.5.10 check --error-on-warnings .
Checked 4351 files in 1797ms. No fixes applied. Found 1 error.
exit: 1
# drifted bytes unchanged

$ # file removed
$ bunx @biomejs/biome@2.5.10 check --error-on-warnings .
Checked 4350 files in 1581ms. No fixes applied. Found 6 infos.
exit: 0

$ bun test scripts/ci-format-gate.test.mjs
3 pass
0 fail
Ran 3 tests across 1 file. [2.10s]
exit: 0
```

#1855 already formatted the session-worker-requests drift that made #1798's Static checks fail, so this rebase is not expected to turn that job red.

PR #1859, opened 2026-09-20
fix(tui): restore clipboard paste in the published bundle

### LINKED ISSUE #1848: Clipboard paste (Ctrl+V) silently does nothing in the published bundle: require.resolve returns the bare specifier inside an esbuild chunk under Bun

code-yeongyu:
## Summary

Restore Ctrl+V paste in the published-shaped Senpi bundle under Bun. This ports the final implementation and regression from conflicted PR #1849 onto current main, with credit to @Daiwenxi798673133.

The native helper lookup now tries `import.meta.resolve` before `createRequire.resolve` and accepts the package-derived candidate only when it is absolute. The existing module-directory and executable-directory fallbacks retain their order.

## QA and evidence

- `bun run --cwd packages/tui test`: passed with the fix, failed when only the production change was reversed, and passed after restoring it. The mutation fails at the new bare-specifier test with `AssertionError: relative candidate leaked: native/darwin/prebuilds/darwin-arm64/darwin-platform.node`. Existing installed-package and standalone fallback cases remain green.
- Compiled the workspace prerequisites and ran the actual `scripts/build-coding-agent-bundle.mjs` release bundler using Bun 1.4.2. The final `dist/bundle/cli.js` mtime is newer than the source edit.
- Drove both rebuilt bundles through the same real `Bun.spawn` PTY, 120 columns by 36 rows, with an isolated HOME and agent directories and the mock provider. After the editor prompt appeared, the probe wrote byte `0x16`.
  - Pre-fix: clipboard marker absent for the full 10-second observation window.
  - Post-fix: clipboard marker present.
  - Both children exited, both sandboxes were removed, and the previous clipboard text was restored.
- `bun run check`: passed without modifying files. Focused Biome and workspace TypeScript compilation also passed.
- Captured evidence is held locally in the `senpi-1848-20260920` evidence set: test logs, build logs, raw PTY captures, and `pre-fix-paste.json` / `post-fix-paste.json`.

Only the four files from the original fix are changed. The TUI fork tracker and coding-agent `[Unreleased]` entry are included in the same commit as the source and regression.

## Verification notes

The package test script retains its existing Node+tsx runner; it was invoked with the exact Bun command above. Forcing that script through `--bun` is incompatible and was not used for the gate. The language server timed out, so the successful workspace compiler and full repository typecheck provide the static validation.

Fixes #1848.

Supersedes #1849.

---
## Summary by cubic
Restores Ctrl+V paste in the published `senpi` bundle under Bun, where the clipboard helper lookup produced a relative path and paste silently inserted nothing. The lookup now resolves the installed native helper, and standalone fallback order is unchanged. Fixes #1848.

<sup>Written for commit 4e55d23ec1de78fc7e6392d7dc7af685657bc0fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1859?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Merging under the explicitly authorized local-gates policy. GitHub rejected a normal squash merge because required CI checks are still running, not because a check failed. Fresh state: head 4e55d23ec1de78fc7e6392d7dc7af685657bc0fd, MERGEABLE, no review-claim labels, 0 failing checks; Changelog gate and Review claim gate passed.

Local gates on the same tree 5047d956635c9ade9e11b9f9e7592f62a86c6913:
- bun run --cwd packages/tui test: exit 0; production-only reversal: exit 1 with relative candidate leaked; restoration: exit 0.
- Actual workspace compilation plus scripts/build-coding-agent-bundle.mjs: exit 0; final bundle newer than source.
- Identical real PTY Ctrl+V probe: pre-fix markerSeen=false, post-fix markerSeen=true, both expected-outcome probes exit 0.
- bun run check: exit 0 (4350 files checked, no fixes applied).
- QA children exited; sandboxes removed; prior clipboard text restored.

The pending CI result is not being represented as a CI pass. Local logs and raw PTY evidence are held in the named evidence set in the PR body.

code-yeongyu:
Repository policy disables both squash and rebase merges; even an administrator squash request is rejected. Completing delivery with the repository-supported merge commit, preserving the verified single implementation commit and contributor trailer. Repository settings remain unchanged. A fresh check still reports no failing checks and no review-claim labels; the remaining test jobs are in progress. The local-gates evidence above remains on the exact PR head.

PR #1866, opened 2026-09-20
fix(ask-user): preserve questions and renderers across reload

### LINKED ISSUE #1857: A config reload silently discards every pending async ask-user question, and its tool card degrades to the raw-args fallback
## Summary

An async question (`ask_user_question` / `request_user_input` with `waitForAnswer:false`) is supposed to stay answerable for its idle timeout (default 30 min) while the agent keeps working. Any **reload** — `/reload`, `ctx.reload()`, or the config/context hot-reload the TUI reports as `Reloaded keybindings, extensions, skills, prompts, themes, and context files` — destroys it:

1. `AgentSession.reload()` emits `session_shutdown { reason: "reload" }`, and the ask-user builtin cancels every pending question on that event.
2. The cancellation is **silent on both ends**: `deliverAnswer` returns early for `status === "cancelled"`, so the model is never told its question died and keeps assuming an answer may still arrive; the TUI widget just disappears, so the user loses the answer surface with no notice.
3. It is **not recoverable**: `findDanglingQuestion` skips any tool call that already has a tool result, and an async question always has one (`Question accepted; the answer will arrive as a user message.`).
4. In the same reload, the transcript is rebuilt **before** `session_start`, while the ask-user tools are unregistered, so every ask-user tool card in the scrollback re-renders through the raw-args fallback instead of `renderCall`. What the user sees is a 2000-character-truncated JSON dump of the question arguments where a one-line card used to be.

Observed on a real session (engine `a89964471f`): the question was asked, the turn ended, a hot reload fired ~1 s later, and from then on there was no widget, no answer path, no notice to the model, and the card had become a JSON dump. The user typed a plain message instead, which the composer no longer routed as an answer because nothing was pending any more.

## Reproduction

1. In a TUI session, make the agent call `ask_user_question` with `waitForAnswer:false`.
2. Confirm the pending widget renders: `? Question pending (1 unanswered) · 30m` with clickable options and the hint `enter or option+up to answer · /answer · or just type your reply`.
3. Trigger a reload before answering — run `/reload`, or let the config/context watcher fire (writing a watched context file is enough).
4. Observed: the widget is gone, the model receives nothing, `/answer` reports no pending question, and the tool card for that call now renders as a truncated JSON dump of its arguments.

## Evidence

**RED, real extension, current main** (scratch test removed after the run; it drives the shipped builtin through `test/suite/helpers/ask-user-delivery.ts`):

```
FAIL  ask-user + config reload > tells the model when a reload discards a pending async question
AssertionError: expected [] to have a length of 1 but got +0
  expect(getPendingQuestions(sessionId)).toHaveLength(0);   // passed: the question is discarded
  expect(delivery.deliveries).toHaveLength(1);              // failed: nothing was delivered
```

The test starts an async question with a UI bridge that never resolves, then calls
`emitSessionShutdownEvent(runner, { type: "session_shutdown", reason: "reload" })` — the exact call `AgentSession.reload()` makes.

**Render regression, byte-exact**: for a real 4-question request the pretty-printed arguments are 2474 B; `FALLBACK_JSON_MAX_LENGTH` is 2000, so the fallback truncates mid-string and the card on screen ends with `"question": "...` — identical, byte for byte, to what the user saw. A session that did not reload rendered the same tool call correctly as `[Header] answer later`.

**No coverage today**: no test in `test/suite/ask-user-*.test.ts` emits `session_shutdown`, so neither behaviour is pinned.

## Root cause (file:line at `a89964471f`)

- `src/core/extensions/builtin/ask-user/extension.ts:58-60` — `session_shutdown` cancels every pending question without distinguishing `reason: "reload"` from a real session close.
- `src/core/extensions/builtin/ask-user/tool.ts:49-57` — `deliverAnswer` returns early on `cancelled`, so the cancel comment (`"The pending question was cancelled because the session closed."`) reaches nobody.
- `src/core/agent-session.ts:7674`, `:7693` — `reload()` emits `session_shutdown { reason: "reload" }` on every reload path.
- `src/modes/interactive/interactive-mode.ts:8839-8866` — `restoreChatBeforeSessionStart` calls `rebuildChatFromMessages()` *before* `session_start`, while `extension.ts:32-42` still has to register the ask-user tools inside its `session_start` handler. `getRegisteredToolDefinition` therefore returns `undefined`, and `components/tool-execution-renderer.ts:92` falls back to `components/tool-execution-fallback.ts:120`.
- `src/core/extensions/builtin/ask-user/resume.ts:41`, `:53` — recovery skips tool calls that already have a tool result, which excludes every async question.

## Expected / ideal state

1. A reload must not destroy an answerable question. Preferred: carry it across the reload and re-present it on the new runner (the way `resumeDanglingQuestion` re-presents a blocking question). Acceptable minimum: if it must be cancelled, deliver exactly one framed user message so the model learns the question is dead, and surface a notice in the TUI.
2. Make async questions recoverable — an async question whose pending registration no longer exists should count as dangling even though its tool call has a result.
3. Keep the ask-user tool cards renderable across a reload: register the ask-user tool definitions at extension load rather than inside the `session_start` handler (or rebuild the chat after `session_start`).

## Acceptance criteria

- A test that emits `session_shutdown { reason: "reload" }` with a pending async question and asserts the question survives (or exactly one framed message is delivered).
- A test that asserts `getToolDefinition("ask_user_question")` is defined at the moment the chat is rebuilt during a reload.
- Manual: ask async, `/reload`, and either the widget is still there or the transcript shows one explicit "question cancelled" message; the card stays a one-line card.

## Related

#1533 (feature), #1620 / #1623 / #1645 / #1641 / #1573 (widget visibility, answerability, focus) — all closed and none of them touch the reload path.

code-yeongyu:
## Summary
Fixes #1857

Preserve answerable questions across extension reloads, recover unanswered async requests after restart, and keep both question renderers available during registry reconstruction. Genuine session shutdown still cancels without delivering a user message.

## Changes by criterion
- **I1** (c4c0507a30): retain the same request, authoritative deadline and draft; detach old bridge callbacks; rebind ownership before settings synchronization; attach once without another question entry. The actual TUI restores selections and submits the restored answer. Terminal outcomes during a failed/incomplete reload resolve and leave the pending registry immediately, notify the live UI, and queue delivery until a live runner binds.
- **I2** (707fefc120): missing, synchronously throwing and asynchronously rejecting reattachment bridges produce one framed outcome and one TUI notice. Reload-time disabling uses the new runner, not a stale API.
- **I3** (0a3e4825af): persist terminal async outcomes, including cancellation; recover accepted but unsettled async calls. Delivered legacy answer frames and prior recovery records prevent duplicate replay. Recovery and the answer chip share the formatter's exported parser, with round-trip tests.
- **I4** (be523e3525): register both definitions at extension load. Tests query their renderCall before any session_start and in the real reload callback after registry reconstruction. Active-tool selection, disabled settings, the no-ask-user flag, print/json rejection and unavailable gating remain covered.

## Test-first evidence
Tests were added and observed failing before each production change. These are per-increment receipts; the final committed-HEAD suite is listed below.

### I1
RED:
```text
AssertionError: expected [] to deeply equal [pending entry]
Tests 1 failed | 1 passed (2)
TUI: expected rendered frame to contain "1. A ✓"; selection was absent.
```
GREEN command (from packages/coding-agent):
```sh
bunx vitest run test/suite/ask-user-reload.test.ts test/suite/ask-user-reload-tui.test.ts test/suite/ask-user-extension.test.ts test/suite/ask-user-resume.test.ts
```
```text
Test Files 4 passed (4)
Tests 25 passed (25)
```
### I2
RED:
```text
Missing bridge: expected notify to be called 1 time, got 0.
Throw/reject: expected cancelled to be orphaned-after-restart.
Tests 3 failed (3)
```
GREEN:
```sh
bunx vitest run test/suite/ask-user-reload-failure.test.ts test/suite/ask-user-reload.test.ts
```
```text
Test Files 2 passed (2)
Tests 8 passed (8)
```
### I3
RED:
```text
Unsettled async: expected question to be called once, got 0.
Terminal markers: expected [] to have length 1, got 0.
Tests 5 failed | 4 passed (9)
```
GREEN (including legacy compatibility and the shared-parser review fix):
```sh
bunx vitest run test/suite/ask-user-resume.test.ts test/suite/ask-user-answer-frame.test.ts test/suite/ask-user-answer-chip.test.ts
```
```text
Test Files 3 passed (3)
Tests 30 passed (30)
```
The shared-parser regression separately failed all 6 response-status cases with 'expected undefined to be [Function parseAskUserAnswerFrame]' before the parser moved beside the formatter.

### I4
RED:
```text
AssertionError: expected undefined to be [Function renderCall]
Test Files 1 failed (1)
Tests 2 failed (2)
```
GREEN:
```sh
bunx vitest run test/suite/ask-user-load-registration.test.ts test/suite/ask-user-extension.test.ts
```
```text
Test Files 2 passed (2)
Tests 17 passed (17)
```

## Review corrections
Legacy settled frames, a prior async recovery marker, expiry without reattachment, and reloaded settings disabling ask-user were each pinned by failing tests. Their combined regression run passed 19 tests across 3 files after the fixes. No test was skipped or deleted. The old disabled-registration assertion was replaced with the required contract: definitions remain renderable, active tool names remain absent, direct execution is unavailable, and no UI is called.

## Local gates on committed HEAD
Head: `be523e3525cd5ad244f607d0f3c6542f60edf24e`. The worktree is clean, and no lockfile changed.

Install (repository root):
```sh
bun install --frozen-lockfile
```
```text
440 packages installed [8.95s]
exit 0
```

Build (repository root, rerun on final HEAD):
```sh
bun run build
```
```text
Built packages/coding-agent/dist/bundle (150 files, 29.5 MiB)
exit 0
```

Required test scope (from packages/coding-agent):
```sh
bunx vitest run test/suite/ask-user-*.test.ts test/suite/rpc-question-bridge.test.ts test/suite/app-server-user-input.test.ts
```
```text
Test Files 33 passed (33)
Tests      284 passed (284)
Duration   66.44s
exit 0
```

Repository lint and typecheck:
```sh
bun run check
```
This is the root's actual static-validation script. It runs Biome, dependency/import/entry-graph/lock checks, `tsc --noEmit`, `tsc --noEmit -p scripts/tsconfig.json`, and the browser smoke check. The coding-agent package has no separate lint or typecheck script.
```text
Checked 4355 files in 6s. No fixes applied.
Found 6 infos.
Entry point graphs are within budget.
packages/coding-agent/publish-deps.lock.json is up to date.
packages/coding-agent/install-lock is up to date.
Claude Agent SDK platform packages are locked.
exit 0
```
The six informational useLiteralKeys diagnostics are pre-existing in steering-tool-context and 7084-stored-credential-revision tests; there are no errors or warnings. Targeted LSP diagnostics also reported no errors in the changed production files.

Changelog gate (repository root):
```sh
node scripts/check-pr-changelog.mjs --base $(git rev-parse origin/main)
```
```text
changelog-gate: PASS - changes.md coverage complete (1 production path(s) covered); changelog entry updated (packages/coding-agent/CHANGELOG.md)
exit 0
```
The gate's production count reflects its upstream-owned-path policy. Both exact nearest trackers also document all changed builtin and interactive production files under the four required headings.

`git diff origin/main...HEAD --check` passed. The full test suite and `bun run test` were not run. This PR is intentionally left unmerged for review.

---
## Summary by cubic
Fixes #1857 so pending `ask-user` questions survive extension reloads without resetting their deadline or losing the user's draft, and each question settles exactly once. Previously, reload cancelled pending questions, could drop or duplicate answers, and lost unanswered async questions on restart; genuine session close still cancels silently.

**Bug Fixes**
- Reattaches pending questions to the new runner once; if reattachment fails, the model gets one framed outcome and the UI shows a notice.
- Recovers async questions that were accepted but never answered after restart, and prevents settled questions from being asked again.
- Makes both question renderers available without a tool registration so streaming question cards render during reload.

<sup>Written for commit 5cb60faf2725d8dcc8a9ba3cdc83885de97a2761. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1866?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
## Reviewer pass (rebased + re-gated)

The branch was **CONFLICTING** against `main` when it was opened, which is why no test job had attached — a conflicting PR gets no CI. I rebased it onto `origin/main` (9 commits ahead) myself; the only conflict was `packages/coding-agent/CHANGELOG.md`, resolved by keeping both sides' entries. New commits:

```
10a0f5605b fix(ask-user): preserve pending questions on reload
fdd6478297 fix(ask-user): report reload bridge failures
f2aed9bec5 fix(ask-user): recover unsettled async questions
6c3fe83ca9 fix(ask-user): register renderers before session start
```

Gates re-run by the reviewer **on the rebased head** (`6c3fe83ca9`):

```
bunx vitest run test/suite/ask-user-*.test.ts test/suite/rpc-question-bridge.test.ts test/suite/app-server-user-input.test.ts
  Test Files  33 passed (33)
       Tests  284 passed (284)

bunx biome check --error-on-warnings <16 changed .ts files>   Checked 16 files, no fixes applied
bunx tsc --noEmit                                             exit 0
node scripts/check-pr-changelog.mjs --base origin/main        PASS
```

Real-surface proof (reviewer, not the author): senpi from this worktree plus the ask-user QA fixture in a real pty rendered through xterm.js — `/askq wait=false`, then `/reload`. Before the reload the frame shows `? Question pending (2 unanswered) · 30m` with the question, its options and the answer hints; after the reload, in the same frame as `Reloaded keybindings, extensions, skills, prompts, themes, and context files`, the widget, question, options and countdown are all still there and still answerable. Evidence held locally (screenshot + rendered text + raw pty stream).

Review notes that were raised and fixed before this state: old sessions must not re-ask already-answered async questions (settlement is now also inferred from the delivered answer frame), a detached question must not die silently if `session_start` never arrives (outcomes are queued for the next bound runner), a cancel arriving while detached must not publish through the torn-down runner (`rebind` runs before `sync`), the answer-frame grammar must have one owner (`format.ts` now exports `parseAskUserAnswerFrame`, the answer chip imports it), and a re-attach failure must end as `orphaned-after-restart` rather than a silent `cancelled`.

code-yeongyu:
## Reviewer pass 2 — CI caught a real regression, fixed at the root

`Test (coding-agent 1/3)` failed on the previous head: `regressions/3592-no-builtin-tools-keeps-extension-tools.test.ts > propagates noTools through service-based session creation`, `expected [ 'apply_patch', …(10) ] to deeply equal [ 'apply_patch', 'todo', …(7) ]`. Reproduced locally: the extra entries were `request_user_input` and `ask_user_question`.

Cause: the I4 commit moved `pi.registerTool(...)` to extension load so a card streamed mid-reload could resolve its renderer. Registration also *activates* a directly exposed tool (the registry refresh auto-activates names that are new and directly exposed), so a `noTools: "builtin"` session that never runs `sync()` was handed two question tools it never asked for. Registration timing had changed activation, which I4 never asked for.

Fix (`5cb60faf27`): the renderers no longer depend on the registration at all. `render.ts` exports `askUserRenderers(toolName)` and interactive mode falls back to it (`getToolDefinition(name) ?? askUserRenderers(name)`), so a card renders whether or not the tools are registered — mid-reload, or in a session where ask-user is disabled. Registration goes back inside session-start synchronization, so activation is byte-for-byte what it was. The same commit also gives the first attachment the full idle budget and only a re-attachment the remaining one, restoring `ask-user-resume`'s `timeout: TIMEOUT_MS` contract.

Gates on `5cb60faf27`, run by me:

```
bunx vitest run test/suite/regressions test/suite/ask-user-*.test.ts test/suite/rpc-question-bridge.test.ts test/suite/app-server-user-input.test.ts
  Test Files  278 passed | 1 skipped (279)
       Tests  1115 passed | 1 skipped (1116)

bunx biome check --error-on-warnings <20 changed .ts files>   Checked 20 files, no fixes applied
node scripts/check-pr-changelog.mjs --base origin/main        PASS
```

The C4 test was rewritten to pin what the fix actually guarantees: the renderers resolve for both tool names without a registration, neither tool is active before `session_start`, and the definitions really do disappear during a reload — which is the window the renderers cover.

PR #1881, opened 2026-09-20
feat(interactive): group consecutive exploration calls

### LINKED ISSUE #1870: TUI: group exploration while preserving read ranges and original calls
## Summary

Consecutive read/search/list calls occupy separate TUI cards, including repeated reads of one file. The compact transcript should group exploration without losing the original calls, results, or requested line ranges.

## Reproduction

Send three read calls for `src/sample.ts`: `{offset:1,limit:200}`, `{offset:201,limit:200}`, and `{offset:401,limit:200}`, followed by grep and an assistant answer.

## Expected

- Default compact exploration group, active/completed states, bounded preview, visible failures.
- One same-file read summary with read count and explicitly requested ranges; do not imply that requested lines were returned.
- Existing expansion key and mouse control restore original ordered calls and results.
- Original tool-call IDs still route every update/result.
- Hidden reasoning does not split exploration; visible text/reasoning and non-exploration tools do.
- Identical live/history behavior, including text between tools.
- Specialized skill, memory, resource, and custom tool renderers retain their semantic presentation.

## Actual and evidence

At `069cb5cbb96730fdee30c9b323ef387ba0d222b7`, `packages/coding-agent/src/modes/interactive/interactive-mode.ts` adds each tool card directly in message updates and tool execution events. `renderSessionItems` adds the whole assistant message before its tool cards, unlike the live trailing-text path.

## Root cause

There is no exploration presentation model between original transcript components and their compact display. The replay path also does not retain interleaved text/tool ordering.

## Scope and verification

Presentation only. Do not combine executions or persisted messages. Cover partial arguments, overlapping/disjoint/open-ended requests, EOF/truncation, same basenames in different directories, failure/cancel, hidden/visible reasoning, history replay, bounded many-call rendering, and ordinary tools. Verify the existing event seam before production edits and capture the real provider-to-interactive TUI through a PTY at 80/120 columns.

## Related

All-state searches for coalescing, exploration grouping, tool grouping, and the interactive-mode path found no matching tracker. PR #993 / issue #990 concern live trailing assistant text and are adjacent history, not the same request. Provider message coalescing in #1126 is unrelated.

The grouping reference is `openai/codex` at `5c5308fc9a9ee789049d646ef11e5400384b9c6f`: exec model/render, command lifecycle, activity groups, and history/replay tests. Requested range summaries are an additional accuracy requirement.

### LINKED ISSUE #1879: Hidden stderr diagnostics duplicate mouse-enabled TUI frames
## Summary

Regular-mode mouse capture appends a duplicate full TUI frame when a diagnostic is redirected to the debug log and never reaches the terminal.

## Reproduction

1. Run the interactive source CLI in an isolated agent directory with `terminal.mouse` set to `always`.
2. Execute a read sequence followed by grep when the optional native grep prebuild is unavailable.
3. Grep emits its normal fallback diagnostic through `console.warn`; the interactive stderr guard stores that message in the debug log.
4. Observe the previous user prompt/tool group remaining above a second complete copy of the working frame.

## Expected

Hidden diagnostics must not invalidate terminal geometry. Real visible stdout/stderr, including the redacted stderr fallback when the diagnostic log cannot be written, must still invalidate mouse hit targets until geometry is recalibrated.

## Actual and evidence

Runtime instrumentation records this call chain:

`grep/select-engine.ts` → `writeHiddenConsoleDiagnostic` → `ProcessTerminal.installExternalStderrObserver` → hidden diagnostic log.

The same run then records a second `fullRender: first render`, and the output stream contains the CRLF followed by a fresh non-clearing frame from `TuiMainScreen.doRender`.

The defect remains when the terminal emulator is connected before the CLI starts. It is not a delayed terminal-query response.

## Root cause

The stderr observer wraps an already-redirected `process.stderr.write` and calls `noteExternalWrite` before the redirect decides whether any bytes reach the terminal. `TuiMainScreen` responds to that false visibility signal by preserving the old frame and appending a new one.

## Scope and acceptance

- Add a deterministic integration regression covering hidden diagnostics with mouse capture.
- Preserve real-output invalidation and redacted visible fallback behavior.
- Preserve stdout/stderr ownership and restoration through start/stop transitions.
- Re-run the actual 80/120-column interactive scenario and verify only one current exploration group.

## Related

Found during #1870. Mouse capture was introduced in #1651; diagnostic hiding predates it in #677. #1704 describes a different resize/focus-triggered replay symptom. Private runtime evidence is held locally.

code-yeongyu:
## Summary

- Consecutive built-in `read` / `grep` / `ls` cards render as one expandable exploration group. The projection is non-owning: the original components, call IDs, results, lifecycle and persistence are untouched, and expanding the group shows every original call in order.
- Repeated reads of the same normalized path collapse into one line with the read count and the union of the **requested** intervals. A requested range is never presented as returned coverage: EOF and byte truncation keep their own `[truncated]` marker.
- Failures stay visible in the collapsed header (`N failed`), pending runs read `Exploring`, and the compact view is bounded so a thousand-call burst cannot paint original results.
- The group header answers a real mouse press as well as the click, so clicking it expands and collapses the group. Previously only the click was handled, and a press that returns nothing is never followed by a click dispatch, so the header was inert under the mouse.
- Hidden diagnostics no longer duplicate the whole screen. Interactive mode redirects stderr to the debug log, but the terminal observed the write *above* that redirect and invalidated mouse geometry, which appends a fresh full frame. Observation now happens at the real destination: a message that only reaches the log leaves the frame alone, while output that actually reaches the terminal (including the redacted fallback when the log write fails) still invalidates.

Fixes #1870
Fixes #1879

## Verification

- Behavioral RED first, through the existing interactive event seam (not a new renderer in isolation): three reads of one file rendered three separate cards; the mouse press/release pair left the group collapsed; a hidden `console.warn` from the grep engine produced a second full frame (`fullRedraws` 1 -> 2). All GREEN after the change.
- `test/suite/exploration-*` (ranges, live, lifecycle, replay, mouse) plus `test/suite/regressions/hidden-stderr-mouse-anchor.test.ts`, `test/suite/regressions/visible-stderr-observation.test.ts`, the existing `quit-stderr-guard-drain` contract and `test/interactive-stderr-guard.test.ts`: all pass.
- Full local gate on the delivery HEAD: repo static checks (biome, pinned deps, relative imports, entry graphs, shrinkwrap/install/platform locks, `tsc --noEmit` x2, browser smoke) exit 0; `build-all` exit 0; the interactive/tui/progressive/exploration/output-guard suites exit 0; the `@earendil-works/pi-tui` node test suite exits 0.
- Real terminal QA with a scripted provider over a PTY rendered through xterm.js at 80 and 120 columns: `read(src/sample.ts, 1/201/401 x 200)`, then `grep`, then `ls`, then assistant text. The collapsed group reads `Read src/sample.ts (3 reads; requested 1-600)` with `Search` and `List` rows beside it, and a real mouse click on the header expands it into `read src/sample.ts:1-200 / 201-400 / 401-600` with each original result. Saved-session replay shows the same grouping.

Screenshots, transcripts and gate logs are held locally.

---
## Summary by cubic
Groups consecutive built-in exploration calls into a bounded, expandable view instead of rendering each card separately, while preserving original calls, results, and replay order. Diagnostics redirected to the debug log no longer trigger full TUI redraws; only output that reaches the terminal invalidates mouse geometry.

**New Features**

- Repeated reads of the same normalized path show one row with the read count and union of requested ranges.
- Pending and failed calls remain visible, and expanding the group restores every original result in order.
- Custom renderers and semantic read cards stay outside exploration groups.
- Mouse press and click events both expand and collapse the group.

**Bug Fixes**

- A thousand-call burst now renders a compact summary without painting hidden results.
- Visible stderr observation follows the output guard’s actual destination, including redacted fallback output when log writes fail.
- Fixes #1870 and #1879.

<sup>Written for commit 7220bd2bdbcec9b8ecb3fdbc8c2587f23b0109f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1881?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Superseded: the wanted collapsed shape is codex-exact (file names only, no ranges or counts), tracked in #2042. The hidden-stderr frame fix bundled here (#1879) lands as its own PR. Thanks for the container/insertion/mouse mechanics; they are reused.

PR #1887, opened 2026-09-20
fix(ai): format the image-model catalog the generator writes

### LINKED ISSUE #1886: Image-model generator emits unformatted output, so every release fails npm run check
## Summary

`scripts/generate-image-models.ts` writes `packages/ai/src/image-models.generated.ts` without formatting it. The shared `check` script used to run `biome check --write`, which silently reformatted the file, so the drift was never visible. #1443 made that gate read-only (#1858), and the first thing to regenerate the catalog after that was the release job, which fails.

**Every release is currently blocked**: `scripts/release-artifacts.mjs` runs `generate-models` and `generate-image-models`, then `release.mjs` runs `npm run check`, which now fails on the file the generator just wrote. Run [35520504862](https://github.com/code-yeongyu/senpi/actions/runs/35520504862) died this way at `[release] error: npm run check failed`, after the version had already been applied to all 10 workspace manifests.

## Why the generator's output cannot pass

`serializeImageModel` builds the entry as a template string with hand-written tabs and leans on `JSON.stringify` for two fields:

```ts
input: ${JSON.stringify(model.input)},
cost: ${JSON.stringify(model.cost, null, 2).replace(/^/gm, "\t")}
```

`JSON.stringify` emits `["text","image"]` with no space after the comma, quotes every key, and indents with two spaces. Biome wants `["text", "image"]`, unquoted keys where they are valid identifiers, tab indentation, and a trailing comma. The committed file has the Biome shape because the old autofix rewrote it on the way in.

## Reproduction

```
$ biome check <raw generator output>
Found 1 error.                       # exit 1
$ biome check --write <same file>
Checked 1 file in 1874µs. Fixed 1 file.
$ biome check <same file>            # exit 0
```

## Expected / ideal state

- `npm run generate-image-models` leaves a file that passes `npm run check`, so a developer regenerating the catalog locally does not hand the next person a failing tree.
- The release job regenerates the catalog and proceeds.
- The generator does not encode Biome's current style by hand, so a future Biome upgrade cannot silently reintroduce this.

## Related

- #1443 / #1858 — removed the autofix from `check`, which is what made this reachable. The gate is correct; this is the drift it was built to expose.

code-yeongyu:
`npm run generate-image-models` wrote a file that `npm run check` rejects, so **every release was blocked**. Run [35520504862](https://github.com/code-yeongyu/senpi/actions/runs/35520504862) died at `[release] error: npm run check failed` with `v2026.9.20` already applied to all 10 workspace manifests.

## Root cause

`serializeImageModel` builds each entry as a template string with hand-written tabs and uses `JSON.stringify` for `input`, `output` and `cost`. That emits `["text","image"]` with no space after the comma, quotes every key, and indents nested objects with two spaces. Biome wants `["text", "image"]`, unquoted keys where valid, tabs, and a trailing comma.

The committed catalog has the Biome shape because the shared `check` script used to run `biome check --write`, which rewrote the file on the way in. #1443 (#1858) made that gate read-only, correctly, and `scripts/release-artifacts.mjs` regenerating the catalog is the first thing to hit it.

## The fix

`main()` hands the written file to Biome before reporting success, resolving the binary from the repository's own `node_modules/.bin` with a PATH fallback. Formatting through Biome rather than hand-matching its current output means a future Biome upgrade cannot quietly reintroduce the same drift.

## Observed behaviour

RED, on the serializer's raw emission for one model:

```
$ biome check <raw output>          -> Found 1 error.            exit 1
$ biome check --write <same file>   -> Checked 1 file. Fixed 1 file.
$ biome check <same file>           -> exit 0
```

The formatted result matches the committed catalog's shape exactly: `input: ["text", "image"]`, unquoted `openrouter` key, tab-indented `cost` with a trailing comma.

GREEN, running the real generator against the live OpenRouter API in this worktree:

```
Fetched 54 image models from OpenRouter
Checked 1 file in 11ms. Fixed 1 file.
Generated .../packages/ai/src/image-models.generated.ts

$ biome check packages/ai/src/image-models.generated.ts   -> exit 0
$ git diff --stat -- packages/ai/src/image-models.generated.ts
(empty)
```

Two things worth reading there. `Fixed 1 file` is the new step doing real work, so it is load-bearing rather than a no-op. And the regenerated catalog is **byte-identical to the committed one**, which is the strongest available evidence: the generator now reproduces the file the repository already ships, where before it produced one the gate rejects.

## QA

- `npm run check` from the repository root: **exit 0** (the exact gate that failed the release).
- No change to `packages/ai/src/image-models.generated.ts` is included; the release job resets and regenerates it anyway, so this PR is the generator plus its tracker and changelog entries.

Fixes #1886

---
## Summary by cubic
Fixes the image-model generator so the catalog it writes passes `npm run check`, which was blocking releases. The generator’s hand-formatted output didn’t match Biome, and the shared `check` script used to rewrite it silently until that gate became read-only. The generator now formats the file with Biome before reporting success, and regenerating against the live API produces a byte-identical file to the committed catalog. Fixes #1886.

<sup>Written for commit 6ff2913e11a783347f3c82e91dfa9b2dbc06a0e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/code-yeongyu/senpi/pull/1887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


code-yeongyu:
Additional evidence: the **full release sequence** was reproduced locally, not just the image generator in isolation. `release-artifacts.mjs` runs both generators before `release.mjs` calls `npm run check`, so the whole chain is what matters for the re-dispatch.

```
$ npm --prefix packages/ai run generate-models          # full catalog, live network
  bai: 56 models
  deepseek: 2 models
  ...
  azure-openai-responses: 39 models
  exit 0

$ npm --prefix packages/ai run generate-image-models
  Fetched 54 image models from OpenRouter
  Checked 1 file in 26ms. Fixed 1 file.
  exit 0

$ npm run check
  ...
  Claude Agent SDK platform packages are locked.
  exit 0
```

Two details worth noting.

`Fixed 1 file` appears on every run of the image generator, so the new format step is doing real work each time rather than being a no-op that happened to pass once.

After both generators ran, `git status` showed changes only in `packages/ai/src/providers/data/*.json` — live catalog drift, which is intended release content. Neither `models.generated.ts` nor `image-models.generated.ts` was modified, meaning both generators now reproduce the committed files exactly. Before this change the image catalog came back in a shape `npm run check` rejects, which is what stopped run 35520504862.

This is the specific gate that failed the release, run in the order the release runs it, so the re-dispatch should clear it.

